PR #17, opened 2018-01-30
Add gitignore

matthewfeickert:
The Jupyter gitignore
c.f. https://github.com/jupyter/notebook/blob/master/.gitignore


.gitignore
```diff
@@ -0,0 +1,40 @@
+MANIFEST
+build
+dist
+_build
+docs/man/*.gz
+docs/source/api/generated
+docs/source/config.rst
+docs/gh-pages
+notebook/i18n/*/LC_MESSAGES/*.mo
```
lukasheinrich: the notebook/* lines are specific to the jupyter/notebook repo and probably should be removed

.gitignore
```diff
@@ -0,0 +1,40 @@
...
+notebook/i18n/*/LC_MESSAGES/nbjs.json
+notebook/static/components
+notebook/static/style/*.min.css*
+notebook/static/*/js/built/
+notebook/static/*/built/
+notebook/static/built/
+notebook/static/*/js/main.min.js*
+notebook/static/lab/*bundle.js
+node_modules
```
lukasheinrich: since we don't have any javascript modules (yet), this can also be removed

.gitignore
```diff
@@ -0,0 +1,40 @@
...
+*.egg-info
+*~
+*.bak
+.ipynb_checkpoints
+.tox
+.DS_Store
+\#*#
+.#*
+.coverage
+src
```
lukasheinrich: same as above

lukasheinrich:
also can you rebase to the current master? it seems your branch still hast the old `.travis.yml`

matthewfeickert:
@lukasheinrich unless you think we should also take out the `docs/*` lines then I think this is okay now.

lukasheinrich:
I think the docs lines should be fine, since hopefully we'll have some actual docs as well :)

LGTM

PR #44, opened 2018-02-01
Beautify to PEP 8 standards

matthewfeickert:
In preparation of adding many docstrings to the code, autopep8 is used to enforce the PEP 8 style standards.

Refinements can be made to taste as needed.

c.f.:
  https://pypi.python.org/pypi/autopep8
  https://www.python.org/dev/peps/pep-0008/


lukasheinrich:
Thanks @matthewfeickert !

PR #54, opened 2018-02-05
Add binder

matthewfeickert:
Add support for [Binder](https://mybinder.org/) using a `requirements.txt` file.


matthewfeickert:
@lukasheinrich let me know if you think there is any other packages that we should have pip installed. I've left out PyTorch for the time being as if you currently try to pip install it you run into some dependency issues and it just tells you to use Anaconda instead. Binder works with Anaconda, but I figured that is an exercise for a later day.

lukasheinrich:
Hi @matthewfeickert can you move the `requirements.txt` into a directory called `binder`. repo2docker looks there as well, and it's a cleaner separation that this file is needed for binder (not to be confused with a `requirements.txt` that has dependencies needed for `pyhf` which would not include `pyhf`). See docs here http://repo2docker.readthedocs.io/en/latest/usage.html#preparing-your-repository

matthewfeickert:
@lukasheinrich [It would appear so](https://github.com/jupyterhub/binderhub/issues/381#issuecomment-354245154). Let me do that now.

matthewfeickert:
@lukasheinrich Should be ready to :shipit: now.

lukasheinrich:
LGTM! Thanks @matthewfeickert

PR #61, opened 2018-02-06
pytorch backend

lukasheinrich:


lukasheinrich:
needs:

* pytorch tests
* pyflakes passage
* coverage back to high 90s

matthewfeickert:
With PR #65 we have passing `pyflakes`, so I think we can rebase against `master` and then start writing some PyTorch tests.

lukasheinrich:
standard tests are passing.. need to add pytorch tests now

PR #78, opened 2018-02-09
Fix link to HistFactory paper on CDS

matthewfeickert:
The link previously 404'd, and it now directs to the proper entry on
CDS --- that being CERN-OPEN-2012-016, _HistFactory: A tool for creating
statistical models for use with RooFit and RooStats_,
https://cds.cern.ch/record/1456844


lukasheinrich:
Thanks!

Issue #80, opened 2018-02-09
MXNet tensor backend


cranmer:
https://mxnet.incubator.apache.org

lukasheinrich:
@matthewfeickert added this and it was merged.

matthewfeickert:
Closed by #83

PR #83, opened 2018-02-10
Add MXNet backend

matthewfeickert:
Add in a backend for [MXNet](https://mxnet.incubator.apache.org) as described in Issue #80.

TODO:
- [x] Finish the `mxnet_backend.py` methods
- [x] Get all tests passing
- [x] Refactor methods to use as few dependencies beyond `mxnet` as possible
- [x] Add [Sphinx compliant](https://pythonhosted.org/an_example_pypi_project/sphinx.html#function-definitions) doc strings
- [x] Add an example Jupyter Notebook
- [x] Add [Binder support](https://mybinder.readthedocs.io/en/latest/sample_repos.html#conda-environment-with-environment-yml)

This was made easier by the ["PyTorch to MXNet" README](https://github.com/zackchase/mxnet-the-straight-dope/blob/master/cheatsheets/pytorch_gluon.md) from the online book "[Deep Learning - The Straight Dope](https://github.com/zackchase/mxnet-the-straight-dope)"

### DIFF
--- a/pyhf/tensor/mxnet_backend.py
+++ b/pyhf/tensor/mxnet_backend.py
@@ -1,6 +1,5 @@
 from mxnet import nd
 import logging
-import itertools  # Hack fix for mxnet_backend.outer()
 log = logging.getLogger(__name__)
 
 
@@ -25,7 +24,7 @@ class mxnet_backend(object):
 
     def outer(self, tensor_in_1, tensor_in_2):
         """
-        The outer product of two tensors: u.v^T
+        The outer product of two tensors
 
         Args:
             tensor_in_1: tensor object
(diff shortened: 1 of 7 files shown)


pyhf/tensor/mxnet_backend.py
```diff
@@ -0,0 +1,183 @@
...
+from mxnet import nd
+import logging
+log = logging.getLogger(__name__)
+
+
+class mxnet_backend(object):
+    """Backend for MXNet"""
+
+    def __init__(self, **kwargs):
+        self.session = kwargs.get('session')
```
lukasheinrich: seems like MXNet does not have a concept of a session (at least I can't find a reference to it. So I think this can be removed?

matthewfeickert: Yeah, that was the result of some sloppy copypasta and me not checking things late at night. I'll fix that up when I get back to this after finishing work tonight.

pyhf/tensor/mxnet_backend.py
```diff
@@ -165,19 +179,85 @@ def exp(self, tensor_in):
...
+        for arg in args:
+            broadcast.append(self.astensor(arg)
+                             if len(arg) > 1 else arg * self.ones(max_dim))
+        return broadcast
 
     def poisson(self, n, lam):
-        pass
+        return self.normal(n, lam, self.sqrt(lam))
 
     def normal(self, x, mu, sigma):
```
lukasheinrich: You can probably use these APIs

https://mxnet.incubator.apache.org/api/python/symbol/symbol.html#mxnet.symbol.random.normal

lukasheinrich: but for poisson, let's stick for now to a gaussian approximation with mu = lambda , sigma = sqrt(lambda) since we need continuous poissons

matthewfeickert: Hm, maybe I'm misunderstanding what you mean, but we want the pdf of the distribution evaluated at `x`, yes? The MXNet distribution generators return sampling of the distributions given the parameters you pass, so [`mx.nd.random.normal()`](https://mxnet.incubator.apache.org/api/python/ndarray/random.html#mxnet.ndarray.random.normal) is akin to [`np.random.normal()`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.random.normal.html). I might have missed another API though.

Yeah, I follow you RE: the Poissons.

lukasheinrich: ah right, i didn't look to closely at the api seems like it more close to np.random stan scipy.stats.. sorry about that!

matthewfeickert: No worries! Just wanted to make sure I wasn't being dumb. :)

matthewfeickert:
@lukasheinrich Please don't accept this yet even though it passes. There are some things I want to fixup in it.

matthewfeickert:
@lukasheinrich I think this is ready for review, and if no comments then to be merged in.

lukasheinrich:
LGTM! thanks a lot!

matthewfeickert:
Closes Issue #80

Issue #77, opened 2018-02-09
create benchmarking code


lukasheinrich:
we'll be interested in scaling behaviour w.r.t number of bins / channels / systematics / etc.. as a start it should not be too hard to measure e.g. pdf evaluation time for a simple hepdata like model

e.g. this code snippet runs sets up a 2-bin likelihood, and we can just generate N-bin likelihoods with a easy loop that spits outs these JSON specs

@matthewfeickert maybe this is something for you?

```
source = {
  "binning": [2,-0.5,1.5],
  "bindata": {
    "data":    [120.0, 180.0],
    "bkg":     [100.0, 150.0],
    "bkgerr":     [10.0, 10.0],
    "sig":     [30.0, 95.0]
  }
}

from pyhf.simplemodels import hepdata_like
pdf  = hepdata_like(source['bindata']['sig'], source['bindata']['bkg'], source['bindata']['bkgerr'])
data = source['bindata']['data'] + pdf.config.auxdata

#now the call we want to benchmark:
pdf.logpdf(pdf.config.suggested_init(), data)
```

the timeit module will be useful for this

https://docs.python.org/2/library/timeit.html

lukasheinrich:
ideally i'd have something similar to `pytest` where we can just run a series of benchmarks under certaing conditions (e.g. tensor backend etc..) which produces a summary on the cmd line or as a txt file

matthewfeickert:
@lukasheinrich So, using [`pytest-benchmark`](http://pytest-benchmark.readthedocs.io/en/latest/index.html) it seems (to me so far at least) that we need to define a test function for every benchmark we want. I can check into this more (maybe ask the devs).

If we cant to use this (which I think we should) then to start off with some benchmarks on bins to expand upon the most important thing is just the number of bins, not necessarily the content (as long as it is somewhat reasonable), correct? Let me know if my question is confusing (it isn't phrased perhaps the best).

matthewfeickert:
Actually I think that [this Issue](https://github.com/ionelmc/pytest-benchmark/issues/48) might help out. I'll follow up tonight.

lukasheinrich:
@matthewfeickert yeah, i think a simple test that just clones the same single bin N times should be sufficient to observe a 0th approximation of the scaling behaviour. later on we might fuzz this out and e.g. have randomized tests

matthewfeickert:
[...]

Issue #86, opened 2018-02-12
implement MXNet autograd based optimizer (newton's method)


lukasheinrich:
now that we have a mxnet backend #83 we can continue to get a optimizer for that backend.

PR #92, opened 2018-02-27
Add benchmarks using pytest-benchmark

matthewfeickert:
Add initial benchmakrs using the features of [`pytest-benchmark`](https://github.com/ionelmc/pytest-benchmark), as outlines in Issue #77.

- [x] Tests are passing


matthewfeickert:
It seems that `test_benchmark.py` is modifying the testing environment in some way. This can be seen as enforcing an order in the way pytest executes the tests causes differing results.

The tests all pass if run individually, but
```
pytest tests/test_import.py tests/test_benchmark.py
```
passes, while
```
pytest tests/test_benchmark.py tests/test_import.py
```
fails.

I'll need to look into this.

lukasheinrich:
hi @matthewfeickert 
what does the local output look like? on travis I just see 

```
tests/test_benchmark.py ............................................     [ 66%]
```

can we extract some more detailed diagnostics locally?

matthewfeickert:
@lukasheinrich That's weird. In the Travis logs I can [see the normal output](https://travis-ci.org/diana-hep/pyhf/jobs/346786418#L2110) (starting at line 2110).

RE: running locally: yes! you get some nicer color coded features that will show the min and max for each feature column. So you should get similar output as shown in the [`pytest-benchmark` "normal run" screenshot](https://github.com/ionelmc/pytest-benchmark#screenshots) if you just run `pytest` locally.

lukasheinrich:
@matthewfeickert huh.. I see it now on reload. Maybe Travis got stuck loading all the outputlines.  Looks nice! have you tried running with `--benchmark-histogram`?

matthewfeickert:
@lukasheinrich I haven't yet, but that's going to be what I do tonight after doing a few more tweaks. I think that should be really nice to have some visualizations. :+1:

lukasheinrich:
@matthewfeickert yeah, we could even have this as part of the docs building and push the results to the static webpage

matthewfeickert:
@lukasheinrich I really like that idea. You might need to show me some things about Sphinx, but I imagine that should be pretty easy to do as well.

matthewfeickert:
@lukasheinrich The default `.svg` output has tooltips too. :+1: 
benchmark_2018-02-27

lukasheinrich:
Hi @matthewfeickert can you add zero-padding to the names, this might make the plot nicer as it would order by the number the bins correctly

sth like `'tensorflow-{0:03d}'.format(nbins)`

matthewfeickert:
@lukasheinrich Unfortunately I don't think that will work. The test results are ordered along the x-axis of the box-plot graph in the same order they would be printed to the screen during a normal run, which is in ascending order with respect to the `--benchmark-sort` [command line option](http://pytest-benchmark.readthedocs.io/en/stable/usage.html#commandline-options) (default is `Min` --- the minimum time for the test to complete any of the rounds it is put through).

I'm currently treating all the backends as parameters of a test, instead as being unique tests (c.f. [`pytest-benchmark` Issue #101](https://github.com/ionelmc/pytest-benchmark/issues/101)). If you think that it makes more sense to break them out into having individual benchmark tests by backends (instead of parameterizing everything into a single test) I can easily do that. Thoughts?

tests/test_benchmark.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def generate_source(n_bins):
+    """
+    Create the source structure for the given number of bins.
+    Currently the data that are being put in is all the same
+    but in the future it should be generated pseudodata.
+    """
+    binning = [n_bins, -0.5, 1.5]
+    data = [120.0]
```
lukasheinrich: This could be implemented more cleanly. e.g. you could do something like

```
data = [120.] * n_bins
bkg = [100.] * n_bins
bkgerr = [10.0] * n_bins
sig = [30.0] * n_bins
```

lukasheinrich: maybe we can have random varations but with a fixed see, so that it's reproducible

tests/test_benchmark.py
```diff
@@ -0,0 +1,69 @@
...
+import pytest
+
+
+def generate_source(n_bins):
+    """
+    Create the source structure for the given number of bins.
+    Currently the data that are being put in is all the same
+    but in the future it should be generated pseudodata.
+    """
+    binning = [n_bins, -0.5, 1.5]
```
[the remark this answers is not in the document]
lukasheinrich: for the fit it's irrelevant, this is mostly for plotting in the end

matthewfeickert:
@lukasheinrich After our discussion on 2018-04-01 I am seeing that when using `pytest-benchmark`'s fixtures that the DAG backends are extremely slow. I'm going to do my own tests with `timeit` to compare them and see if there is a considerable difference. If so, then it might be worth having me spend time seeing what `pytest-benchmark` is actually doing and if it is throttling the parallelism.

matthewfeickert:
Rebased against `master`

matthewfeickert:
@lukasheinrich Ignoring the weirdness from Travis described in Issue #103, everything is passing. Further related questions:
- Given the benchmarking `runOnePoint` tests that are run with `test_benchmark.py` and `test_backend_consistency.py` the CI takes about 16 minutes to fully complete. Do we want to pair down any of the nuisance parameter/bin tests to speed things up in the CI as this will take even longer once the MXNet optimizer is finished, or leave it as is?
- Should we add in the [`WS_Builder_Numpy` benchmarks](https://github.com/diana-hep/pyhf/issues/77#issuecomment-380572211) here, or should that be a seperate PR?
- If no to all of the above, should we merge this is now, or should I finish Issue #86 and then once that is merged in rebase this against `master` and then merge to catch those new tests?

lukasheinrich:
@matthewfeickert I think we can merge this now. We could think of only testing performance on `master` builds via build filters

https://docs.travis-ci.com/user/conditional-builds-stages-jobs/

ideally we'd have something that would be quick in a PR setting, and just before merging we could trigger a new build that tests performance. But we can cross that bridge when it comes unbearable.

I would really like to see WS_builder_Numpy added, but agree it should be a separate PR.

tldr: LGTM

matthewfeickert:
Okay. I'll add the build filter into this MR as well, and then we can merge it this evening. I'll open up an issue for the WS_builder_Numpy.

kratsg:
@matthewfeickert a little nitpicky, but can you indent the .travis.yml file?

matthewfeickert:
@kratsg absolutely. I just don't run my beautifier on other people's code without asking in advance. Do you have a particular one you'd like? Mine tends to be a bit "aggressive" with indenting (see what I mean?)
aggressive_beautifier

kratsg:
Oh, I guess I meant more like

```
python:
  - '2.7'
  - '3.5'
  - '3.6'
...
```

I think your beautifier is just aligning on colons.

matthewfeickert:
@lukasheinrich I've believe that I've got things setup so that for PRs we only test (and we only test the `pr` builds not the `push` builds) and then when PRs are then actually merged into `master` we test, benchmark, build the docs, and deploy to PyPI. This brings running the CI for a commit in the PR to around 10 minutes (still bottlenecked by having to install Conda for PyTorch). Sound okay, or are there other stages beyond test that we always want to run?

@kratsg, maybe once this PR is merged into `master` you can also look at the `.travis.yml` and improve it in a new PR (both in terms of beautification and functionality --- I don't think my instance is perhaps the "best")?

Something that in general confuses/concerns me is that _sometimes_ the tests fail in Python 3.5 but _not_ in Python 3.6. Restarting the job seems to usually fix this, but the nondeterministic nature of it bothers me. I'll reopen Issue #103.

lukasheinrich:
@matthewfeickert i'll merge this. but I do remember seeing something similar in yadage wit python 3.5 (also 3.5 doesn't seem like the most popular version). I'll try to remember

matthewfeickert:
Damnit, I messed it up the YAML as the merged PR is only running the tests. I'll open a new one to fix it, sorry.

matthewfeickert:
Okay, fixed in PR #112, so things are working as intended now. :+1:

Example: Travis CI [build 336](https://travis-ci.org/diana-hep/pyhf/builds/365891896?utm_source=github_status&utm_medium=notification)

Issue #99, opened 2018-04-01
Bug Report: Optimizer not set when backend is changed


matthewfeickert:
# Description

If using a backend other than `numpy_backend` currently we have to manually set the optimizer. However, this should be done automatically when the backend is changed.

Otherwise this causes an error, as shown below with `pyhf.runOnePoint()`.

# Expected Behavior

When changing `pyhf.tensorlib` have this change be automatically detected and also change `pyhf.optimizer` appropriately.

# Actual Behavior

When changing `pyhf.tensorlib` `pyhf.optimizer` must be manually set. If it is not, then errors can be caused, as seen below.

# Steps to Reproduce

```python
import pyhf
from pyhf.tensor.tensorflow_backend import tensorflow_backend
from pyhf.simplemodels import hepdata_like
import tensorflow as tf

if __name__ == '__main__':
    default_backend = pyhf.tensorlib
    pyhf.tensorlib = tensorflow_backend(session=tf.Session())

    n_bins = 5
    binning = [n_bins, -0.5, n_bins + 0.5]
    data = [120.0] * n_bins
    bkg = [100.0] * n_bins
    bkgerr = [10.0] * n_bins
    sig = [30.0] * n_bins
    source = {
        'binning': binning,
        'bindata': {
            'data': data,
            'bkg': bkg,
            'bkgerr': bkgerr,
            'sig': sig
        }
    }

    pdf = hepdata_like(source['bindata']['sig'],
                       source['bindata']['bkg'],
                       source['bindata']['bkgerr'])
    data = source['bindata']['data'] + pdf.config.auxdata

    pyhf.runOnePoint(1.0, data, pdf,
                     pdf.config.suggested_init(),
                     pdf.config.suggested_bounds())

    # Reset backend
    pyhf.tensorlib = default_backend
```
**Traceback:**
```python-traceback
Traceback (most recent call last):
  File "/home/mcf/anaconda3/envs/pyhf/lib/python3.6/site-packages/scipy/optimize/slsqp.py", line 380, in _minimize_slsqp
    fx = float(np.asarray(fx))
TypeError: float() argument must be a string or a number, not 'Tensor'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "runOnePoint_example.py", line 36, in <module>
    pdf.config.suggested_bounds())
  File "/home/mcf/Code/GitHub/pyhf/pyhf/__init__.py", line 412, in runOnePoint
    pdf, init_pars, par_bounds))
[...]

PR #100, opened 2018-04-01
Change the optimizer with the backend

matthewfeickert:
`pyhf.set_backend()` detects the backend that is being set and then globally sets
the backend and the corresponding optimizer. This is necessary to avoid
errors in data types and would normally have to be done manually.

This solves Issue #99 and will be fully implemented for all optimizers once Issue #86 is finished.

- [x] Tests are passing

### DIFF
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -20,7 +20,9 @@ except ImportError:
 
 try:
     from .tensor.mxnet_backend import mxnet_backend
+    # from .optimize.opt_mxnet import mxnet_optimizer
     assert mxnet_backend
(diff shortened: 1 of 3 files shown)


lukasheinrich:
Hi @matthewfeickert, thanks for the PR. LGTM. In the future they might be a many-to-many relationship between optimizers (e.g. multiple methods) but for now I think the one-to-one is fine.

matthewfeickert:
@lukasheinrich Yeah, I wasn't envisioning this as a fix all, but more of a setting defaults. As we move to many-to-many we can add the abilities for users to specify an optimizer with a backend.

PR #101, opened 2018-04-03
Install PyTorch with pip

### LINKED ISSUE #67: speed up CI tests (do we need all conda packages?)
By using Conda, unfortunately the setup phase of the CI jobs have become a bit slower than without conda, maybe we can look into speeding them up again by checking whether we need all the packages that we install during CI

matthewfeickert:
Now that [PyTorch is on PyPI](https://twitter.com/pytorch/status/981198007352705025) Conda is no longer required to get an easy install of stable builds. As PyTorch was the only reason Conda was being installed during CI Miniconda no longer needs to be part of the Travis YAML file. This should result in a speedup of the CI.

Resolves #67 

- [x] Tests are passing


matthewfeickert:
Still getting the memory error after all tests pass:
```
=================== 24 passed, 26 warnings in 156.84 seconds ===================
*** Error in `/home/travis/virtualenv/python3.6.3/bin/python': double free or corruption (!prev): 0x000000000137f0f0 ***
/home/travis/.travis/job_stages: line 57:  4879 Aborted                 (core dumped) pytest --ignore tests/benchmarks/
```
So further work needs to be done.

matthewfeickert:
Using the [Python TravisCI Docker testing image](https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image) I was able to see that the reason that in the past the builds have succeeded on the Python 3.6 tests in Travis but failed in Python 2 is that the Travis Python 2 version is too old.

```
$ pip install -U --process-dependency-links -e .[develop]
...
Collecting tornado>=4.0 (from ipykernel->jupyter->pyhf==0.0.8)
  Using cached https://files.pythonhosted.org/packages/cf/d1/3be271ae5eba9fb59df63c9891fdc7d8044b999e8ac145994cdbfd2ae66a/tornado-5.0.2.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-Wk7Aqw/tornado/setup.py", line 146, in <module>
        raise ImportError("Tornado requires an up-to-date SSL module. This means "
    ImportError: Tornado requires an up-to-date SSL module. This means Python 2.7.9+ or 3.4+ (although some distributions have backported the necessary changes to older versions).
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-Wk7Aqw/tornado/
You are using pip version 9.0.1, however version 10.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
$ python --version
Python 2.7.6
```

kratsg:
Can you report this to Travis-CI? I don't see an existing issue for it. This seems like a weird problem though.

matthewfeickert:
Having to force upgrade `numpy` after everything else has been installed feels like a really dumb patch to get this everything to build. I'd like to understand what is going on better before this gets finished up.

matthewfeickert:
@lukasheinrich @kratsg  I think this is ready for review. push CI builds now run in under 6 minutes, and pull CI builds should run in around 10 minutes.

If there are no comments or change requests please squash and merge this in.

lukasheinrich:
Hi @matthewfeickert can you remove the `--process-dependency-links` in the setup.py? I think I carried this over but it's actually not a recommended practice. Also we don't have any custom dependency_links anyways.

matthewfeickert:
@lukasheinrich:
> can you remove the `--process-dependency-links` in the `setup.py`>

Removed in 5904994

.travis.yml
```diff
@@ -4,25 +4,19 @@ python:
...
-  - conda info -a
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION pytorch scipy -c pytorch
-  - source activate test-environment
-  - pip install -U --process-dependency-links -e .[develop]
-  - pip install --upgrade pytest
+- sudo apt-get update
+- pip install --upgrade pip
+- pip install -U -q -e .[develop]
+# numpy upgrade needed for Python 2.7 only
+- pip install --upgrade pytest numpy
```
lukasheinrich: what version of numpy is needed? Can we encode this in `setup.py`?

matthewfeickert: Will check and try to fix this there.

matthewfeickert: @lukasheinrich Fixed in commit a7d3c3a

.travis.yml
```diff
@@ -4,25 +4,17 @@ python:
...
-  - export PATH="$HOME/miniconda/bin:$PATH"
-  - hash -r
-  - conda config --set always_yes yes --set changeps1 no
-  - conda update -q conda
-  - conda info -a
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION pytorch scipy -c pytorch
-  - source activate test-environment
-  - pip install -U --process-dependency-links -e .[develop]
-  - pip install --upgrade pytest
+- sudo apt-get update
```
lukasheinrich: seems like we can remove this

.travis.yml
```diff
@@ -42,25 +34,15 @@ jobs:
     python: '3.6'
     install:
       - sudo apt-get update
```
lukasheinrich: dito

.travis.yml
matthewfeickert: Removed in commit 6311608

lukasheinrich:
awesome -- this is a great improvement. thanks a lot @matthewfeickert

PR #104, opened 2018-04-10
Add starter json schema for spec

kratsg:
This PR adds a starter JSON schema that currently validates against all specs I could find in the pyhf code at the moment.

An interesting item was raised, during the writing of the spec, that the spec may need to be changed slightly in terms of modifications. Currently, modifications are defined like so:

```python
                'mods': [
                    {'name': 'mu', 'type': 'normfactor', 'data': None}
                ]   
```

however, this introduces the case where two modifications of different types might have the same name! This could easily be solved (where a JSON schema cannot validate such conflicts) by restructuring like so

```python
                'mods': {
                    'mu': {'type': 'normfactor', 'data': None}
                }   
```

which has the added benefit of making the structure a little bit more intuitive. Instead of modifications being literally a list of modification "objects" (the constraint, etc...), the modifications then becomes a list of Nuisance Parameter names (e.g. true, unknown `alpha` values), which gains some speed up from python's dict-lookups as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,8 @@ setup(
        'tensorflow',
        'mxnet>=1.0.0',
        'graphviz',
-       'sphinx'
+       'sphinx',
+       'jsonschema>=2.6.0'
     ]
   },
   entry_points = {
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -76,7 +76,7 @@ def test_core_pdf_broadcasting():
     assert np.all(naive_python  == broadcasted)
 
 def test_pdf_integration_histosys():
-    schema = json.load(file('validation/spec.json'))
+    schema = json.load(open('validation/spec.json'))
     source = json.load(open('validation/data/2bin_histosys_example2.json'))
     spec = {
         'singlechannel': {
@@ -122,7 +122,7 @@ def test_pdf_integration_histosys():
 
 
 def test_pdf_integration_normsys():
-    schema = json.load(file('validation/spec.json'))
+    schema = json.load(open('validation/spec.json'))
     source = json.load(open('validation/data/2bin_histosys_example2.json'))
     spec = {
         'singlechannel': {
@@ -154,7 +154,7 @@ def test_pdf_integration_normsys():
     assert pdf.expected_data(pars, include_auxdata = False).tolist()   == [100*0.9,150*0.9]
 
 def test_pdf_integration_shapesys():
-    schema = json.load(file('validation/spec.json'))
+    schema = json.load(open('validation/spec.json'))
     source = json.load(open('validation/data/2bin_histosys_example2.json'))
     spec = {
         'singlechannel': {
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -2,6 +2,7 @@ import pyhf
 import pyhf.simplemodels
 import numpy as np
 import json
+import jsonschema
 
 def test_interpcode_0():
     f = lambda x: pyhf._hfinterp_code0(at_minus_one = 0.5, at_zero =1, at_plus_one = 2.0, alphas = x)
@@ -75,6 +76,7 @@ def test_core_pdf_broadcasting():
     assert np.all(naive_python  == broadcasted)
 
 def test_pdf_integration_histosys():
+    schema = json.load(file('validation/spec.json'))
     source = json.load(open('validation/data/2bin_histosys_example2.json'))
     spec = {
         'singlechannel': {
@@ -93,6 +95,7 @@ def test_pdf_integration_histosys():
             }
         }
     }
+    jsonschema.validate(spec, schema)
     pdf  = pyhf.hfpdf(spec)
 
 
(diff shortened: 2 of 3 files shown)


matthewfeickert:
@kratsg I schema validator seems like it would indeed be useful. Thanks for the PR! I'll let @lukasheinrich review this in case he can think of anything else that we might want in it.

lukasheinrich:
hi @kratsg 

thanks for the PR! Usually I think it is considered a best practice to have deterministic property names (in fact I was actually considering switching from a dictionary to lists in channels and samples) and let only the value types be dynamic. This allows for e.g. for easier processing of JSON documents.

E.g. to filter all JES-type systematics you could do 

```
{
  "singlechannel": {
    "background": {
      "data": [1,2,3,4],
      "mods": [
        {
          "type": "shapesys",
          "name": "mod_JES1",
          "data": [1,2,3,4]
        },
        {
          "type": "shapesys",
          "name": "mod_JES2",
          "data": [1,2,3,4]
        },
        {
          "type": "shapesys",
          "name": "mod_FlavTag",
          "data": [1,2,3,4]
        }
      ]
    },
    "signal": {
      "data": [1,2,3,4],
      "mods": [
        {
          "type": "normfactor",
          "name": "mu",
          "data": null
        }
      ]
    }
  }
}
```

i.e. to get all variations that contain "JES" in the name

```
jq '.singlechannel.background.mods[]|select(.name|contains("JES"))'
```

while with dynamic property names it's a bit more awkward

```
jq '[.singlechannel.background.mods|to_entries[]|select(.key|contains("JES"))]|from_entries'
```

we hit is problem for example when designing the yadage JSON schemas for the CERN analysis preservation project, b/c e.g. a lot of web-forms can't handle dynamic property names very well. I do agree that it adds/hardcodes a nice semantic check on the modification names. Also, I suspect we probably would need to do more validation beyond syntactic correctness anyways (e.g. that all "data" arrays in the samples have the same length etc", so one could add these semantic checks there.

My opinion is not too strong on this, happy to hear your thoughts

lukasheinrich:
one example for a system that does not support `patternProperties` is for example the OpenAPI / Swagger spec: https://swagger.io/docs/specification/data-models/keywords/. E.g. Once likelihoods are pure JSON docs, it would be quiet natural for a service  (think: HepData) to support e.g. POST'ing a likelihood but with dynamic names one could e.g. not use this natively with OpenAPI

kratsg:
> one example for a system that does not support patternProperties is for example the OpenAPI / Swagger spec: https://swagger.io/docs/specification/data-models/keywords/. E.g. Once likelihoods are pure JSON docs, it would be quiet natural for a service (think: HepData) to support e.g. POST'ing a likelihood but with dynamic names one could e.g. not use this natively with OpenAPI

You can get around this with

```
type: object
additionalProperties: 
  type: string
```

at least, that's how I've done it before with an OpenAPI spec. patternProperties made its way into the JSON schema (draft v4). That said, this is a bit of a tricky problem between JSON API specification and intuitiveness. It does seem like we both agree on having a dict of key=alpha names, and values= configurations for the name.

For the spec I've written, this is probably the point of the PR to design the first draft of a live document, that can be used with something like https://github.com/cloudflare/doca. Another issue I see is when you have two different modifications with the same name and two different types in different samples... 

Moving forward, I'll file a issue to add spec proposals/structuring since we'll want to design it well. This PR should currently describe all current structuring.

kratsg:
Also, the deterministic version of the above you posted would then be...

```
[
  {"name": "singlechannel",
   "type": "channel",
   "samples": [
      {
        "name": "background",
        "data": [1,2,3,4],
        "mods": [
          {
            "type": "shapesys",
            "name": "mod_JES1",
            "data": [1,2,3,4]
          },
        {
          "type": "shapesys",
          "name": "mod_JES2",
          "data": [1,2,3,4]
        },
        {
          "type": "shapesys",
          "name": "mod_FlavTag",
          "data": [1,2,3,4]
        }
    ]
  },
  {
    "name": "signal",
    "data": [1,2,3,4],
    "mods": [
        {
          "type": "normfactor",
          "name": "mu",
          "data": null
        }
      ]
    }
  }
}
```

lukasheinrich:
Hi @kratsg, 

yes I think it's perfectly fine to add a JSON schema for the current specs and build out the validation infrastructure etc while at the same time discussing what the final schema should be. I am indeed partial to the "deterministic version" you posted https://github.com/diana-hep/pyhf/pull/104#issuecomment-380075751 which I think is most explicit.

But yes the fully-explicit JSON might not be the easiest / most intuitive to work with. That's why I was experimenting e.g. with "shortcut spec factories" like `pyhf.simplemodels.hepdata_like` which have a simpler API and spit out fully-defined pyhf specs

also in the tests we translate e.g. a "simple JSON" like 

https://github.com/diana-hep/pyhf/blob/master/validation/data/1bin_histosys.json

into  a full json (see e.g. https://github.com/diana-hep/pyhf/blob/master/tests/test_pdf.py#L78)

there is some prior work on "full JSON spec" / "lite JSON spec" from 

https://vega.github.io/vega/examples/
https://vega.github.io/vega-lite/

that we could riff off of

kratsg:
Ok, I suggest we merge this in, and I will file an issue to continue this discussion as the PR is probably not the *right* place for it.

lukasheinrich:
LGTM

Issue #105, opened 2018-04-10
JSON Schema / Spec discussion


kratsg:
As initiated in #104, there are questions raised about the spec and the way forward with two overarching goals in mind:

- intuitive and clean for the user
- fully-specified and documentable by a schema for an API

There are two main issues raised as described below.

# Fully-Specified

A schema like

```
{
  "singlechannel": {
    "background": {
      "data": [1,2,3,4],
      "mods": [...]
    },
    "signal": {
      "data": [1,2,3,4],
      "mods": [...]
    }
  }
}
```

is not fully-specified as it contains a dictionary with variable key-names (`singlechannel`, `background`, `signal`). A more fully-specified spec looks like so

```
[
  {
    "name": "singlechannel",
    "type": "channel",
    "samples": [
      {
        "name": "background",
            "data": [1,2,3,4]
        ],
        "mods": [...]
      },
      {
        "name": "signal",
        "data": [1,2,3,4],
        "mods": [...]
      }
    ]
  }
]
```

where an array of channels, and samples are specified. This is a first proposal, but still has a nested array which may or may not be useful for many -- and flattening the array is a possibility, through a process of denormalization (see [firebase docs](https://firebase.google.com/docs/database/web/structure-data)).

# Intuitive-ness

Currently, modifications are defined as an array

```
        "mods": [
          {"type": "shapesys", "name": "mod_JES1", "data": [1,2,3,4]},
          {"type": "shapesys", "name": "mod_JES2", "data": [1,2,3,4]},
          {"type": "shapesys", "name": "mod_FlavTag", "data": [1,2,3,4]}
        ]
```

however, one of the draw-backs is that it makes a user think of each modification as an entire "object". That is, this should define three modification objects, which is not necessarily true. In spirit, a modification refers to a nuisance parameter, such as `mod_JES1` along with configurations for that. 

## first proposal

A first proposal to make this more intuitive was to structure the modifications as a dictionary, with each key name referring to the nuisance parameter that is of interest

```
        "mods": {
          "mod_JES1": {"type": "shapesys", "data": [1,2,3,4]},
          "mod_JES2": {"type": "shapesys", "data": [1,2,3,4]},
          "mod_flavTag": {"type": "shapesys", "data": [1,2,3,4]}
        ]
```

A drawback is that we now have configurable dictionary key names, which does not help with JSON Schema / API specification.

## second proposal

which separates the nuisance parameter from the actual modifications for a given sample/channel

```
        "NPs": [
[...]

PR #113, opened 2018-04-14
feature: exclusive specifications

kratsg:
Moving towards a fully-specified spec as described in #105. This requires updating:

Fully-specified channels:
- [x] `tests/*.py` (to make sure tests have right structure)
- [x] `validation/spec.json` (to document new spec)
- [x] `pyhf/__init__.py` (pyhf code to parse different spec structure)
- [x] `pyhf/readxml.py` (so that XML parse to the new structure)
- [x] `pyhf/simplemodels.py` (so that the simplemodel has right structure)
- [x] update notebooks
- [x] tests pass

Fully-specified samples:
- [x] `tests/*.py` (to make sure tests have right structure)
- [x] `validation/spec.json` (to document new spec)
- [x] `pyhf/__init__.py` (pyhf code to parse different spec structure)
- [x] `pyhf/readxml.py` (so that XML parse to the new structure)
- [x] `pyhf/simplemodels.py` (so that the simplemodel has right structure)
- [x] update notebooks
- [x] tests pass


kratsg:
The first half of the fully-specified spec changes are in. The code is very well-written btw which made this easier than expected 👍 Have a look over and see if you're ok with what I did for `samples` and I'll do the same for channels soon.

Side note: in cases where I could simplify function calls by passing in a sample object, instead of `sample_name`, `sample_def` separately, I combined them into a single `sample` argument which is generally cleaner. One side-effect is that some functions are passed in `channel` (the name of channel) without needing it anymore because you don't need to look up the sample definitions from `self.channels[channel]`.

validation/spec.json
```diff
@@ -60,7 +56,7 @@
 
 	"type": "object",
```
lukasheinrich: can you move this to the top, this will make it more intuitive that this is the toplevel object. Also I think the indenting is off.

validation/spec.json
```diff
@@ -60,7 +56,7 @@
...
 	"patternProperties": {
-		"^[a-zA-Z]+$": {
+		"^[a-zA-Z0-9]+$": {
```
lukasheinrich: just a comment: The XML schema allows generic CDATA for names, so something like `hello_world_my_channel` is also possible (much to my annoyance since RooFit uses `_` also as separator in its function names).. so we should be aware of that with an eye towards ROOT HF compatibility, but i'm fine with extending this regex on a when-needed basis

validation/spec.json
```diff
@@ -4,23 +4,19 @@
...
       "type": "object",
       "properties": {
-        "data": { "$ref": "#/definitions/data"},
-        "mods": {
-          "type": "array",
-          "items": {"$ref": "#/definitions/modification"}
-        }
+        "name": { "type": "string" },
+        "data": { "$ref": "#/definitions/data" },
+        "mods": { "type": "array", "items": {"$ref": "#/definitions/modification"} }
```
lukasheinrich: should we be more expressive and call his `modification` / `modifier`?

validation/spec.json
kratsg: JSON schema tends to put definitions up top, and the top-level object at the bottom. It's a de-facto standard.. but this will actually be split into multiple separate schema files for each "object" later on. Up to you. I'll fix indenting when I get back.

validation/spec.json
```diff
@@ -60,7 +56,7 @@
```
lukasheinrich: this is a comment towards the `lo_hist`, `hi_hist` above..  it would be more consistent to move away from the `histogram` methaphor as we did in the main nominal histogram, where we named it `data`, so maybe this can be changed to `lo_data`, `hi_data` and be propagated throughout

lukasheinrich:
Hi @kratsg, thanks a lot for the PR. I think this looks good to me, and please do proceed to the channels. I added some minor comments that would be nice to be included as drive-by changes

matthewfeickert:
@kratsg This looks good to me too, though I'm still trying to educate myself on the spec and standards, so I'll defer to Lukas on all of this. Thanks for the PR and all the nice work.

validation/spec.json
kratsg: Noted.

validation/spec.json
lukasheinrich: ok

validation/spec.json
```diff
@@ -1,68 +1,63 @@
...
+                "data": { "$ref": "#/definitions/data"}
+            },
+            "additionalProperties": false
         },
-        {
-          "type": "object",
-          "required": ["lo", "hi"]
+        "data": {
+            "oneOf": [
+                {
```
lukasheinrich: should we maybe group those definitions into it's separate location in the definitions and have something like

```yaml
oneOf:
- {$ref: #/definitions/modifier_definitions/histosysdata}
- {$ref: #/definitions/modifier_definitions/normsysdata}
- {$ref: #/definitions/modifier_definitions/shapesysdata}
```

kratsg: Yeah, let's file an issue for this, as this is out of the scope of fully-specified channels/samples.

kratsg:
@lukasheinrich @matthewfeickert this should be all ready to go.

kratsg:
A side note, `channel_order` is completely removed, since migrating from dictionary keys to lists of dictionaries now means the order is preserved just by how python lists work.

matthewfeickert:
Oops. [`test_import.py` is failing now](https://travis-ci.org/diana-hep/pyhf/jobs/366537123#L2116).

kratsg:
> Oops. test_import.py is failing now.

Forgot the last commit. Sorry. The current `test_import` needs to be refactored a little bit better, I suspect. Right now, because it's lists of dicts, it's not as trivial to ask if the spec has a specific channel name or a specific sample in a specific channel. So it's generating a temporary list of names.

lukasheinrich:
ok this looks good to me. @matthewfeickert any idea why we don't have anymore travis tests for the branch and the merge commit? e.g. we used to have 2 builds per PR like here

https://github.com/diana-hep/pyhf/pull/101

matthewfeickert:
@lukasheinrich Yeah, I [changed it in PR #92](https://github.com/diana-hep/pyhf/pull/92/files#diff-354f30a63fb0907d4ad57269548329e3R7) to only do `pr` builds. My thought on this was inspired by xAH, but also along the fact that `push` builds only build against what is in the branch of the PR, but what we're more interested in is the `pr` builds which are also checking how those commits will work if they are merged into the target branch (in this case `master`).

If I'm missing an obvious reason why we would want the `push` build in the PR as well (and not just when it is merged into `master`) then I can change it back. It just seemed like a redundant build to me.

PR #119, opened 2018-04-14
Use pytest fixtures in tests

### LINKED ISSUE #66: create proper test fixtures
Many of the tests have large blocks of setup code that could be shared as proper `@pytest.fixtures` 

https://docs.pytest.org/en/latest/fixture.html

matthewfeickert:
This resolves #66.

- [x] Identify all tests that fixtures could be used in:
   - `test_optim.py `
   - `test_validation.py` (not so much as data isn't shared, but can make the validation part a function)
- [x] Determine if a single `conftest.py` can be used across all tests (not so helpful as data isn't shared across tests)
- [x] Beautify the spec for readability (use [PEP8](https://www.python.org/dev/peps/pep-0008/) and ad hoc choice of [standard-json](https://github.com/standard/standard-json) style)

- [x] Tests are passing

### DIFF
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -40,7 +40,9 @@ def set_backend(backend):
         None
 
     Example:
-        pyhf.set_backend(pyhf.tensor.tensorflow_backend(session=tf.Session()))
+        import pyhf.tensor as tensor
+        import tensorflow as tf
+        pyhf.set_backend(tensor.tensorflow_backend(session=tf.Session()))
     """
     global tensorlib
     global optimizer
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -24,8 +24,9 @@ def get_backend():
     return tensorlib, optimizer
 
 # modifiers need access to tensorlib
-#   make sure import is below get_backend()
+# make sure import is below get_backend()
 from . import modifiers
+
 
 def set_backend(backend):
     """
@@ -39,7 +40,7 @@ def set_backend(backend):
         None
 
     Example:
-        pyhf.set_backend(tensorflow_backend(session=tf.Session()))
+        pyhf.set_backend(pyhf.tensor.tensorflow_backend(session=tf.Session()))
     """
     global tensorlib
     global optimizer
@@ -47,10 +48,10 @@ def set_backend(backend):
     tensorlib = backend
     if isinstance(tensorlib, tensor.tensorflow_backend):
         optimizer = optimize.tflow_optimizer(tensorlib)
-    elif isinstance(tensorlib,tensor.pytorch_backend):
+    elif isinstance(tensorlib, tensor.pytorch_backend):
         optimizer = optimize.pytorch_optimizer(tensorlib=tensorlib)
     # TODO: Add support for mxnet_optimizer()
-    # elif isinstance(tensorlib, mxnet_backend):
+    # elif isinstance(tensorlib, tensor.mxnet_backend):
     #     optimizer = mxnet_optimizer()
     else:
         optimizer = optimize.scipy_optimizer()
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -4,7 +4,7 @@ import pytest
 
 
 @pytest.fixture
-def source():
+def source(scope='module'):
     source = {
         'binning': [2, -0.5, 1.5],
         'bindata': {
@@ -19,30 +19,36 @@ def source():
 
 
 @pytest.fixture
-def spec(source):
+def spec(source, scope='module'):
     spec = {
-        'channels': [{
-            'name': 'singlechannel',
-            'samples': [{
+        'channels': [
+            {
+                'name': 'singlechannel',
+                'samples': [
+                    {
                         'name': 'signal',
                         'data': source['bindata']['sig'],
-                        'modifiers': [{
+                        'modifiers': [
+                            {
                                 'name': 'mu',
                                 'type': 'normfactor',
                                 'data': None
-                        }]},
-                        {
+                            }]
+                    },
+                    {
                         'name': 'background',
                         'data': source['bindata']['bkg'],
-                        'modifiers': [{
+                        'modifiers': [
+                            {
                                 'name': 'bkg_norm',
                                 'type': 'histosys',
                                 'data': {
                                     'lo_data': source['bindata']['bkgsys_dn'],
                                     'hi_data': source['bindata']['bkgsys_up']
                                 }
-                        }]}]
-        }]
+                            }]
+                    }]
+            }]
     }
     return spec
 
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -70,6 +70,9 @@ def test_optim(source, spec, backend):
     pyhf.set_backend(pyhf.tensor.tensorflow_backend())
     pyhf.tensorlib.session = tf.Session()
     optim = pyhf.optimizer
+    if isinstance(pyhf.tensorlib, tensorflow_backend):
+        tf.reset_default_graph()
+        pyhf.tensorlib.session = tf.Session()
 
     result = optim.unconstrained_bestfit(
         pyhf.loglambdav, data, pdf, init_pars, par_bounds)
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -57,6 +57,13 @@ def spec(source):
     return spec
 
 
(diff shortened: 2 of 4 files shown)


lukasheinrich:
Hi @matthewfeickert ,

thanks for the PR. We can in addition use fixtures everytime we have something like `json.load(open('validation/data/*.json'))` 

e.g. 

```
@pytest.fixture
def 1bin_example1_source():
    return json.load(open('validation/data/1bin_example1.json'))

@pytest.fixture
def 1bin_example1_spec(1bin_example1_source):
   ##code to produce the full verbose spec based on source data

def test_XXXX(1bin_example1_source, 1bin_example1_spec)
     p = hfpdf(1bin_example1_spec)
     p.logpdf(1bin_example1_source['data'])
```

this should simplify the tests in

lukasheinrich:
@matthewfeickert yes, so the idea is that we have a list of test likelihoods that should try to probe all corners of the functionality of pyhf. Each test might probe a specific piece of functionality, but it would probe it on an array of test likelihoods that are sensitive to that piece of functionality

e.g.

if we test histosys we might have test specs that

* only have histosys in a single channel
* have histosys but also multiple channels
* have histosys but also other systematics

etc.. so these fixtures represent the test population from which each test can pick the ones applicable to it

tests/test_validation.py
```diff
@@ -415,30 +428,136 @@ def test_validation_2bin_2channel_coupledshapefactor():
...
+                                         source['bindata']['bkg'],
+                                         source['bindata']['bkgerr'])
+
+    data = source['bindata']['data'] + pdf.config.auxdata
+
+    assert len(pdf.config.suggested_init()) == 2
+    assert len(pdf.config.suggested_bounds()) == 2
+    validate_runOnePoint(pdf, data, 1.0, expected_result)
+
+
```
matthewfeickert: @lukasheinrich There is still cleaning and refactoring that needs to be done, but is the below along the lines of what you were thinking? Or by moving everything into one parameterized function did I miss the point?

lukasheinrich: hi @matthewfeickert yes this is exactly what I meant. possibly the fixture can just include both the spec and the expected results as say a dictionary, but no strong opinion on this

lukasheinrich:
Hi @matthewfeickert .. just checking in, should I review this or are you expecting more commits?

matthewfeickert:
@lukasheinrich I think this is probably ready to get reviewed by you and @kratsg (I didn't ping people right away as it was late when I submitted the last commit and the CI hadn't run yet). Beyond functionality, any review comments that you might have on simplifying/cleaning the structure of `test_validation.py` would of course be welcome.

pyhf/__init__.py
```diff
@@ -38,18 +39,18 @@ def set_backend(backend):
...
 
     tensorlib = backend
     if isinstance(tensorlib, tensor.tensorflow_backend):
         optimizer = optimize.tflow_optimizer(tensorlib)
-    elif isinstance(tensorlib,tensor.pytorch_backend):
+    elif isinstance(tensorlib, tensor.pytorch_backend):
         optimizer = optimize.pytorch_optimizer(tensorlib=tensorlib)
     # TODO: Add support for mxnet_optimizer()
-    # elif isinstance(tensorlib, mxnet_backend):
+    # elif isinstance(tensorlib, tensor.mxnet_backend):
```
kratsg: should say `pyhf.tensor.mxnet_backend`

tests/test_optim.py
```diff
@@ -1,119 +1,25 @@
...
-        "sig":     [30.0, 95.0]
-      }
+        'binning': [2, -0.5, 1.5],
+        'bindata': {
+            'data': [120.0, 180.0],
+            'bkg': [100.0, 150.0],
+            'bkgsys_up': [102, 190],
+            'bkgsys_dn': [98, 100],
+            'sig': [30.0, 95.0]
+        }
```
kratsg: JSON uses double-quotes. Is there a reason to prefer single-quotes in python?

matthewfeickert: No, that was my linter. I'll change it everywhere.

tests/test_optim.py
```diff
@@ -123,32 +29,64 @@ def test_optim_tflow():
...
+        pyhf.tensorlib.session = tf.Session()
 
-    result = optim.unconstrained_bestfit(pyhf.loglambdav, data, pdf, init_pars, par_bounds)
+    result = optim.unconstrained_bestfit(
+        pyhf.loglambdav, data, pdf, init_pars, par_bounds)
     assert pyhf.tensorlib.tolist(result)
 
-    result = optim.constrained_bestfit(pyhf.loglambdav, 1.0, data, pdf, init_pars, par_bounds)
+    result = optim.constrained_bestfit(
+        pyhf.loglambdav, 1.0, data, pdf, init_pars, par_bounds)
```
lukasheinrich: mu could be a test parameter

pyhf/__init__.py
kratsg: Well, a few lines above and below it, you use `pyhf.tensor....` so just be consistent.

lukasheinrich:
super-minor comment from my side otherwise LGTM. @kratsg ?

kratsg:
> super-minor comment from my side otherwise LGTM. @kratsg ?

Same. Do we need the spec-schema validation in these tests too? It would help catch if the spec wasn't written correctly.

tests/test_optim.py
lukasheinrich: hm, i would keep single quotes tbh. it's not a JSON string, it's a python dictionary (which might have been the result from a JSON parsing, but this is irrelevant I think), and the canonical quote for strings in python is single-quote

pyhf/__init__.py
matthewfeickert: No, the only place `pyhf.tensor` appears in the file is in the import and in the doc string example:
```python
    """
    Set the backend and the associated optimizer

    Args:
        backend: One of the supported pyhf backends: NumPy,
                 TensorFlow, PyTorch, and MXNet

    Returns:
        None

    Example:
        pyhf.set_backend(pyhf.tensor.tensorflow_backend(session=tf.Session()))
    """
```
It appears in the doc string as I am not making assumptions on what someone has imported. Do you want me to change the doc string instead of the import?

kratsg: Yes, that would be better.

tests/test_optim.py
kratsg: > hm, i would keep single quotes tbh. it's not a JSON string, it's a python dictionary (which might have been the result from a JSON parsing, but this is irrelevant I think), and the canonical quote for strings in python is single-quote

Is there really a convention? Or is this just a code-style that we agree upon? If so, it's fine to leave it as single quotes. Just note that if it was in double-quotes, one could just copy/paste directly into a JSON without having to change anything (my 2 cents).

pyhf/__init__.py
matthewfeickert: So you want something like
```
    Example:
        import pyhf.tensor as tensor
        pyhf.set_backend(tensor.tensorflow_backend(session=tf.Session()))
    """
```
?
It seems to me that doc string example really should not make any assumptions as it will most likely not be ready by someone reading the code, but rather in docs or in the suggestion/info box of an editor. I'm missing why this is better, so can you give an example?

kratsg: > It seems to me that doc string example really should not make any assumptions as it will most likely not be ready by someone reading the code, but rather in docs or in the suggestion/info box of an editor. I'm missing why this is better, so can you give an example?

Yes. At some point, the docstrings will need to get tested, so if we just start writing good examples now and clean up the structure later -- that makes it easier.

matthewfeickert: Ah, now that I understand that for doctest each doc-block is tested in a new python shell with only the code `pip install -e` the point makes sense. Thanks for pointing this out!

matthewfeickert: Fixed in bdc74a4

tests/test_optim.py
matthewfeickert: Good point. Fixed in 9ba95f1

kratsg:
LGTM.

matthewfeickert:
@lukasheinrich @kratsg, the spec-schema validation checks are in for all the spec generation fixtures, so I think this is ready for another (final?) review.

kratsg:
Lukas had a LGTM, the last thing was from my end so we're good 👍

PR #120, opened 2018-04-15
WIP: feature: modifiers-module

kratsg:
# Description

- [x] Migrate existing constrained modifiers (`shapesys`, `normsys`, `histosys`) to `pyhf/modifiers/*.py`
- [ ] Create new modifiers for `shapefactor`, `normfactor`
- [x] Create a modifier registry to keep track of loaded modifiers
- [x] Create decorator and meta-class for modifier registry
- [ ] Update `hfpy/__init__.py` to use modifier registry
- [ ] Get tests passing (need to load modifiers in tests that use it so they load in the registry for the code to see it)


lukasheinrich:
hi @kratsg ,

we shold take up this PR as an opportunity to clean up the terminology a bit. So currently the setup is

* each sample defines what modifiers it participates in via the spec
* on spec reading we add modiiers to the configuration of the pdf `hfpdf.config`

but now something perhaps confusing happens  in  `add_modifier ` at https://github.com/diana-hep/pyhf/blob/master/pyhf/__init__.py#L237

* for modifiers that are *constraints*  we add an object
* for modifiers that are not constraints `shapefactor`, `normfactor`

we do this because we separate out the

1) implementation of the modification
2) additions to the statistical model implied by the modification (only constraints have those, the `*factors` only affects the mean rates in the main poissonian piece of the pdf while the constraints do that *but also* add new constraint pdf terms (the gaussians)

the implementation of the modification happens here 

for normsys
* https://github.com/diana-hep/pyhf/blob/master/pyhf/__init__.py#L291

for histosys
* https://github.com/diana-hep/pyhf/blob/master/pyhf/__init__.py#L306

for shapesys, normfactor and shapefactor 
* https://github.com/diana-hep/pyhf/blob/master/pyhf/__init__.py#L286

these can probably also be factored out of the main hfpdf class, but in some sense your checkbox to add modifier classes for the factors currently a no-op

I discussed a bit with @cranmer on whether to group both 1) and 2) into the same object, but in light new variations we might want to add, it might be useful to keep those separate

pyhf/__init__.py
```diff
@@ -229,54 +156,28 @@ def add_modifier(self, name, npars, modifier, suggested_init, suggested_bounds):
...
-            self.add_modifier(
-                modifier_def['name'],
-                npars=1,
-                modifier=modifier,
-                suggested_init=[1.0],
-                suggested_bounds=[[-5, 5]])
-            self.modifier(modifier_def['name']).add_sample(channel, sample, modifier_def['data'])
+        # get the class associated with the modifier
+        modifier_cls = modifiers.registry.get(modifier_def['type'], None)
+        if modifier_cls is None:
```
lukasheinrich: I think this is not the same logic as it used to be

what was intended in the original code is:

* all samples call add_modifier for the modifiers they want to participate in

on the first add of a modifier with a given name it is added to `self.parmap`, on any subsequent add, we need to decide what to do. In principle all should be able to be shared, but we don't necessarily have that supported.

on the contrary this just seems to check for the existence of the modifier class

pyhf/__init__.py
```diff
@@ -390,7 +290,7 @@ def logpdf(self, pars, data):
         cut = int(data.shape[0]) - len(self.config.auxdata)
         actual_data, aux_data = data[:cut], data[cut:]
         lambdas_data = self.expected_actualdata(pars)
-        summands = tensorlib.log(_poisson_impl(actual_data, lambdas_data))
+        summands = tensorlib.log(tensorlib.poisson(actual_data, lambdas_data))
```
lukasheinrich: a rebase to master should get equivalent changes in. see #122

pyhf/__init__.py
```diff
@@ -204,20 +133,18 @@ def set_poi(self,name):
         assert s.stop-s.start == 1
         self.poi_index = s.start
 
-    def add_modifier(self, name, npars, modifier, suggested_init, suggested_bounds):
-        is_constraint = type(modifier) in [histosys_constraint, normsys_constraint, shapesys_constraint]
+    def add_modifier(self, name, modifier):
+        is_constrained = getattr(modifier, 'is_constrained', False)
```
lukasheinrich: `getattr` seems a bit unpythonic, maybe the modifiers should have `is_constrained()` in which they return their class-level attribute

pyhf/__init__.py
```diff
@@ -229,54 +156,28 @@ def add_modifier(self, name, npars, modifier, suggested_init, suggested_bounds):
...
-                suggested_init=[1.0],
-                suggested_bounds=[[-5, 5]])
-            self.modifier(modifier_def['name']).add_sample(channel, sample, modifier_def['data'])
+        # get the class associated with the modifier
+        modifier_cls = modifiers.registry.get(modifier_def['type'], None)
+        if modifier_cls is None:
+            raise RuntimeError('shared systematic not implemented yet (processing {0:s})'.format(modifier_def['type']))
+
+        # use inspect to figure out how to call the modifier
+        args = inspect.getargspec(modifier_cls.__init__).args
```
lukasheinrich: ideally, i'd like to try to not use `inspect` if we factor the modifiers out we should try to find a uniform API

pyhf/__init__.py
kratsg: Well, if we don't want abstract classes -- then can't really rely on duck-typing. We can just do `modifier.is_constrained` and this raises an error if it's not set. I guess that's fine.

pyhf/__init__.py
kratsg: done.

pyhf/__init__.py
kratsg: yeah, this is a monkey-patch until I get around to that part.

pyhf/__init__.py
kratsg: Yes, this logic is still in place. The refactor will actually merge `add_modifier_from_def` and `add_modifier` together as they're not really significantly different now that modifiers are separated out.

pyhf/__init__.py
kratsg: This is now fixed up.

pyhf/__init__.py
kratsg: I've added tests for this, so at least for our modifiers, tests will catch this.

pyhf/__init__.py
```diff
@@ -2,10 +2,29 @@
...
 import pyhf.tensor as tensor
 
-
 log = logging.getLogger(__name__)
 tensorlib = tensor.numpy_backend()
+default_backend = tensorlib
 optimizer = optimize.scipy_optimizer()
+default_optimizer = optimizer
+
+def get_backend():
```
lukasheinrich: hi @kratsg,

why is this needed? Shouldn't `pyhf.tensorlib` and `pyhf.optimizer` always point to the current backend?

pyhf/__init__.py
```diff
@@ -171,79 +126,57 @@ def set_poi(self,name):
...
+        modifier_cls = modifiers.registry.get(modifier_def['type'])
+
+        # check if modifier of the given name already exists
+        if modifier_def['name'] in self.par_map:
+            if modifier_cls.is_constrained:
+                log.info('accepting existing {0:s} (type: {1:s})'.format(modifier_def['name'], modifier_cls.__name__))
+            else:
+                log.info('accepting existing unconstrained factor')
+            return self.par_map[modifier_def['name']]['modifier']
+
```
lukasheinrich: This seems to *always* accept any shared modifier (e.g. if the first `if` clause passes). Before we had a filter to only accept shared modifiers if they are eiter a 

* histosys constrained modifier
* normsys constrained modifier
* any type of unconstrained modifier (e.g. shapefactor, normfactor)

notably shapesys should *not* be shared. Perhaps we should add this as a test

pyhf/modifiers/normfactor.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def alphas(self, pars):
+        raise NotImplementedError
+
+    def expected_data(self, pars):
+        raise NotImplementedError
+
+    def pdf(self, a, alpha):
+        raise NotImplementedError
+
```
lukasheinrich: the `alphas`,`expected data` and `pdf` methods only pertain to the constraint term. As these are unconstrained modifiers, they should not only not be implemented but not be methods at all, as they're not part of the interface for unconstarined modifiers

pyhf/modifiers/shapefactor.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def alphas(self, pars):
+        raise NotImplementedError
+
+    def expected_data(self, pars):
+        raise NotImplementedError
+
+    def pdf(self, a, alpha):
+        raise NotImplementedError
+
```
lukasheinrich: same as above

pyhf/__init__.py
```diff
@@ -171,79 +126,57 @@ def set_poi(self,name):
...
+        # did not return, so create new modifier and return it
+        modifier = modifier_cls(sample['data'], modifier_def['data'])
+        npars = modifier.n_parameters
+        suggested_init = modifier.suggested_init
+        suggested_bounds = modifier.suggested_bounds
+
```
lukasheinrich: I think through the refactoring this has become superfluous, can just to this directly in the `par_map` dict in line 173-174

pyhf/__init__.py
kratsg: Yeah, it turns out that the class definitions in a separate subpackage or somehow caching or seeing the old value of `tensorlib`, even if it changes... If I use `tensorlib` outright and not `get_backend()`, the tests do not pass.

I think `pyhf.tensorlib` will always be right, but we're using relative imports for this so if we're keeping the import style consistent... then the cache'ing happens with relative imports.

pyhf/__init__.py
kratsg: Yes, I noticed this. I refactored it this way, because the refactoring is going to need to take a few steps. I want to switch to using singletons which might be more elegant here (http://python-3-patterns-idioms-test.readthedocs.io/en/latest/Singleton.html) as this removes the actual logic of knowing which modifier is shared and which is not, from `hfpdf` as it should not care -- this is modifier-specific behavior, no? I think something like the `Borg` version is the most elegant via something like

```
class Modifier(object):
  pass

class UnsharedModifier(Modifier):
  def __init__(self, name):
      pass

class SharedModifier(Modifier):
  _shared_state = {}
  def __init__(self, name):
    self.__dict__ = _shared_state.setdefault(name, {}).get(name)

class ShapeSys(UnsharedModifier):
  def __init__(self, name):
    super(self.__class__, self).__init__(name)
    self.name = name

class HistoSys(SharedModifier):
  def __init__(self, name):
    super(self.__class__, self).__init__(name)
    self.name = name
```

then the modifiers, themselves, encode whether or not they're shared -- and the `hfpdf` code doesn't care at all [and it shouldn't].

pyhf/modifiers/normfactor.py
kratsg: ok. will fix.

pyhf/modifiers/shapefactor.py
kratsg: ok. will fix.

pyhf/__init__.py
kratsg: Thinking again, the refactor should preserve existing functionality. I will file a new issue about refactoring later for singletons (if so desired). And I will fix this for now.

lukasheinrich: yes I think we should always keep master functional. thanks!

pyhf/modifiers/shapefactor.py
kratsg: done

pyhf/modifiers/normfactor.py
kratsg: done

pyhf/__init__.py
kratsg: done

pyhf/__init__.py
kratsg: done

kratsg:
@lukasheinrich all requested changes are now implemented. Please review :)

pyhf/__init__.py
```diff
@@ -173,79 +126,54 @@ def set_poi(self,name):
...
+        Example:
+            modifier = instance.add_or_get_modifier(channel, sample, modifier_def)
+        """
+        # check if modifier type is implemented
+        if modifier_def['type'] not in modifiers.registry:
+            raise RuntimeError('modifier type not implemented yet (processing {0:s})'.format(modifier_def['type']))
+
+        # get modifier class associated with modifier type
+        modifier_cls = modifiers.registry.get(modifier_def['type'])
+
```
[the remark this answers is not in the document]
kratsg: done. I check for a KeyError now, this is a good catch.

pyhf/__init__.py
```diff
@@ -173,79 +126,54 @@ def set_poi(self,name):
...
+        # check if modifier of the given name already exists
+        if modifier_cls.is_shared:
+            if modifier_def['name'] in self.par_map:
+                if modifier_cls.is_constrained:
+                    log.info('accepting existing {0:s} (type: {1:s})'.format(modifier_def['name'], modifier_cls.__name__))
+                else:
+                    log.info('accepting existing unconstrained factor')
+                return self.par_map[modifier_def['name']]['modifier']
```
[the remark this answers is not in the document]
kratsg: done.

kratsg:
> thanks @kratsg, just some more nit-picky things (hope you don't mind). and I think we should be ready to merge (will have final look then)

Done. Go ahead.

Issue #121, opened 2018-04-15
Back-end submodule structuring should be less convoluted?


kratsg:
# Description

The current way of importing backends is a little redundant:

```python
from pyhf.tensor.numpy_backend import numpy_backend
from pyhf.tensor.pytorch_backend import pytorch_backend
from pyhf.tensor.tensorflow_backend import tensorflow_backend
from pyhf.tensor.mxnet_backend import mxnet_backend
```

Something more pythonic would be

```python
from pyhf.tensor.backends import numpy_backend, pytorch_backend, tensorflow_backend, mxnet_backend
```

Or is there something technically difficult about doing it this way?

lukasheinrich:
I'd lke to keep all of these backend in different files, but probably with `__init__.py` and `__all__` we could get to the import semantic you lay out

matthewfeickert:
I think this can be closed now. @kratsg please reopen if I'm wrong.

PR #123, opened 2018-04-15
simplify import structure

lukasheinrich:
* moves tensor and optimzer import into their respective __init__.py


lukasheinrich:
this addresses #121

matthewfeickert:
@lukasheinrich Seeing this I now realize this is also going to affect `test_backend_consistency.py` and `test_benchmark.py`. Can you include them as well?

lukasheinrich:
added thanks for the suggestion

PR #126, opened 2018-04-16
global function-scope fixture to set backend back to default

### LINKED ISSUE #125: test_backend_consistency not resetting to default backend if test fails unexpectedly
# Description

A cascading error is observed when test_backend_consistency fails, which keeps the backend as tensorflow and causes all the other tests to erroneously fail.

    
screenshot 2018-04-15 20 45 50

Easy to reproduce, run `pytest` and see `test_pdf.py` fail. Run `pytest tests/test_pdf.py` and see that it's fine (as in screenshot).

kratsg:
# Description

Resolves #125.

- add a `default_backend` and `default_optimizer` to `pyhf/__init__.py` for resetting purposes
- update tests to add a global fixture to `conftest.py` that resets backend to default.

This is also related to #66.

### DIFF
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -5,7 +5,9 @@ import pyhf.tensor as tensor
 
 log = logging.getLogger(__name__)
 tensorlib = tensor.numpy_backend()
+default_backend = tensorlib
 optimizer = optimize.scipy_optimizer()
+default_optimizer = optimizer
 
 def set_backend(backend):
     """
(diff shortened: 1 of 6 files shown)


matthewfeickert:
@kratsg This also should be applied in `test_benchmark.py`. Can you add that as well?

kratsg:
> @kratsg This also should be applied in test_benchmark.py. Can you add that as well?

What do you mean? It's applied globally to all functions in tests.

matthewfeickert:
@kratsg Thanks, given your latest commit you understood what I meant.

matthewfeickert:
@kratsg Thanks for this nice solution. I'm going to merge this in so that the review of PR #120 can move forward.

PR #132, opened 2018-04-18
fix non existent modifier handling in pdf construction

lukasheinrich:
# Description

the behaviour for non-existing modifiers is wrong `.get(name)` should be `[name]`

### DIFF
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -147,7 +147,7 @@ class modelconfig(object):
             modifier_cls = modifiers.registry[modifier_def['type']]
         except KeyError:
             log.exception('Modifier type not implemented yet (processing {0:s}). Current modifier types: {1}'.format(modifier_def['type'], modifiers.registry.keys()))
-            raise
+            raise modifiers.InvalidModifier()
 
         # if modifier is shared, check if it already exists and use it
         if modifier_cls.is_shared and modifier_def['name'] in self.par_map:
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -144,10 +144,10 @@ class modelconfig(object):
         """
         # get modifier class associated with modifier type
         try:
-            modifier_cls = modifiers.registry.get(modifier_def['type'])
+            modifier_cls = modifiers.registry[modifier_def['type']]
         except KeyError:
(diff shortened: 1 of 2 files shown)


kratsg:
Good catch, d'oh on me.

pyhf/__init__.py
```diff
@@ -144,10 +144,10 @@ def add_or_get_modifier(self, channel, sample, modifier_def):
         """
         # get modifier class associated with modifier type
         try:
-            modifier_cls = modifiers.registry.get(modifier_def['type'])
+            modifier_cls = modifiers.registry[modifier_def['type']]
         except KeyError:
-            raise RuntimeError('Modifier type not implemented yet (processing {0:s}). Current modifier types: {1}'.format(modifier_def['type'], modifiers.registry.keys()))
-
+            log.exception('Modifier type not implemented yet (processing {0:s}). Current modifier types: {1}'.format(modifier_def['type'], modifiers.registry.keys()))
+            raise
```
kratsg: This should just be `raise modifiers.InvalidModifier` since we have that [error class available](https://github.com/diana-hep/pyhf/blob/master/pyhf/modifiers/__init__.py#L7-L8).

tests/test_pdf.py
```diff
@@ -75,6 +76,25 @@ def test_core_pdf_broadcasting():
...
+                    {
+                        'modifiers': [
+                            {'name': 'a_name', 'type': 'this_should_not_exist', 'data': None}
+                        ]
+                    },
+                ]
+            }
+        ]
+    }
+    with pytest.raises(KeyError):
```
kratsg: `pytest.raises(modifiers.InvalidModifier)`

lukasheinrich: good point, in general should we have a `pyhf.exceptions` module?

kratsg: yeah, I was thinking same.

pyhf/__init__.py
lukasheinrich: done

tests/test_pdf.py
lukasheinrich: done. is probably out of scope for this PR, but good issue

Issue #85, opened 2018-02-12
implement staterror


lukasheinrich:
should be mostly straight forward but requires additional bookkeeping which samples participate. Essentially one additional constraint term per bin

lukasheinrich:
this is the relevant section..

<img width="666" alt="screenshot" src="https://user-images.githubusercontent.com/2318083/36130515-79fe5b10-106d-11e8-8bf7-c411cc74edf6.png">

kratsg:
Need to essentially add a `Poisson()` term for each channel in the constraint PDF (around here: https://github.com/diana-hep/pyhf/blob/master/pyhf/__init__.py#L394). This would be useful to get the [MBJ analysis](https://arxiv.org/abs/1711.01901) up and running in HistFactory.

lukasheinrich:
for basic mechanics of doing 2D contour check cell [48]

https://github.com/diana-hep/pyhf/blob/master/examples/notebooks/multiBinPois.ipynb

lukasheinrich:
i.e. do add some more info

the general HiFa setup is:

* there are multiple channels
* each channel has multiple samples

each sample can declare whether or not it participates in stat. flucuations. e.g. in the JSON we could have 


```
name: 'some sample'
mods: [...]
enable_staterror: true / false
```

for each bin in the channel, you go through the samples add up all the mean rates of the samples that do participate. the mechanics of the auxiliary measurement is the essentially the same like the shapesys here

https://github.com/diana-hep/pyhf/blob/master/pyhf/__init__.py#L142

so for each channel one gets 1 Pois() term per bin (one can use broadcasting to compute this in one go using the tensorlib backends)

ill add more info later

PR #135, opened 2018-04-18
staterror support

lukasheinrich:
# Description

Starting to work on #85 .

ROOT implementation links: 

https://root.cern.ch/doc/v610/HistoToWorkspaceFactoryFast_8cxx_source.html#l02424

this looks more convoluted in ROOT than it is (at least for basic support), but keeping this as reference

Todo:

- [x] add sufficient information (just need nominal counts and uncertainties for each sample)
- [x] calculate overall uncertainty from individual sample uncertainties (~addition in quadrature)
- [x] implement constraint term (default seems gaussian, using this for now)

- [ ] Tests are passing


kratsg:
Don't forget to update `modifiers/__init__.py` towards the bottom to make sure it's auto-loaded/auto-registered (by default it won't be).

lukasheinrich:
the implementation seems ok.. just need to add some tests.. comparigon with ROOT (we really need #3, this overlay business is silly :-p )

screenshot

lukasheinrich:
@kratsg I added a comment and also a test to make sure the computation is right. Maybe you can take a look, the tests should pass and from my side, i'm happy w/ th epR

pyhf/modifiers/staterror.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def alphas(self, pars):
+        return pars  # nuisance parameters are also the means of the
+
+    def pdf(self, a, alpha):
+        tensorlib, _ = get_backend()
+        # this computes sum_i uncertainty_i for all samples
+        # (broadcastted for all bins in the channel)
+        # for each bin, the relative uncert is the width of a gaussian
+        # which is the constraint pdf; Prod_i Gaus(x = 1, mu = alpha_i, sigma = relunc_i)
```
kratsg: Shouldn't this be `x=a`?

lukasheinrich: done

kratsg:
Remove WIP and LGTM.

PR #137, opened 2018-04-18
Add shapefactor to accepted modifications

matthewfeickert:
Needed to complete #119 (`spec_2bin_2channel_coupledshapefactor`)

- [x] Tests are passing

### DIFF
--- a/validation/spec.json
+++ b/validation/spec.json
@@ -34,7 +34,7 @@
                 },
                 "type": {
                     "type": "string",
-                    "enum": ["shapesys","normfactor","histosys","normsys"]
+                    "enum": ["shapesys","normfactor","histosys","normsys","shapefactor"]
                 },
                 "data": { "$ref": "#/definitions/data"}
             },


matthewfeickert:
I'm assuming that there wasn't anything else needed, but I thought I'd have you check first instead of just throwing it into #119.

matthewfeickert:
@lukasheinrich, as @kratsg has approved this and was the one to write the schema I'm going to merge this in so we can get #119 done tonight. This is bad manners to not wait for your feedback, but my assumption is that you won't mind in this situation.

kratsg:
> This is bad manners/teamwork on my end to not wait for your feedback, but my assumption is that you won't mind in this situation.

Tbh, it was probably a small oversight on my part. I know @lukasheinrich and I have plans at least later to migrate the schema into separate definitions for each modifier, but this is ok for now.

matthewfeickert:
Sounds good. If this is going to get all cleaned up later that's great.

lukasheinrich:
no worries. looks fine

PR #139, opened 2018-04-19
[WIP] Move modifier factor calculables into modifiers

### LINKED ISSUE #127: Refactor modifier-specific code from hfpy/__init__.py to modifier code
# Description

Refactor the following functions under `hfpdf` class:

- `_normsysfactor`
- `_histosysdelta`
- `_multiplicative_factors`

into either of the following signatures on the modifier classes

```
def apply(self, channel, sample, pars)
# or
def __call__(self, channel, sample, pars)
```

#120 started the refactor, it will be merged in, and this issue to keep track of the next refactor of modifier code to complete refactor.

kratsg:
# Description

This will resolve #127.  The goal is to try and get the modifier code to calculate the relevant (multiplicative) factor that goes into the PDF calculation. This is very tricky as one needs to split up information specific to the modifier from information specific to the spec.

- [ ] Tests are passing

### DIFF
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -58,33 +58,6 @@ def set_backend(backend):
     else:
         optimizer = optimize.scipy_optimizer()
 
-def _hfinterp_code0(at_minus_one, at_zero, at_plus_one, alphas):
-    at_minus_one = tensorlib.astensor(at_minus_one)
-    at_zero = tensorlib.astensor(at_zero)
-    at_plus_one = tensorlib.astensor(at_plus_one)
-
-    alphas = tensorlib.astensor(alphas)
-
-    iplus_izero  = at_plus_one - at_zero
-    izero_iminus = at_zero - at_minus_one
-
-    mask = tensorlib.outer(alphas < 0, tensorlib.ones(iplus_izero.shape))
-    return tensorlib.where(mask, tensorlib.outer(alphas, izero_iminus), tensorlib.outer(alphas, iplus_izero))
-
-def _hfinterp_code1(at_minus_one, at_zero, at_plus_one, alphas):
-    at_minus_one = tensorlib.astensor(at_minus_one)
-    at_zero = tensorlib.astensor(at_zero)
-    at_plus_one = tensorlib.astensor(at_plus_one)
-    alphas = tensorlib.astensor(alphas)
-
-    base_positive = tensorlib.divide(at_plus_one,  at_zero)
-    base_negative = tensorlib.divide(at_minus_one, at_zero)
-    expo_positive = tensorlib.outer(alphas, tensorlib.ones(base_positive.shape))
-    mask = tensorlib.outer(alphas > 0, tensorlib.ones(base_positive.shape))
-    bases = tensorlib.where(mask,base_positive,base_negative)
-    exponents = tensorlib.where(mask, expo_positive,-expo_positive)
-    return tensorlib.power(bases, exponents)
-
 class modelconfig(object):
     @classmethod
     def from_spec(cls,spec,poiname = 'mu'):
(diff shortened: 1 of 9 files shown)


kratsg:
I think for now, this is all I want to refactor up to. The code is a bit cleaner and a major piece of the logic is back into the modifier where it most likely belongs.

pyhf/__init__.py
```diff
@@ -180,59 +153,44 @@ def __init__(self, spec, **config_kwargs):
...
+            modifier, modpars = self.config.modifier(m['name']), pars[self.config.par_slice(m['name'])]
+            results[m['type']].append(modifier.apply(channel, sample, modpars))
+
+
+        # start building the entire set of factors
+        factors = []
+        factors += results['shapesys'] + results['normfactor'] + results['shapefactor']
+        # normsysfactor(nom_sys_alphas)   = 1 + sum(interp(1, anchors[i][0],
+        # anchors[i][0], val=alpha)  for i in range(nom_sys_alphas))
+        factors += [tensorlib.product(results['normsys'])]
```
lukasheinrich: I think this could be 


```
factors += [results['normsys']]
```

instead of

```
factors += [tensorlib.product(results['normsys'])]
```

this would treat each normfactor separately than all at once, and we might save some computations since `tensorlib.product` is called at the end

lukasheinrich:
Hi @kratsg thanks for the PR. I think I'm fine with the overall refactor (there is some issue if all modifiers can fit the same bill, but for now this works and we can tackle other cases later)

Only thing is, I think the interpolation codes I think should stay explicitly named (perhaps even in an `interpolate.py` module?) since they relate directly to one of several interpolation strategies in https://cds.cern.ch/record/1456844/files/CERN-OPEN-2012-016.pdf and a more feature-complete implementation of HF would allow a user-option on which interpolation to use

kratsg:
> Only thing is, I think the interpolation codes should stay explicitly named instead of _apply (perhaps even in an interpolate.py module?) since they relate directly to one of several interpolation strategies in https://cds.cern.ch/record/1456844/files/CERN-OPEN-2012-016.pdf and a more feature-complete implementation of HF would allow a user-option on which interpolation to use

Ok, these are static. I can definitely make an interpolation.py module here for this.

pyhf/__init__.py
```diff
@@ -180,63 +153,46 @@ def __init__(self, spec, **config_kwargs):
...
-        factors += self._multiplicative_factors(channel, sample, pars)
-
-        # print('sample factors {}'.format(factors))
-
+        factors += results['shapesys']
+        factors += results['normfactor']
+        factors += results['shapefactor']
+        factors += results['staterror']
+        factors += results['normsys']
+
```
lukasheinrich: there are two times of factors: vectors and scalars that get broadcasted, perhaps it's useful in the code to have comments which are which

`shapesys`, `shapefactor`, staterror` are vectors

`normfactor` , `normsys` ae scalars

kratsg: Can you explain this more? I don't get the broadcasting, and I don't see how `[a, [b,c], d]` works correctly? Is broadcasting sort of trying to flatten things out or do something else more magically?

kratsg:
Added an interpolation module in the meantime and separated out the tests.

pyhf/__init__.py
lukasheinrich: I think we lost the documentation on this when we introduce the tensor backends. The idea is that we compute all bin-values at once.. each bin is a product of various factors, but sum are per-channel the other per-channel

```
b1 = shapesys_1   |      shapef_1   |
b2 = shapesys_2   |      shapef_2   |
...             normfac1    ..     normfac2
                (broad)            (broad)
bn = shapesys_n   |      shapef_1   |
```

this can be achieved by `numpy`'s `broadcast_arrays` and `np.product`

```
np.product(np.broadcast_arrays([2],[3,4,5],[6],[7,8,9]),axis=0)
```

is `[ 2*3*6*7, 2*4*6*8, 2*5*6*9]`

PR #142, opened 2018-04-21
[WIP] Add multibjets multibin as part of ROOT-based validation

kratsg:
# Description

See #3 for related issue. This adds some changes to `readxml.py` so that the (giant) workspace for MBJ is imported correctly.

The goal is to have a test against an existing (published) CERN analysis that we can use as part of validation. In order to make that happen

- [x] get the workspace imported (and data loaded up)
- [x] validate the generated spec against 'spec.json'
- [x] get a working test for it
- [x] check the results against HiFa to ensure that `pyhf` returns same results
- [x] update tests to encode the correct answer from HiFa

- [x] Tests are passing


kratsg:
Whew, this incorporates an actual physics use case.

matthewfeickert:
@kratsg Is `scan-mbj.pdf` used as a figure in the notebook somewhere? If not, do we need to add it?

kratsg:
> @kratsg Is scan-mbj.pdf used as a figure in the notebook somewhere? If not, do we need to add it?

Nope, I'll remove it.

> Beyond that quibble, LGTM, once the Python 3.6 test_import.py fail is fixed.

@lukasheinrich this is a problem. The python 2 and python 3 evaluation are returning significantly different numbers...

Issue #95, opened 2018-03-13
make test-statistics tensor-aware


lukasheinrich:
# Description
 
we want to be able to do e.g. `tf.Session().run(qmu)`

matthewfeickert:
As part of this, since in `runOnePoint()` `sqrtqmuA_v` will now be a tensor as a result of 

```python
sqrtqmuA_v = tensorlib.sqrt(qmuA_v)
```
then

```python
    sigma = muTest / sqrtqmuA_v if sqrtqmuA_v > 0 else None
```
will have to be tensorized, as tensors can not be used as `bools` in TensorFlow
```
TypeError: Using a `tf.Tensor` as a Python `bool` is not allowed.
```
As a result of this, I'm going to add tensorlib functions `greater` and `less` for the backends so that they can be safely used with `tensorlib.where`.

matthewfeickert:
Relevant to PR #149, once I figure out how to get PyTorch v0.4 properly installed through Conda, is the ["Merging Tensor and Variable classes" section of the release notes](https://github.com/pytorch/pytorch/releases/tag/v0.4.0).

matthewfeickert:
As a result of PR #149, can now do:
```python
>>> import pyhf
>>> import pyhf.simplemodels
>>> import tensorflow as tf
>>> pyhf.set_backend(pyhf.tensor.tensorflow_backend(session=tf.Session()))
>>> pdf = pyhf.simplemodels.hepdata_like([10.], [50.], [7.])
>>> data = [50.] + pdf.config.auxdata
>>> q_mu = pyhf.qmu(1.0, data, pdf, pdf.config.suggested_init(), pdf.config.suggested_bounds())
>>> q_mu
<tf.Tensor 'add_9:0' shape=(1,) dtype=float32>
>>> pyhf.tensorlib.session.run(q_mu)
array([1.0174799], dtype=float32)
```

PR #148, opened 2018-04-23
Add tensor clipping

### LINKED ISSUE #147: add tensor clipping to shared API of tensorlib
# Description

in order to get #95 it would be useful to have support in the tensor libraries for tensor clipping in order to nicely implement 

```
    if -1e-6 < qmu < 0:
        log.warning('WARNING: qmu negative: %s', qmu)
        return 0.0
```

in a more tensor-native way. all backends support this

http://pytorch.org/docs/master/torch.html?#torch.clamp
https://www.tensorflow.org/api_docs/python/tf/clip_by_value
https://docs.scipy.org/doc/numpy/reference/generated/numpy.clip.html
https://mxnet.incubator.apache.org/api/python/ndarray/ndarray.html#mxnet.ndarray.clip

so this should be an easy one. I would suggest `clip` as the name

matthewfeickert:
Resolves #147 

- [x] Tests are passing

### DIFF
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -314,9 +314,7 @@ def qmu(mu, data, pdf, init_pars, par_bounds):
     qmu = tensorlib.tolist(loglambdav(mubhathat, data, pdf) - loglambdav(muhatbhat, data, pdf))[0]
     if muhatbhat[pdf.config.poi_index] > mu:
         return 0.0
-    if -1e-6 < qmu < 0:
-        log.warning('WARNING: qmu negative: %s', qmu)
-        return 0.0
+    qmu = tensorlib.tolist(tensorlib.clip([qmu], 0, max=None))[0]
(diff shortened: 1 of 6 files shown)


lukasheinrich:
Hi @matthewfeickert 

thanks! this looks goot to me. Do you want to tackle #95 completely in this PR, or should we make this a separate one? (Either way is fine to me, if that format we'd need

* re-write if statement in `qmu(..)` via `tensorlib.where`
* make clipping mandatory not only on exception
* `math.sqrt` -> `tensorlib.sqrt`
* tensorize `pvals_from_teststat`

matthewfeickert:
@lukasheinrich I'd say let's just do #95 in a seperate PR, but I can also start that tonight.

@lukasheinrich and @kratsg, if you have no comments or suggested edits can you please squash and merge this one in?

PR #149, opened 2018-04-24
Add tensor aware test stats

### LINKED ISSUE #95: make test-statistics tensor-aware

matthewfeickert:
Resolves #95 

Outline:
- [x] re-write if statement in `qmu(..)` via `tensorlib.where`
- [x] make clipping mandatory not only on exception (needed for TensorFlow as is a non-eager (lazy) evaluation)
- [x] `math.sqrt` -> `tensorlib.sqrt`
- [x] ~add `tensorlib` function `greater`~
- [x] ~add `tensorlib` function `less`~
- [x] tensorize `pvals_from_teststat`
- [x] tensorize `runOnePoint`
- [x] Check that fully tensorized, e.g., `tf.Session().run(qmu)`

- [x] Tests are passing


matthewfeickert:
The start of the PR will fail for the [reasons described in Issue #149](https://github.com/diana-hep/pyhf/issues/95#issuecomment-383772983), but this is know and just to get the PR going.

pyhf/__init__.py
```diff
@@ -336,17 +361,17 @@ def runOnePoint(muTest, data, pdf, init_pars, par_bounds):
...
     try:
-        sqrtqmuA_v = math.sqrt(qmuA_v)
+        sqrtqmuA_v = tensorlib.sqrt(qmuA_v)
     except ValueError:
         log.warning('WARNING: qmuA_v negative: %s', qmuA_v)
         qmuA_v = tensorlib.tolist(tensorlib.clip(qmuA_v, 0, max=None))
-        sqrtqmuA_v = math.sqrt(qmuA_v)
+        sqrtqmuA_v = tensorlib.sqrt(qmuA_v)
 
     sigma = muTest / sqrtqmuA_v if sqrtqmuA_v > 0 else None
```
lukasheinrich: actually I think we can just remove this, we don't actually use this anywhere, I thin it's just a relic from some tests I did

matthewfeickert: Ah, okay. I'll remove `sigma` everywhere in `pvals_from_teststat` then

lukasheinrich: I think it's just in the line where it's defined and the return line

matthewfeickert: Yeah, that's right. The tests will of course still fail after 9ea1264 as I haven't tensorized `pvals_from_teststat` yet.

lukasheinrich:
I guess we need a `cdf_normal` function in tensorlib which wraps

* https://www.tensorflow.org/api_docs/python/tf/distributions/Normal
* https://docs.scipy.org/doc/scipy-0.16.1/reference/generated/scipy.stats.norm.html
* http://pytorch.org/docs/master/distributions.html#torch.distributions.normal.Normal.cdf
* mxnet??

matthewfeickert:
This is almost done. I'm just trying to review and understand how the changes to `runOnePoint` have affected how `invert_interval()` in the notebooks is constructing the shape of `yvals`. The interpolation scheme requires (obviously) the same shape of the lists representing `x` points and `y` points, but `yvals` has a shape of `(testmus.shape, 1)` where we need it to have shape `testmus.shape`. This can be easily caught and fixed with a `try/excpet` block that asserts and reshapes, but this should actually be understood and fixed properly.

Additionally, [PyTorch v0.4 is out today](https://twitter.com/PyTorch/status/988883720114507776) :tada:  and once it is in Conda this will include the fix for the hack in e9278fb (after a small rewrite).

pyhf/__init__.py
```diff
@@ -308,52 +309,76 @@ def loglambdav(pars, data, pdf):
...
+    """
+    mubhathat = tensorlib.tolist(optimizer.constrained_bestfit(
+        loglambdav, mu, data, pdf, init_pars, par_bounds))
+    muhatbhat = tensorlib.tolist(optimizer.unconstrained_bestfit(
+        loglambdav, data, pdf, init_pars, par_bounds))
+    qmu = tensorlib.tolist(
+        loglambdav(mubhathat, data, pdf) - loglambdav(muhatbhat, data, pdf))[0]
+    qmu = tensorlib.tolist(
+        tensorlib.where(muhatbhat[pdf.config.poi_index] > mu, [0], qmu))
     return qmu
```
lukasheinrich: Hi @matthewfeickert, why do the above lines have `.tolist`? could we just keep them as tensors?

matthewfeickert: Indeed. That is just the result of me being sloppy while working on this. I've removed them now, thanks for the catch.

pyhf/__init__.py
```diff
@@ -1,8 +1,9 @@
-import logging
 import pyhf.optimize as optimize
 import pyhf.tensor as tensor
 from . import exceptions
 
+import logging
+
```
lukasheinrich: can we move this up again :) I think it's good to have the stdlib imports on top

pyhf/__init__.py
```diff
@@ -308,52 +309,76 @@ def loglambdav(pars, data, pdf):
...
 
-import math
+
 def runOnePoint(muTest, data, pdf, init_pars, par_bounds):
-    asimov_mu = 0.0
-    asimov_data = tensorlib.tolist(generate_asimov_data(asimov_mu, data,
-                                   pdf, init_pars, par_bounds))
+    asimov_mu = 0.
+    asimov_data = tensorlib.tolist(
+        generate_asimov_data(asimov_mu, data, pdf, init_pars, par_bounds))
```
lukasheinrich: probably also don't need `.tolist` here

pyhf/tensor/mxnet_backend.py
```diff
@@ -339,3 +350,31 @@ def log_prob(value, loc, scale):
...
+
+        Returns:
+            MXNet NDArray: The CDF
+        """
+        # This is currently using SciPy stats until a better way can be found to
+        # do this in MXNet
+        x = self.astensor(x).asnumpy().tolist()[0]
+        mu = self.astensor(mu).asnumpy().tolist()[0]
+        sigma = self.astensor(sigma).asnumpy().tolist()[0]
+        return self.astensor([norm.cdf(x, loc=mu, scale=sigma)])
```
lukasheinrich: `norm` should be able to deal with numpy arrays, so probably don't need `tolist`

pyhf/__init__.py
matthewfeickert: If we get rid of the `.tolist` here then we need to change other things. `asimov_data` gets passed to `qmu()` which then passes it to (for our supported optimizers) `optimizer.constrained_bestfit()`. This fails for PyTorch, as it requires a list for data. I would suggest removing the `.tolist` here and simply adding 

```python
data = self.tensorlib.tolist(data)
```
to `unconstrained_bestfit` and `constrained_bestfit` in `opt_pytorch`. Thoughts?

pyhf/__init__.py
matthewfeickert: Absolutely! I will try to make a mental note of this and always put those imports first. :+1:

pyhf/__init__.py
lukasheinrich: thanks for the explanation. where exactly does that factor in? I only see it being passed to the objective function which is `pyhf.loglambdav` / `pdf.logpdf`  which I would have thought are fine with tensors

matthewfeickert: Very good point. Indeed, it is fine with tensors (NumPy and TensorFlow handle it just fine). The problem arises in that for the PyTorch backend, in `generate_asimov_data()`, `pdf.expected_data()` is returning type `torch.autograd.variable.Variable` instead of `torch.FloatTensor`. 

If I stick a cast to `torch.FloatTensor` 

```python
data = torch.FloatTensor(self.tensorlib.tolist(data))
```

in just inside `constrained_bestfit` and `unconstrained_bestfit` this goes away, so this indicates that for the PyTorch backend `astensor()` is not giving the desired result.

```python
    def astensor(self, tensor_in):
        if isinstance(tensor_in, torch.autograd.Variable):
            v = tensor_in
        else:
            try:
                v = torch.autograd.Variable(torch.Tensor(tensor_in))
            except TypeError:
                try:
                    v = torch.autograd.Variable(tensor_in)
                except TypeError:
                    v = torch.autograd.Variable(torch.Tensor([tensor_in]))
                except RuntimeError:
                    v = torch.autograd.Variable(torch.Tensor([tensor_in]))
        return v.type(torch.FloatTensor)
```

matthewfeickert: To make this a bit more easy to follow, I've pushed 305e561 (knowing this will fail the CI) so that the code that I'm talking about is at least in the PR.

lukasheinrich: hi @matthewfeickert I tested this quick on `master` and see this. So I wonder how this breaks. Does that snippet work in your branch?

screenshot

matthewfeickert: It does.
```python
>>> import pyhf
>>> import pyhf.simplemodels
>>> pdf = pyhf.simplemodels.hepdata_like([10.], [50.], [7.])
>>> data = [50.] + pdf.config.auxdata
>>> asimov = pyhf.generate_asimov_data(0., data, pdf, pdf.config.suggested_init(), pdf.config.suggested_bounds())
>>> pdf.logpdf(pdf.config.suggested_init(), asimov)
Variable containing:
-6.7963
[torch.FloatTensor of size 1]
```

lukasheinrich: just for posterity's sake: solution was to add `retain_graph` in backward() calls of the fit

lukasheinrich:
hi @matthewfeickert, what are the open points for the PR?

matthewfeickert:
@lukasheinrich Just to harmonize the notebooks, as the output shape is different, so that all the tests pass. This shouldn't take much time but I've been trying to move on analysis so I might not get to it until tonight or tomorrow.

pyhf/tensor/mxnet_backend.py
matthewfeickert: Fixed in 0b76619

pyhf/__init__.py
```diff
@@ -307,53 +308,75 @@ def generate_asimov_data(asimov_mu, data, pdf, init_pars, par_bounds):
...
-    asimov_data = tensorlib.tolist(generate_asimov_data(asimov_mu, data,
-                                   pdf, init_pars, par_bounds))
+    asimov_mu = 0.
+    asimov_data = generate_asimov_data(asimov_mu, data, pdf, init_pars, par_bounds)
 
-    qmu_v  = qmu(muTest, data, pdf, init_pars, par_bounds)
+    qmu_v = qmu(muTest, data, pdf, init_pars, par_bounds)
     qmuA_v = qmu(muTest, asimov_data, pdf, init_pars, par_bounds)
 
     try:
```
matthewfeickert: @lukasheinrich If I remember right, you had suggested that for the blocks below we always `clip` and not test first. Is this correct, or did I misunderstand you?

matthewfeickert:
Coverage has gone down. Is this something we should look at a bit more, or it is okay? I think this is mostly a result of me adding `try`/`except` blocks.

pyhf/__init__.py
lukasheinrich: Hi, yes the try except only works for the eager backends anyways (e.g. not TF), so I think we can just make this mandatory, and separately thinkg about how to deal with problematic situations

pyhf/__init__.py
```diff
@@ -307,53 +308,68 @@ def generate_asimov_data(asimov_mu, data, pdf, init_pars, par_bounds):
...
     CLsb, CLb, oneOverCLs = pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v)
 
     oneOverCLs_exp = []
     for nsigma in [-2, -1, 0, 1, 2]:
         sqrtqmu_v_sigma = sqrtqmuA_v - nsigma
-        oneOverCLs_exp.append(pvals_from_teststat(sqrtqmu_v_sigma, sqrtqmuA_v)[-1])
-    return qmu_v, qmuA_v, sigma, CLsb, CLb, oneOverCLs, oneOverCLs_exp
+        oneOverCLs_exp.append(
+            pvals_from_teststat(sqrtqmu_v_sigma, sqrtqmuA_v)[-1])
+    return qmu_v, qmuA_v, CLsb, CLb, oneOverCLs, oneOverCLs_exp
```
lukasheinrich: thinking whether we should wrap `tensorlib.astensor(oneOverCLs_exp)`, this will make the expected CLs values a (5,1) tensor, which is a bit nicer to handle than 5 (1,) tensors

matthewfeickert: Sounds good to me.

pyhf/tensor/mxnet_backend.py
```diff
@@ -339,3 +350,31 @@ def log_prob(value, loc, scale):
...
+
+        Returns:
+            MXNet NDArray: The CDF
+        """
+        # This is currently using SciPy stats until a better way can be found to
+        # do this in MXNet
+        x = self.astensor(x).asnumpy()
+        mu = self.astensor(mu).asnumpy()
+        sigma = self.astensor(sigma).asnumpy()
+        return self.astensor(norm.cdf(x, loc=mu, scale=sigma))
```
lukasheinrich: I'm thinking whether or not this warrants a `log.warning`. Since this will definitely break autodifferentiability. We don't use this backend properly anyways, so adding this should poolute any logs.

pyhf/__init__.py
matthewfeickert: Fixed in commit 0691e21

pyhf/tensor/mxnet_backend.py
matthewfeickert: Removed in 745b87a

lukasheinrich: oh I didn't mean in `astensor`, but for calls of `normal_cdf`

matthewfeickert: Sorry, it was late and I misunderstood what you mean. You're saying if we should have the log warning on the comment I had to indicate that this is not fully MXNet, correct?

matthewfeickert: Okay, should be actually fixed in 8b58fde now.

pyhf/tensor/mxnet_backend.py
```diff
@@ -82,7 +83,16 @@ def astensor(self, tensor_in):
...
+        try:
+            tensor = nd.array(tensor_in)
+        except ValueError:
+            try:
+                # Guard against a float being passed
+                tensor = nd.array([tensor_in])
+            except Exception as err:
+                log.warning('ERROR: astensor fails with: %s', err)
+                raise
+        return tensor
```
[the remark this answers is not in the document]
matthewfeickert: Removed in commit 3274724

pyhf/tensor/pytorch_backend.py
```diff
@@ -44,12 +44,19 @@ def astensor(self, tensor_in):
...
-                v = torch.autograd.Variable(tensor_in)
+                try:
+                    v = torch.autograd.Variable(tensor_in)
+                # Guard against not passing in something that is not a list
+                # or a tensor and causing a TypeError or RuntimeError
+                except TypeError:
+                    v = torch.autograd.Variable(torch.Tensor([tensor_in]))
+                except RuntimeError:
+                    v = torch.autograd.Variable(torch.Tensor([tensor_in]))
         return v.type(torch.FloatTensor)
```
lukasheinrich: sorry for the piecemeal requests, can we make this

```
except (TypeError,RuntimeError):
    v = torch.autograd.Variable(torch.Tensor([tensor_in]))
```

lukasheinrich:
Hi @matthewfeickert sorry I just stumbled on one last requests. After that I think we can merge

pyhf/tensor/pytorch_backend.py
matthewfeickert: No problem! I truly appreciate the fact that you do a good review and this save us time overall. :) Fixed in commit 3f64c92.

lukasheinrich:
ok this LGTM. merged. Thanks agains for the PR! 🎉

PR #153, opened 2018-05-02
fix histosys suggested init

lukasheinrich:
# Description

The suggested "nominal" value for HistoSys is 0

- [x] Tests are passing

### DIFF
--- a/pyhf/modifiers/histosys.py
+++ b/pyhf/modifiers/histosys.py
@@ -9,7 +9,7 @@ from ..interpolate import interpolator
 class histosys(object):
     def __init__(self, nom_data, modifier_data):
         self.n_parameters = 1
-        self.suggested_init = [1.0]
+        self.suggested_init = [0.0]
         self.suggested_bounds = [[-5, 5]]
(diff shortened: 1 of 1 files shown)


lukasheinrich:
Sorry, I should have added more details. The "nominal" values for systematics are different depending on the systematic. I couldn't find any reference in the original paper, so it's probably even more necessary to explain.

For HistoSys and OverallSys (called NormSys in pyhf) the actual effect (say `eta`) of the modifier is computed from the nuisance parameters

```
Pois(eta(alpha)..) ... Constraint(auxdata | alpha)
```

and by convention `eta(alpha = 0) = 1` e.g. what we would consider the nominal setting, where the modifier does not affect the main pdf.  , (and `eta(alpha = +1)` is the "up variation" and `eta(alpha=-1)` is the "down variation")

For StatError and ShapeSys, the nuisance parameter *is the same thing as the effect*

```
Pois(alpha..) ... Constraint(auxdata | alpha)
```

so the nominal value is `alpha=1`

PR #157, opened 2018-05-06
Start setting up documentation more

kratsg:
# Description

This moves our notebooks inside documentation and changes the theme to look a little bit more modern. 

Originally started with: screenshot 2018-05-06 09 57 41

Now looks more like: 
screenshot 2018-05-06 09 58 01

Can pick any of the bootswatch 3 themes from here (https://bootswatch.com/3/) -- do we want a lighter theme or a darker theme? I'm currently using `flatly` which seems fine.

Also added additional functionality to incorporate notebooks straight into our documentation: 
screenshot 2018-05-06 09 59 34

and updated the copyright to include my name as one of the authors as well (that ok?).

- [X] Tests are passing
- [X] "WIP" removed from the title of the pull request


matthewfeickert:
I'm more just interested in having something that allows for [Google Style docstrings](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google). To my mind they are much more readable while looking at the actual code, which is a plus.

As an example, PyTorch uses them with the [Read The Docs Sphinx theme](https://github.com/rtfd/sphinx_rtd_theme). So their docstrings (like this one for [`torch.distributions.bernoulli`](http://pytorch.org/docs/master/_modules/torch/distributions/bernoulli.html#Bernoulli)) end up looking [like this when rendered](http://pytorch.org/docs/master/distributions.html#bernoulli).

I truly don't know how theme dependent this is, so if this can be supported in bootswatch 3 themes then that's fine be me too.

kratsg:
> I'm more just interested in having something that allows for Google Style docstrings. To my mind they are much more readable while looking at the actual code, which is a plus. As an example, PyTorch uses them with the Read The Docs Sphinx theme. So their docstrings (like this one for torch.distributions.bernoulli) end up looking like this when rendered.
> 
> I truly don't know how theme dependent this is, so if this can be supported in bootswatch 3 themes then that's fine be me too.

This isn't theme-dependent. That's just a functionality baked into Sphinx itself.

kratsg:
Using `sphinx_rtd_theme` now, and these are example of what it looks like:

    
screenshot 2018-05-06 12 24 12
screenshot 2018-05-06 12 21 38

lukasheinrich:
Hi, thanks, we need to work on the docs. I do like bootswatch theme, and it should support any docstring style we want, no?

kratsg:
> Hi, thanks, we need to work on the docs. I do like bootswatch theme, and it should support any docstring style we want, no?

It should but I disliked the formatting of the actual documentation itself.. Let's just stick with the tried/true RTD theme right now, merge it in, and you can look at the other ones locally? If we need to switch the theme later, it's trivial.

kratsg:
@lukasheinrich  here's a comparison -- you can see why I have problems with the bootswatch one -- particularly with font sizes and the separate/distinction between classes/methods/attributes..

screenshot 2018-05-06 17 10 13
screenshot 2018-05-06 17 10 34

kratsg:
The other problem is that with bootswatch, we add two extra clicks to go from one modifier to another as the other modifiers are grouped fine in Sphinx, but are not revealed in bootswatch.

lukasheinrich:
ok. LGTM, let's hash out the details later.

Issue #143, opened 2018-04-21
How slow is too slow for evaluating CLs?


kratsg:
# Description

Here I describe the different steps as part of #3 for the Multi-b-jet analysis which I'm setting up with 34 channels. It seems to take forever to evaluate the CLs at a single mu_SIG value, so I link a few of the profiler.log visualizations generated based on call stacks.

This is the profiler output for 34 channels when calling:

- [`pdf.pdf(init_pars, data)`](https://cdn.rawgit.com/kratsg/a555d77519acb7b53b0bcc357058e444/raw/0e47fc4dfe2263491b44c0b5132c13cf1f66fd03/profiler_34channels.html)
- [`pyhf.optimizer.unconstrained_bestfit(pyhf.loglambdav,data,pdf,init_pars,par_bounds)`](https://cdn.rawgit.com/kratsg/3a7006fe2c0fdee9afe3686f42a52f1b/raw/31056cb339da391f75caf214b555f41a11876426/unconstrained_bestfit.html)

It seems that some optimization could be done to improve the times here. It's certainly not *that* slow, but there is probably some caching that could be done with some of the function calls.

--------

To make these, I have a python file with something like this

```python
import cProfile
pr = cProfile.Profile()
pr.enable()
#pdf.pdf(init_pars, data)
pyhf.optimizer.unconstrained_bestfit(pyhf.loglambdav,data,pdf,init_pars,par_bounds)
pr.disable()
pr.dump_stats('profiler.log')
```

with this `profiler.log`, I run `snakeviz profiler.log` (`pip install snakeviz`) and then save the HTML of that page. Follow the instructions [here](https://gist.github.com/jiffyclub/6b5e0f0f05ab487ff607) to convert that into a usable HTML that can be saved in a [gist](https://gist.github.com) and then a URL with the right content-type headers can be generated with https://rawgit.com/.

kratsg:
For the evaluation of a single CLs point `mu_SIG = 0.0`, I tried to see how slow it was as we keep adding channels. This evaluates

```python
CLsOnePoint(0.0, data, pdf, init_pars, par_bounds)
```

and the results so far are

```
[1 channel] 1 loop, best of 3: 148 ms per loop
[2 channel] 1 loop, best of 3: 521 ms per loop
[3 channel] 1 loop, best of 3: 1.32 s per loop
[4 channel] 1 loop, best of 3: 3.3 s per loop
[5 channel] 1 loop, best of 3: 5.45 s per loop
[...]

PR #159, opened 2018-05-06
Fix simple_broadcast()

### LINKED ISSUE #158: simple_broadcast is broken in numpy?
# Description

simple_broadcast does not work the same in different backends

# Expected Behavior

Arrays got broadcasted correctly in numpy

# Actual Behavior

Arrays did not get broadcast correctly in numpy

# Steps to Reproduce

```python
>>> import pyhf
>>> pyhf.set_backend(pyhf.tensor.mxnet_backend())
>>> pyhf.tensorlib.simple_broadcast(pyhf.tensorlib.astensor([1]), pyhf.tensorlib.astensor([2, 2]), pyhf.tensorlib.astensor([3, 3, 3]))

[[1. 1. 1.]
 [2. 2. 2.]
 [3. 3. 3.]]
<NDArray 3x3 @cpu(0)>
```

```python
>>> import pyhf
>>> pyhf.set_backend(pyhf.tensor.numpy_backend())
>>> pyhf.tensorlib.simple_broadcast(pyhf.tensorlib.astensor([1]), pyhf.tensorlib.astensor([2, 2]), pyhf.tensorlib.astensor([3, 3, 3]))
>>> pyhf.tensorlib.simple_broadcast(pyhf.tensorlib.astensor([1]), pyhf.tensorlib.astensor([2, 2]), pyhf.tensorlib.astensor([3, 3, 3]))
  File "/Users/kratsg/pyhf/pyhf/tensor/numpy_backend.py", line 90, in simple_broadcast
    return np.broadcast_arrays(*args)
  File "/Users/kratsg/.virtualenvs/pyhf/lib/python2.7/site-packages/numpy/lib/stride_tricks.py", line 249, in broadcast_arrays
    shape = _broadcast_shape(*args)
  File "/Users/kratsg/.virtualenvs/pyhf/lib/python2.7/site-packages/numpy/lib/stride_tricks.py", line 184, in _broadcast_shape
    b = np.broadcast(*args[:32])
ValueError: shape mismatch: objects cannot be broadcast to a single shape
```

# Checklist

matthewfeickert:
Resolves #158

- [x] Fix `simple_broadcast()` for PyTorch backend
- [x] Fix `simple_broadcast()` for TensorFlow backend
- [x] Fix `simple_broadcast()` for MXNet backend
- [x] Add test for `tensorlib.simple_broadcast()`
- [x] Add tests that fail when number of non-lenght-1 arrays (that don't match `max_dim`) is greater than 1
- [x] ~Add decorator for tensorization of inputs~ [Moved to Issue #163]
- [x] ~Apply decorator in all backends~ [Moved to Issue #163]
 
- [x] Tests are passing

### DIFF
--- a/pyhf/tensor/mxnet_backend.py
+++ b/pyhf/tensor/mxnet_backend.py
@@ -305,14 +305,18 @@ class mxnet_backend(object):
         Returns:
             MXNet NDArray: The sequence broadcast together.
         """
+        args = [self.astensor(arg) for arg in args]
         max_dim = max(map(len, args))
-        broadcast = []
-        for arg in args:
-            if len(arg) < max_dim:
-                broadcast.append(nd.broadcast_axis(
-                    arg[0], axis=len(arg.shape) - 1, size=max_dim))
-            else:
-                broadcast.append(arg)
+        try:
+            assert len([arg for arg in args if 1 < len(arg) < max_dim]) == 0
+        except AssertionError as error:
+            log.error(
+                'ERROR: The arguments must be of compatible size: 1 or %i', max_dim)
+            raise error
+
+        broadcast = [arg if len(arg) > 1
+                     else nd.broadcast_axis(arg[0], axis=len(arg.shape) - 1, size=max_dim)
+                     for arg in args]
         return nd.stack(*broadcast)
 
     def poisson(self, n, lam):
--- a/pyhf/tensor/mxnet_backend.py
+++ b/pyhf/tensor/mxnet_backend.py
@@ -79,7 +79,7 @@ class mxnet_backend(object):
 
     def astensor(self, tensor_in):
         """
-        Convert a tensor to an MXNet NDArray.
+        Convert to a MXNet NDArray.
 
         Args:
             tensor_in (Number or Tensor): Tensor object
@@ -291,12 +291,12 @@ class mxnet_backend(object):
             >>> pyhf.set_backend(pyhf.tensor.mxnet_backend())
             >>> pyhf.tensorlib.simple_broadcast(
             ...   pyhf.tensorlib.astensor([1]),
-            ...   pyhf.tensorlib.astensor([2, 2]),
-            ...   pyhf.tensorlib.astensor([3, 3, 3]))
+            ...   pyhf.tensorlib.astensor([2, 3, 4]),
+            ...   pyhf.tensorlib.astensor([5, 6, 7]))
             <BLANKLINE>
             [[1. 1. 1.]
-             [2. 2. 2.]
-             [3. 3. 3.]]
+             [2. 3. 4.]
+             [5. 6. 7.]]
             <NDArray 3x3 @cpu(0)>
 
         Args:
--- a/pyhf/tensor/mxnet_backend.py
+++ b/pyhf/tensor/mxnet_backend.py
@@ -293,7 +293,6 @@ class mxnet_backend(object):
             ...   pyhf.tensorlib.astensor([1]),
             ...   pyhf.tensorlib.astensor([2, 2]),
             ...   pyhf.tensorlib.astensor([3, 3, 3]))
-            ...
             <BLANKLINE>
             [[1. 1. 1.]
              [2. 2. 2.]
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -1,14 +1,5 @@
 import logging
 
-import pyhf.optimize as optimize
-import pyhf.tensor as tensor
-from . import exceptions
-
-log = logging.getLogger(__name__)
-tensorlib = tensor.numpy_backend()
-default_backend = tensorlib
-optimizer = optimize.scipy_optimizer()
-default_optimizer = optimizer
 
 
 def get_backend():
@@ -28,9 +19,32 @@ def get_backend():
     global optimizer
     return tensorlib, optimizer
 
-# modifiers need access to tensorlib
+
+def tensorize(func):
+    """
+    Decorator to ensure all arguments passed to a function are tensors
+
+    Args:
+        func: pyhf.tensor class method with arguments (self, *args)
+    """
+    def wrapper(*args):
+        return func(args[0], *map(tensorlib.astensor, args[1:]))
+    return wrapper
+
+
+# modifiers and tensor needs access to tensorlib
 # make sure import is below get_backend()
+from . import tensor
+from . import optimize
+from . import exceptions
 from . import modifiers
+
+log = logging.getLogger(__name__)
+
+tensorlib = tensor.numpy_backend()
+default_backend = tensorlib
+optimizer = optimize.scipy_optimizer()
+default_optimizer = optimizer
 
 
 def set_backend(backend):
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -1,5 +1,14 @@
 import logging
 
+import pyhf.optimize as optimize
+import pyhf.tensor as tensor
+from . import exceptions
+
+log = logging.getLogger(__name__)
+tensorlib = tensor.numpy_backend()
+default_backend = tensorlib
+optimizer = optimize.scipy_optimizer()
+default_optimizer = optimizer
 
 
 def get_backend():
@@ -19,32 +28,9 @@ def get_backend():
     global optimizer
     return tensorlib, optimizer
 
-
-def tensorize(func):
-    """
-    Decorator to ensure all arguments passed to a function are tensors
-
-    Args:
-        func: pyhf.tensor class method with arguments (self, *args)
-    """
-    def wrapper(*args):
-        return func(args[0], *map(tensorlib.astensor, args[1:]))
-    return wrapper
-
-
-# modifiers and tensor needs access to tensorlib
+# modifiers need access to tensorlib
 # make sure import is below get_backend()
-from . import tensor
-from . import optimize
-from . import exceptions
 from . import modifiers
-
-log = logging.getLogger(__name__)
-
-tensorlib = tensor.numpy_backend()
-default_backend = tensorlib
-optimizer = optimize.scipy_optimizer()
-default_optimizer = optimizer
 
 
 def set_backend(backend):
@@ -76,7 +62,6 @@ def set_backend(backend):
     #     optimizer = mxnet_optimizer()
     else:
         optimizer = optimize.scipy_optimizer()
-
 
 class modelconfig(object):
     @classmethod
--- a/pyhf/tensor/numpy_backend.py
+++ b/pyhf/tensor/numpy_backend.py
@@ -40,8 +40,17 @@ class numpy_backend(object):
         tensor_in_2 = self.astensor(tensor_in_2)
         return np.outer(tensor_in_1,tensor_in_2)
 
-    def astensor(self,tensor_in):
-        return np.array(tensor_in)
+    def astensor(self, tensor_in):
+        """
+        Convert to a NumPy array.
+
+        Args:
+            tensor_in (Number or Tensor): Tensor object
+
+        Returns:
+            `numpy.ndarray`: A multi-dimensional, fixed-size homogenous array.
+        """
+        return np.asarray(tensor_in)
 
     def sum(self, tensor_in, axis=None):
         tensor_in = self.astensor(tensor_in)
--- a/pyhf/tensor/numpy_backend.py
+++ b/pyhf/tensor/numpy_backend.py
@@ -89,6 +89,28 @@ class numpy_backend(object):
         return np.concatenate(sequence)
 
     def simple_broadcast(self, *args):
+        """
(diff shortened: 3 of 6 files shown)


matthewfeickert:
This makes the assumption that the input arguments that are not of length 1 are repeated entries of the same number (i.e., `[3,3,3]`). However, as Lukas [mentioned](https://github.com/diana-hep/pyhf/issues/158#issuecomment-386919916) that
> The only type of broadcasting we need is the scenario of taking a set of `(1,)`- or `(N,)`-shape arrays and normalizing them to `(N,)`-shape arrays.

I think this is okay.

Let me know if there are any changes that people would like to see, and I can make them in the morning.

lukasheinrich:
yeah my understanding is that `np.broadcast_array` is the canonical solution. But we don't need to fully implement its functionality in the other backends, just the scenario @kratsg explained. I also think an example where the non-scalar/non-single-item lists are lists of changing numbers makes the docstring more understandable.

E.g.

```
simple_broadcast([1,[1,2,3]] ) = [ [1,1,1], [1,2,3] ] # turn scalar to list and handle example 2
simple_broadcast([[1],[1,2,3]]) = [ [1,1,1], [1,2,3] ] # expand (1,) list to (N,) list
simple_broadcast([[1,1,1],[1,2,3]]) = [ [1,1,1], [1,2,3] ] # noop
```

i'll also note that the broadcasting can be a candidate for optimization since in #143 I saw that it might be a significant source of CPU time

matthewfeickert:
Okay, I can update the doc strings and also add tests that ensure that check the scalar to list is done correctly as well.

Sure, I didn't clearly communicate this, but my point was at the moment this PR has this functionality
```python
>>> pyhf.tensorlib.simple_broadcast(pyhf.tensorlib.astensor([1]),
    pyhf.tensorlib.astensor([2, 3]),
    pyhf.tensorlib.astensor([3, 4, 5]))
[array([1, 1, 1]), array([2, 2, 2]), array([3, 4, 5])]
```
so is this okay, or it is better in these cases where the shapes are incompatible to not make assumptions for the user and to just fail?

lukasheinrich:
Hi @matthewfeickert, hm I think it's best to perhaps let it fail.  E.g. check shapes of all input arrays and if number of non-lenght-1 arrays is greater than 1 fail.

matthewfeickert:
Okay sounds good. I'll rework all of this this evening.

matthewfeickert:
@lukasheinrich As I'm going to be reworking this, then an additional question RE: the API. Do we want to totally match the `np.broadcast_arrays` API in that lists will be accepted? In the other backends the user is expected to pass in a tensor at the moment.
```
>>> import pyhf
>>> import numpy as np
>>> pyhf.tensorlib.simple_broadcast(
    pyhf.tensorlib.astensor([1]),
    pyhf.tensorlib.astensor([2, 3, 4]),
    pyhf.tensorlib.astensor([5, 6, 7]))
[tensor([ 1.,  1.,  1.]), tensor([ 2.,  3.,  4.]), tensor([ 5.,  6.,  7.])]
>>> pyhf.tensorlib.simple_broadcast([[1], [1, 2, 3], [3, 4, 5]])
ValueError: expected sequence of length 1 at dim 1 (got 3)
>>> np.broadcast_arrays([[1], [1, 2, 3], [3, 4, 5]])
[array([list([1]), list([1, 2, 3]), list([3, 4, 5])], dtype=object)]
```

kratsg:
> @lukasheinrich As I'm going to be reworking this, then an additional question RE: the API. Do we want to totally match the np.broadcast_arrays API in that lists will be accepted? In the other backends the user is expected to pass in a tensor at the moment.

Not entirely sure one would want to accept lists. This will break the lazy evaluation such as in tensorflow which builds up tensors for passing around, no? We should encourage tensors where possible to keep consistency.

pyhf/tensor/mxnet_backend.py
```diff
@@ -293,7 +293,6 @@ def simple_broadcast(self, *args):
             ...   pyhf.tensorlib.astensor([1]),
             ...   pyhf.tensorlib.astensor([2, 2]),
             ...   pyhf.tensorlib.astensor([3, 3, 3]))
-            ...
```
kratsg: double-check that this actually needs to be removed. When you run in a python interpreter, there's always one more blank `...` in a block before you close it out. This is mapping that.

kratsg: Yeah, I mean, you usually see something like

```python
>>> def test():
...   print "1"
... 
```

matthewfeickert: Yeah, exactly. Hence the need for the `...` in the `with` block in the TensorFlow backend.

kratsg: Well, I meant the trailing `...` here which you deleted, but ok. If the doctest is fine with it either way (e.g. just run pytest) then it's fine w/ me.

matthewfeickert:
> Not entirely sure one would want to accept lists. This will break the lazy evaluation such as in tensorflow which builds up tensors for passing around, no? We should encourage tensors where possible to keep consistency.

@kratsg Yeah, I mean we would wrap the inputs in `astensor`, so that should keep things from breaking (right?). I don't have strong feelings on this, as I assume that anywhere it is going to get called tensor objects will already have been constructed, so I'll just implement whatever you and @lukasheinrich agree on.

kratsg:
> @kratsg Yeah, I mean we would wrap the inputs in astensor, so that should keep things from breaking (right?). I don't have strong feelings on this, as I assume that anywhere it is going to get called tensor objects will already have been constructed, so I'll just implement whatever you and @lukasheinrich agree on.

Would it be easy to provide a decorator for the functions that require all tensor inputs to just do something like

```python
def tensorizable(func):
  def _wrapper(*args):
    return func(*map(pyhf.tensorlib.astensor, args))
  return _wrapper
```

@lukasheinrich ?

pyhf/tensor/mxnet_backend.py
matthewfeickert: We're in agreement in understanding about what is being talked about, even though we seem to both not be able to convey this to each other in text. Of course, if my REPL is different then the doc test for some reason it will go back.

lukasheinrich:
Hi,

something like the decorator might be nice also for other reasons (e.g. when optimizing code for #143 I found even np.asarray being an overhad (presumably because it does some type checking) ) and maybe if you have an decorator one could have an "optimized" mode in which the `.asensor` calls are skipped and you just blindly trust that the inputs work. But the decorator also adds an additional layer of indirection and given that the tensorlibs are very core and get calld *a ton* of times we really need to make sure this stays performant.

lukasheinrich:
@matthewfeickert no, I don't think we need to re-implement the functionality of `np.broadcast_arrays` exhaustively. That was part of the motivation why I called this method `simple_broadcast` since it just does the reduced use-case we need for pyhf

matthewfeickert:
@lukasheinrich 

> I don't think we need to re-implement the functionality of `np.broadcast_arrays` exhaustively. 

Sure. I guess I should have just said "Do we want to support lists and floats as inputs, or just tensors?". However, as

> something like the decorator might be nice

then I guess this is irrelevant as the decorator will take care of all of that. :+1: 

Naive question: Is there an easy way to monitor the performance effect of using the decorator? Or is this something that you can look at using the checks that you've been developing for Issue #143?

kratsg:
> Naive question: Is there an easy way to monitor the performance effect of using the decorator? Or is this something that you can look at using the checks that you've been developing for Issue #143?

`cProfile` will do it and combine it with a visualizer like `snakeviz`: https://jiffyclub.github.io/snakeviz/

matthewfeickert:
So in commit fa95141 I was able to add a `@tensorize` decorator, but it isn't what we want yet. The problem is caching, in that it doesn't know to reset the backend.

Example, for the `test_tensor.py` tests to pass one needs to add
```
pyhf.set_backend(tb)
```
at the start of the loop.

Obviously the solution is to just have tensorize look like
```python
def tensorize(func):
    tensorlib, _ = get_backend()
    def wrapper(*args):
        return func(args[0], *map(tensorlib.astensor, args[1:]))
    return wrapper
```
However, I am having trouble figuring out how this all should work in the deceleration and import order in `pyhf/__init__.py`. Comments in the code review would be highly appreciated.

Additionally, I would think that if we could move `tensorize` into `tensor/__init__.py` it would be better given that it is needed for those modules. Thoughts?

kratsg:
> So in commit fa95141 I was able to add a @tensorize decorator, but it isn't what we want yet. The problem is caching, in that it doesn't know to reset the backend.
> 
> Example, for the test_tensor.py tests to pass one needs to add

This, I suspect is due partially to the fact that relative imports are being cache'd. This issue wouldn't exist (I think) if one just did `pyhf.tensorlib` to access the current tensorlib. That said, it's still not clear to me why relative imports suffer this issue -- hence why I had introduced `get_backend()` to fix this thing...

matthewfeickert:
So I can move `tensorize` into `tensor/__init__.py` by having the first lines be

```python
import pyhf


def tensorize(func):
    def wrapper(*args):
        return func(args[0], *map(pyhf.tensorlib.astensor, args[1:]))
    return wrapper
```

This import feels like wrong though, and it doesn't really solve anything: 
- `test_tensor.py` still fails unless the `set_backend()` is at the top of the loop
- `get_backend()` is still unable to be used in `tensorized` as `get_backend()` needs to be defined in `pyhf` before `pyhf.tensor` is imported to be used in the tensor modules, but `get_backend()` also relies on the names `tensorlib`, and `optimizer` being defined but `tensorlib` requires the tensor modules to be used. So it seems we're in a "nice" circle.

I don't really understand how to fix this yet, so this might take a while to conclude all of this.

kratsg:
> test_tensor.py still fails unless the set_backend() is at the top of the loop

Remember that the tests need the set_backend(). but https://github.com/diana-hep/pyhf/blob/master/tests/conftest.py should be doing it for yu -- not sure why it isn't...

matthewfeickert:
@kratsg 
> Remember that the tests need the `set_backend()` but `conftest.py` should be doing it for you

Sorry, I didn't clarify what I mean. `conftest.py` is resetting the backend to the default just fine. What I meant is the loop in `test_common_tensor_backends` requires explicitly setting the backend desired
```python
#...
    for tb in [
        numpy_backend(),
        pytorch_backend(),
        tensorflow_backend(session=tf_sess),
        mxnet_backend()
    ]:
        pyhf.set_backend(tb)
```
which wasn't required in the past (current `master`).

matthewfeickert:
Since this decorator is starting to take up a lot of time, I am tempted to just revert the decorator commits and finish fixing the `simple_broadcast()` for the backends and then address the decorator in another PR. Anyone against this? I can open an Issue that details all the discussion that we've had so far.

lukasheinrich:
Yeah I think that's a good idea @matthewfeickert

matthewfeickert:
The doc stage of the PR build is failing within the first 2 minutes with a message of 

```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
The job has been terminated
```
and then continuing to hang. This is curious to me as it is just downloading packages, so I'm not sure why it should have hit 4 MB as it isn't printing that much out.

kratsg:
Well, there's good news and bad news. The good news is that the travis is working as expected.. the bad news is that even if this had succeeded -- your benchmark tests most likely would've failed when merging with master: https://github.com/diana-hep/pyhf/blob/master/.travis.yml#L55

@lukasheinrich any idea why this `pip install` command is so bad? [the docs need the development environment in order to be able to handle adding the jupyter notebooks as part of documentation....].

matthewfeickert:
It is a bit of a bandaid rather than a fix, but doing a quiet `pip install` fixes this for now.

@lukasheinrich @kratsg I think this is finally ready for review. If there aren't any comments or requested changes then we can squash and merge this (and it would be nice to clean up the commit message log a bit as I don't think that having revert statements is in anyway helpful).

matthewfeickert:
@kratsg One thing that @lukasheinrich pointed out is that the logic for checking to make sure that broadcasts aren't attempted with wrong shapes might could potentially slow things down. If you have time could you profile this? If you're busy I can look at it, but as I need to setup `cProfile` and `snakeviz` it might be faster if you have the time to check.

lukasheinrich:
I think we can keep the checks and potential castings for correctness and just be mindful that they can become a bottleneck, since we call these functions so much. We should indeed set up some profiling in CI to keep track of this. (e.g. while optimzing for #143 I saw that I mistakenly used `np.array` instead of `np.asarray` which did indeed cause many unneeded copies)

@matthewfeickert  I think this is a LGTM for me, or are you planning to add some more to this PR?

matthewfeickert:
@lukasheinrich If we open up a separate Issue for adding profiling to CI (would you mind opening it?) then I think this PR can be done, squashed and merged. :shipit:

PR #164, opened 2018-05-12
Return CLs instead of 1/CLs

### LINKED ISSUE #140: figure out why we do `CLb / CLsb`
# Description

we have this snippet:

```python
from scipy.stats import norm
def pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v):
    CLsb = 1 - norm.cdf(sqrtqmu_v)
    CLb = norm.cdf(sqrtqmuA_v - sqrtqmu_v)
    CLs = CLb / CLsb
    return CLsb, CLb, CLs
```

as @matthewfeickert correctly observed this seems off at first sight. But I seem to remember there was a specific reason I did it this way (perhaps while keeping around wrong variable names) 

i think it has something to do with this line in ROOT

```c++
 virtual Double_t CLb() const { return !fBackgroundIsAlt ? NullPValue() : AlternatePValue(); }
```

https://root.cern.ch/doc/v606/HypoTestResult_8h_source.html#l00138

matthewfeickert:
Follows up on work from PR #141 and resolves #140 

- [x] Have `pvals_from_teststat` return `CLs` rather than `1/CLs`
- [x] Correct this in `test_validation.py`
- [x] ~Find a way to version the Jupyter notebooks so that changes are reflected without creating a notebook wide diff~ (this is actually really hard, so just manually change everything without running if I don't have to)
- [x] Correct this in all Jupyter notebooks
- [x] Expand the docstring to give more information on the equations in [arXiv:1007.1727](https://arxiv.org/abs/1007.1727) (note to self: double check that these doc strings are right)
- [x] Ensure that the LaTeX style looks good for the docs (build them locally in addition to the doc tests)

- [x] Tests are passing


docs/examples/notebooks/ShapeFactor.ipynb
```diff
@@ -258,8 +247,8 @@
     "    plt.ylim(0,1)\n",
     "\n",
     "def CLsOnePoint(muTest, data, pdf, init_pars, par_bounds):\n",
-    "    clsobs, cls_exp = pyhf.runOnePoint(muTest, data, pdf, init_pars,par_bounds)[-2:]\n",
-    "    return 1./clsobs, [1./x for x in cls_exp]\n",
+    "    cls_obs, cls_exp = pyhf.runOnePoint(muTest, data, pdf, init_pars,par_bounds)[-2:]\n",
+    "    return cls_obs, cls_exp\n",
     "\n",
```
lukasheinrich: I think with this diff we can completely remove this `CLsOnePoint` function and just use `pyhf.runOnePoint`

docs/examples/notebooks/StatError.ipynb
```diff
@@ -201,8 +201,8 @@
     "from pyhf.simplemodels import hepdata_like\n",
     "\n",
     "def CLsOnePoint(muTest, data, pdf, init_pars,par_bounds):\n",
-    "    clsobs, cls_exp = runOnePoint(muTest, data,pdf,init_pars,par_bounds)[-2:]\n",
-    "    return 1./clsobs, [1./x for x in cls_exp]\n",
+    "    cls_obs, cls_exp = runOnePoint(muTest, data,pdf,init_pars,par_bounds)[-2:]\n",
+    "    return cls_obs, cls_exp\n",
     "\n",
```
lukasheinrich: dito

docs/examples/notebooks/ShapeFactor.ipynb
matthewfeickert: Good point. I'll apply this to everything.

docs/examples/notebooks/binderexample/StatisticalAnalysis.ipynb
```diff
@@ -297,8 +297,8 @@
     "    ax.set_ylim(0,1)\n",
     "\n",
     "def CLsOnePoint(muTest, data, pdf, init_pars, par_bounds):\n",
-    "    qmu,_,_,_,clsobs, cls_exp = pyhf.runOnePoint(muTest, data, pdf, init_pars,par_bounds)\n",
-    "    return qmu, 1./clsobs, [1./x for x in cls_exp]\n",
+    "    qmu,_,_,_,cls_obs, cls_exp = pyhf.runOnePoint(muTest, data, pdf, init_pars,par_bounds)\n",
+    "    return qmu, cls_obs, cls_exp\n",
     "\n",
```
lukasheinrich: dito (need to add `[-2:]` though

docs/examples/notebooks/ShapeFactor.ipynb
matthewfeickert: This (and all other instances) should now be fixed in commit b4a17d8

matthewfeickert:
This is what the docs should now look like given commit af1d23d 

sphinx_pyhf_docs

Let me know if there are style changes desired in the doc string presentation.

matthewfeickert:
I think this is now ready for review. If there aren't any questions or edits, I'll let @lukasheinrich squash and merge this in.

lukasheinrich:
thanks @matthewfeickert , looks great, esp. the documentation. merged.

PR #167, opened 2018-05-13
Fix headings of histogrammar notebook

### LINKED ISSUE #166: Fix Histogrammar Jupyter Notebook headings
The heading for "Creating a Statistical model based on the Histograms (HistFactory)" in the Histogrammar Jupyter notebook currently has the wrong level (level 1 instead of 2). As a result, Sphinx thinks that it is its own notebook and creates a separate entry in the Examples sidebar.
histogrammar_problem

matthewfeickert:
Resolves #166 

If a notebook has multiple first level headings Sphinx will think that it should create a separate nav bar level entry for it, treating it like its own notebook. This enforces a single first level heading.

This is what the fix should look like:
histogrammar_fix

- [x] Tests are passing

### DIFF
--- a/docs/examples/notebooks/histogrammar.ipynb
+++ b/docs/examples/notebooks/histogrammar.ipynb
@@ -9,7 +9,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 111,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [
     {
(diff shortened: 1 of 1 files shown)


lukasheinrich:
thanks LGTM. (i'm merging this, though I don't see us installing hg in the develop env, so i'm confused how it works :-p )

matthewfeickert:
Yeah, as you point out this _doesn't_ work (out of the box for us). If this notebook was in the CI then this would totally fail. However, at the moment we only test a few core notebooks, but I guess this is a good time to open an Issue about testing _all_ the notebooks. :grimacing:

PR #171, opened 2018-05-14
RECAST example

lukasheinrich:
# Description

Adds a nice example how easy Reinterpretations are once we have likelihoods expressed as JSON. Just apply a JSON-patch!

(also adds `jsonpatch` as a dev package)

- [ ] Tests are passing


lukasheinrich:
hmm I pushed the wrong NB

matthewfeickert:
Ah okay, that would explain why I didn't see any cells using `jsonpatch` :?

PR #175, opened 2018-05-14
Add Binder example notebook to docs examples

### LINKED ISSUE #173: Add binderexample/StatisticalAnalysis.ipynb to docs examples
At the moment the Binder example notebook `binderexample/StatisticalAnalysis.ipynb` is not being included in the build of the docs. Lukas has put some really nice examples in there and it would be nice to have this in the docs in the even that people don't check out the Binder.

This is also somewhat relevant to Issue #168

matthewfeickert:
Resolves #173 

This is what the docs look like with the addition of the Binder example
examples_with_pr

It should be noted that the interactive plotting doesn't display (as expected as these are static pages) but I don't think that is a big enough detractor from the high quality of the notebook to keep it out of the docs.

- [x] Also add in the [RECAST example notebook](https://github.com/diana-hep/pyhf/blob/master/docs/examples/notebooks/Recast.ipynb) to the docs examples TOC

- [x] Tests are passing

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -94,7 +94,7 @@ language = None
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This patterns also effect to html_static_path and html_extra_path
-exclude_patterns = ['_build', '**.ipynb_checkpoints', 'examples/experiments', 'examples/notebooks/binderexample']
+exclude_patterns = ['_build', '**.ipynb_checkpoints', 'examples/experiments']
 
 # The reST default role (used for this markup: `text`) to use for all
 # documents.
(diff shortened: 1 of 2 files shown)


matthewfeickert:
> It would be more straightforward to just have a link for creating a binder from a given notebook page on the static docs, no?

Sure, my only thought is that there might be times that you just would want a static version. For example, there are times when Binder needs to rebuild the image and repo2docker can be pretty slow at times. So if you just want to look at the docs _now_ then this could be useful then just the "launch Binder" link. Thoughts?

kratsg:
I mean, have both. Like a link to GH page, but for the jupyter notebooks to launch a binder demo as well. Either have the link as part of the notebook or, as part of the docs.

matthewfeickert:
@kratsg You're thinking something like this? Or something else?
examples_with_binder_badge

lukasheinrich:
LGTM !

PR #176, opened 2018-05-14
better default options

lukasheinrich:
# Description

This adds some sane defaults for init and par bounds since we have access to the pdf config in the function anyways.

One could also think about automatically extending the data with the aux measurements, since right now we have this pattern

```
obs_data = [...]
pyhf.runOnePoint(1.0, obs_data + pdf.config.auxdata, pdf ...)
```

which could be handled for the user as well to be

```
obs_data = [...]
pyhf.runOnePoint(1.0, obs_data, pdf, ...)
```

 but either we make this automatic / "smart" or the default would change, which would be backward-compat. breaking (requiring a modification of notebooks etc)

- [x] Tests are passing

### DIFF
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -380,7 +380,33 @@ def pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v):
     return CLsb, CLb, CLs
 
 
-def runOnePoint(muTest, data, pdf, init_pars, par_bounds):
+def runOnePoint(muTest, data, pdf, init_pars = None, par_bounds = None):
+    r"""
+    Computes test statistics (and expected statistics) for a single value
+    of the parameter of interest
+
+    Args:
+        muTest (Number or Tensor): The value of the parameter of interest (POI)
+        data (Number or Tensor): The root of the calculated test statistic given the Asimov data, :math:`\sqrt{q_{\mu,A}}`
+        init_pars (Array or Tensor): the initial parameter values to be used for minimization
+        par_bounds (Array or Tensor): the parameter value bounds to be used for minimization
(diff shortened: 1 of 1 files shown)


pyhf/__init__.py
```diff
@@ -380,7 +380,33 @@ def pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v):
...
+    Args:
+        muTest (Number or Tensor): The value of the parameter of interest (POI)
+        data (Number or Tensor): The root of the calculated test statistic given the Asimov data, :math:`\sqrt{q_{\mu,A}}`
+        init_pars (Array or Tensor): the initial parameter values to be used for minimization
+        par_bounds (Array or Tensor): the parameter value bounds to be used for minimization
+
+    Returns:
+        Tuple of Floats: a tuple containing (qmu, qmu_A, CLsb, CLb, CLs, CLs_exp)
+                         where qmu and qmu_A are the test statistics for the
+                         observerd and asimov datasets respectivleyself.
```
matthewfeickert: This looks good, and is nice to have. Also great to have a doc string for runOnePoint now. :+1: However, since this is going into the docs, I think there are some tiny typos here. I think this line should be "observed and Asimov datasets respectively."

matthewfeickert:
LGTM! :shipit:

matthewfeickert:
> but either we make this automatic / "smart" or the default would change, which would be backward-compat. breaking (requiring a modification of notebooks etc)

This sounds like a doable thing that just will require some late at night busy work cleaning of stuff, so if you want to open an Issue for this in the future we can definitely do it (might be nicer in the long run?).

PR #181, opened 2018-07-11
Use pip cache in Travis

matthewfeickert:
Caching pip speeds up builds
https://docs.travis-ci.com/user/caching/#pip-cache

- [ ] Tests are passing


matthewfeickert:
So the question is does anyone see a problem with having a cache (I'm not sure how long it persists before it gets cleared automatically, but we can manually clean it from the [Travis CI settings for the repo](https://travis-ci.org/diana-hep/pyhf/caches)) to get some speedup in the CI? When I first made this commit the CI (with no cache yet) took O(9 minutes) to finish. I did a `git commit --amend --no-edit` and force pushed and the time came down to O(6 minutes). Though comparing to some older builds that isn't that big of a deal.

Thoughts?

kratsg:
> So the question is does anyone see a problem with having a cache (I'm not sure how long it persists before it gets cleared automatically, but we can manually clean it from the Travis CI settings for the repo) to get some speedup in the CI?

I don't know the details of cache invalidation for pip, but this should be fine on the default. If you find it's pulling in old dependencies later, this will presumably just break the tests since the dependencies would only be updated for very specific reasons, rather than simply requiring the min version of a dep.



> When I first made this commit the CI (with no cache yet) took O(9 minutes) to finish. I did a git commit --amend --no-edit and force pushed and the time came down to O(6 minutes). Though comparing to some older builds that isn't that big of a deal.

Do we have any guidelines on how fast this should be? Do we think O(5 minutes) or O(10 minutes) is reasonable? Or are we expecting O(1 minute)?

matthewfeickert:
> I don't know the details of cache invalidation for pip, but this should be fine on the default.

Cool.

> Do we have any guidelines on how fast this should be? 

The bulk of the time is coming from the testing suite, so I think with just the general overhead of getting the Docker image setup and going things are going to have a lower bound of O(5+ minutes).

lukasheinrich:
LGTM. feel free to merge.

PR #186, opened 2018-08-17
fix up some of the package versioning. add docs on development installation

### LINKED ISSUE #162: Updated setup.py for pytorch > 0.4 dependency
# Description

I had 0.3.1 for Torch and that caused issues with some of the doctesting as the distributions did not have `cdf` methods. I forced an upgrade pytorch and things are fine now.

kratsg:
# Description

This PR adds a `development` docs with a short blurb about how to `pip` install for full-suite development

```
pip install -e .[develop,tensorflow,mxnet,torch]
```

and also updates `setup.py` to freeze down some of the necessary versions to ensure that things are installed correctly.

Closes #162.

- [x] Tests are passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - '3.6'
 install:
 - pip install --upgrade pip
-- pip install -U -q -e .[develop]
+- pip install -U -q -e .[develop,tensorflow,mxnet,torch]
 script:
   - pyflakes pyhf
   - pytest --ignore tests/benchmarks/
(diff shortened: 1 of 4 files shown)


lukasheinrich:
LGTM.

lukasheinrich:
should we recommend using pipenv?

kratsg:
> should we recommend using pipenv?

I'll file an issue.

kratsg:
An important note is that this change will remove **any/all** backends from being installed by default. The reason for this is that it should really be up to the user to choose their backend. E.G. they may already have tensorflow/torch/mxnet installed so this is installed on top assuming that backend exists.

They may not have a backend and can choose to install one via something like `pip install pyhf[tensorflow]` or `pyhf[mxnet,torch]` which is a lot cleaner, I think.

matthewfeickert:
LGTM. :+1:

PR #188, opened 2018-08-20
Feature: spec validation on init, update logging, add staterror to schema

### LINKED ISSUE #118: Add spec validation on hfpdf.__init__
# Description

we now have a spec, that should allow us to do a lint-type check for validity of the spec. passing that doesn't guarantee correctness, but just that the spec is well-formed.

we might implement a validation routine whose first step would be linting and later we can add other checks (e.g. check same-length for all sample data)

kratsg:
# Description

This resolves #118. This adds specification validation to `hfpdf.__init__()` calls. This will simplify the tests used throughout, and all testing relevant to making sure the schema validation catches common pitfalls is being separated into `tests/test_schema.py` (for want of a better name).

This also sets up class-specific logging functionality to help filter out the logging a bit better as well in the future.

Also adds `staterror` to schema -- somehow missed this.

- [x] Tests are passing

### DIFF
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -1,8 +1,11 @@
 import logging
+logging.basicConfig()
+import pkg_resources
 
 import pyhf.optimize as optimize
 import pyhf.tensor as tensor
 from . import exceptions
+from . import utils
 
 log = logging.getLogger(__name__)
 tensorlib = tensor.numpy_backend()
@@ -153,8 +156,10 @@ class modelconfig(object):
 
 class hfpdf(object):
     def __init__(self, spec, **config_kwargs):
-        self.config = modelconfig.from_spec(spec,**config_kwargs)
         self.spec = spec
+        self.schema = config_kwargs.get('schema', pkg_resources.resource_filename(__name__,'../validation/spec.json'))
+        utils.validate(self.spec, self.schema)
+        self.config = modelconfig.from_spec(spec,**config_kwargs)
 
     def expected_sample(self, channel, sample, pars):
         """
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -1,5 +1,4 @@
 import logging
-logging.basicConfig()
 import pkg_resources
 
 import pyhf.optimize as optimize
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -1,5 +1,4 @@
 import logging
-import pkg_resources
 
 import pyhf.optimize as optimize
 import pyhf.tensor as tensor
@@ -156,7 +155,7 @@ class modelconfig(object):
 class hfpdf(object):
     def __init__(self, spec, **config_kwargs):
         self.spec = spec
-        self.schema = config_kwargs.get('schema', pkg_resources.resource_filename(__name__,'data/spec.json'))
+        self.schema = config_kwargs.get('schema', utils.get_default_schema())
         utils.validate(self.spec, self.schema)
         self.config = modelconfig.from_spec(spec,**config_kwargs)
 
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -156,7 +156,7 @@ class modelconfig(object):
 class hfpdf(object):
     def __init__(self, spec, **config_kwargs):
         self.spec = spec
-        self.schema = config_kwargs.get('schema', pkg_resources.resource_filename(__name__,'../validation/spec.json'))
+        self.schema = config_kwargs.get('schema', pkg_resources.resource_filename(__name__,'data/spec.json'))
         utils.validate(self.spec, self.schema)
         self.config = modelconfig.from_spec(spec,**config_kwargs)
 
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -65,6 +65,9 @@ def set_backend(backend):
         optimizer = optimize.scipy_optimizer()
 
 class modelconfig(object):
+    # set up the logging using the name of this class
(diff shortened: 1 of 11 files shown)


kratsg:
A current blocker to this, right now, is https://github.com/Julian/jsonschema/issues/454 because I've noticed that we're not currently catching invalid modifiers right now. This should be caught at the schema level (and not just inside the `hfpdf` code which is what happens right now).

pyhf/__init__.py
```diff
@@ -1,8 +1,11 @@
 import logging
+logging.basicConfig()
```
lukasheinrich: as a library, we should probably not be configuring logging. if we add a CLI tool at some point we might do it there

pyhf/__init__.py
```diff
@@ -153,8 +156,10 @@ def add_or_get_modifier(self, channel, sample, modifier_def):
 
 class hfpdf(object):
     def __init__(self, spec, **config_kwargs):
-        self.config = modelconfig.from_spec(spec,**config_kwargs)
         self.spec = spec
+        self.schema = config_kwargs.get('schema', pkg_resources.resource_filename(__name__,'../validation/spec.json'))
```
lukasheinrich: not sure what the best practice here is. to me `pkg_resources.resource_filename('pyhf','validation/spec.json'))` looks a bit nicer

pyhf/utils.py
```diff
@@ -0,0 +1,24 @@
+import json, jsonschema
```
lukasheinrich: indenting seems off in this file

pyhf/__init__.py
kratsg: Unfortunately, it'll do it with respect to `__file__` which means we need an extra `../` to get to `validation/`. Similarly, `__name__ == 'pyhf'` for all intents and purposes. I don't know whether to be explicit here or not.

pyhf/utils.py
kratsg: Noted.

pyhf/__init__.py
lukasheinrich: if you rebase to `master`  `pkg_resources.resource_filename('pyhf','validation/spec.json'))` should now work

pyhf/utils.py
kratsg: This is done.

lukasheinrich:
Hi @kratsg should we have class-specific logging? I kind of like the module-level pattern. Perhaps this indicates more that the modelconfig and hfpdf classes should be in separate modules?

kratsg:
> Hi @kratsg should we have class-specific logging? I kind of like the module-level pattern. Perhaps this indicates more that the modelconfig and hfpdf classes should be in separate modules?

Good question. I err on having much more detailed logging -- only because even though both are under `pyhf` -- we have extra granularity if we add in the class itself. I can always change it out later -- but we do need to separate it out like this when you eventually multi-thread some pieces of it.

lukasheinrich:
Yeah maybe it's time to clean out `__init__.py` and move everything into modules. I recall @dguest being confused where the code was :)

kratsg:
> Yeah maybe it's time to clean out __init__.py and move everything into modules. I recall @dguest being confused where the code was :)

Yeah, new issue for that.

Second issue here is that now that the validation/schema is actually truly fixed -- we're losing coverage in our tests since the schema validation is catching the errors first ;) 

What shall we do?

dguest:
Ha, yeah seriously, who writes their code in `__init__.py`? Crazy people, that's who.

Anyway, keep up the good work, this is one of the more awesome projects I've seen in HEP recently.

kratsg:
> Ha, yeah seriously, who writes their code in __init__.py? Crazy people, that's who.

#192 is filed for this.

pyhf/utils.py
```diff
@@ -0,0 +1,27 @@
...
+import pkg_resources
+
+from .exceptions import InvalidSpecification
+
+DEFAULT_SCHEMA = pkg_resources.resource_filename(__name__,'data/spec.json')
+def get_default_schema():
+  global DEFAULT_SCHEMA
+  return DEFAULT_SCHEMA
+
```
lukasheinrich: why can this not simply be

```
def get_default_schema():
    return pkg_resources.resource_filename(__name__,'data/spec.json')
```

pyhf/utils.py
```diff
@@ -0,0 +1,27 @@
...
+SCHEMA_CACHE = {}
+def load_schema(schema):
+    global SCHEMA_CACHE
+    try:
+        return SCHEMA_CACHE[schema]
+    except KeyError:
+        pass
+
```
lukasheinrich: do we really want to swallow this error?

kratsg: This error occurs when the provided schema wasn't cached. In this case, it's because it wasn't loaded in the cache yet -- so load the json, cache it, and return it. This is just a try/except to be more pythonic.

pyhf/utils.py
kratsg: Yeah, probably can change to just that. I wasn't sure about the repetitive function calls to `pkg_resources` etc... but it's a silly optimization here.

pyhf/utils.py
kratsg: This is to handle the case where the schema is very large and there's also no reason why we shouldn't cache the schemas used for a given instance of `pyhf`.

lukasheinrich: ah sorry I was confused by the naming. I would rename this to `load_schema_from_cache` or `schema_from_cache`

lukasheinrich: scatch that. I just looked at the code. it's fine.

Issue #192, opened 2018-08-21
Migrate things from __init__ to separate modules


kratsg:
# Description

`__init__.py` is crazy cluttered and needs to be cleaned up.

PR #193, opened 2018-08-21
Reorganize/clean up init file, rename hfpdf, update API docs

kratsg:
# Description

We're not crazy people, cref: #192.

- [x] Tests are passing


kratsg:
Any ideas on where to put `hfpdf` and `modelconfig`? Can also rename them, but I haven't decided because of how the import statements should look? `from pyhf import pdf`?

matthewfeickert:
>Any ideas on where to put `hfpdf` and `modelconfig`? 

I don't have any strong feelings on this, so I'll defer to you and @lukasheinrich.

>Can also rename them, but I haven't decided because of how the import statements should look? 

As you suggested, `from pyhf import pdf` sounds fine to me. It keeps things simple and clear.

Given that there is going to be quite a bit of renaming going on it might be worth creating a release `v0.0.9` after this PR is in. Thoughts?

lukasheinrich:
yeah `class pdf` in a `pdf.py` file sounds good to me. we can then import it in `__init__.py` and add it to `__all__`

matthewfeickert:
For future reference, the link that I added in my commit makes the `pyhf.pdf.Model` docs look like the following

hf_paper_link_in_docs

matthewfeickert:
I am really used to the `scipy.stats` style to the point where uppercase starts to look weird to me, but I do acknowledge that goes against [PEP 8](https://www.python.org/dev/peps/pep-0008/#class-names) which I'm usually a pedant on.

However, I think I might lean towards uppercase, as if we look towards the [TensorFlow distributions](https://www.tensorflow.org/api_docs/python/tf/distributions) and [PyTorch distributions](https://pytorch.org/docs/stable/distributions.html) they also go uppercase.

lukasheinrich:
alright, let's go with uppercase then. 

Merged.

lukasheinrich:
actually i'll let @kratsg merge so that he can give a commit message for the squashed commit

matthewfeickert:
@lukasheinrich Are you also on board with making release `v0.0.9` off of this PR? Or do you want to wait a bit more?

lukasheinrich:
@matthewfeickert yeah I can cut the release

PR #194, opened 2018-08-23
fix benchmark tests

kratsg:
# Description

The current master is breaking because of #193 moving `runOnePoint` to `utils.py` and forgetting to update the one line in the test for fixing it.

- [x] Tests are passing


matthewfeickert:
I guess we should also bump to `v0.0.11` so that we don't deploy a latest release with failing tests(?).

lukasheinrich:
kicked of another deploy. should be bumped on pypi soon once travis
finishes.

PR #196, opened 2018-08-23
Fix deploy

### LINKED ISSUE #195: fix deploy (on tags only)
# Description

The auto PyPI deploy seems broken as we 

* filter only on master
* filter only on tags

which block each other

https://travis-ci.org/diana-hep/pyhf/builds/419568619

I think we should only deploy on tagged commits.

lukasheinrich:
Resolves #195 

- [x] Tests are passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ stages:
   - name: docs
     if: (branch = master)
   - name: deploy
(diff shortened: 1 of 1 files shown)


matthewfeickert:
@lukasheinrich Thanks for noticing and fixing this!

PR #197, opened 2018-08-23
Feature: command line parse xml

kratsg:
# Description

This allows someone to use `pyhf` from command-line to parse the XML workspaces using the `readxml.parse` functionality. This uses `click` to set up options/arguments.

This also updates `readxml` with some slight reorganization to allow for a `tqdm` progress bar that can be enabled (default: disabled) for reading in the channels in various XML files.

```
pyhf_xml2json --entrypoint-xml validation/multibin_multibjets/config/NormalMeasurement.xml --workspace validation/multibin_multibjets/ --output_file test.json
```

screenshot 2018-08-23 10 45 23

NB: both `click` and `tqdm` are already installed via our dependencies, but I've explicitly listed them in `setup.py` now.

- [x] Tests are passing

### DIFF
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -1,13 +1,16 @@
 import logging
 logging.basicConfig()
+log = logging.getLogger(__name__)
 
 import click
 import json
 from . import readxml
 
 @click.command()
-@click.option('--entrypoint-xml', required=True, prompt='Top-level XML', help='The top-level XML file for the workspace definition.')
-@click.option('--workspace', required=True, prompt='Workspace directory', help='The location of workspace.')
-def xml2json(entrypoint_xml, workspace):
-    spec = readxml(entrypoint_xml, workspace)
-    import pdb; pdb.set_trace()
+@click.option('--entrypoint-xml', required=True, prompt='Top-level XML', help='The top-level XML file for the workspace definition.', type=click.Path(exists=True))
+@click.option('--workspace', required=True, prompt='Workspace directory', help='The location of workspace.', type=click.Path(exists=True))
+@click.option('--output-file', required=True, prompt='Output file', help='The location of the output json file. If not specified, prints to screen.', type=click.Path(exists=False))
+def xml2json(entrypoint_xml, workspace, output_file=None):
+    spec = readxml.parse(entrypoint_xml, workspace, enable_tqdm=True)
+    json.dump(spec, open(output_file, 'w+'), indent=4, sort_keys=True)
+    log.info("Written to {0:s}".format(output_file))
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -6,7 +6,11 @@ import click
 import json
 from . import readxml
 
-@click.command()
+@click.group(context_settings=dict(help_option_names=['-h', '--help']))
+def pyhf():
+    pass
+
+@pyhf.command()
 @click.option('--entrypoint-xml', required=True, prompt='Top-level XML', help='The top-level XML file for the workspace definition.', type=click.Path(exists=True))
 @click.option('--workspace', required=True, prompt='Workspace directory', help='The location of workspace.', type=click.Path(exists=True))
 @click.option('--output-file', required=True, prompt='Output file', help='The location of the output json file. If not specified, prints to screen.')
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -10,7 +10,8 @@ from . import readxml
 @click.option('--entrypoint-xml', required=True, prompt='Top-level XML', help='The top-level XML file for the workspace definition.', type=click.Path(exists=True))
 @click.option('--workspace', required=True, prompt='Workspace directory', help='The location of workspace.', type=click.Path(exists=True))
 @click.option('--output-file', required=True, prompt='Output file', help='The location of the output json file. If not specified, prints to screen.')
-def xml2json(entrypoint_xml, workspace, output_file=None):
-    spec = readxml.parse(entrypoint_xml, workspace, enable_tqdm=True)
+@click.option('--tqdm/--no-tqdm', default=True)
+def xml2json(entrypoint_xml, workspace, output_file, tqdm):
+    spec = readxml.parse(entrypoint_xml, workspace, enable_tqdm=tqdm)
     json.dump(spec, open(output_file, 'w+'), indent=4, sort_keys=True)
     log.info("Written to {0:s}".format(output_file))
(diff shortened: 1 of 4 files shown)


pyhf/commandline.py
```diff
@@ -0,0 +1,17 @@
...
+logging.basicConfig()
+log = logging.getLogger(__name__)
+
+import click
+import json
+from . import readxml
+
+@click.command()
+@click.option('--entrypoint-xml', required=True, prompt='Top-level XML', help='The top-level XML file for the workspace definition.', type=click.Path(exists=True))
+@click.option('--workspace', required=True, prompt='Workspace directory', help='The location of workspace.', type=click.Path(exists=True))
```
lukasheinrich: I think `workspace` is a bit of a misnomer here. The workspace is the result of parsing the xml and root files via `hist2workspace`. Should we use the same name as the `parse` function, i.e. `--rootdir`

tests/test_scripts.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+import json
+import shlex
+
+import pyhf
+
+# see test_import.py for the same (detailed) test
+def test_import_prepHistFactory(tmpdir, script_runner):
```
lukasheinrich: where is this fixture defined?

pyhf/commandline.py
```diff
@@ -0,0 +1,17 @@
+import logging
+logging.basicConfig()
+log = logging.getLogger(__name__)
+
+import click
+import json
+from . import readxml
+
+@click.command()
```
lukasheinrich: what do you think about a toplevel `pyhf` command and `xml2json` could be one of the subcommands

```
> pyhf --help
> pyhf xml2json --help
```


```
@click.group():
@click.option(...) #some global opts
def pyhf():
    pass

@pyhf.command()
@click.option('--entrypoint-xml')
...
def xml2json(...)
```

setup.py
```diff
@@ -52,6 +55,7 @@
     ]
   },
   entry_points = {
+      'console_scripts': ['pyhf_xml2json=pyhf.commandline:xml2json']
```
lukasheinrich: possibly renamed after comment above

pyhf/commandline.py
lukasheinrich: one global option could be the logging verbosity

```
@click.option('verbosity', default='INFO')
def pyhf(verbosity):
    logging.basicConfig(level=getattr(logging, verbosity), format=LOGFORMAT)
 ```

btw I kinda like using log formats with fixed width sections like, thoughts?

```
LOGFORMAT = '%(asctime)s | %(name)20.20s | %(levelname)6s | %(message)s'
```

File: pyhf/commandline.py (hunk 1 above)
Comment (kratsg) [reply pyhf/commandline.py]: maybe top-level directory?

File: tests/test_scripts.py (hunk 2 above)
Comment (kratsg) [reply tests/test_scripts.py]: Comes from `pytest-console-scripts` (#198) which adds the fixture [(see readme)]( https://github.com/kvas-it/pytest-console-scripts/#usage)

File: pyhf/commandline.py (hunk 3 above)
Comment (kratsg) [reply pyhf/commandline.py]: I thought about that too. I'm fine with doing that as well.

**Comment by kratsg** [comment]
This now spawns three lines of progress bars when running from command line. The first one is overall progress (how many channels left to process), the next line is the number of samples for the given channel, and the last (third) line is the modifiers for the given sample for the given channel.

screenshot 2018-08-23 18 08 41

File: pyhf/readxml.py [hunk 5]
```diff
@@ -11,6 +11,7 @@ def import_root_histogram(rootdir, filename, path, name):
     #import pdb; pdb.set_trace()
     #assert path == ''
     # strip leading slashes as uproot doesn't use "/" for top-level
+    if path is None: path = ''
```
kratsg: This was needed to handle situations where `HistoPath` wasn't included -- and in these cases, it's equivalent to `''`. This code does need to be fixed up more to normalize the XMLs better...

tests/test_scripts.py
lukasheinrich: ah ok click has some built in testing capabilities from `click.testing`. Could use that and avoid the dependency unless `pytest-console-scripts` adds some nice features (haven't used it)

example usage
https://github.com/yadage/yadage/blob/master/tests/test_maincli.py#L5

pyhf/readxml.py
lukasheinrich: maybe 

```
path = path or ''
```

is more pythonic?

pyhf/commandline.py
```diff
@@ -0,0 +1,21 @@
...
+from . import readxml
+
+@click.group(context_settings=dict(help_option_names=['-h', '--help']))
+def pyhf():
+    pass
+
+@pyhf.command()
+@click.option('--entrypoint-xml', required=True, prompt='Top-level XML', help='The top-level XML file for the PDF definition.', type=click.Path(exists=True))
+@click.option('--basedir', required=True, prompt='Base directory', help='The base directory for the XML files to point relative to.', type=click.Path(exists=True))
+@click.option('--output-file', required=True, prompt='Output file', help='The location of the output json file. If not specified, prints to screen.')
```
lukasheinrich: what do you think about making the `entrypoint-xml` be a `click.argument` there is really not way to convert without input, so `pyhf xml2json input.xml` seems to be a good cmd line

`--basedir` could default to `os.getcwd()`

also, maybe it's somewhat more unixy to print to stdout if the output file is not provided?

`pyhf xml2json input.xml > test.json` ?

pyhf/commandline.py
```diff
@@ -0,0 +1,21 @@
...
+@click.option('--tqdm/--no-tqdm', default=True)
```
lukasheinrich: suggestion: `--track/--no-track` or `--track-progress/--no-track-progress`

pyhf/readxml.py
```diff
@@ -26,7 +28,7 @@ def import_root_histogram(rootdir, filename, path, name):
 
     raise KeyError('Both {0:s} and {1:s} were tried and not found in {2:s}'.format(name, os.path.join(path, name), os.path.join(rootdir, filename)))
 
-def process_sample(sample,rootdir,inputfile, histopath, channelname):
+def process_sample(sample,rootdir,inputfile, histopath, channelname, enable_tqdm=False):
     if 'InputFile' in sample.attrib:
```
lukasheinrich: here and in the other cases, I'd also suggest renaming to `track_progress` instead of `enable_tqdm`

pyhf/commandline.py
kratsg: > pyhf xml2json input.xml > test.json ?

The good part is `tqdm` is part of `stderr` so we can definitely do that.

tests/test_scripts.py
kratsg: CliRunner doesn't isolate stdout/stderr. It's probably only specific to running click-enabled commands. The pytest-console-scripts is much more generic (runs any script). I would use CliRunner if I spent more time figuring out stderr extraction.

lukasheinrich: ok yes, testing stdout/stderr separately is important, especially if we want to do e.g. ` > bla.json` where we must ensure that the stdout is json deserializable. let's go with `pytest-consolte-scripts` then.

lukasheinrich:
looks good to me. we have a cli! ✨

Issue #204, opened 2018-08-24
pyhf xml2json|pyhf cls / pyhf cls wspace.json 


lukasheinrich:
# Description

subject to #203 

but this would be nice

```
import pyhf.utils
import json
d = json.load(open('./test.json'))
from pyhf import Model
p = Model({'channels':d['channels']}, poiname='SigXsecOverSM')
result = pyhf.utils.runOnePoint(1.0, d['data']['channel1'] + p.config.auxdata, p)
json.dumps({'CLs_obs': result[-2].tolist()[0], 'CLs_exp': result[-1].ravel().tolist()})
```
returns 

```
'{"CLs_obs": 0.5250899403989365, "CLs_exp": [0.005398954300520573, 0.02413688765551562, 0.0955643537176678, 0.30007471275202563, 0.6452061679174199]}'
```

once workspaces are on hepdata we could do

```
curl http://hepdata.net/ins12345/workspace.json | pyhf cls
```

and to reinterpret for RECAST

```
curl http://hepdata.net/ins12345/workspace.json | pyhf cls --patch newsignal.json
```

PR #205, opened 2018-08-24
add cls command to cli

lukasheinrich:
# Description

addresses #204 

both the pipe and the read from file work

- [x] Tests are passing

### DIFF
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -38,7 +38,7 @@ def cls(workspace,output_file):
     if len(measurements) > 1:
         log.warning('multiple measurements definded. Taking the first %s', measurements[0]['name'])
     p = Model({'channels':d['channels']}, poiname=measurements[0]['config']['poi'])
-    result = runOnePoint(1.0, d['data']['channel1'] + p.config.auxdata, p)
+    result = runOnePoint(1.0, sum((d['data'][c['name']] for c in d['channels']),[]) + p.config.auxdata, p)
     result = {'CLs_obs': result[-2].tolist()[0], 'CLs_exp': result[-1].ravel().tolist()}
     if output_file is None:
         print(json.dumps(result, indent=4, sort_keys=True))
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -7,6 +7,8 @@ import json
 import os
 
 from . import readxml
+from .utils import runOnePoint
+from .pdf import Model
 
 @click.group(context_settings=dict(help_option_names=['-h', '--help']))
 def pyhf():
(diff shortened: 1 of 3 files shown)


pyhf/commandline.py
```diff
@@ -25,3 +27,18 @@ def xml2json(entrypoint_xml, basedir, output_file, track_progress):
...
         log.info("Written to {0:s}".format(output_file))
+
+@pyhf.command()
+@click.argument('workspace', default = '-')
+@click.option('--output-file', help='The location of the output json file. If not specified, prints to screen.', default=None)
+def cls(workspace,output_file):
+    specstream = click.open_file(workspace)
+    d = json.load(specstream)
+    p = Model({'channels':d['channels']}, poiname=d['toplvl']['measurements'][0]['config']['poi'])
+    result = runOnePoint(1.0, d['data']['channel1'] + p.config.auxdata, p)
```
lukasheinrich: this is hardcoded to take the 'first' measurement. in my experience not a lot of analyses actually use the measurement feature all that much..99% probably only have one

pyhf/commandline.py
```diff
@@ -25,3 +27,18 @@ def xml2json(entrypoint_xml, basedir, output_file, track_progress):
     else:
         json.dump(spec, open(output_file, 'w+'), indent=4, sort_keys=True)
         log.info("Written to {0:s}".format(output_file))
+
+@pyhf.command()
+@click.argument('workspace', default = '-')
+@click.option('--output-file', help='The location of the output json file. If not specified, prints to screen.', default=None)
+def cls(workspace,output_file):
+    specstream = click.open_file(workspace)
```
lukasheinrich: not sure how to check for existence in case of non `-` in a click-y way

lukasheinrich:
to make this work #200 must be merged (includes some fixes)

lukasheinrich:
https://asciinema.org/a/lyQ8xsxCfeSxeirSQlNSIoU4G

lukasheinrich:
this also works

```
cat recast.json|pyhf cls workspace.json --patch -
cat recast.json|pyhf cls <(curl -s http://physics.nyu.edu/~lh1132/workspace.json) --patch -
```

pyhf/commandline.py
kratsg: Linux-y would also allow someone to pipe-in the JSON output into this function, no? `pyhf xml2json .... | pyhf cls`

pyhf/commandline.py
kratsg: Should spit a warning if more than one `poiname` detected -- and also do some logging to indicate which one is being used.

pyhf/commandline.py
lukasheinrich: this should already work. my point is that for it I had to remove the `type = click.Path`

lukasheinrich: the reason it works is that click is smart about `-` so wherever it is '-' the stdin (if mode 'r') or stdout (if mode 'w') is written to

kratsg:
This also addresses #207.

PR #206, opened 2018-08-24
Add link to GitHub folder for generated source code

### LINKED ISSUE #165: Add link back to GitHub on webpage
At the moment there is no link on [the docs webpage](http://diana-hep.org/pyhf/) back to the pyhf GitHub page. It would be nice to add one to make navigation between the two easier.

Perhaps a ["Fork me on GitHub" ribbon](https://blog.github.com/2008-12-19-github-ribbons/)?

kratsg:
# Description

This adds an "Edit on GitHub" link on top of generated API pages for the code -- but to the folder/module for the code rather than the file directly (`autosummary` doesn't keep track of the file generating the code at the moment and we're not consistent in class/file names). This should close #165.

- [x] Tests are passing

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -18,7 +18,7 @@ Contents:
 
 .. raw:: html
 
-   <a class="github-fork-ribbon left-bottom fixed" href="https://github.com/diana-hep/pyhf/fork" data-ribbon="Fork me on GitHub" title="Fork me on GitHub">Fork me on GitHub</a>
+   <a class="github-fork-ribbon right-top fixed" href="https://github.com/diana-hep/pyhf/fork" data-ribbon="Fork me on GitHub" title="Fork me on GitHub">Fork me on GitHub</a>
 
 Indices and tables
 ==================


kratsg:
screenshot 2018-08-24 12 32 03

ribbon on bottom left as well

docs/index.rst
```diff
@@ -16,6 +16,10 @@ Contents:
    development
    api
 
+.. raw:: html
+
+   <a class="github-fork-ribbon left-bottom fixed" href="https://github.com/diana-hep/pyhf/fork" data-ribbon="Fork me on GitHub" title="Fork me on GitHub">Fork me on GitHub</a>
```
matthewfeickert: I personally prefer `right-top` over `left-bottom`, but I think this is just a matter of personal preference rather then being right/wrong.
fork_banner_top-right

kratsg: I did too, but it'll overlap with the top `view page source` link on smaller windows.

kratsg: I can just do it since it's what most people expect anyway.

matthewfeickert: Hm, I shrunk down a window and I see what you mean. However, the "[view page source](http://diana-hep.org/pyhf/_sources/index.rst.txt)" link is just to the `index.rst` so I don't think that a huge number of people are going to be clicking that (and the link is still accessible to click from under the banner even if you really shrink down the screen size).

Issue #208, opened 2018-08-24
pyhf loading `histosys` modifier for `normsys` definition


kratsg:
# Description

Filing this to look at in a little while. I've got a local JSON of MBJ from an exclusion fit multibin, Gtt, that is validated against the schema. When inflating the `Model`, I get a crash. 

Clearly, there's some sort of mismatch where `histosys` is being given a `normsys` definition, or we're loading `histosys` for a `normsys` definition. Not sure where this mismatch occurs but will dig into it soon.

# Expected Behavior

`Model` is inflated correctly.

# Actual Behavior

A crash occurs.

```python
>>> import pyhf
>>> import json
>>> d = json.load(file('3b_tag21.2.27-1_RW_ExpSyst_79800_multibin_excl_Gtt_2400_5000_800.json'))
>>> pyhf.utils.validate({'channels': d['channels']}, pyhf.utils.get_default_schema())
>>> try:
...   pdf = pyhf.Model({'channels': d['channels']}, poiname='mu_SIG')
... except:
...   import traceback, sys, code
...   type, value, tb = sys.exc_info()
...   traceback.print_exc()
...   last_frame = lambda tb=tb: last_frame(tb.tb_next) if tb.tb_next else tb
...   frame = last_frame().tb_frame
...   ns = dict(frame.f_globals)
...   ns.update(frame.f_locals)
...   code.interact(local=ns)
... 

Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "/Users/kratsg/pyhf/pyhf/pdf.py", line 106, in __init__
    self.config = _ModelConfig.from_spec(spec,**config_kwargs)
  File "/Users/kratsg/pyhf/pyhf/pdf.py", line 19, in from_spec
    modifier.add_sample(channel, sample, modifier_def)
  File "/Users/kratsg/pyhf/pyhf/modifiers/histosys.py", line 34, in add_sample
    self.at_minus_one.setdefault(channel['name'], {})[sample['name']] = modifier_def['data']['lo_data']
KeyError: 'lo_data'
Python 2.7.14 (default, Mar 22 2018, 15:04:47) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> self
<pyhf.modifiers.histosys.histosys object at 0x119392c90>
>>> modifier_def
[...]

PR #209, opened 2018-08-25
qualify modifier names by modifier type

lukasheinrich:
# Description

Apparently ROOT allows modifiers to reuse names (without sharing) across modifiers. I would suggest that in  `pyhf` we qualify the modifier names as `type/name` which need to be unique

addresses #208 

- [ ] Tests are passing

### DIFF
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -9,18 +9,19 @@ from . import utils
 
 class _ModelConfig(object):
     @classmethod
-    def from_spec(cls,spec,poiname = 'mu'):
+    def from_spec(cls,spec,poiname = 'mu', qualify_names = False):
         # hacky, need to keep track in which order we added the constraints
         # so that we can generate correctly-ordered data
         instance = cls()
         for channel in spec['channels']:
             for sample in channel['samples']:
                 for modifier_def in sample['modifiers']:
-                    modifier_def = copy.deepcopy(modifier_def)
-                    fullname = '{}/{}'.format(modifier_def['type'],modifier_def['name'])
-                    if modifier_def['name'] == poiname:
-                        poiname = fullname
-                    modifier_def['name'] = fullname
+                    if qualify_names:
+                        modifier_def = copy.deepcopy(modifier_def)
+                        fullname = '{}/{}'.format(modifier_def['type'],modifier_def['name'])
+                        if modifier_def['name'] == poiname:
+                            poiname = fullname
+                        modifier_def['name'] = fullname
                     modifier = instance.add_or_get_modifier(channel, sample, modifier_def)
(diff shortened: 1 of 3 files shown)


pyhf/pdf.py
```diff
@@ -15,6 +16,11 @@ def from_spec(cls,spec,poiname = 'mu'):
         for channel in spec['channels']:
             for sample in channel['samples']:
                 for modifier_def in sample['modifiers']:
+                    modifier_def = copy.deepcopy(modifier_def)
```
lukasheinrich: I copy here in order to not changes the original object being passed in by the user

lukasheinrich:
I switched this a bit, such that `pyhf.pdf.Model` has the ability to automatically qualify the modifier names by `type`, but it does not do so automatically. Users should not be punished to have unqieldy `normsys/mu` etc names if they already did the work and named their modifiers such that the names are unique.

pyhf/pdf.py
```diff
@@ -97,13 +106,13 @@ def add_or_get_modifier(self, channel, sample, modifier_def):
...
     def __init__(self, spec, **config_kwargs):
-        self.spec = spec
+        self.spec = copy.deepcopy(spec) #may get modified by config
         self.schema = config_kwargs.get('schema', utils.get_default_schema())
         # run jsonschema validation of input specification against the (provided) schema
         log.info("Validating spec against schema: {0:s}".format(self.schema))
         utils.validate(self.spec, self.schema)
         # build up our representation of the specification
-        self.config = _ModelConfig.from_spec(spec,**config_kwargs)
+        self.config = _ModelConfig.from_spec(self.spec,**config_kwargs)
```
kratsg: I prefer that the `copy.deepcopy()` is on this line. Throw away the `deepcopy`.

lukasheinrich:
+1 on needing tests. Regarding catching and encouraging, I tried suggesting it in the exception that is thrown.

https://github.com/diana-hep/pyhf/pull/209/files#diff-0e8e9106451dbaea56a5ff43a27335edR87

kratsg:
> +1 on needing tests. Regarding catching and encouraging, I tried suggesting it in the exception that is thrown.

Let's do a custom error, rather than a `RuntimeError`, such as `InvalidSharedModifier` or something. This issue only comes up w/ shared modifiers, I think.

lukasheinrich:
it should probably be a `InvalidSpecification` error. This can be checked at the validation stage, but we would have to introduce validation beyond just checking the schema, since this is a semantic and not a structural error

pyhf/pdf.py
lukasheinrich: not sure I follow, do you want the deepcopy removed or moved here?

PR #215, opened 2018-08-26
Add hello world example and check to tests

### LINKED ISSUE #214: Add check for Hello World in README
I just realized that I don't think that we actually have a check anywhere for the ["Hello World" example](https://github.com/diana-hep/pyhf#hello-world) that is in the README. 

```
>>> import pyhf
>>> import pyhf.simplemodels
>>> p = pyhf.simplemodels.hepdata_like(signal_data=[12.,11.], bkg_data=[50.,52.], bkg_uncerts=[3.,7.])
>>> *_, CLs_obs,CLs_exp = pyhf.runOnePoint(1.0, [51, 48] + p.config.auxdata, p)
>>> print('obs: {} exp: {}'.format(CLs_obs, CLs_exp[2]))
obs: [0.05290116] exp: [0.06445521]
```

If something was going to break the README example it would break elsewhere as well, but we wouldn't also realize that we need to update the README example. For example, the above should now be [`pyhf.utils.runOnePoint`](https://github.com/diana-hep/pyhf/blob/75e321eda4d50bbcb6e04b690bdb9880da6857eb/pyhf/utils.py#L117).

Unless there is an alternative suggestion I'll open a PR that will add a notebook to both the CI and to the docs.

matthewfeickert:
Resolves #214 

- [x] Tests are passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ jobs:
       - sudo apt-get -qq install pandoc
       - pip install -U -q -e .[develop,tensorflow,mxnet,torch]
     script:
+    - python -m doctest README.md
     - cd docs && make html && cd -
     - touch docs/_build/html/.nojekyll
     deploy:
--- a/README.md
+++ b/README.md
@@ -11,17 +11,19 @@
 The HistFactory p.d.f. template [[CERN-OPEN-2012-016](https://cds.cern.ch/record/1456844)] is per-se independent of its implementation in ROOT and sometimes, it's useful to be able to run statistical analysis outside
 of ROOT, RooFit, RooStats framework.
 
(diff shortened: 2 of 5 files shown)


matthewfeickert:
Yeah, that's exactly what I say in [this commit](https://github.com/diana-hep/pyhf/pull/215/commits/1775a61c7c9a0f36ea5d224bedba95567ead0bc6).

However, I would prefer to not have Python2 code in the very first examples people see ([for](https://pythonclock.org/) [a few](https://github.com/python/devguide/pull/344) [reasons](https://python3statement.org/)) so I've just switched to explicitly calling `doctest` in the docs CI. Is that okay with people?

lukasheinrich:
nice. thanks @matthewfeickert

Issue #29, opened 2018-01-30
write HistFactory XML + ROOT <-> JSON Importer/exporter


lukasheinrich:
we should be able to run something like `hist2workspace` but such that it dumps a JSON which can be fed to the `pyhf.hfpdf` ctor

lukasheinrich:
I guess the other way around would also be useful. Take JSON, create xml + root

lukasheinrich:
partly adressed by https://github.com/lukasheinrich/pyhf/pull/42

lukasheinrich:
superseded by #115

Issue #217, opened 2018-08-27
ATLAS-CONF-2018-041 is not validated


kratsg:
# Description

This is the bug report to keep track of the validation attempt for [ATLAS-CONF-2018-041](https://atlas.web.cern.ch/Atlas/GROUPS/PHYSICS/CONFNOTES/ATLAS-CONF-2018-041/). The goal is to have `pyhf` calculate the same expected and observed [CLs](https://en.wikipedia.org/wiki/CLs_method_(particle_physics)) that is reported by `HistFitter` -- which was used to perform the statistical fitting in the CONF.

To do so, we are validating a single signal point that the exclusion fit was performed on: `Gtt_2400_5000_800` (gluino mass = 2.4 TeV, neutralino mass = 800 GeV, and stop mass is integrated out by setting to 5 TeV). The workspace name for this is `3b_tag21.2.27-1_RW_ExpSyst_79800_multibin_excl_Gtt_2400_5000_800` and all the necessary XML/ROOT/JSON files are in the branch [pyhf@validation/atlas-conf-2018-041](https://github.com/diana-hep/pyhf/tree/validation/atlas-conf-2018-041/).

From `HistFitter`, we have the following `CLs` from running `run_cls.py`:

```json
{
    "CLs_exp": [
        0.07408094,
        0.1682635,
        0.3514716,
        0.6266292,
        0.8772564
    ],
    "CLs_obs": 0.1938868,
}
```

From `pyhf`, we have the following:

```json
{
    "CLs_exp": [
        0.10241684820963048,
        0.21210550275772358,
        0.4067405266774231,
        0.6746761281349481,
        0.8995358465168168
    ],
    "CLs_obs": 0.27743067969367186
}
```

# Expected Behavior

We expect `pyhf` to match what `HistFitter` reports.

# Actual Behavior

`pyhf` does not match the `HistFitter` `CLs` values.

# Steps to Reproduce

This uses the branch, `run_cls.py`, and `pyhf xml2json`. All of this is used in combination with `hist2workspace` to generate all the necessary files and calculate the `CLs`.

# Checklist

- [x] Run `git fetch` to get the most up to date version of `master`
- [x] Searched through existing Issues to confirm this is not a duplicate issue
- [x] Filled out the Description, Expected Behavior, Actual Behavior, and Steps to Reproduce sections above or have edited/removed them in a way that fully describes the issue

kratsg:
[...]

PR #218, opened 2018-08-28
stub for XML writing

lukasheinrich:
# Description

Partially adresses #29  https://github.com/diana-hep/pyhf/issues/170

This is an incomplete stub to seed code for XML writing.

* `pyhf.writexml` module with XML generation code
* adds `pyhf json2xml command such that`

```
pyhf xml2json config/example.xml|pyhf json2xml > new.xml
pyhf xml2json new.xml
```

goes through without error by producing valid JSON and XML. (though it does not produce the right XML just yet)

- [x] Tests are passing

### DIFF
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -7,6 +7,7 @@ import json
 import os
 
 from . import readxml
+from . import writexml
 from .utils import runOnePoint
 from .pdf import Model
 
@@ -27,6 +28,16 @@ def xml2json(entrypoint_xml, basedir, output_file, track_progress):
     else:
         json.dump(spec, open(output_file, 'w+'), indent=4, sort_keys=True)
         log.info("Written to {0:s}".format(output_file))
+
+@pyhf.command()
+@click.argument('workspace', default = '-')
+@click.argument('xmlfile', default = '-')
+def json2xml(workspace,xmlfile):
+    specstream = click.open_file(workspace)
+    outstream = click.open_file(xmlfile,'w')
+    d = json.load(specstream)
+
+    outstream.write(writexml.writexml(d,'','','').decode('utf-8'))
 
 @pyhf.command()
 @click.argument('workspace', default = '-')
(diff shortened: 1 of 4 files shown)


lukasheinrich:
maybe we can proceed as with the reading and merge the stub in (with the understanding that it dosn't crash / break anything else but is not functional in itself) and add functionality as we go?

@matthewfeickert @kratsg

lukasheinrich:
this will be useful for closure tests / validation as in #217

matthewfeickert:
> maybe we can proceed as with the reading and merge the stub in (with the understanding that it doesn't crash / break anything else but is not functional in itself) and add functionality as we go?

Sounds reasonable.

pyhf/commandline.py
```diff
@@ -28,6 +29,18 @@ def xml2json(entrypoint_xml, basedir, output_file, track_progress):
         json.dump(spec, open(output_file, 'w+'), indent=4, sort_keys=True)
         log.info("Written to {0:s}".format(output_file))
 
+@pyhf.command()
+@click.argument('workspace', default = '-')
+@click.argument('xmlfile', default = '-')
```
kratsg: workspace and xmlfile can't both be stdin. How do you differentiate the two?

pyhf/commandline.py
```diff
@@ -28,6 +29,18 @@ def xml2json(entrypoint_xml, basedir, output_file, track_progress):
...
+@click.option('--specroot', default = click.Path(exists = True))
+@click.option('--dataroot', default = click.Path(exists = True))
```
kratsg: specroot and dataroot should just be config/data/results by default -- no? It's hard-coded for `HistFitter` and the goal is to at least be compatible with that (even though it's on top of HistFactory).

pyhf/commandline.py
lukasheinrich: so but they are both tty streams by default. Click differentiates between the two by the mode you open them with 

'r' -> stdin
'w' -> stdout

pyhf/commandline.py
lukasheinrich: in HistFactory it's usually `data` and `config`

lukasheinrich: `results` is for the resulting workspace root file, normally.

PR #221, opened 2018-08-28
Feature: specify a modifier's pdf type

kratsg:
# Description

This PR attempts to optimize the internal calculations being made by grouping the modifiers into *meta-modifiers* instead to perform the calculations. This is discussed somewhat heavily in #150.

- [ ] Tests are passing

### DIFF
--- a/pyhf/modifiers/__init__.py
+++ b/pyhf/modifiers/__init__.py
@@ -27,7 +27,7 @@ or if the modifier does not have the right structure.
 '''
 def add_to_registry(cls, cls_name=None, constrained=False, shared=False, pdf_type='normal'):
     global registry
-    cls_name = cls_name if cls_name else cls.__name__
+    cls_name = cls_name or cls.__name__
     if cls_name in registry: raise KeyError('The modifier name "{0:s}" is already taken.'.format(cls_name))
     # validate the structure
     validate_modifier_structure(cls, constrained)
--- a/pyhf/modifiers/__init__.py
+++ b/pyhf/modifiers/__init__.py
@@ -15,7 +15,7 @@ def validate_modifier_structure(modifier, constrained):
 
     for method in required_methods + required_constrained_methods*constrained:
         if not hasattr(modifier, method):
-          raise exceptions.InvalidModifier('Expected {0:s} method on {1:s}constrained modifier {2:s}'.format(method, '' if constrained else 'un', modifier.__name__))
+            raise exceptions.InvalidModifier('Expected {0:s} method on {1:s}constrained modifier {2:s}'.format(method, '' if constrained else 'un', modifier.__name__))
     return True
 
 '''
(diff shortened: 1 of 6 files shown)


pyhf/modifiers/__init__.py
```diff
@@ -93,27 +116,28 @@ def modifier(*args, **kwargs):
     name = kwargs.pop('name', None)
     constrained = bool(kwargs.pop('constrained', False))
     shared = bool(kwargs.pop('shared', False))
+    pdf_type = str(kwargs.pop('pdf_type', 'normal'))
```
lukasheinrich: we hardcode the default here as well as in `add_to_registry` , should we pick one or the other/

one way is to have 

`    pdf_type = str(kwargs.pop('pdf_type'))`
in the decorator and 

```
def ..(...pdf_type = None)
pdf_type = pdf_type or 'normal'
```

in the function

pyhf/modifiers/normsys.py
```diff
@@ -31,7 +31,7 @@ def expected_data(self, pars):
 
     def pdf(self, a, alpha):
         tensorlib, _ = get_backend()
-        return tensorlib.normal(a, alpha, tensorlib.astensor([1]))
+        return getattr(tensorlib, self.pdf_type)(a, alpha, tensorlib.astensor([1]))
```
lukasheinrich: here we see that the signature of the pdf_type obviously depends on the pdf used, we can just trust that this is correct and introduce checks later?

pyhf/modifiers/__init__.py
kratsg: `pdf_type = str(kwargs.pop('pdf_type'))` fails if `pdf_type` isn't specified.

pyhf/modifiers/normsys.py
kratsg: Good point, I can always grab the current tensorlib and validate at the time of modifier definition.

lukasheinrich: how woud you validate it? seems like you'd need to analyze the implementation of `def pdf` to check if the signature called matches the signature that's defined

pyhf/modifiers/__init__.py
lukasheinrich: just realized we're doing the same with the kwargs above -- it's not a big issue so feel free to ignore.

lukasheinrich:
the current diff LGTM, should we keep this open until we see an actual improvement in speed through the use of meta-modifiers or do you want to merge it in @kratsg

kratsg:
> should we keep this open until we see an actual improvement in speed through the use of meta-modifiers or do you want to merge it in @kratsg

multiple small diffs is better. I'll rename this PR.

kratsg:
> @kratsg still [WIP]?

My bad.

PR #225, opened 2018-09-01
add Dockerfile

lukasheinrich:
# Description

Adds Dockerfile

- [ ] Tests are passing


matthewfeickert:
> not sure how this interacts with binder (binder should prob not use this Dockerfile, but rather it's buildpacks)

@lukasheinrich I'm not 100% sure on this, but I think that as there is the `binder` directory it will just always look in there, so we should be good.

I'm assuming that the idea of the Dockerfile is the same as [the repo I made a while back](https://github.com/matthewfeickert/pyhf-testing-env-docker/issues/1)? If so, then I'm assuming we should also setup automated builds for it, yes?

matthewfeickert:
@lukasheinrich actually, I take that back. From the [Binder docs](https://mybinder.readthedocs.io/en/latest/using.html#id8):

> With Dockerfiles, a regular Docker build will be performed. **If a Dockerfile is present, all other configuration files will be ignored.**

So we might have to follow up with the Binder team if we want to have the Dockerfile live in this repo.

matthewfeickert:
@lukasheinrich @minrk [says that](https://gitter.im/jupyterhub/binder?at=5b8d663060f9ee7aa4d0ea09)

> Binder directory has precedence.

So as long as we have `binder/` then the Dockerfile in the repo root won't get picked up.

The Dockerfile you have in there right now doesn't install any backends as it is just
```
pip install -e .
```
Is there a reason to keep it lightweight instead of distributing all the backends in the image?

kratsg:
Also, silly question -- but why not have a `docker/` folder as well?

matthewfeickert:
> Also, silly question -- but why not have a `docker/` folder as well?

I don't think that's silly at all. The only reason I could think of is that if you want to have Binder Hub do automated builds (I *think*) you can only have one Dockerfile in the repo (named `Dockerfile` at least) and that it expects it to be in the root of the repo. However, if we instead wanted to have multiple Dockerfiles and have the CI build them and then push them to Docker Hub then I would think it seems like a very good idea to have a directory for that.

kratsg:
> Docker Hub do automated builds (I think) you can only have one Dockerfile in the repo (named Dockerfile at least) and that it expects it to be in the root of the repo.

docker file location

https://docs.docker.com/docker-hub/builds/#create-an-automated-build doesn't seem like it.

matthewfeickert:
Ah yeah. That's a good point. I've always just looked at the branch for my Docker Hub builds and ignored the fact that there was a path as well (oops).

matthewfeickert:
@kratsg Ah, [this is the problem that I was remembering](https://stackoverflow.com/a/51953331/8931942). So unless this has gotten fixed sometime since August 2018 I still don't think there is a way for Docker Hub to *publish* multiple Docker images from an automated build, and that if you want to do so you are limited to one Dockerfile per repo.

minrk:
For a bit of reference: The binder directory mainly exists to avoid conflicts if you already have a Dockerfile or other spec at the top-level of your repo, but that's not what Binder should use. It gives you a place to explicitly communicate with repo2docker about what files should be considered in the event that you have files Binder would find in default locations that you don't want it to use.

matthewfeickert:
> We don't really need multiple images, just the one in the `docker/` location, no?

I again seemed to have misunderstand what you meant. I thought you were suggesting that we have a publish a Docker image for each setup of the environment that a user might want (base, tensorflow, pytorch, mxnet, all) and that we have them all in the built with every push to `master` (I still might do that over in [this repo](https://github.com/matthewfeickert/pyhf-testing-env-docker) and have it watch pyhf's `master`). @kratsg You just want it to not be in the repo root but in `docker/` for clarity or to keep the repo root only for things that really should or [_have to be_](https://github.com/travis-ci/travis-ci/issues/5519) there? Just asking to make sure we're on the same page, but I think the idea is good.

lukasheinrich:
done

matthewfeickert:
@lukasheinrich Can you rebase against `master`? The `push` CI is failing for the same reasons (I think) as #228 was.

lukasheinrich:
rebased but didn't pull on `master` before...

PR #230, opened 2018-09-04
bump coverage

lukasheinrich:
# Description

seems like pytest has a hard dep on coverage (which was pinned due to coveralls, but checking now if that's still relevant)

- [ ] Tests are passing


lukasheinrich:
hm I thought I had rebased #225 -- let me check

PR #232, opened 2018-09-05
Feature: maintain union of specification items as part of the model

kratsg:
# Description

This is useful for #231 and similar where we can access a union of the names of all the various modifiers, samples, and channels involved in configuration of the model (pdf). This PR allows for the following to be possible:

```python
>>> spec = {  ... }
>>> model = pyhf.Model(spec)
>>> model.channels
['firstchannel']
>>> model.samples
['mu', 'bkg1', 'bkg2', 'bkg3']
>>> model.modifiers
['mu', 'stat_firstchannel']
```

- [x] Tests are passing

### DIFF
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -18,9 +18,9 @@ class _ModelConfig(object):
         # so that we can generate correctly-ordered data
         instance = cls()
         for channel in spec['channels']:
-            channels.append(channel)
+            channels.append(channel['name'])
             for sample in channel['samples']:
-                samples.append(sample)
+                samples.append(sample['name'])
                 for modifier_def in sample['modifiers']:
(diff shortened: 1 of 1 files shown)


lukasheinrich:
this LGTM, but I would wait with merging a bit to see how #231 develops so that we understand what we will need specifically (e.g. the `make_cube` function keeps indices on the tuple etc).

kratsg:
> this LGTM, but I would wait with merging a bit to see how #231 develops so that we understand what we will need specifically (e.g. the make_cube function keeps indices in the form of index tuples that point into the cube etc).

Not sure why we wouldn't want this, imo. It gives a nice overview of what sort of structure the underlying model will have without full-blown details. But sure.

PR #235, opened 2018-09-05
Install NumPy through SciPy's dependency on it

### LINKED ISSUE #224: loosen numpy requirements for non-extra installs
# Description

we are pretty restrictive in the numpy version range due to trying to conform to TF's valid range, but TF is only one of the backends. If just installing `pip install pyhf` we should not force users to a speciic range unless we require the APIs

`numpy>=1.14.0` should be enough unless i'm missing something. @kratsg since you changed this last, any reason you see to restrict numpy further?

matthewfeickert:
Resolves #224. See [this SO question](https://stackoverflow.com/questions/52192492/possible-to-have-setuptools-extras-require-override-requirement-from-install-req) for more details.

As `scipy` is an installation requirement of `pyhf` then as `scipy` requires `numpy` the latest release of `numpy` from PyPI will be installed if `pyhf` is installed in the default manner. However, if an optional setup is requested through `extra_require` and that setup has `numpy` constraints those constraints will be implemented.

So

```
pip install .
```

installs the latest `numpy` release

and

```
pip install .[tensorflow]
```

installs `numpy` as constrained by `tensorflow`.

- [x] Tests are passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ python:
   - '3.6'
 install:
 - pip install --upgrade pip
-- pip install -U -q -e .[develop,tensorflow,mxnet,torch]
+- pip install -U -q -e .[tensorflow,torch,mxnet,develop]
+- pip install "numpy<=1.14.5,>=1.14.0"
+- pip freeze | grep numpy
 script:
   - pyflakes pyhf
   - pytest --ignore tests/benchmarks/
@@ -34,14 +36,14 @@ jobs:
     python: '3.6'
     install:
       - pip install --upgrade pip
-      - pip install -U -q -e .[develop,tensorflow,mxnet,torch]
+      - pip install -U -q -e .[tensorflow,torch,mxnet,develop]
     script: pytest --benchmark-sort=mean tests/benchmarks/
   - stage: docs
     python: '3.6'
     install:
       - sudo apt-get update
       - sudo apt-get -qq install pandoc
-      - pip install -U -q -e .[develop,tensorflow,mxnet,torch]
+      - pip install -U -q -e .[tensorflow,torch,mxnet,develop]
     script:
     - python -m doctest README.md
     - cd docs && make html && cd -
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ install:
 - pip install --upgrade pip
 - pip install -U -q -e .[tensorflow,torch,mxnet,develop]
 - pip install "numpy<=1.14.5,>=1.14.0"
-- pip freeze | grep numpy
 script:
   - pyflakes pyhf
   - pytest --ignore tests/benchmarks/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,7 @@ If you have found a bug please report it by filling out the [bug report template
 You can install the development environment (which includes a number of extra) libraries via `pip`:
 
 ```
-pip install -e.[develop]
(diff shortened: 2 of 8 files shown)


matthewfeickert:
This is currently failing the CI, so I need to check to see what I missed and then fix it (or this might not work out as I thought).

kratsg:
Extra spaces need to be added. It needs to match exactly as a string. Doctesting is done by comparing stdouts.

matthewfeickert:
Yes. I pushed just before leaving to catch the bus, so I hadn't checked the logs yet.

matthewfeickert:
Everything is passing for Python 3.6 (:+1:) but Python 2.7 is now failing on the papermill tests

```
RuntimeError: module compiled against API version 0xc but this version of numpy is 0xb
```

kratsg:
> Everything is passing for Python 3.6 (👍) but Python 2.7 is now failing on the papermill tests

Conflicting python package installations? E.G. a fix when I've seen that before is to do `pip install --upgrade numpy`.

matthewfeickert:
> A fix when I've seen that before is to do `pip install --upgrade numpy`

Yes, that's what everyone seems to recommend, but I'm trying to understand _why_ that is. Though that might not be a good use of time. I've cleared all the caches just to make sure this isn't an issue there, and if that still fails then I'll try adding the upgrade thing to the CI (reluctantly).

kratsg:
> Yes, that's what everyone seems to recommend, but I'm trying to understand why that is. Though that might not be a good use of time. I've cleared all the caches just to make sure this isn't an issue there, and if that still fails then I'll try adding the upgrade thing to the CI (reluctantly).

conflicts of versions between `pip install .` and `pip install .[develop,....]`?

matthewfeickert:
> conflicts of versions between `pip install .` and `pip install .[develop,....]`?

I don't think so, as we only ever do

```
pip install -U -q -e .[develop,tensorflow,mxnet,torch]
```
in the CI.

lukasheinrich:
that's pretty clever. if we can get tests to pass

matthewfeickert:
Hm, so Python 2.7 on Travis continues to be a bit confusing as if numpy is upgraded, then it is upgraded from `1.14.5` to `1.15.1`, which then causes the RuntimeError above to be resolved (though now the version is in conflict with the latest accepted for `tensorflow` and `mxnet`) but then the doctest fails. :upside_down_face: I guess this going to involve some more thinking about how Travis has things setup.

matthewfeickert:
Should be ready for review now. Let me know if you have questions.

matthewfeickert:
The above confusion I was having is resolved in PR #237.

lukasheinrich:
this made the docker image work

```
%  cat << 'EOF' | docker run -i pyhf/pyhf pyhf cls
{               
    "channels": [
        {
            "name": "channel1",
            "samples": [
                {"name": "sig", "data": [ 5.0], "modifiers": [{"name": "mu","data": null, "type": "normfactor"}]},
                {"name": "bkg", "data": [50.0], "modifiers": []}
            ]
        }
    ],
    "data": {"channel1": [51.0] },
    "toplvl": {
        "measurements": [
          {"config": {"poi": "mu"}, "name": "HelloWorld"}
        ]
    }
}
EOF
{
    "CLs_exp": [
        0.15948986298223164, 
        0.29000867372064876, 
        0.4934558602497132, 
        0.741297207488922, 
        0.9268610639479048
    ], 
    "CLs_obs": 0.5268665657158867
}
```

thanks @matthewfeickert

PR #237, opened 2018-09-06
Install with --ignore-installed to ensure proper NumPy

matthewfeickert:
If this is not done, then the default Travis installation of `numpy==1.13.3` will be used throughout instead of installing a later requested version unless it is explicitly asked for in `setup.py`'s `install_requires` (which is not done for details explained in PR #235). This allows for pyhf to be installed in Travis CI in the same way that a user or developer would, without having to force any version of NumPy.

This resolves the [confusion that I was having in PR #235](https://github.com/diana-hep/pyhf/pull/235#issuecomment-418910900) as well about the difference in NumPy releases being installed in my Docker containers and in Travis CI. In retrospect this difference should have been an obvious indicator that Travis CI already had a version of NumPy installed, so sorry about missing about.

- [x] Tests are passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,11 @@ cache: pip
 python:
   - '2.7'
   - '3.6'
+before_install:
+- pip install --upgrade pip setuptools
+- pip uninstall -y numpy  # As numpy 1.13.3 is install by default in Travis it will be used unless removed
 install:
-- pip install --upgrade pip
 - pip install -U -q -e .[tensorflow,torch,mxnet,develop]
-- pip install "numpy<=1.14.5,>=1.14.0"
 script:
   - pyflakes pyhf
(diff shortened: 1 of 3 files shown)


kratsg:
I didn't realize it had one pre-installed. Can we just use a `--ignore-installed` instead of explicitly uninstalling?

lukasheinrich:
so this means that by relying on scipy to install numpy we are not guaranteed to get the right numpy since scipy has looser restriction on numpy than we do, correct?

matthewfeickert:
> so this means that by relying on scipy to install numpy we are not guaranteed to get the right numpy since scipy has looser restriction on numpy than we do, correct?

This is correct, if there is already an installed NumPy.

matthewfeickert:
Nevermind, caching works even with `--ignore-installed`.

Issue #246, opened 2018-09-08
add interp code for modifiers (generic labels infrastructure?)


lukasheinrich:
# Description

some modifiers implement their modification as an interpolation between user-provided data (e.g. `histosys` and `normsys`). As these computations are typically expensive we want to vectorize these as much as possible #231 (even across multiple modifiers -- the meta-modifier idea). To do this we need a way to group these computations by e.g. interpolation type. 

comment:
This is not a generic property of modifiers and we are now adding more and more decorators to them (`is_constrained`, `pdf_type`, `interp_code`..) so maybe at some point we might consider a generic `labels` key-value infrastructure instead of adding properties one by one

kratsg:
Given the way #285 is going -- this is going to be slowly irrelevant as we've moved the interpolation into the `paramset` labeling and other properties like `is_constrained` will be going away -- and overall, the modifier decorator is most likely going to disappear.

Issue #219, opened 2018-08-28
set up validation testbench


lukasheinrich:
# Description

## Is your feature request related to a problem? Please describe.

Currently testing / comparing the outputs of `pyhf` and `roohf` (i'll coin that term) is hard and allows manually (or even visually) comparing outputs

### Describe the solution you'd like

A testbench suite that allows for for a given input to 

1. generate pyhf JSON
2. generate roohf XML+ROOT (via #218) and then workspace via `hist2workspace`
3. run limit setting for both
4. compare / print report

something like this that tests both formats (XML/JSON) and both implementations
```
                  .___  JSON __ json2xml -- hist2workspace -- CLs
                  |___  JSON __ pyhf CLs
source of truth --|        
                  |___ XML + ROOT __ hist2workspace __ CLs
                  |___ XML + ROOT __ xml2json __ pyhf CLs
```
# Relevant Issues and Pull Requests

- #217 
- #218

lukasheinrich:
as part of #231 I wrote this bit that might be useful

```
def makespec(nchans,nsamps,nbins,nsysts):
    channels = []
    for cc in range(nchans):
        backgrounds = []
        for ss in range(nsamps):
            mods = []
            for nn in range(nsysts):
                mods.append(
                    {'name': 'syst_{}_{}_{}'.format(cc,ss,nn), 'type': 'shapesys', 'data': [7.]*nbins}
                )
                mods.append(
                    {'name': 'h_syst_{}_{}_{}'.format(cc,ss,nn), 'type': 'histosys', 'data': {'hi_data': [51.]*nbins, 'lo_data': [49.]*nbins}}
                )
                mods.append(
                    {'name': 'n_syst_{}_{}_{}'.format(cc,ss,nn), 'type': 'normsys', 'data': {'hi': 0.95, 'lo': 1.05}}
                )
                mods.append(
                    {'name': 'sf_{}_{}_{}'.format(cc,ss,nn), 'type': 'shapefactor', 'data': None}
                )
            backgrounds.append(
                {'name': 'background_{}_{}'.format(cc,ss),'data': [50.0]*nbins,'modifiers': mods}
            )
        c = {
            'name': 'channel_{}'.format(cc),
            'samples': [
               {'name': 'signal','data': [5.0]*nbins, 'modifiers': [{'name': 'mu', 'type': 'normfactor', 'data': None}]},
            ] + backgrounds
        }
        channels.append(c)
    spec = {'channels': channels}
    return spec
[...]

PR #251, opened 2018-09-09
Feature: tensorize the interpolation codes (more general)

kratsg:
# Description

The overarching theme for this PR is to get the interpolation codes to be tensorizable. This is one of the small pieces needed for #231 and we'll be doing this through a series of smaller PRs to break this up a bit more.

Related: #231, #246, #219.

A notebook is added describing the process of tensorizing these functions to aid with future understanding.

Additionally, I've added a (first draft of a) neat decorator in `utils.py` that tensorizes the inputs to the interpolator functions if they're not tensorized.

- [x] ~add `tile()` to backends~ (done by #262)
- [x] ~add `einsum()` to backends~ (done by #262)
- [x] tensorize the interpolation code
- [x] add/check tests and coverage

- [x] Tests are passing

### DIFF
--- a/pyhf/interpolate.py
+++ b/pyhf/interpolate.py
@@ -3,7 +3,6 @@ log = logging.getLogger(__name__)
 
 from . import get_backend
 from . import exceptions
-from . import utils
 
 def _slow_hfinterp_looper(histogramssets, alphasets, func):
     all_results = []
--- a/pyhf/interpolate.py
+++ b/pyhf/interpolate.py
@@ -25,19 +25,24 @@ def _kitchensink_looper(histogramssets, alphasets, func):
 def _hfinterp_code0(histogramssets, alphasets):
     tensorlib, _ = get_backend()
 
-    allset_all_histo_deltas_up = histogramssets[:,:,2] - histogramssets[:,:,1]
-    allset_all_histo_deltas_dn = histogramssets[:,:,1] - histogramssets[:,:,0]
-    allset_all_histo_nom = histogramssets[:,:,1]
+    # these three variables can be pre-computed at PDF time
+    allset_allhisto_nom = histogramssets[:,:,1]
+    allset_allhisto_deltas_up = histogramssets[:,:,2] - histogramssets[:,:,1]
+    allset_allhisto_deltas_dn = histogramssets[:,:,1] - histogramssets[:,:,0]
 
-    #x is dummy index
-    allsets_all_histos_alphas_times_deltas_up = tensorlib.einsum('sa,sxu->sxau',alphasets,allset_all_histo_deltas_up)
-    allsets_all_histos_alphas_times_deltas_dn = tensorlib.einsum('sa,sxu->sxau',alphasets,allset_all_histo_deltas_dn)
     where_alphasets_positive = tensorlib.where(alphasets > 0, tensorlib.ones(alphasets.shape), tensorlib.zeros(alphasets.shape))
-    allsets_all_histos_masks = tensorlib.einsum('sa,sxu->sxau', where_alphasets_positive, tensorlib.ones(allset_all_histo_deltas_dn.shape))
 
-    allsets_all_histos_deltas = tensorlib.where(allsets_all_histos_masks,allsets_all_histos_alphas_times_deltas_up, allsets_all_histos_alphas_times_deltas_dn)
-    allsets_all_histos_noms_repeated = tensorlib.einsum('sa,sxu->sxau',tensorlib.ones(alphasets.shape),allset_all_histo_nom)
-    set_results = allsets_all_histos_deltas + allsets_all_histos_noms_repeated
+    # s: set under consideration (i.e. the modifier)
+    # a: alpha variation
+    # h: histogram affected by modifier
+    # b: bin of histogram
+    allsets_allhistos_alphas_times_deltas_up = tensorlib.einsum('sa,shb->shab',alphasets,allset_allhisto_deltas_up)
+    allsets_allhistos_alphas_times_deltas_dn = tensorlib.einsum('sa,shb->shab',alphasets,allset_allhisto_deltas_dn)
+    allsets_allhistos_masks = tensorlib.einsum('sa,shb->shab', where_alphasets_positive, tensorlib.ones(allset_allhisto_deltas_dn.shape))
+
+    allsets_allhistos_deltas = tensorlib.where(allsets_allhistos_masks,allsets_allhistos_alphas_times_deltas_up, allsets_allhistos_alphas_times_deltas_dn)
+    allsets_allhistos_noms_repeated = tensorlib.einsum('sa,shb->shab',tensorlib.ones(alphasets.shape),allset_allhisto_nom)
+    set_results = allsets_allhistos_deltas + allsets_allhistos_noms_repeated
     return set_results
 
 def _kitchensink_code0(histogramssets, alphasets):
@@ -55,29 +60,28 @@ def _kitchensink_code0(histogramssets, alphasets):
 @utils.tensorize_args
 def _hfinterp_code1(histogramssets, alphasets):
     tensorlib, _ = get_backend()
-    #warning: alphasets must be ordered
-    at_plus_one, at_zero, at_minus_one = histogramssets[0][0]
 
-    up_variation = tensorlib.divide(at_plus_one, at_zero)
-    down_variation = tensorlib.divide(at_minus_one, at_zero)
+    # these three variables can be pre-computed at PDF time
+    allset_allhisto_nom = histogramssets[:,:,1]
+    allset_allhisto_deltas_up = tensorlib.divide(histogramssets[:,:,2], allset_allhisto_nom)
+    allset_allhisto_deltas_dn = tensorlib.divide(histogramssets[:,:,0], allset_allhisto_nom)
 
-    positive_parameters = alphasets[alphasets >= 0]
-    negative_parameters = alphasets[alphasets < 0]
+    allsets_allhistos_masks = tensorlib.where(alphasets > 0, tensorlib.ones(alphasets.shape), tensorlib.zeros(alphasets.shape))
 
-    bases_negative = tensorlib.tile(down_variation, negative_parameters.shape+(1,)*len(down_variation.shape))
-    bases_negative = tensorlib.einsum('ij->ji', bases_negative)
+    # s: set under consideration (i.e. the modifier)
+    # a: alpha variation
+    # h: histogram affected by modifier
+    # b: bin of histogram
+    bases_up = tensorlib.einsum('sa,shb->shab', tensorlib.ones(alphasets.shape), allset_allhisto_deltas_up)
+    bases_dn = tensorlib.einsum('sa,shb->shab', tensorlib.ones(alphasets.shape), allset_allhisto_deltas_dn)
+    exponents = tensorlib.einsum('sa,shb->shab', tensorlib.abs(alphasets), tensorlib.ones(allset_allhisto_deltas_up.shape))
+    masks = tensorlib.einsum('sa,shb->shab', allsets_allhistos_masks, tensorlib.ones(allset_allhisto_deltas_up.shape))
+    allsets_allhistos_noms_repeated = tensorlib.einsum('sa,shb->shab', tensorlib.ones(alphasets.shape), allset_allhisto_nom)
 
-    bases_positive = tensorlib.tile(up_variation, positive_parameters.shape+(1,)*len(up_variation.shape))
-    bases_positive = tensorlib.einsum('ij->ji', bases_positive)
-
-    expo_positive = tensorlib.tile(positive_parameters, up_variation.shape+(1,)) #was outer
-    expo_negative = -tensorlib.tile(negative_parameters, down_variation.shape+(1,)) #was outer
-
-    res_neg = tensorlib.power(bases_negative, expo_negative)
-    res_pos = tensorlib.power(bases_positive, expo_positive)
-
-    result = tensorlib.concatenate([res_neg, res_pos], axis=-1)
-    return tensorlib.einsum('ijk->ikj', result)
+    bases = tensorlib.where(masks, bases_up, bases_dn)
+    allsets_allhistos_deltas = tensorlib.power(bases, exponents)
+    set_results = allsets_allhistos_deltas * allsets_allhistos_noms_repeated
+    return set_results
 
 def _kitchensink_code1(histogramssets, alphasets):
     def product(down, nom, up, alpha):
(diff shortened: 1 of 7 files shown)


lukasheinrich:
just dumping here what we want to ultimately do:

```
def kitchen_sink(histogramssets,alphasets):
    all_results = []
    for histoset, alphaset in zip(histogramssets,alphasets):
        all_results.append([])
        set_result = all_results[-1]
        for histo in histoset:
            set_result.append([])
            histo_result = set_result[-1]
            for alpha in alphaset:
                alpha_result = []
                for down,nom,up in zip(histo[0],histo[1],histo[2]):
                    delta_up = up - nom
                    delta_down = nom - down
                    if alpha > 0:
                        delta =  delta_up*alpha
                    else:
                        delta =  delta_down*alpha
                    v = nom + delta
                    alpha_result.append(v)
                histo_result.append(alpha_result)
    return all_results


histogramssets = [ 
    [#all histos affected by syst1 
        [#sample 1, syst1
            [10,11,12],
            [15,16,19],
            [19,20,25]
        ],
        [#sample 2, syst1
            [10,11,12],
            [15,16,19],
            [19,20,25]
        ]
    ]
]
alphasets = [
    [-1,0,1] #set of alphas for which to interpolate syst1 histos
]
kitchen_sink(histogramssets,alphasets)
```

matthewfeickert:
As @kratsg has pointed out, MXNet doesn't currently support `einsum` [though there is a PR out on it](https://github.com/apache/incubator-mxnet/issues/10840). Until that PR is in I would suggest that we disable all MXNet support by commenting out or removing the invocations.

Please don't remove the actual files as it would be nice to easily add MXNet back in once support is added by the community.

matthewfeickert:
To get a request/whinge that will come up later in now, can we please rename `_kitchensink_<X>`? That name doesn't really convey any helpful information.

kratsg:
> To be slightly more helpful then just complaining, an example of names (not saying they should be used) would be `_kitchensink_looper` → `_interpolate_systematics`, `_kitchensink_code0` → `_interp_code0_linear`, `_kitchensink_code1` → `_interp_code1_exponential`. To be clear, these are just suggestions that I came up with after scanning the code. A more careful reading may reveal these to be bad suggestions.

Note that the `kitchensink` codes are the poorly optimized versions of the actual interpcodes that I added from @lukasheinrich which are more optimized. The idea is that the slow, loop manually and calculate, is the poor version that will always be the correct calculation -- and we just want the tensorizable calculations to match them (which makes testing significantly easier).

lukasheinrich:
my latest stab at it

these should be equivalent

```
def kitchen_sink(histogramssets,alphasets):
    all_results = []
    for histoset, alphaset in zip(histogramssets,alphasets):
        all_results.append([])
        set_result = all_results[-1]
        for histo in histoset:
            set_result.append([])
            histo_result = set_result[-1]
            for alpha in alphaset:
                alpha_result = []
                for down,nom,up in zip(histo[0],histo[1],histo[2]):
                    delta_up = up - nom
                    delta_down = nom - down
                    if alpha > 0:
                        delta =  delta_up*alpha
                    else:
                        delta =  delta_down*alpha
                    v = nom + delta
                    alpha_result.append(v)
                histo_result.append(alpha_result)
    return all_results

def new_kitchen_sink(histogramssets,alphasets):
    histogramssets = np.asarray(histogramssets)
    alphasets      = np.asarray(alphasets)

    allset_all_histo_deltas_up = histogramssets[:,:,2] - histogramssets[:,:,1]
    allset_all_histo_deltas_dn = histogramssets[:,:,1] - histogramssets[:,:,0]
    allset_all_histo_nom = histogramssets[:,:,1]
    
    allsets_all_histos_alphas_times_deltas_up = np.einsum('sa,s...u->s...au',alphasets,allset_all_histo_deltas_up)
    allsets_all_histos_alphas_times_deltas_dn = np.einsum('sa,s...u->s...au',alphasets,allset_all_histo_deltas_dn)
    allsets_all_histos_masks = np.einsum('sa,s...u->s...au',alphasets > 0,np.ones_like(allset_all_histo_deltas_dn))
    
    allsets_all_histos_deltas = np.where(allsets_all_histos_masks,allsets_all_histos_alphas_times_deltas_up, allsets_all_histos_alphas_times_deltas_dn)
    allsets_all_histos_noms_repeated = np.einsum('sa,shn->shan',np.ones_like(alphasets),allset_all_histo_nom)
    set_results = allsets_all_histos_deltas + allsets_all_histos_noms_repeated
    return set_results
```

[better.zip](https://github.com/diana-hep/pyhf/files/2383512/better.zip)

lukasheinrich:
this is without ellipses

```
def new_kitchen_sink(histogramssets,alphasets):
    histogramssets = np.asarray(histogramssets)
    alphasets      = np.asarray(alphasets)

    allset_all_histo_deltas_up = histogramssets[:,:,2] - histogramssets[:,:,1]
    allset_all_histo_deltas_dn = histogramssets[:,:,1] - histogramssets[:,:,0]
    allset_all_histo_nom = histogramssets[:,:,1]
    
    #x is dummy index
    
    allsets_all_histos_alphas_times_deltas_up = np.einsum('sa,sxu->sxau',alphasets,allset_all_histo_deltas_up)
    allsets_all_histos_alphas_times_deltas_dn = np.einsum('sa,sxu->sxau',alphasets,allset_all_histo_deltas_dn)
    allsets_all_histos_masks = np.einsum('sa,sxu->sxau',alphasets > 0,np.ones_like(allset_all_histo_deltas_dn))
    
    allsets_all_histos_deltas = np.where(allsets_all_histos_masks,allsets_all_histos_alphas_times_deltas_up, allsets_all_histos_alphas_times_deltas_dn)
    allsets_all_histos_noms_repeated = np.einsum('sa,sxu->sxau',np.ones_like(alphasets),allset_all_histo_nom)
    set_results = allsets_all_histos_deltas + allsets_all_histos_noms_repeated
    return set_results
```

```
histogramsets = [
    [
        [
            [0.5],
            [1.0],
            [2.0]
        ]
    ]
]

alphasets = [
    [-2,-1,0,1,2]
]

new_kitchen_sink(histogramsets,alphasets)[0]
```

gives 
```
array([[[0. ],
        [0.5],
        [1. ],
        [2. ],
        [3. ]]])
```

lukasheinrich:
for the record a notebook on how we arrived at this highly vectorized code
[StepByStep.ipynb.zip](https://github.com/diana-hep/pyhf/files/2383804/StepByStep.ipynb.zip)

pyhf/tensor/tensorflow_backend.py
```diff
@@ -157,9 +174,39 @@ def generic_len(a):
             raise error
 
         broadcast = [arg if generic_len(arg) > 1 else
-                     tf.tile(tf.slice(arg, [0], [1]), tf.stack([max_dim])) for arg in args]
+                     self.tile(tf.slice(arg, [0], [1]), tf.stack([max_dim])) for arg in args]
         return broadcast
 
+    def tile(self, A, reps):
```
lukasheinrich: I think we can remove `tile` and thus reduce surface-area we need to sync between backends

pyhf/tensor/tensorflow_backend.py
```diff
@@ -113,8 +119,19 @@ def where(self, mask, tensor_in_1, tensor_in_2):
         tensor_in_2 = self.astensor(tensor_in_2)
         return mask * tensor_in_1 + (1-mask) * tensor_in_2
 
-    def concatenate(self, sequence):
-        return tf.concat(sequence, axis=0)
+    def concatenate(self, sequence, axis=0):
```
lukasheinrich: I don't think we're using the axis anymore, so can be removed until we need it

pyhf/interpolate.py
```diff
@@ -3,38 +3,105 @@
...
+    # b: bin of histogram
+    bases_up = tensorlib.einsum('sa,shb->shab', tensorlib.ones(alphasets.shape), allset_allhisto_deltas_up)
+    bases_dn = tensorlib.einsum('sa,shb->shab', tensorlib.ones(alphasets.shape), allset_allhisto_deltas_dn)
+    exponents = tensorlib.einsum('sa,shb->shab', tensorlib.abs(alphasets), tensorlib.ones(allset_allhisto_deltas_up.shape))
+    masks = tensorlib.einsum('sa,shb->shab', allsets_allhistos_masks, tensorlib.ones(allset_allhisto_deltas_up.shape))
+    allsets_allhistos_noms_repeated = tensorlib.einsum('sa,shb->shab', tensorlib.ones(alphasets.shape), allset_allhisto_nom)
+
+    bases = tensorlib.where(masks, bases_up, bases_dn)
+    allsets_allhistos_deltas = tensorlib.power(bases, exponents)
+    set_results = allsets_allhistos_deltas * allsets_allhistos_noms_repeated
```
lukasheinrich: make sure we remove the multiplcation with nominal at the end

matthewfeickert:
> Note that the kitchensink codes are the poorly optimized versions of the actual interpcodes that I added from @lukasheinrich which are more optimized. The idea is that the slow, loop manually and calculate, is the poor version that will always be the correct calculation -- and we just want the tensorizable calculations to match them (which makes testing significantly easier).

Yup, that was clear from the presence of a `do_optimal` bool. 

My point is that the name "kitchensink" conveys that "this does everything" but that conveys no useful information to what they do. If the idea is that the `_kitchensink<X>` functions are slow checks then why not name them something along the `_slowcheck<X>` instead? Or am I missing some reason why this should be informative naming?

kratsg:
> My point is that the name "kitchensink" conveys that "this does everything" but that conveys no useful information to what they do. If the idea is that the `_kitchensink<X>` functions are slow checks then why not name them something along the `_slowcheck<X>` instead? Or am I missing some reason why this should be informative naming?
> 
> To be clear, this isn't pressing. This is just something I have unnecessarily strong and verbose opinions on (that I'd like to think are loosely held enough to be swayed). :P

Yeah, I'll rename these functions more clearly in a bit.

matthewfeickert:
From taking a quick scan over the notebook it is going to be super helpful. :+1: Super great! :)

lukasheinrich:
@matthewfeickert I agree the naming should be improved. But the new naming should convey that it does indeed "do everything" 

the functions implement the application of multiple modifiers of the same type to all affected samples for a given set of nuisance parameter values. So perhaps something like `applyModfiers` (i.e. instead of looping over modifiers and calling`.apply()` we pass a list of modifiers.

lukasheinrich:
the slow versions are only a naive implementation of it. We could only define it in the tests (or in `utils.py`)

pyhf/modifiers/histosys.py
```diff
@@ -46,7 +46,18 @@ def pdf(self, a, alpha):
...
+        results = interpolator(0)(
+            tensorlib.astensor([
+                [
+                    [
+                        self.at_minus_one[channel['name']][sample['name']],
+                        self.at_zero[channel['name']][sample['name']],
+                        self.at_plus_one[channel['name']][sample['name']]
+                    ]
+                ]
+            ]), tensorlib.astensor([tensorlib.tolist(pars)])
```
lukasheinrich: why to we call `tolist` here?

pyhf/modifiers/normsys.py
```diff
@@ -36,7 +36,18 @@ def pdf(self, a, alpha):
...
+        results = interpolator(1)(
+            tensorlib.astensor([
+                [
+                    [
+                        self.at_minus_one[channel['name']][sample['name']],
+                        self.at_zero,
+                        self.at_plus_one[channel['name']][sample['name']]
+                    ]
+                ]
+            ]), tensorlib.astensor([tensorlib.tolist(pars)])
```
lukasheinrich: dito

pyhf/modifiers/histosys.py
kratsg: This gets factored out soon enough. It's a bandage for this inconsistency:

```
pars = [1.0]
np.array([pars]) == [[1.0]]
tf.Tensor([pars]) == [[1.0]]
torch([pars]) == [1.0]
```

pyhf/modifiers/normsys.py
kratsg: This gets factored out soon enough. It's a bandage for this inconsistency:

```
pars = [1.0]
np.array([pars]) == [[1.0]]
tf.Tensor([pars]) == [[1.0]]
torch([pars]) == [1.0]
```

kratsg:
> > Currently, the tests that fail on pytorch/tensorflow are due to issues with single float precision versus double float precision versus what python does
> 
> what do we do with regards to that^?

There's a couple of ways we can go. 

- relax the approximation checks for tensorflow/pytorch since numpy passes at a stricter level. This works because we already check the consistency of the backends with each other in a different set of tests, so we should be reasonably confident that it will be fine here
- force numpy to use single float precision (this requires casting everything to float32 dtype and making that configurable in the numpy backend)
- force the slow interpolator to do calculations on 32-bit numbers for pytorch/tensorflow and 64-bit for numpy.

The main issue we're seeing is that when we calculate the results of the slow interpolation (in vanilla python) and compare against the backend, we're comparing python's double/arbitrary float precision against the single precision of the pytorch/tensorflow backends.

kratsg:
> > > Currently, the tests that fail on pytorch/tensorflow are due to issues with single float precision versus double float precision versus what python does
> > 
> > 
> > what do we do with regards to that^?
> 
> There's a couple of ways we can go.
> 
> * relax the approximation checks for tensorflow/pytorch since numpy passes at a stricter level. This works because we already check the consistency of the backends with each other in a different set of tests, so we should be reasonably confident that it will be fine here
> * force numpy to use single float precision (this requires casting everything to float32 dtype and making that configurable in the numpy backend)
> * force the slow interpolator to do calculations on 32-bit numbers for pytorch/tensorflow and 64-bit for numpy.
> 
> The main issue we're seeing is that when we calculate the results of the slow interpolation (in vanilla python) and compare against the backend, we're comparing python's double/arbitrary float precision against the single precision of the pytorch/tensorflow backends.

@matthewfeickert @lukasheinrich The last commit here 4e23366 is representing my perspective on this:

- numpy can calculate using double float precision which mirrors how python normally does the calculations
- tensorflow/pytorch can only calculate using single float precision which does not mirror how python normally does the calculations

The goal of the slow interpolation code (effectively python-only loops) is to make sure that the fast interpolation code (tensorized form) returns the same results. Therefore, the slow interpolation code should be reduce to single-float precision in the situation when we only have single-float precision.

This will make the tests pass. It can be an issue we file later for a feature to configure the numpy backend as single-float precision, and that simply means we can pass in `dtype=np.float32` in all the functions for the numpy backend. For now, I chose not to do this, since I think that's not what people will want to expect -- especially because ROOT does the calculations in double-float precision and if we need to validate against it, use NumPy which has double-float precision.

lukasheinrich:
I see @kratsg. so do I understand correctly, that we did not need to loosen any tolerances and forcing the slow implementation to use single precision is sufficient to make the tests pass?

PR #258, opened 2018-09-13
add patch CLI

lukasheinrich:
# Description

This adds a `--patch` flag to `pyhf cls`

```
cat newsignal.json|pyhf cls original.json --patch -
```

```
pyhf cls original.json --patch newsignal.json
```

```
pyhf cls original.json --patch newsignal.json
```

```
cat original.json pyhf cls - --patch newsignal.json
```

- [x] Tests are passing

### DIFF
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -5,11 +5,13 @@ log = logging.getLogger(__name__)
 import click
 import json
 import os
+import jsonpatch
 
 from . import readxml
 from . import writexml
 from .utils import runOnePoint
 from .pdf import Model
+
 
 @click.group(context_settings=dict(help_option_names=['-h', '--help']))
 def pyhf():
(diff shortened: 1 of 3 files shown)


kratsg:
Saw this in your slides! Neat :)

pyhf/commandline.py
```diff
@@ -5,12 +5,14 @@
 import click
 import json
 import os
+import jsonpatch
```
kratsg: We need the additional setup.py requirements for this dependency, no?

lukasheinrich: added. schould be fine. only sub-dependency is jsonpinter which is in turn python-stdlib only

lukasheinrich:
need to add tests

matthewfeickert:
I think once tests are in this will be an awesome addition! :)

tests/test_scripts.py
```diff
@@ -68,3 +68,38 @@ def test_import_and_export(tmpdir, script_runner):
 
     command = 'pyhf json2xml {0:s} --specroot {1:s} --dataroot {1:s}'.format(temp.strpath,str(tmpdir))
     ret = script_runner.run(*shlex.split(command))
+    assert ret.success
```
lukasheinrich: drive-by fix

matthewfeickert:
Seems there is some pesky Python 2.7 failure still though :disappointed:  (he said annoyingly stating the obvious). Python 3 is perfectly happy!

tests/test_scripts.py
matthewfeickert: [pytest is still giving some errors in Python 2.7](https://travis-ci.org/diana-hep/pyhf/jobs/428334799#L510-L512)

```
>       assert ret.success
E       assert False
E        +  where False = <pytest_console_scripts.RunResult object at 0x7f80a9d50e50>.success
```

lukasheinrich: something weird in the way the `pytest` file tmp file objects work in 2.7. They should be file-like but they are not. Changed it to `io.StringIO`

lukasheinrich: see line 91

Issue #259, opened 2018-09-13
Research: Dask integration


lukasheinrich:
# Description

as mentioned in my ATLAS-internal talk, Dask is an interesting option for distributed computation.

Main question: does it hook in at the same level as any other tensor backend, where we replace numpy arrays with dask arrays?

lukasheinrich:
i.e. maybe translation is `np` -> `da`

first issue 
https://github.com/dask/dask/issues/732

lukasheinrich:
minimal example: 

this should be quite quick I believe

```
import dask.array as da
a = da.asarray([1,2,3])
proxy = da.sum(a)
proxy.compute()
```

matthewfeickert:
@lukasheinrich What's the status on this in terms of what still needs to be done/what would you like added on PR #260 (it is still WIP at the moment)? Are there additional things that would be nice to add or test? Or were you wanting for PR #251 to get finished before PR #260 was merged in?

lukasheinrich:
we definitely need to add the new methods in #262. Apart from that we might want to understand Dask a bit better, but it all works, I think.

PR #260, opened 2018-09-13
simple pdf eval works

lukasheinrich:
# Description

Adresses #259 this was really a drop in replacement (except or tensorlib.ones(), where one needed to 
add chunks)

simple logpod evaluation works

```
import pyhf
import pyhf.tensor

import pyhf.tensor.dask_backend
pyhf.tensor.dask_backend.dask_backend()
db = pyhf.tensor.dask_backend.dask_backend()
pyhf.set_backend(db)

import pyhf.simplemodels
pdf = pyhf.simplemodels.hepdata_like(signal_data=[7.], bkg_data=[50.], bkg_uncerts=[7.])


testdata = pdf.expected_data(pdf.config.suggested_init())

v = pdf.logpdf(pdf.config.suggested_init(), testdata)

print(v.compute())
```

- [ ] Tests are passing


pyhf/tensor/dask_backend.py
```diff
@@ -0,0 +1,150 @@
+import dask.array as da
+import logging
+from scipy.special import gammaln
+from scipy.stats import norm
```
lukasheinrich: hm not sure how standard `scipy` interfaces with non-evaluated arrays. somehow it works but not sure why

lukasheinrich:
interestingly some of the basic `tensorlib` tests fail

```
>       assert np.std(values) < 1e-6
E       assert 0.2802208533607308 < 1e-06
E        +  where 0.2802208533607308 = <function std at 0x1155a41e0>([-16.948276294321396, -16.948274612426758, -16.94827651977539, -16.948274612426758, -17.648827643136507])
E        +    where <function std at 0x1155a41e0> = np.std
```

a) good thing we have tests :)
b) probably good to add more tests for each of the tensorlib methods so that we now better which one is failing

pyhf/tensor/dask_backend.py
lukasheinrich: it probably works because dask imlements generalized ufuncs .. 

http://dask.pydata.org/en/latest/array-gufunc.html

matthewfeickert:
@lukasheinrich Can you verify that the updated backend + optimizer table is correct?

matthewfeickert:
I'm going to rebase this against `origin/master` to bring in the work from PR #262

matthewfeickert:
[From Issue #259](https://github.com/diana-hep/pyhf/issues/259#issuecomment-421547714):
>  we definitely need to add the new methods in #262. Apart from that we might want to understand Dask a bit better

@lukasheinrich `abs`, `zeros`, `concatenate`, and `einsum` are copied in now. As you point out, at the moment we are just using Dask as a NumPy clone for the most part. Are there explicit things that you wanted to try to explore for this PR or should we first do more research on Dask's capabilities and then come back to this?

PR #262, opened 2018-09-14
Fix concatenate and add missing functionality to backends:

kratsg:
# Description

Add the following functionality to backends in preparation for tensorizing the interpolation code as in #251 

- einsum
- zeros
- abs

NB: mxnet does not support einsum. In lieu of this, I added a `raise NotImplementedError` when einsum is called.

- [ ] Tests are passing

### DIFF
--- a/tests/test_tensor.py
+++ b/tests/test_tensor.py
@@ -61,14 +61,23 @@ def test_einsum():
     backends = [numpy_backend(poisson_from_normal=True),
                 pytorch_backend(),
                 tensorflow_backend(session=tf_sess),
-                # mxnet_backend() #no einsum in mxnet
+                mxnet_backend() #no einsum in mxnet
                 ]
-    for b in backends:
+
+
+
+    for b in backends[:-1]:
         pyhf.set_backend(b)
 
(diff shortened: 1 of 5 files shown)


lukasheinrich:
we should add some tests for `einsum` and axis-aware concat

matthewfeickert:
Yeah, I agree. This all looks really nice (:+1:), so once tests are in it should be ready to :shipit:.

tests/test_tensor.py
```diff
@@ -54,6 +54,7 @@ def test_common_tensor_backends():
             == [[1, 1, 1], [2, 3, 4], [5, 6, 7]]
         assert tb.tolist(tb.ones((4,5)))  == [[1.]*5]*4
         assert tb.tolist(tb.zeros((4,5))) == [[0.]*5]*4
+        assert tb.tolist(tb.abs(tb.astensor([-1,-2]))) == [1,2]
```
kratsg: no need tb.astensor() here. you put it in the tb.abs() call.

lukasheinrich:
ok, i'll ignore the .1% decreasein coverage and merge

Issue #238, opened 2018-09-07
CuPy tensor backend


lukasheinrich:
# Description

following the discussion in #231 it should be quite easy to get a CuPy backend into pyhf

https://cupy.chainer.org/

it's advertised as a drop-in replacement for numpy. This will give us GPU acceleration (like TF and PyTorch, but separate from autodifferentiation -- which might be nice to have)

@matthewfeickert has some experience in writing these backends and iirc access to a CUDA enabled machine.

matthewfeickert:
Sweet! That looks really cool. I'll poke at their docs over the weekend and then we can see if it is any more work to add in then normal.

matthewfeickert:
This is looking promising, given this section of the docs:

[How to write CPU/GPU agnostic code](https://docs-cupy.chainer.org/en/stable/tutorial/basic.html#how-to-write-cpu-gpu-agnostic-code)

lukasheinrich:
maybe relevant https://github.com/cupy/cupy/issues/1196

matthewfeickert:
Ah, so I'll need to set things up at my cluster to actually do this, as the [requirements to even install include CUDA](https://docs-cupy.chainer.org/en/v5.0.0b3/install.html#requirements). So this isn't just CUDA enabled, but actually built using CUDA (this actually sounds like a good thing to me).

matthewfeickert:
As an update, I'm still working on this, but I'm waiting to hear back on the HPC admins on my cluster. For technical reasons they might make me submit a list of all software that I want and they will setup the testing environment for me. This could be nice, but might result in a few days delay.

lukasheinrich:
ones we get a installation of CuPy it would be *very* interesting to get a feel of what the hardware speed up looks like. Even before implementing the backend we could run this notebook

[BetterKitchenSink.ipynb.zip](https://github.com/diana-hep/pyhf/files/2384771/BetterKitchenSink.ipynb.zip)

probably only the `tb = np` lines need to be adapted if I understood the link 

https://docs-cupy.chainer.org/en/stable/tutorial/basic.html#how-to-write-cpu-gpu-agnostic-code

correctly

```
numpy: 127 µs ± 12.2 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
naive: 56.1 ms ± 4.55 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

matthewfeickert:
[...]

Issue #79, opened 2018-02-09
autograd for numpy backend?


cranmer:
https://github.com/HIPS/autograd

cranmer:
FYI: @matthewfeickert @kratsg

matthewfeickert:
I am in the process of finishing #86, but once that is done I will start on this if @kratsg doesn't beat me to it.

kratsg:
you probably will beat me to it. @cranmer do you know if they implement runge-kutta or something else? It looked like runge-kutta from a first glance.

cranmer:
That tools is just providing the autodiff as far as I know. They have some examples that do various things with those derivatives, but I don't think they are focusing on providing minimization. They do have adam and rmsprop:

https://github.com/HIPS/autograd/blob/master/autograd/misc/optimizers.py

Probably would be best to do Newton's method to be comparable to the other backends, just use autograd for the derivatives

matthewfeickert:
Given that a contributor to autograd has [suggested that people check out JAX](https://github.com/HIPS/autograd/issues/46#issuecomment-475910293) I think it is safe to say that this Issue can be closed as Issue #416 will cover it.

PR #268, opened 2018-09-17
[WIP] Use distributions in ML backends for better Poisson

### LINKED ISSUE #267: Find better Poisson approximation in ML backends
# Description

As I forgot in Issue #266, the different backends have different approximations of the the Poisson distribution. In the [NumPy backend](https://github.com/diana-hep/pyhf/blob/725aa63e091aa59f2fb260580c3b0fb72abb60e4/pyhf/tensor/numpy_backend.py#L159-L163) a Gamma function based approximation is used. However, in the [TensorFlow](https://github.com/diana-hep/pyhf/blob/725aa63e091aa59f2fb260580c3b0fb72abb60e4/pyhf/tensor/tensorflow_backend.py#L198-L200), [PyTorch](https://github.com/diana-hep/pyhf/blob/725aa63e091aa59f2fb260580c3b0fb72abb60e4/pyhf/tensor/pytorch_backend.py#L171-L172), and [MXNet](https://github.com/diana-hep/pyhf/blob/725aa63e091aa59f2fb260580c3b0fb72abb60e4/pyhf/tensor/mxnet_backend.py#L344-L358) backends a Normal function approximation is used.

It would be good if possible to try and have a better approximation than the Normal approximation that is more consistent across backends. I forgot this, and so it won't be obvious to new users.

matthewfeickert:
# Description

Resolves #267 

- [ ] Tests are passing

### DIFF
--- a/pyhf/tensor/tensorflow_backend.py
+++ b/pyhf/tensor/tensorflow_backend.py
@@ -1,10 +1,13 @@
 import logging
 import tensorflow as tf
+import tensorflow_probability as tfp
 
 log = logging.getLogger(__name__)
 
 
 class tensorflow_backend(object):
+    """TensorFlow backend for pyhf"""
+
     def __init__(self, **kwargs):
         self.session = kwargs.get('session')
 
@@ -18,10 +21,10 @@ class tensorflow_backend(object):
             >>> import tensorflow as tf
             >>> sess = tf.Session()
             ...
-            >>> pyhf.set_backend(pyhf.tensor.tensorflow_backend())
+            >>> pyhf.set_backend(pyhf.tensor.tensorflow_backend(session=sess))
             >>> a = pyhf.tensorlib.astensor([-2, -1, 0, 1, 2])
             >>> with sess.as_default():
-            ...   pyhf.tensorlib.clip(a, -1, 1).eval()
+            ...   sess.run(pyhf.tensorlib.clip(a, -1, 1))
             ...
             array([-1., -1.,  0.,  1.,  1.], dtype=float32)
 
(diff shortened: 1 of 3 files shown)


lukasheinrich:
@matthewfeickert something similar for pytorch?

matthewfeickert:
> something similar for pytorch?

I'm doing them all. :) Just after dinner.

lukasheinrich:
awesome! ✨

lukasheinrich:
I think i'd be fine with that. Let's cut is as close to the vest as possible with a tol of 5e-5? @kratsg and I are wrestling with a similar issue in the interp code. At some point we migh need to figure out how to really compare this properly (by e.g. forcing single precision in numpy?)

matthewfeickert:
@lukasheinrich @kratsg Given that this PR removes the need for using the Normal approximation of the Poisson distribution, is there any reason to keep the 

```python
self.pois_from_norm = kwargs.get('poisson_from_normal',False)
```

in the [NumPy backend](https://github.com/diana-hep/pyhf/blob/725aa63e091aa59f2fb260580c3b0fb72abb60e4/pyhf/tensor/numpy_backend.py#L8-L9) anymore? Or can I remove that everywhere?

kratsg:
> Or can I remove that everywhere?

Remove. The only reason it was there was to ameliorate differences and make it easier to compare cross-backend.

matthewfeickert:
I seem to have broken something in the CI. The test suite all passes and then [it core dumps](https://travis-ci.org/diana-hep/pyhf/jobs/429780599#L1166-L1168). :cry: 

Things all pass fine on my local machine with Python 3.6.6, so I'll need to follow up on this. I'm guessing it has something to do with the huge numbers of `divide by zero encountered in log` in the NumPy backend.

lukasheinrich:
yeah +1 on removing the normal approx

matthewfeickert:
@kratsg @lukasheinrich If you have any thoughts on this `corrupted size vs. prev_size` and core dump issue that Travis is having, but that I can't reproduce locally I'd be very interested in hearing them.

kratsg:
> @kratsg @lukasheinrich If you have any thoughts on this `corrupted size vs. prev_size` and core dump issue that Travis is having, but that I can't reproduce locally I'd be very interested in hearing them.

Running out of memory. You're allocated some amount of memory to run python and you've exceeded it.

matthewfeickert:
> Running out of memory

Yeah, that part I understood. Sorry, I should have been more explicit. Do you have any ideas on _why_ changing this approximation seems to have caused (what I'm assuming is) NumPy to cause the memory to grow out of the allowed limit?

lukasheinrich:
@matthewfeickert can you try adding each backend one by one? it might be only a single one that is misbehaving

matthewfeickert:
For reasons not fully understood, using [TensorFlow Probability](https://www.tensorflow.org/probability/) causes Travis CI to use too much memory and a core dump occurs.

So until this is understood all use of tfp should be reverted to a combination of [`tf.distributions`](https://www.tensorflow.org/api_docs/python/tf/distributions) and `tf.contrib.distributions` (for the [Poisson](https://www.tensorflow.org/api_docs/python/tf/contrib/distributions/Poisson)).

This is really annoying, but at least a temporary fix that doesn't break all functionality and gains. However, even with this still enough memory is being eaten up that I needed to break the notebook tests off into their own parallel tests in the `test` stage of the CI.

For additional reference on this issue c.f.
- pytest-dev/pytest#3527
- travis-ci/travis-ci#9153

matthewfeickert:
@lukasheinrich @kratsg With the exception of the coverage, everything is passing now.

This is a large and strange PR, sorry about that, and it is not perfect in that there will need to be additional fixes in the future to actually get `tfp` in and working. It also strongly motivates me making time to do more research on Issue #79 and how that would play with Issue #238. However, it does give us the ability to get rid of the the Normal approximation for the NumPy backend.

Let me know your thoughts and if you want additional changes.

matthewfeickert:
This may want to wait until PR #251 is in to make the rebase easier then this one going first. That being said, if there are no changes it would be nice to have this for tomorrow's talk.

matthewfeickert:
Okay, now that PR #251 (huzzah! :D) is in then I can resolve conflicts here and then if there are no other PRs to go in before then this will be ready for a review.

If people don't mind I'd like to merge this in once it is ready to go so that I can write a somewhat detailed commit log.

matthewfeickert:
> The diff looks a bit busy, (not sure if the file opening contexts are a result of debugging travis) but feel free to merge as you see fit

Yeah, that eliminates some I/O warnings that arise in the tests. This came up when I was trying to debug memory issues in Travis, as you were saying. I don't think they are **required** for this PR, but in general I think anything we can do to keep the Travis log as clean as possible is probably a good idea all around. 

If people don't mind them but don't want it in this PR we can move them to another one later.

matthewfeickert:
Rebased against master to bring in the changes from PR #251

matthewfeickert:
@lukasheinrich @kratsg Given that I cherry-picked in 36e9b70b288e955002669a0466cf48a3748e663f to get the tests [working with `uproot` `v3.0.0` and `uproot-methods` `v0.0.8`](https://github.com/diana-hep/pyhf/pull/276#issuecomment-422980772) do you want me to revert that and wait for a commit to master that has `uproot<3`? Or should I keep this and then moving forward we implement `uproot>3.0.0` along with other guards on our core dependencies?

tests/test_import.py
```diff
@@ -1,12 +1,15 @@
 import pyhf
+from pyhf.tensor.numpy_backend import numpy_backend
 import pyhf.readxml
 import json
 import pytest
 import numpy as np
 
+
 def test_import_prepHistFactory():
+    pyhf.set_backend(numpy_backend())
```
matthewfeickert: @kratsg Given what we discussed in chat RE: a03edbdfc9d8308c42e82b12e53d3bc450441be5 do you want me to keep these in? Or do you want me to wait until you add in something that universally protects against this and then rebase against that and remove these `pyhf.set_backend(numpy_backend())` calls?

tests/test_import.py
```diff
@@ -48,9 +52,11 @@ def test_import_prepHistFactory():
     assert pdf.expected_data(
         pars, include_auxdata=False).tolist() == [140, 120]
 
+
 def test_import_histosys():
+    pyhf.set_backend(numpy_backend())
```
matthewfeickert: @kratsg same as the above.

matthewfeickert:
Rebased against master to prepare to merge with PR #280

matthewfeickert:
Thanks for the support on getting this in @kratsg. :+1:

PR #269, opened 2018-09-17
precomputes all modifications at once

lukasheinrich:
# Description

This PR refactors the pdf to compute all modifications needed in one go. This should be possible independent of #231 #251 but help with the overall refactoring by collecting all necessary computations in a single function that can then be optimized

- [ ] Tests are passing

### DIFF
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -233,11 +233,17 @@ class Model(object):
 
         for channel in self.spec['channels']:
             for sample in channel['samples']:
-                #sum of lists is concat
+                #concatenate list of lists using sum() and an initial value of []
+                #to get a list of all prefactors that should be multiplied to the 
+                #base histogram (the deltas below + the nominal)
                 factors = sum([
                     all_results.get(x,{}).get(channel['name'],{}).get(sample['name'],[])
                     for x in factor_mods
                 ],[]) 
+
+                #concatenate list of lists using sum() and an initial value of []
+                #to get a list of all deltas that should be addded to the 
+                #nominal values
                 deltas  = sum([
                     all_results.get(x,{}).get(channel['name'],{}).get(sample['name'],[])
                     for x in delta_mods
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -154,10 +154,14 @@ class Model(object):
     def _expected_sample(self, nominal, factors, deltas):
         tensorlib, _ = get_backend()
 
+
+        all_deltas = tensorlib.sum(tensorlib.stack(deltas), axis=0)
+        nominal_plus_deltas = tensorlib.stack(
+                    [nominal,all_deltas]
+        )
         basefactor = [
             tensorlib.sum(
-                tensorlib.stack(
-                    [nominal,tensorlib.sum(tensorlib.stack(deltas), axis=0)]),
+                nominal_plus_deltas,
                 axis=0)
                 if len(deltas) > 0
                 else nominal
@@ -177,11 +181,6 @@ class Model(object):
         total_factors = tensorlib.product(stacked_factors_binwise, axis=0)
         return total_factors
 
-        # return tensorlib.product(
-        #             tensorlib.stack(
-        #                   tensorlib.simple_broadcast(*factors)
-        #             ),
-        #        axis=0)
 
         #the base value for each bin is either the nominal with deltas applied
         #if there were any otherwise just the nominal
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -157,15 +157,13 @@ class Model(object):
         nominal_plus_deltas = None
         if len(deltas):
             all_deltas = tensorlib.sum(tensorlib.stack(deltas), axis=0)
-            nominal_plus_deltas = tensorlib.stack(
-                        [nominal,all_deltas]
-            )
+            nominal_and_deltas  = tensorlib.stack([nominal,all_deltas])
+            nominal_plus_deltas = tensorlib.sum(nominal_and_deltas,axis=0)
+
         basefactor = [
-            tensorlib.sum(
-                nominal_plus_deltas,
-                axis=0)
-                if len(deltas) > 0
-                else nominal
+            nominal_plus_deltas
+            if len(deltas) > 0
+            else nominal
         ]
         factors += basefactor
 
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -153,18 +153,23 @@ class Model(object):
 
     def _expected_sample(self, nominal, factors, deltas):
         tensorlib, _ = get_backend()
-
-        nominal_plus_deltas = None
+        
+        #the base value for each bin is either the nominal with deltas applied
+        #if there were any otherwise just the nominal
         if len(deltas):
+            #stack all bin-wise shifts in the yield value (delta) from the modifiers
+            #on top of each other and sum through the first axis
+            #will give us a overall shift to apply to the nominal histo
             all_deltas = tensorlib.sum(tensorlib.stack(deltas), axis=0)
+
+            #stack nominal and deltas and sum through first axis again
+            #to arrive at yield value after deltas (but before factor mods)
             nominal_and_deltas  = tensorlib.stack([nominal,all_deltas])
             nominal_plus_deltas = tensorlib.sum(nominal_and_deltas,axis=0)
+            basefactor = [nominal_plus_deltas]
+        else:
+            basefactor = [nominal]
 
-        basefactor = [
-            nominal_plus_deltas
-            if len(deltas) > 0
-            else nominal
-        ]
         factors += basefactor
 
         #multiplicative modifiers are either a single float that should be broadcast
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -185,36 +185,6 @@ class Model(object):
         total_factors = tensorlib.product(stacked_factors_binwise, axis=0)
         return total_factors
 
-
-        #the base value for each bin is either the nominal with deltas applied
-        #if there were any otherwise just the nominal
-        if len(deltas):
-            #stack all bin-wise shifts in the yield value (delta) from the modifiers
-            #on top of each other and sum through the first axis
-            #will give us a overall shift to apply to the nominal histo
-            overall_delta_from_modifiers = tensorlib.sum(tensorlib.stack(deltas), axis=0)
-
-            #stack nominal and deltas and sum through first axis again
-            #to arrive at yield value after deltas (but before factor mods)
-            nominal_and_deltas  = tensorlib.stack([nominal,overall_delta_from_modifiers])
-            basefactor = tensorlib.sum(nominal_and_deltas,axis=0)
-        else:
-            basefactor = nominal
-        factors += [basefactor]
-
-        #multiplicative modifiers are either a single float that should be broadcast
-        #to all bins or a list of floats (one for each bin of the histogram)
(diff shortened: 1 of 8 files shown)


lukasheinrich:
this is passing tests. the decrease in coverage is because this refactoring exposed that the 

```
pdf.expected_sample(...)
```

method is not tested separately (it was only traversed as part of the other tests). so it should

1. either be removed
2. as part of the public API be tested explicitly (i'll open a issue for that)

maybe you guys can already have a look

lukasheinrich:
to summarize this PR refactors the `expected_actualdata` method such that all modifications are computed in one go 

* instead of computing the modificatio within `expected_sample` we pre-compute it into a structure `all_modifications` (the expensive part)
* the internal `_expected_sample` really only implements how these modifications and the nominal data get amalgated into a singe sample histogram


* the `all_modifications` computation is performed by modifier type `_mtype_results` which will be useful in follow-up PRs where for some types we can replace the naive loops in `_mtype_results` with a fully vectorized calculation

pyhf/pdf.py
```diff
@@ -30,6 +31,7 @@ def from_spec(cls,spec,poiname = 'mu', qualify_names = False):
                     modifier = instance.add_or_get_modifier(channel, sample, modifier_def)
                     modifier.add_sample(channel, sample, modifier_def)
                     modifiers.append(modifier_def['name'])
+                    sample['modifiers_by_type'].setdefault(modifier_def['type'],[]).append(modifier_def['name'])
         instance.channels = list(set(channels))
```
lukasheinrich: we need to book-keep a list of modifiers by type, so that we can loop over them on a type-by-type bais

pyhf/pdf.py
```diff
@@ -126,7 +128,44 @@ def __init__(self, spec, **config_kwargs):
...
+                for mname in sample['modifiers_by_type'].get(mtype,[]):
+                    modifier = self.config.modifier(mname)
+                    modpars  = pars[self.config.par_slice(mname)]
+                    mtype_results.setdefault(channel['name'],
+                            {}).setdefault(sample['name'],
+                            []).append(
+                                modifier.apply(channel, sample, modpars)
+                            )
+        return mtype_results
+
```
lukasheinrich: This method implements the computation off the modifier application for a given modifier type and within that type for each channel and sample..

in a follow up PR it will be further refactored to reverse the order of the three loops, such that the outer-most loop is over modifiers (which is the structure we are aiming for in #251

pyhf/pdf.py
```diff
@@ -126,7 +128,44 @@ def __init__(self, spec, **config_kwargs):
...
     def expected_sample(self, channel, sample, pars):
+        tensorlib, _ = get_backend()
+        #public API only, not efficient
+        all_modifications = self._all_modifications(pars)
+        return self._expected_sample(
+            tensorlib.astensor(sample['data']), #nominal
+            *all_modifications[channel['name']][sample['name']] #mods
+        )
+
```
lukasheinrich: the comment reflects the fact that we do not expect this to be fast. we compute *all* modification for all samples in this method even though we are only interested in the modifications in a single sample. Alternatively the `_all_modifications` could take a list of channel/samples for which it should compute mods

pyhf/pdf.py
```diff
@@ -126,7 +128,44 @@ def __init__(self, spec, **config_kwargs):
...
+                tensorlib.stack(
+                    [nominal,tensorlib.sum(tensorlib.stack(deltas), axis=0)]),
+                axis=0)
+                if len(deltas) > 0 
+                else nominal
+        ]
+        factors += basefactor
+
+        return tensorlib.product(tensorlib.stack(tensorlib.simple_broadcast(*factors)), axis=0)
+
```
lukasheinrich: this just implements how the modifications get combined with the nominal data. Up to now we have 2 types of modifications factors, and shifts (deltas) 

first we sum all deltas with the nominal and then apply all factors on that sum

pyhf/pdf.py
```diff
@@ -181,36 +220,33 @@ def expected_sample(self, channel, sample, pars):
...
+                deltas  = sum([
+                    all_results.get(x,{}).get(channel['name'],{}).get(sample['name'],[])
+                    for x in delta_mods
+                ],[])
+                
+                all_modifications.setdefault(
+                    channel['name'],{})[
+                    sample['name']
+                ] = (factors, deltas)
+        return all_modifications
```
lukasheinrich: this function implements the calculation of *all* modifications by looping over all possible modifications and calling `_mtype_results ` for each one. We then acumulate the results in a nested dict-like structure to keep track of factors/deltas that is then used by `expected_actualdata`/`expected_sample`

lukasheinrich: the names / roles of the modifiers are hardcoded for now, if we can get more expressive labels / attributes on the modifiers that signify what type of modifier they are (factor / delta) this hard-coding could be removed

pyhf/pdf.py
```diff
@@ -231,8 +267,17 @@ def expected_actualdata(self, pars):
...
         for channel in self.spec['channels']:
-            data.append(tensorlib.sum(tensorlib.stack([self.expected_sample(channel, sample, pars) for sample in channel['samples']]),axis=0))
+            sample_stack = [
+                self._expected_sample(
+                    tensorlib.astensor(sample['data']), #nominal
+                    *all_modifications[channel['name']][sample['name']] #mods
+                )
+                for sample in channel['samples']
+            ]
+            data.append(tensorlib.sum(tensorlib.stack(sample_stack),axis=0))
```
lukasheinrich: this diff looks bigger than it is, it's just unpacking the busy one-liner and changing the call signature of `_expected_sample`

lukasheinrich:
rebased

lukasheinrich:
this is ready for review I think

lukasheinrich:
the next PR on top of this is then something along the lines of #273

pyhf/pdf.py
```diff
@@ -30,6 +31,7 @@ def from_spec(cls,spec,poiname = 'mu', qualify_names = False):
                     modifier = instance.add_or_get_modifier(channel, sample, modifier_def)
                     modifier.add_sample(channel, sample, modifier_def)
                     modifiers.append(modifier_def['name'])
+                    sample['modifiers_by_type'].setdefault(modifier_def['type'],[]).append(modifier_def['name'])
```
kratsg: How is this affected by `qualify_names`?

pyhf/pdf.py
```diff
@@ -181,36 +220,33 @@ def expected_sample(self, channel, sample, pars):
...
-          tensorlib.sum(tensorlib.stack(results['histosys']), axis=0)
-        ]), axis=0) if len(results['histosys']) > 0 else nominal]
 
-        return tensorlib.product(tensorlib.stack(tensorlib.simple_broadcast(*factors)), axis=0)
+        all_modifications = {}
+        factor_mods = ['normfactor','normsys','shapesys','shapefactor','staterror']
+        delta_mods  = ['histosys']
+
+        all_results = {}
+        for mtype in factor_mods + delta_mods:
```
kratsg: `for mtype in pyhf.modifiers.__all__` or something similar.

pyhf/pdf.py
```diff
@@ -181,36 +220,33 @@ def expected_sample(self, channel, sample, pars):
...
+            all_results[mtype] = self._mtype_results(mtype,pars)
+
+        for channel in self.spec['channels']:
+            for sample in channel['samples']:
+                #sum of lists is concat
+                factors = sum([
```
kratsg: is this supposed to be multiplication and not summation?

matthewfeickert:
Reviewing this quickly I think that I agree with everything that @kratsg put down. So if those requests get resolved then I think this looks great. I'll review this again in more depth, but thanks very much for taking time to put in review comments in advance @lukasheinrich &mdash; they were quite nice.

pyhf/pdf.py
lukasheinrich: once we decide on a new name there is no difference viz. a spec that already had unique names (not we update the modifer_def structure `modifier_def['name'] = fullname`)

pyhf/pdf.py
lukasheinrich: the refactored code will look like this

```
    def _mtype_results(self,mtype,pars):
        mtype_results = {}
        if mtype in self.combined_mods.keys():
            return self.combined_mods[mtype].apply(pars)
        for channel in self.spec['channels']:
            for sample in channel['samples']:
                for mname in sample['modifiers_by_type'].get(mtype,[]):
                    modifier = self.config.modifier(mname)
                    modpars  = pars[self.config.par_slice(mname)]
                    mtype_results.setdefault(channel['name'],
                            {}).setdefault(sample['name'],
                            []).append(
                                modifier.apply(channel, sample, modpars)
                            )
        return mtype_results
```

pyhf/pdf.py
lukasheinrich: what this does is  compute

```
summands = [delta1,delta2,... nominal]
basefactor = np.sum(summands)
factors = [f1,f2,f3...,basefactor]
np.product(factors)

but in a follow up PR this entire function will be removed and `expected_actualdata` will look like this


```
        all_modifications = self._all_modifications(pars)
        delta_field  = np.zeros(self.cube.shape)
        factor_field = np.ones(self.cube.shape)
        for cname,smods in all_modifications.items():
            for sname,(factors,deltas) in smods.items():
                ind = self.hm[cname][sname]['index']
                for f in factors:
                    factor_field[ind] = factor_field[ind] * f
                for d in deltas:
                    delta_field[ind]  = delta_field[ind] + d
        combined = factor_field * (delta_field + self.cube)
```

pyhf/pdf.py
kratsg: `qualify_names` was needed because we had conflicts between modifiers of different types having the same names. I guess when we fix this, then `qualify_names` is actually not needed anymore, since we were trying to fix what we thought was a bug, but is apparently a feature?

pyhf/pdf.py
lukasheinrich: this is a concatenation of lists .. sum([list1,list3,...],[]) concatenates

pyhf/pdf.py
kratsg: Isn't this before refactoring? It's what you have right now in this commit.

pyhf/pdf.py
kratsg: Hrmm, the way the code is currently written is verrrrry opaque to me. It looks like you sum up a bunch of summands, and then multiply those summand results together. Especially because you do `factors += basefactor`. Can you do sum and products separately, and return the combination of them rather than doing internal bookkeeping?

pyhf/pdf.py
kratsg: That's not very clear from the code, or the comment. But ok.

pyhf/pdf.py
lukasheinrich: check the `if` clause which uses the combined mods which are coming.. it might be that e will have combined mods for each possible mtype so the else might never be reached

pyhf/pdf.py
lukasheinrich: that's right

kratsg: #275 will handle the deprecation.

lukasheinrich:
@kratsg PTAL

pyhf/pdf.py
kratsg: I see.

lukasheinrich:
rebased

pyhf/pdf.py
kratsg: Not hard-coded anymore ;)

PR #272, opened 2018-09-19
add interesting validation case

lukasheinrich:
# Description

In the process of validating MBJ #217 we notices that  roohf allows multiple systematics **of different type** to share the same nuis parameter (with I guess the implicit understanding that the structure of the nuis parameters must be compatible)

The primary example is a sample that has both 

1.  a HistoSys
2. a OverallSys

both are functions of a single nuisance parameter. with `--qualify-names` pyhf forces the names to be different even though they are the same in the XML, but roohf actually allows this.

This also adds a test script that runs a single point using roohf

- [ ] Tests are passing


lukasheinrich:
we are already have some here

https://github.com/diana-hep/pyhf/tree/master/validation/xmlimport_input2/data

and they should all be quite small. Since they are not part of the main package I think they should not end up on the PyPI distribution, so this should only affect devs

matthewfeickert:
>  Since they are not part of the main package I think they should not end up on the PyPI distribution, so this should only affect devs

yeah, as long as we don't go crazy with the binaries then I think it is okay.

PR #276, opened 2018-09-19
Feature: (automatically) set backends through a fixture

kratsg:
# Description

This PR tries to make testing a lot easier for us moving forward as we start having complexity problems with testing combinations of backends+optimizers and so on and lots of duplicated code. In particular, the fixture is called `backend` so adding a `backend` parameter to any test function you have will automatically wrap that function in setting up the backend correctly, and breaking it down after the test runs. This can be useful with certain backends like tensorflow that need a new session or its internal graph cleared.

Additionally, some extra decorations (`pytest.mark`s) are specified to make this even easier to fine-tune which backends run. Right now, we add 5: `numpy`, `tensorflow`, `pytorch`, `mxnet`, `numpy_minuit`. The last one is numpy with the minuit optimizer.

One can specify, for example, to only run `numpy` and `numpy_minuit`

```python
@pytest.mark.only_numpy
@pytest.mark.only_numpy_minuit
```

or

```python
@pytest.mark.skip_tensorflow
@pytest.mark.skip_pytorch
@pytest.mark.skip_mxnet
```

will be equivalent. The latter will not skip any new backends that get added after this PR, but the former will still only consider those two.

Similarly, if you expect a particular backend to fail for any reason, you can mark that specific backend as failing via

```python
@pytest.mark.fail_tensorflow
```

for example.

- [x] Tests are passing

### DIFF
--- a/pyhf/readxml.py
+++ b/pyhf/readxml.py
@@ -21,7 +21,7 @@ def extract_error(h):
     Returns:
         list: The uncertainty for each bin in the histogram
     """
-    err = h.fSumw2[1:-1] if h.fSumw2 else h.numpy[0]
+    err = h.variances if h.variances else h.numpy[0]
     return np.sqrt(err).tolist()
 
 def import_root_histogram(rootdir, filename, path, name):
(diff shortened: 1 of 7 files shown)


kratsg:
Tests are breaking currently, see scikit-hep/uproot#130.

kratsg:
Fixed with scikit-hep/uproot-methods#1.

PR #277, opened 2018-09-20
Use scipy.special.xlogy in poisson computation for numpy backend

wiso:
This is a micro-optimization on the computation of the poisson in the numpy backend.

# Description

When the observed events are equal to zero the log is not computed in the expression

    np.exp(n*np.log(lam)-lam-gammaln(n+1.))

using [`scipy.special.xlogy(n, lambda)`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.xlogy.html), which return `0` if `n==0` (also when `lambda==0`) without computing the log.
In principle further optimization is possible, returning just `np.exp(-lam)` when `n == 0` but probably this require some python control flow you want to avoid (and `np.where` is not lazy).

Quick performance test:

    def f2(l, n):   # present one
        return np.exp(n*np.log(l)-l-gammaln(n+1.))

    def f3(l, n):   # new branch
        return np.exp(xlogy(n, l)-l-gammaln(n+1.))

There is a 40% speed improvement when the observed data are 0

    L = np.random.randint(1, 100, size=1000000)
    N = np.zeros(1000000)

    %timeit f2(L, N)  # present
    10 loops, best of 3: 63.4 ms per loop

    %timeit f3(L, N)  # new branch
    10 loops, best of 3: 40.4 ms per loop

when N>1 there is no difference:

    L = np.random.randint(1, 100, size=1000000)
    N = np.random.randint(1, 100, size=1000000)

    --> 105 (f2) / 109 ms (f3)

as a bonus five warnings in the test are removed.

- [x] Tests are passing

When running on master I have 

    ============== 8 failed, 98 passed, 70 warnings in 444.06 seconds ==============

on this branch

    ============== 8 failed, 98 passed, 65 warnings in 474.10 seconds ==============

Five RuntimeWarning are removed in this branch (all the same)
```
/home/turra/pyhf/pyhf/tensor/numpy_backend.py:163: RuntimeWarning: divide by zero encountered in log
return np.exp(n*np.log(lam)-lam-gammaln(n+1.))
```

### DIFF
--- a/pyhf/tensor/numpy_backend.py
+++ b/pyhf/tensor/numpy_backend.py
@@ -1,6 +1,6 @@
 import numpy as np
 import logging
-from scipy.special import gammaln
+from scipy.special import gammaln, xlogy
 from scipy.stats import norm
 log = logging.getLogger(__name__)
 
@@ -160,7 +160,7 @@ class numpy_backend(object):
         n = np.asarray(n)
         if self.pois_from_norm:
             return self.normal(n,lam, self.sqrt(lam))
-        return np.exp(n*np.log(lam)-lam-gammaln(n+1.))
+        return np.exp(xlogy(n, lam) - lam - gammaln(n + 1.))
 
     def normal(self, x, mu, sigma):
         return norm.pdf(x, loc=mu, scale=sigma)


matthewfeickert:
@wiso Thanks very much for the PR!

kratsg:
I'll point out that while the tests pass, we currently have a regression here where the backends will provide different results

```python
>>> import pyhf
>>> pyhf.tensorlib.poisson(0, 0)
1.0
>>> pyhf.set_backend(pyhf.tensor.mxnet_backend())
>>> pyhf.tensorlib.poisson(0, 0)

[nan]
<NDArray 1 @cpu(0)>
>>> pyhf.tensorlib.poisson(0, 0)
tensor([nan])
```

matthewfeickert:
We should probably revert this and add in a test

wiso:
Sorry about that, I haven't checked the other backends.

By the way if I well understood the implementation for the poisson in the other backends is a gaussian, e.g.

https://github.com/diana-hep/pyhf/blob/cbb899fc134b7d156550dfdc9130154f264a5188/pyhf/tensor/tensorflow_backend.py#L198-L200

So I don't think you want to implement such a check:

```
In [2]: pyhf.tensorlib.poisson(3, 2)
Out[2]: 0.18044704431548356

In [3]: pyhf.set_backend(pyhf.tensor.mxnet_backend())

In [4]: pyhf.tensorlib.poisson(3, 2)
Out[4]: 

[0.21969564]
<NDArray 1 @cpu(0)>

In [5]: pyhf.set_backend(pyhf.tensor.pytorch_backend())

In [6]: pyhf.tensorlib.poisson(3, 2)
Out[6]: tensor([0.2197])
```

I don't know how the nan vs 1 problem can affect the optimization (probably you should reason in that way and not from the mathematical point of view (e.g. Poisson not defined for lam=0), in the same way as you are using a form of continuous Poissonian). By the way if we observe n=0 then the formula become `exp(-lam)` and its limit is 1 for `lam->0`, which is also the maximum. This is true also for the (discrete) Poisson distribution.

kratsg:
@wiso -- so we merged in #268 to fix up this approximation issue (it was on our list of things to merge at the same time) which actually fixes up this approximation problem now -- so you should have compatible results from numpy/pytorch/tensorflow/mxnet (this also added tests to make sure they were equally compatible!)

https://github.com/diana-hep/pyhf/blob/49a9e87eec056d472a013655652b7428819ce9e7/tests/test_tensor.py#L55-L56

- For n=0 and lambda=0 -- this is ill-defined (NaN)
- For n=0 and lambda=1 -- this is defined
- For n=1 and lambda=0 -- this is 0 (should it be 0.37?)
- For n=1 and lambda=1 -- this is defined

At least all the backends are self-consistent. Whether they're consistent with the math is a separate question.

kratsg:
Since this is a simple change, feel free to pull from master, apply the change, and run `pytest tests/test_tensor.py` to see the self-compatibility which fails for all `numpy` backends with the `xlogy` change.

```
>       assert tb.tolist(pyhf.tensorlib.poisson([0, 0, 1, 1], [0, 1, 0, 1])) == pytest.approx([np.nan, 0.3678794503211975, 0.0, 0.3678794503211975], nan_ok=True)
E       assert [1.0, 0.36787944117144233, 0.0, 0.36787944117144233] == approx([nan +- ???, 0.367879450321 +- 3.7e-07, 0.0 +- 1.0e-12, 0.367879450321 +- 3.7e-07])
E        +  where [1.0, 0.36787944117144233, 0.0, 0.36787944117144233] = <bound method numpy_backend.tolist of <pyhf.tensor.numpy_backend.numpy_backend object at 0x1136e6cd0>>(array([1.        , 0.36787944, 0.        , 0.36787944]))
E        +    where <bound method numpy_backend.tolist of <pyhf.tensor.numpy_backend.numpy_backend object at 0x1136e6cd0>> = <pyhf.tensor.numpy_backend.numpy_backend object at 0x1136e6cd0>.tolist
E        +    and   array([1.        , 0.36787944, 0.        , 0.36787944]) = <bound method numpy_backend.poisson of <pyhf.tensor.numpy_backend.numpy_backend object at 0x1136e6cd0>>([0, 0, 1, 1], [0, 1, 0, 1])
E        +      where <bound method numpy_backend.poisson of <pyhf.tensor.numpy_backend.numpy_backend object at 0x1136e6cd0>> = <pyhf.tensor.numpy_backend.numpy_backend object at 0x1136e6cd0>.poisson
E        +        where <pyhf.tensor.numpy_backend.numpy_backend object at 0x1136e6cd0> = pyhf.tensorlib
E        +  and   approx([nan +- ???, 0.367879450321 +- 3.7e-07, 0.0 +- 1.0e-12, 0.367879450321 +- 3.7e-07]) = <function approx at 0x108276b90>([nan, 0.3678794503211975, 0.0, 0.3678794503211975], nan_ok=True)
E        +    where <function approx at 0x108276b90> = pytest.approx
```

wiso:
Here a stupid case where having a number for Poisson[0, 0] may be useful.

```
from pyhf import get_backend
from pyhf.utils import loglambdav
import pyhf.simplemodels

_, optimizer = get_backend()

pdf = pyhf.simplemodels.hepdata_like(signal_data=[1], bkg_data=[0.],
                                     bkg_uncerts=[10])  # last parameter should not count in this case

init_pars = pdf.config.suggested_init()
par_bounds = pdf.config.suggested_bounds()

NOBS = 10
print (optimizer.unconstrained_bestfit(loglambdav, [NOBS] + pdf.config.auxdata,
                                       pdf, init_pars, par_bounds)[0])
```

with this branch, result is 10 as expected. Also, if I put Let put `NOBS=0` as a result I get 0, without any RuntimeWarning.

With master branch (just pulled) with any numbers (I tried 10 or 0):

```
/home/turra/pyhf/pyhf/tensor/numpy_backend.py:186: RuntimeWarning: divide by zero encountered in log
  return np.exp(n * np.log(lam) - lam - gammaln(n + 1.))
/home/turra/pyhf/pyhf/tensor/numpy_backend.py:186: RuntimeWarning: invalid value encountered in multiply
  return np.exp(n * np.log(lam) - lam - gammaln(n + 1.))
     fun: nan
     jac: array([nan, nan])
 message: 'Iteration limit exceeded'
    nfev: 1404
     nit: 101
    njev: 101
  status: 9
 success: False
       x: array([nan, nan])
  File "mytest.py", line 15, in <module>
    pdf, init_pars, par_bounds)[0])
  File "/home/turra/pyhf/pyhf/optimize/opt_scipy.py", line 14, in unconstrained_bestfit
    assert result.success
AssertionError
```

It works only if I put a non-zero background.

PR #280, opened 2018-09-20
Fix: bug in benchmark tests

kratsg:
# Description

The benchmark tests were not running because of forgetting to `pip install minuit` dependency. This was introduced through #276 which imports the minuit optimizer in `tests/conftest.py` which raises the exception.

This also reverts the unfortunate regression introduced by #277 as we don't have equivalent methods to return `0` for the other backends, and whether the `poisson` for `lambda=0` should return 0 or `NaN` is probably not well-defined? (correct me if I'm wrong).

- [x] Tests are passing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ python:
 before_install:
   - pip install --upgrade pip setuptools wheel
 install:
-  - pip install --ignore-installed -U -q -e .[tensorflow,torch,mxnet,minuit,develop]  # Ensure right version of NumPy installed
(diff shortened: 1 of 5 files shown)


matthewfeickert:
Thanks for getting this, @kratsg.

> whether the `poisson` for `lambda=0` should return `0` or `NaN` is probably not well-defined

Now that I think about it `NaN` is really the correct behavior as the Poisson p.m.f. is only defined for `lambda>0`.

kratsg:
rebased against #268 to prepare ~thrusters~ for merge.

PR #281, opened 2018-09-20
introduce cube

lukasheinrich:
# Description

This introduces the "cube-based" computation of the expected yields as first discussed in #231 . Right now this cheats a bit and uses numpy until I figure out a way to to this in generic tensor backends but I wanted to get this diff in so people can have a look.

Note that this still uses the non-vectorized versions of the modifiers. The combined modifiers still need a bit of discussion (the prototype is in this branch https://github.com/diana-hep/pyhf/tree/refactor/vectorize_sampleloop)

- [ ] Tests are passing


kratsg:
Close in favor of #285.

PR #285, opened 2018-09-22
[WIP] Refactor/megachannel

lukasheinrich:
# Description

This is a very much WIP PR that introduces a somewhat radical change:

Instead of keeping multiple channels we "unroll" the channels and create one "mega-channel" in which all channels are concatenated.

In order to do that, the channels must be normalized such that

1. each channel has the same number of samples. For channels in which certain samples are not present a "spurious sample is introduced" that just has zero bin contents 

Further, all modifiers must be normalized into "mega-modifiers" where for bins that are not affected, the modification results in a no-op.

e.g.

*a histosys that only affects one sample in one channel can be expanded to apply to all samples and all channels by just setting `nom=up=lo` for those bins that are not affected

* a normsys that only affects one sample in one channel can be expanded to apply to all samples and all channels by just setting `up=lo=1.0` for those bins that are not affected

* shapesys/shapefactor/staterror/normfactor can be normalized by keeping a mask of which bins are affected and in the result either applying the corresponding factor or 1.0 via `tensorlib.where`

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -131,6 +131,169 @@ class Model(object):
         # build up our representation of the specification
         self.config = _ModelConfig.from_spec(self.spec,**config_kwargs)
 
+        _allmods = []
+        _allsamples = []
+        _allchannels = []
+        _allmods = []
+        channel_nbins = {}
+
+        for c in self.spec['channels']:
+            _allchannels.append(c['name'])
+            for s in c['samples']:
+                channel_nbins[c['name']] = len(s['data'])
+                _allsamples.append(s['name'])
+                for mod in s['modifiers']:
+                    _allmods.append((mod['name'],mod['type']))
+        _allmods = list(set(_allmods))
+        _allsamples = list(set(_allsamples))
+        _allchannels = list(set(_allchannels))
+        self.do_samples  = _allsamples[:]
+        self.do_channels = _allchannels[:]
+        self.do_mods = _allmods[:]
+        self.channel_nbins = channel_nbins
+
+    def _make_mega(self):
+        helper = {}
+        for c in self.spec['channels']:
+            for s in c['samples']:
+                helper.setdefault(c['name'],{})[s['name']] = (c,s)
+
+        mega_mods = {}
+        import copy
+        for m,mtype in self.do_mods:
+            for s in self.do_samples:
+                modspec = {'type': mtype, 'name': m}
+                if mtype == 'histosys':
+                    modspec.setdefault('data',{})['hi_data'] = []
+                    modspec.setdefault('data',{})['lo_data'] = []
+                    modspec.setdefault('data',{})['mask'] = []
+                elif mtype == 'normsys':
+                    modspec.setdefault('data',{})['hi'] = []
+                    modspec.setdefault('data',{})['lo'] = []
+                    modspec.setdefault('data',{})['mask'] = []
+                elif mtype == 'normfactor':
+                    modspec.setdefault('data',{})['mask'] = []
+                elif mtype == 'shapefactor':
+                    modspec.setdefault('data',{})['mask'] = []
+                elif mtype == 'shapesys':
+                    modspec.setdefault('data',{})['mask'] = []
+                elif mtype == 'staterror':
+                    modspec.setdefault('data',{})['uncrt'] = []
+                    modspec.setdefault('data',{})['mask']  = []
+                mega_mods.setdefault(s,{})[m] = copy.deepcopy(modspec)
+                
+        mega_samples = {}
+        for s in self.do_samples:
+            mega_nom = []
+            for c in self.do_channels:
+                defined_samp = helper.get(c,{}).get(s)
+                defined_samp = None if not defined_samp else defined_samp[1]
+                nom = defined_samp['data'] if defined_samp else [0.0]*self.channel_nbins[c]
+                mega_nom += nom
+                defined_mods = {x['name']:x for x in defined_samp['modifiers']} if defined_samp else {}
+                for m,mtype in self.do_mods:
+                    thismod = defined_mods.get(m)
+                    if mtype == 'histosys':
+                        lo_data = thismod['data']['lo_data'] if thismod else nom
+                        hi_data = thismod['data']['lo_data'] if thismod else nom
+                        maskval = True if thismod else False
+                        mega_mods[s][m]['data']['lo_data'] += lo_data
+                        mega_mods[s][m]['data']['hi_data'] += hi_data
+                        mega_mods[s][m]['data']['mask'] += [maskval]*len(nom) #broadcasting
+                        pass
+                    elif mtype == 'normsys':
+                        maskval = True if thismod else False
+                        lo_factor = thismod['data']['lo'] if thismod else 1.0
+                        hi_factor = thismod['data']['hi'] if thismod else 1.0
+                        mega_mods[s][m]['data']['lo'] += [lo_factor]*len(nom) #broadcasting
+                        mega_mods[s][m]['data']['hi'] += [hi_factor]*len(nom)
+                        mega_mods[s][m]['data']['mask'] += [maskval]*len(nom) #broadcasting
+                    elif mtype == 'normfactor':
+                        maskval = True if thismod else False
+                        mega_mods[s][m]['data']['mask'] += [maskval]*len(nom) #broadcasting
+                    elif mtype == 'staterror':
+                        uncrt = thismod['data'] if thismod else [0.0]*len(nom)
+                        maskval = [True if thismod else False]*len(nom)
+                        mega_mods[s][m]['data']['mask']  += maskval
+                        mega_mods[s][m]['data']['uncrt'] += uncrt
+                    elif mtype == 'shapefactor':
+                        maskval = True if thismod else False
+                        mega_mods[s][m]['data']['mask'] += [maskval]*len(nom) #broadcasting
+                    else:
+                        raise RuntimeError
+            sample_dict = {
+                'name': 'mega_{}'.format(s),
+                'nom': mega_nom,
+                'modifiers': list(mega_mods[s].values())
+            }
+            mega_samples[s] = sample_dict
+        self.mega_samples = mega_samples
+        self.mega_mods    = mega_mods
+
+    def prep(self):
+        import numpy as np
+        self.normsys_histoset = np.asarray([
+            [
+                [
+                    self.mega_mods[s][m]['data']['lo'],
+                    [1.]*len(self.mega_samples[s]['nom']),
+                    self.mega_mods[s][m]['data']['hi'],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'normsys' 
+        ])
+        self.normsys_mask = np.asarray([
+            [
+                [
+                    self.mega_mods[s][m]['data']['mask'],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'normsys' 
+        ])
+        self.normsys_default = np.ones(self.normsys_mask.shape)
+
+
+        self.histosys_histoset = np.asarray([
+            [
+                [
+                    self.mega_mods[s][m]['data']['lo_data'],
+                    self.mega_samples[s]['nom'],
+                    self.mega_mods[s][m]['data']['hi_data'],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'histosys' 
+        ])
+        self.histosys_mask = np.asarray([
+            [
+                [
+                    self.mega_mods[s][m]['data']['mask'],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'histosys' 
+        ])
+        self.histosys_default = np.zeros(self.histosys_mask.shape)
+
+
+        self.normfactor_mask = np.asarray([
+            [
+                [
+                    self.mega_mods[s][m]['data']['mask'],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'normfactor' 
+        ])
+        self.normfactor_default = np.ones(self.normfactor_mask.shape)
+        self.staterror_mask = np.asarray([
+            [
+                [
+                    self.mega_mods[s][m]['data']['mask'],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'staterror' 
+        ])
+        self.staterror_default = np.ones(self.staterror_mask.shape)
+
+
     def _mtype_results(self,mtype,pars):
         """
         This method implements the computation of the modifier's application
@@ -350,22 +513,66 @@ class Model(object):
             auxdata = tensorlib.concatenate(tocat)
         return auxdata
 
-    def expected_actualdata(self, pars):
-        tensorlib, _ = get_backend()
-        pars = tensorlib.astensor(pars)
-        data = []
+    def expected_actualdata(self,pars):
+        import numpy as np
+        results_staterr = np.asarray([
+            [
+                [
+                    np.ones_like(self.mega_mods[s][m]['data']['mask'])*pars[self.config.par_slice(m)],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'staterror' 
+        ])
 
-        all_modifications = self._all_modifications(pars)
-        for channel in self.spec['channels']:
-            sample_stack = [
-                self._expected_sample(
-                    tensorlib.astensor(sample['data']), #nominal
-                    *all_modifications[channel['name']][sample['name']] #mods
-                )
-                for sample in channel['samples']
-            ]
-            data.append(tensorlib.sum(tensorlib.stack(sample_stack),axis=0))
-        return tensorlib.concatenate(data)
+
+        results_normfac = np.asarray([
+            [
+                [
+                    np.ones_like(self.mega_mods[s][m]['data']['mask'])*pars[self.config.par_slice(m)],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'normfactor' 
+        ])
+
+        histosys_alphaset = np.asarray([
+            pars[self.config.par_slice(m)] for m,mtype in self.do_mods if mtype == 'histosys'
+        ])
+        normsys_alphaset = np.asarray([
+            pars[self.config.par_slice(m)] for m,mtype in self.do_mods if mtype == 'normsys'
+        ])
+        from .interpolate import _hfinterp_code1,_hfinterp_code0
+        results_norm   = _hfinterp_code1(self.normsys_histoset,normsys_alphaset)
+        results_norm   = np.where(self.normsys_mask,results_norm,self.normsys_default)
+
+        results_histo   = _hfinterp_code0(self.histosys_histoset,histosys_alphaset)
+        results_histo   = np.where(self.histosys_mask,results_histo,self.histosys_default)
+
+        results_staterr = np.where(self.staterror_mask,results_staterr,self.staterror_default)
+        results_normfac = np.where(self.normfactor_mask,results_normfac,self.normfactor_default)
+
+
+        thenom = np.asarray([self.mega_samples[s]['nom'] for s in self.do_samples])
+        thenom = np.asarray(thenom).reshape((1,)+results_histo.shape[1:])
+
+        allsum = np.concatenate([
+            results_histo,
+            thenom
+        ])
+
+        nom_plus_delta = np.sum(allsum,axis=0)
+        nom_plus_delta = nom_plus_delta.reshape((1,)+nom_plus_delta.shape)
+
+        # print(nom_plus_delta.shape,results_histo.shape)
+        allfac = np.concatenate([
+            results_norm,
+            results_staterr,
+            results_normfac,
+            nom_plus_delta
+        ])
+        allfac.shape
+        newbysample = np.product(allfac,axis=0)
+        newresults = np.sum(newbysample,axis=0)
+        return newresults[0] #only one alphas
 
     def expected_data(self, pars, include_auxdata=True):
         tensorlib, _ = get_backend()
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -121,6 +121,12 @@ class _ModelConfig(object):
             self.auxdata_order.append(modifier_def['name'])
         return modifier
 
+def finalize_stats(modifier):
+    tensorlib, _ = get_backend()
+    inquad = tensorlib.sqrt(tensorlib.sum(tensorlib.power(tensorlib.astensor(modifier.uncertainties),2), axis=0))
+    totals = tensorlib.sum(modifier.nominal_counts,axis=0)
+    return tensorlib.divide(inquad,totals)
+
 class Model(object):
     def __init__(self, spec, **config_kwargs):
         self.spec = copy.deepcopy(spec) #may get modified by config
@@ -151,6 +157,9 @@ class Model(object):
         self.do_channels = _allchannels[:]
         self.do_mods = _allmods[:]
         self.channel_nbins = channel_nbins
+
+        self.finalized_stats = {k:finalize_stats(self.config.modifier(k)) for k,v in self.config.par_map.items() if 'staterror' in k}
+
 
     def _make_mega(self):
         helper = {}
@@ -587,9 +596,8 @@ class Model(object):
 
     def constraint_logpdf(self, auxdata, pars):
         tensorlib, _ = get_backend()
-        # iterate over all constraints order doesn't matter....
         start_index = 0
-        summands = None
+        bytype = {}
         for cname in self.config.auxdata_order:
             modifier, modslice = self.config.modifier(cname), \
                 self.config.par_slice(cname)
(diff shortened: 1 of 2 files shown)


lukasheinrich:
I'm pretty sure the new code is correct through random testing on parameter sets within the bounds of the parameter ranges.

screenshot

But somehow now the fit is not converging for MBJ. 

I'll investigate further possibly on smaller models

lukasheinrich:
apparently we are hitting situations more often where the logpdf terms include `-inf` more often (which happens if the actual pdf term is zero. Arguably we should have protected against this earlier, but it's still not clear to me why this appeared in the refactor

pyhf/pdf.py
```diff
@@ -131,6 +137,173 @@ def __init__(self, spec, **config_kwargs):
...
+                    modspec.setdefault('data',{})['mask'] = []
+                elif mtype == 'shapefactor':
+                    modspec.setdefault('data',{})['mask'] = []
+                elif mtype == 'shapesys':
+                    modspec.setdefault('data',{})['mask'] = []
+                elif mtype == 'staterror':
+                    modspec.setdefault('data',{})['uncrt'] = []
+                    modspec.setdefault('data',{})['mask']  = []
+                mega_mods.setdefault(s,{})[m] = copy.deepcopy(modspec)
+
```
lukasheinrich: the mega-channel will consist of mega-samples that subscribe to mega-modifiers. i.e. while in normal histfactory, each sample might be affected by some modifiers and some not, here we change it so that samples are affected by *all modifiers*, but we set up the modifier data such that the application of the modifier does not actually change the bin value for bins that are not originally affected by that modifier

pyhf/pdf.py
```diff
@@ -131,6 +137,173 @@ def __init__(self, spec, **config_kwargs):
...
+                for m,mtype in self.do_mods:
+                    thismod = defined_mods.get(m)
+                    if mtype == 'histosys':
+                        lo_data = thismod['data']['lo_data'] if thismod else nom
+                        hi_data = thismod['data']['hi_data'] if thismod else nom
+                        maskval = True if thismod else False
+                        mega_mods[s][m]['data']['lo_data'] += lo_data
+                        mega_mods[s][m]['data']['hi_data'] += hi_data
+                        mega_mods[s][m]['data']['mask']    += [maskval]*len(nom) #broadcasting
+                        pass
```
lukasheinrich: for histogram systematics, samples that are affected, store `hi_data` `lo_data` we add this to the corresponding `hi_data` and `lo_data` of the mega_histosys of that name. if the sample is not originally affected, we add dummy data where hi_data=lo_data=nom such that the delta of the interpolation will always be zero (we also keep a mask, so that we can actually force it to zero to in case we run into numerical issues

pyhf/pdf.py
```diff
@@ -131,6 +137,173 @@ def __init__(self, spec, **config_kwargs):
...
+                    elif mtype == 'normsys':
+                        maskval = True if thismod else False
+                        lo_factor = thismod['data']['lo'] if thismod else 1.0
+                        hi_factor = thismod['data']['hi'] if thismod else 1.0
+                        mega_mods[s][m]['data']['lo']   += [lo_factor]*len(nom) #broadcasting
+                        mega_mods[s][m]['data']['hi']   += [hi_factor]*len(nom)
+                        mega_mods[s][m]['data']['mask'] += [maskval]  *len(nom) #broadcasting
+                    elif mtype == 'normfactor':
```
lukasheinrich: for normsys systematics, samples that are affected, store `hi` and `lo` factors that should be applied to all bins in the sampl. for samples which are not affected by the normsys, we just force `hi=lo=1.0` which will result in a 1.0 factor from the interpolation code.. in any case we also keep a mask to force if necessary

pyhf/pdf.py
```diff
@@ -131,6 +137,173 @@ def __init__(self, spec, **config_kwargs):
...
+                        mega_mods[s][m]['data']['mask'] += [maskval]*len(nom) #broadcasting
+                    elif mtype == 'staterror':
+                        uncrt = thismod['data'] if thismod else [0.0]*len(nom)
+                        maskval = [True if thismod else False]*len(nom)
+                        mega_mods[s][m]['data']['mask']  += maskval
+                        mega_mods[s][m]['data']['uncrt'] += uncrt
+                    elif mtype == 'shapefactor':
+                        maskval = True if thismod else False
+                        mega_mods[s][m]['data']['mask'] += [maskval]*len(nom) #broadcasting
+                    else:
```
lukasheinrich: for shapefactors and staterrors we just keep a mask as to which bins are affected

pyhf/pdf.py
```diff
@@ -131,6 +137,173 @@ def __init__(self, spec, **config_kwargs):
...
+        self.normsys_histoset = np.asarray([
+            [
+                [
+                    self.mega_mods[s][m]['data']['lo'],
+                    [1.]*len(self.mega_samples[s]['nom']),
+                    self.mega_mods[s][m]['data']['hi'],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'normsys' 
+        ])
```
lukasheinrich: this preps the histoset for the interpolation codes..

pyhf/pdf.py
```diff
@@ -131,6 +137,173 @@ def __init__(self, spec, **config_kwargs):
...
+        self.normsys_mask = np.asarray([
+            [
+                [
+                    self.mega_mods[s][m]['data']['mask'],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'normsys' 
+        ])
+        self.normsys_default = np.ones(self.normsys_mask.shape)
+
```
lukasheinrich: here we create a mask for the affected bins and prep the default values that should be applied to those not affeced

pyhf/pdf.py
```diff
@@ -131,6 +137,173 @@ def __init__(self, spec, **config_kwargs):
...
+        self.histosys_mask = np.asarray([
+            [
+                [
+                    self.mega_mods[s][m]['data']['mask'],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'histosys' 
+        ])
+        self.histosys_default = np.zeros(self.histosys_mask.shape)
+
```
lukasheinrich: same for histosys

pyhf/pdf.py
```diff
@@ -131,6 +137,173 @@ def __init__(self, spec, **config_kwargs):
...
+        self.staterror_mask = np.asarray([
+            [
+                [
+                    self.mega_mods[s][m]['data']['mask'],
+                ]
+                for s in self.do_samples
+            ] for m,mtype in self.do_mods if mtype == 'staterror' 
+        ])
+        self.staterror_default = np.ones(self.staterror_mask.shape)
+
```
lukasheinrich: the default value for normfactor and staterror is a multiplication by 1.0 (a no-op) so we crate masks and the default values

lukasheinrich:
the reason the inf guards were necessary is because I set up the observed data incorrectly (the channel order was wrong). without them the fit converges to the same values as before (yay!)

kratsg:
```
for a sinle channel single sample

we compute


Pois(d | fac(pars) * (delta(pars) + nom) ) * Gaus(a | pars[is_gaus], sigmas) * Pois(a * cfac | pars[is_poi] * cfac)

delta(pars) is the result of an apply(pars) of a combined modifier with additive op_code (histosys)

fac(pars) is the result of apply(pars) of combined multiplicative modifier

pars[is_gaus] are the subset of parameters that are constrained by gauss (with sigmas accrodingly, some of which are cmputed by modifiers)

pars[is_pois] are the poissons and their rates (they come with their own factors "cfac" unrelated to fac(pars) .. which are also computed by the finalize() of the modifier)

so in the end we only make 3 calls to pdfs

1. The main pdf of data and modified rates
2. All Gaus constraint as one call
3. All Pois constraints as one call

```

docs/examples/notebooks/binderexample/StatisticalAnalysis.ipynb
```diff
@@ -129,18 +129,15 @@
      "output_type": "stream",
```
lukasheinrich: do we need this diff in this PR?

.ackrc
```diff
@@ -1,2 +1,8 @@
 --ignore-directory=docs/_generated
 --ignore-directory=htmlcov
+--ignore-directory=.ipynb_checkpoints
+--ignore-directory=.pytest_cache
+--ignore-directory=ext:egg-info
+--ignore-file=ext:coverage
+--ignore-file=ext:log
+
```
lukasheinrich: does this need to be in this PR?

pyhf/modifiers/__init__.py
```diff
@@ -11,7 +11,7 @@
 Check if given object contains the right structure for constrained and unconstrained modifiers
 '''
 def validate_modifier_structure(modifier, constrained):
-    required_methods = ['__init__', 'add_sample', 'apply']
+    required_methods = ['__init__']
```
lukasheinrich: this can probably be entirely removed (checking for existence of `__init__` doesn't make much sense

pyhf/modifiers/histosys.py
```diff
@@ -4,48 +4,16 @@
...
-                    ]
-                ]
-            ]), tensorlib.astensor([tensorlib.tolist(pars)])
-        )
-
-        return results[0][0][0]
+    @classmethod
+    def required_parset(cls, n_parameters):
+        return {
+            'parset': constrained_by_normal,
```
lukasheinrich: i'd rather call this `parset_type`

pyhf/modifiers/normfactor.py
```diff
@@ -6,18 +6,14 @@
...
-    
-    def add_sample(self, channel, sample, modifier_def):
-        pass
-
-    def apply(self, channel, sample, pars):
-        return pars
+    @classmethod
+    def required_parset(cls, n_parameters):
+        return {
+            'parset': unconstrained,
```
lukasheinrich: i'd rather call this `parset_type`

pyhf/modifiers/normsys.py
```diff
@@ -4,50 +4,16 @@
...
-                    ]
-                ]
-            ]), tensorlib.astensor([tensorlib.tolist(pars)])
-        )
-
-        return results[0][0][0]
+    @classmethod
+    def required_parset(cls, n_parameters):
+        return {
+            'parset': constrained_by_normal,
```
lukasheinrich: i'd rather call this `parset_type`

pyhf/modifiers/shapefactor.py
```diff
@@ -6,18 +6,14 @@
...
-
-    def add_sample(self, channel, sample, modifier_def):
-        pass
-
-    def apply(self, channel, sample, pars):
-        return pars
+    @classmethod
+    def required_parset(cls, n_parameters):
+        return {
+            'parset': unconstrained,
```
lukasheinrich: i'd rather call this `parset_type`

pyhf/modifiers/shapesys.py
```diff
@@ -6,29 +6,14 @@
...
-
-    def add_sample(self, channel, sample, modifier_def):
-        pass
-
-    def apply(self, channel, sample, pars):
-        return pars
+    @classmethod
+    def required_parset(cls, n_parameters):
+        return {
+            'parset': constrained_by_poisson,
```
lukasheinrich: i'd rather call this `parset_type`

pyhf/modifiers/staterror.py
```diff
@@ -4,37 +4,16 @@
...
-    def add_sample(self, channel, sample, modifier_def):
-        self.nominal_counts.append(sample['data'])
-        self.uncertainties.append(modifier_def['data'])
-
-    def apply(self, channel, sample, pars):
-        return pars
+    @classmethod
+    def required_parset(cls, n_parameters):
+        return {
+            'parset': constrained_by_normal,
```
lukasheinrich: i'd rather call this `parset_type`

pyhf/paramsets.py
```diff
@@ -6,10 +6,11 @@ def __init__(self, n_parameters, inits, bounds):
         self.suggested_bounds = bounds
 
 class unconstrained(paramset):
-    pass
+    def __init__(self, n_parameters, inits, bounds, auxdata=None, factors=None):
+        super(unconstrained,self).__init__(n_parameters, inits, bounds)
 
 class constrained_by_normal(paramset):
-    def __init__(self, n_parameters, inits, bounds, auxdata):
+    def __init__(self, n_parameters, inits, bounds, auxdata, factors=None):
```
lukasheinrich: why do we add this here? `factors` is not a concept for normal constraints

tests/benchmarks/test_benchmark.py
```diff
@@ -75,6 +75,7 @@ def runOnePoint(pdf, data):
 
 @pytest.mark.parametrize('n_bins', bins, ids=bin_ids)
 @pytest.mark.skip_mxnet
+@pytest.mark.skip_pytorch
```
lukasheinrich: can we remove this `skip`?

tests/test_import.py
```diff
@@ -41,9 +41,9 @@ def test_import_prepHistFactory():
     assert pdf.expected_actualdata(
         pdf.config.suggested_init()).tolist() == [120.0, 110.0]
 
-    assert pdf.config.auxdata_order == ['syst1', 'staterror_channel1', 'syst2', 'syst3']
+    assert pdf.config.auxdata_order == ['staterror_channel1', 'syst1', 'syst2', 'syst3']
```
lukasheinrich: could also change this to `sorted(['syst1', 'staterror_channel1', 'syst2', 'syst3'])` makes the diff smaller and intent is clearer

tests/test_modifiers.py
```diff
@@ -29,7 +29,6 @@ def test_modifiers_structure():
     class myCustomModifier(object):
         def __init__(self): pass
         def add_sample(self): pass
```
lukasheinrich: `add_sample` is not a thing anymore

pyhf/pdf.py
```diff
@@ -70,68 +88,146 @@ def par_slice(self, name):
...
+        parset = modifier_cls.required_parset(len(sample['data']))
+        parset.update({'channel,sample': (channel['name'], sample['name'])})
+
+        if not(parset['is_shared']) and self.paramsets.get(name):
+            raise ValueError("Trying to add unshared-paramset but other paramsets exist with the same name.")
+
+        if parset['is_shared'] and self.paramsets.get(name) and not(self.paramsets.get(name)[0]['is_shared']):
+            raise ValueError("Trying to add shared-paramset but other paramset of same name is indicated to be unshared.")
+
+        self.paramsets.get(name).append(parset)
```
lukasheinrich: can just make this one `self.paramsets.setdefault(name,[]).append(parset)`

pyhf/pdf.py
```diff
@@ -70,68 +88,146 @@ def par_slice(self, name):
...
-
-        # did not return, so create new modifier and return it
-        modifier = modifier_cls(sample['data'], modifier_def['data'])
-        self.register_paramset(modifier_def['name'], modifier.n_parameters, modifier)
-
-        return modifier
+        self.paramsets.setdefault(name, [])
+
+        parset = modifier_cls.required_parset(len(sample['data']))
+        parset.update({'channel,sample': (channel['name'], sample['name'])})
```
lukasheinrich: why do we need this?

pyhf/pdf.py
```diff
@@ -70,68 +88,146 @@ def par_slice(self, name):
...
+        if parset['is_shared'] and self.paramsets.get(name) and not(self.paramsets.get(name)[0]['is_shared']):
+            raise ValueError("Trying to add shared-paramset but other paramset of same name is indicated to be unshared.")
+
+        self.paramsets.get(name).append(parset)
+
+    def combine_paramsets(self):
+        for param_name in list(self.paramsets.keys()):
+            params = self.paramsets[param_name]
+
+            def get_inits(parset, npars, param_matching):
```
lukasheinrich: I would make the inits/bounds/auxdata be part of the requirements returned by the modifier classes.. and here we only check compatibility (e.g. all modifeirs requesting a parmset of a given name need to declare the same bounds, etc)

pyhf/pdf.py
```diff
@@ -70,68 +88,146 @@ def par_slice(self, name):
...
+
+        if not(parset['is_shared']) and self.paramsets.get(name):
+            raise ValueError("Trying to add unshared-paramset but other paramsets exist with the same name.")
+
+        if parset['is_shared'] and self.paramsets.get(name) and not(self.paramsets.get(name)[0]['is_shared']):
+            raise ValueError("Trying to add shared-paramset but other paramset of same name is indicated to be unshared.")
+
+        self.paramsets.get(name).append(parset)
+
+    def combine_paramsets(self):
```
lukasheinrich: I would suggest moving this to parsets.py, the module where we define what parsets are and (then) also define how to properly combine them `combine_paramsets(parset_dict)`

pyhf/pdf.py
```diff
@@ -70,68 +88,146 @@ def par_slice(self, name):
...
+
+            if len(params) == 1:
+                self.paramsets[param_name] = self.paramsets[param_name][0]
+            else:
+                channel_sample = set([])
+                constraint_types = set([])
+                n_parameters = set([])
+                param_matching = set([])
+                mod_types = set([])
+                for param in params:
```
lukasheinrich: not clear to me why those (except n_parameters) are needed (as they're not used by `add_paramsets`)

pyhf/pdf.py
```diff
@@ -70,68 +88,146 @@ def par_slice(self, name):
     def param_set(self, name):
         return self.par_map[name]['parset']
 
-    def modifier(self, name):
-        return self.par_map[name]['modifier']
+    def modifier_type(self, name):
+        return self.par_map[name]['modifier_type']
```
lukasheinrich: if I see things correctly, I think we can remove this

pyhf/pdf.py
```diff
@@ -70,68 +88,146 @@ def par_slice(self, name):
...
     def set_poi(self,name):
-        if name not in self.modifiers:
+        if name not in [x for x,_ in self.modifiers]:
             raise exceptions.InvalidModel("The paramter of interest '{0:s}' cannot be fit as it is not declared in the model specification.".format(name))
         s = self.par_slice(name)
         assert s.stop-s.start == 1
         self.poi_index = s.start
 
-    def register_paramset(self, name, n_parameters, modifier):
+    def register_paramset(self, modifier_type, name, n_parameters, parset):
```
lukasheinrich: if indeed `def modifier_type(self, name):` is never used, we might not need the first arg here anymore

pyhf/pdf.py
```diff
@@ -70,68 +88,146 @@ def par_slice(self, name):
...
-        parset = modifier.parset
-
         sl = slice(self.next_index, self.next_index + n_parameters)
         self.next_index = self.next_index + n_parameters
         self.par_order.append(name)
         self.par_map[name] = {
             'slice': sl,
             'parset': parset,
-            'modifier': modifier,
+            'modifier_type': modifier_type,
```
lukasheinrich: in this case this will also be gone..

pyhf/paramsets.py
```diff
@@ -32,3 +34,35 @@ def expected_data(self, pars):
...
+            for k in param_keys:
+                combined_param.setdefault(k, set([])).add(param[k])
+
+        for k in param_keys:
+            v = combined_param[k]
+            if len(v) != 1:
+                raise exceptions.InvalidNameReuse("Multiple values for '{}' ({}) were found for {}. Use unique modifier names or use qualify_names=True when constructing the pdf.".format(k, list(v), param_name))
+            else:
+                v = list(v)[0]
+                if isinstance(v, tuple): v = list(v)
```
lukasheinrich: what's going on here? why is this needed?

pyhf/pdf.py
```diff
@@ -70,68 +70,49 @@ def par_slice(self, name):
...
-        log.info('adding modifier %s (%s new nuisance parameters)', name, n_parameters)
+    def _bookkeep_paramset(self, channel, sample, name, modifier_type):
+        # get modifier class associated with modifier type
+        try:
+            modifier_cls = modifiers.registry[modifier_type]
+        except KeyError:
+            log.exception('Modifier type not implemented yet (processing {0:s}). Current modifier types: {1}'.format(modifier_type, modifiers.registry.keys()))
+            raise exceptions.InvalidModifier()
+
+        parset = modifier_cls.required_parset(len(sample['data']))
```
lukasheinrich: seems like if we just pass the parset here

```
def _bookkeep_paramset(self, parset):
```

the APi would be cleaner (also see comment below, i'm not sure why we need sample / channel)

pyhf/pdf.py
```diff
@@ -22,34 +24,32 @@ def __init__(self, spec, poiname = 'mu', qualify_names = False):
...
                     if qualify_names:
                         fullname = '{}/{}'.format(modifier_def['type'],modifier_def['name'])
                         if modifier_def['name'] == poiname:
                             poiname = fullname
                         modifier_def['name'] = fullname
-                    modifier = self.add_or_get_modifier(channel, sample, modifier_def)
-                    modifier.add_sample(channel, sample, modifier_def)
-                    self.modifiers.append(modifier_def['name'])
-                    sample['modifiers_by_type'].setdefault(modifier_def['type'],[]).append(modifier_def['name'])
+                    self._bookkeep_paramset(channel, sample, modifier_def['name'], modifier_def['type'])
```
lukasheinrich:
```
parset = modifiers.registry[modifier_def['type']].required_parset(len(sample['data']))
self._bookkeep_paramset(parset_req)
```

pyhf/pdf.py
```diff
@@ -70,68 +70,49 @@ def par_slice(self, name):
...
+            raise ValueError("Trying to add unshared-paramset but other paramsets exist with the same name.")
 
-        sl = slice(self.next_index, self.next_index + n_parameters)
-        self.next_index = self.next_index + n_parameters
+        if parset['is_shared'] and not(self.paramset_requirements.get(name, [{'is_shared': True}])[0]['is_shared']):
+            raise ValueError("Trying to add shared-paramset but other paramset of same name is indicated to be unshared.")
+
+        self.paramset_requirements.setdefault(name,[]).append(parset)
+
+    def _register_paramset(self, name, parset):
```
lukasheinrich: rename to `_create_and_register_paramset`

pyhf/pdf.py
```diff
@@ -70,68 +70,49 @@ def par_slice(self, name):
...
-            if not type(modifier).__name__ == modifier_def['type']:
-                raise exceptions.InvalidNameReuse('existing modifier is found, but it is of wrong type {} (instead of {}). Use unique modifier names or use qualify_names=True when constructing the pdf.'.format(type(modifier).__name__, modifier_def['type']))
-            return modifier
-
-        # did not return, so create new modifier and return it
-        modifier = modifier_cls(sample['data'], modifier_def['data'])
-        self.register_paramset(modifier_def['name'], modifier.n_parameters, modifier)
-
-        return modifier
+    def _register_paramsets(self):
```
lukasheinrich: rename to `_create_and_register_paramsets`

pyhf/pdf.py
```diff
@@ -70,68 +70,49 @@ def par_slice(self, name):
...
+        for param_name, combined_param in reduce_paramset_requirements(self.paramset_requirements).items():
+            constraint = combined_param.pop('constraint')
```
lukasheinrich: why are we popping here? all paramsets should just take `kwargs` now, no?

pyhf/modifiers/histosys.py
```diff
@@ -3,49 +3,68 @@
...
+            'n_parameters': n_parameters,
+            'modifier': cls.__name__,
+            'is_constrained': cls.is_constrained,
+            'is_shared': True,
+            'op_code': cls.op_code,
+            'inits': (0.0,) * n_parameters,
+            'bounds': ((-5., 5.),) * n_parameters,
+            'auxdata': (0.,) * n_parameters,
+            'factors': () * n_parameters
+        }
```
lukasheinrich: now that all paramsets take `kwargs`, we should probably not even list these un-needed attributes here and elsewhere (.. unconstrained mods do not have auxdata.. only poisson constraints have factors

pyhf/modifiers/histosys.py
```diff
@@ -3,49 +3,68 @@
...
-        self.at_zero = {}
-        self.at_minus_one = {}
-        self.at_plus_one = {}
-
+    @classmethod
+    def required_parset(cls, n_parameters):
+        n_parameters = 1
+        return {
+            'constraint': constrained_by_normal,
+            'n_parameters': n_parameters,
```
lukasheinrich: why not just `n_parameters: 1` ?

pyhf/pdf.py
```diff
@@ -22,34 +24,32 @@ def __init__(self, spec, poiname = 'mu', qualify_names = False):
         self.samples = []
         self.parameters = []
         self.modifiers = []
+        # keep track of the width of each channel (how many bins)
         self.channel_nbins = {}
+        # bookkeep all requirements for paramsets we need to build
+        self.paramset_requirements = {}
```
lukasheinrich: make this just a local variable

pyhf/pdf.py
```diff
@@ -22,34 +24,32 @@ def __init__(self, spec, poiname = 'mu', qualify_names = False):
...
-                    self.modifiers.append(modifier_def['name'])
-                    sample['modifiers_by_type'].setdefault(modifier_def['type'],[]).append(modifier_def['name'])
+                    self._bookkeep_paramset(channel, sample, modifier_def['name'], modifier_def['type'])
+                    self.modifiers.append((modifier_def['name'],modifier_def['type']))
         self.channels = list(set(self.channels))
         self.samples = list(set(self.samples))
         self.parameters = list(set(self.parameters))
         self.modifiers = list(set(self.modifiers))
         self.channel_nbins = self.channel_nbins
+        self._register_paramsets()
```
lukasheinrich:
```python
self._register_paramsets(paramset_requirements)
```

pyhf/pdf.py
lukasheinrich: change sig to ```def _create_and_register_paramsets(paramset_requirements)```

pyhf/paramsets.py
kratsg: bounds/inits/etc.. are sets() to add to a set() to check for uniqueness. Lists cannot be added to sets. But then since we need these to be mutable later in the code, we convert the sets to a list.

pyhf/pdf.py
lukasheinrich: right, but if we just craete the parset requirements elsewhere and pass it in, we don't anymore, right?

pyhf/modifiers/normfactor.py
```diff
@@ -3,21 +3,54 @@
...
-@modifier(name='normfactor', shared=True, op_code = 'multiplication')
+@modifier(name='normfactor', op_code = 'multiplication')
 class normfactor(object):
-    def __init__(self, nom_data, modifier_data):
-        self.n_parameters = 1
+    @classmethod
+    def required_parset(cls, n_parameters):
+        n_parameters = 1
+        return {
+            'constraint': unconstrained,
```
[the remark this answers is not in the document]
lukasheinrich: hm but an unconstrained parameter set is not a constraint, so isn't it weird to call it that?

pyhf/paramsets.py
lukasheinrich: ok I was confused by the many casts.. maybe rewrite to 
```
comb_attr = v.pop()
if isinstance(comb_attr, tuple): comb_attr = list(comb_attr)
combined_param[k]  = comb_attr
```

lukasheinrich:
🍾

PR #289, opened 2018-10-01
[WIP] Updates interpolators to allow for precomputation

kratsg:
# Description

This PR is to help get ready for #285 by making it more manageable. Change interpolation functions to interpolation classes at a first step. Define an `Interpolator` class which all interpolator codes will inherit from. At the moment, this parent class will just set up the `__init__()` call for us as a way to duck-type things. It will also call the internal `_precompute()` code that is defined for each interpolator that will precompute various bits and pieces for us.

At the end, this might be split up into `pyhf/interpolate/__init__.py` to break it up more to get ready for more interpolator codes in the future.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyhf/interpolate.py
+++ b/pyhf/interpolate.py
@@ -20,24 +20,38 @@ def _slow_hfinterp_looper(histogramssets, alphasets, func):
                 histo_result.append(alpha_result)
     return all_results
 
+class _hfinterpolator_code0(object):
+    def __init__(self,histogramssets,alphasets_shape):
+        tensorlib, _ = get_backend()
+
+        # these two variables can be pre-computed at PDF time
+        self.allset_allhisto_deltas_up = histogramssets[:,:,2] - histogramssets[:,:,1]
+        self.allset_allhisto_deltas_dn = histogramssets[:,:,1] - histogramssets[:,:,0]
+
+        self.mask_on = tensorlib.ones(alphasets_shape)
+        self.mask_off = tensorlib.zeros(alphasets_shape)
+
+        self.broadcast_helper = tensorlib.ones(self.allset_allhisto_deltas_dn.shape)
+
+    def interpolate(self,alphasets):
+        tensorlib, _ = get_backend()
+        where_alphasets_positive = tensorlib.where(alphasets > 0, self.mask_on, self.mask_off)
+        
+        # s: set under consideration (i.e. the modifier)
+        # a: alpha variation
+        # h: histogram affected by modifier
+        # b: bin of histogram
+        allsets_allhistos_alphas_times_deltas_up = tensorlib.einsum('sa,shb->shab',alphasets,self.allset_allhisto_deltas_up)
+        allsets_allhistos_alphas_times_deltas_dn = tensorlib.einsum('sa,shb->shab',alphasets,self.allset_allhisto_deltas_dn)
+
+        allsets_allhistos_masks = tensorlib.einsum('sa,shb->shab', where_alphasets_positive, self.broadcast_helper)
+
+        return tensorlib.where(allsets_allhistos_masks,allsets_allhistos_alphas_times_deltas_up, allsets_allhistos_alphas_times_deltas_dn)
+
 def _hfinterp_code0(histogramssets, alphasets):
     tensorlib, _ = get_backend()
-
-    # these two variables can be pre-computed at PDF time
-    allset_allhisto_deltas_up = histogramssets[:,:,2] - histogramssets[:,:,1]
-    allset_allhisto_deltas_dn = histogramssets[:,:,1] - histogramssets[:,:,0]
-
-    where_alphasets_positive = tensorlib.where(alphasets > 0, tensorlib.ones(alphasets.shape), tensorlib.zeros(alphasets.shape))
-
-    # s: set under consideration (i.e. the modifier)
-    # a: alpha variation
-    # h: histogram affected by modifier
-    # b: bin of histogram
-    allsets_allhistos_alphas_times_deltas_up = tensorlib.einsum('sa,shb->shab',alphasets,allset_allhisto_deltas_up)
-    allsets_allhistos_alphas_times_deltas_dn = tensorlib.einsum('sa,shb->shab',alphasets,allset_allhisto_deltas_dn)
-    allsets_allhistos_masks = tensorlib.einsum('sa,shb->shab', where_alphasets_positive, tensorlib.ones(allset_allhisto_deltas_dn.shape))
-
-    return tensorlib.where(allsets_allhistos_masks,allsets_allhistos_alphas_times_deltas_up, allsets_allhistos_alphas_times_deltas_dn)
+    i = _hfinterpolator_code0(histogramssets,alphasets.shape)
+    return i.interpolate(alphasets)
 
 def _slow_hfinterp_code0(histogramssets, alphasets):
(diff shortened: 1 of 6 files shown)


pyhf/interpolate.py
```diff
@@ -4,6 +4,13 @@
 from . import get_backend
 from . import exceptions
 
+class Interpolator(object):
+    def __init__(self, histogramssets):
+        self.tensorlib = None
+        self.shape = (0,)
```
lukasheinrich: maybe cleaner to make this `None`?

pyhf/interpolate.py
```diff
@@ -53,24 +87,53 @@ def summand(down, nom, up, alpha):
...
+        tensorlib, _ = get_backend()
+
+        # did things change that we need to recompute?
+        tensor_type_change = tensorlib != self.tensorlib
+        shape_change = shape != self.shape
+
+        if tensor_type_change:
+            # nb: self.histogramssets will never change
+            histogramssets = tensorlib.astensor(self.histogramssets)
+            self.allset_allhisto_deltas_up = tensorlib.divide(histogramssets[:,:,2], histogramssets[:,:,1])
```
lukasheinrich: actually we do not need `histogramssets` to be tensors of the backend, the division can be done in numpy .. the only thing that needs to be an actual tensor of the backend are  `self.X`. This will simplify the computational graph of e.g. ML backends since they will not have to know about `histogramssets` but only about the tensors that are actually used

pyhf/interpolate.py
```diff
@@ -53,24 +87,53 @@ def summand(down, nom, up, alpha):
...
+            self.allset_allhisto_deltas_dn = tensorlib.divide(histogramssets[:,:,0], histogramssets[:,:,1])
+            self.broadcast_helper = tensorlib.ones(self.allset_allhisto_deltas_up.shape)
+
```
lukasheinrich: same here, this can be `np.ones(histogramssets.shape)`

lukasheinrich: discuss in #290 close here

pyhf/interpolate.py
kratsg: discuss in #290 close here

pyhf/tensor/backend.py
```diff
@@ -0,0 +1,8 @@
+class Backend(object):
+    """General backend class for all of our backends"""
+
+    def __eq__(self, other):
+        if isinstance(other, Backend):
+            return self.__class__ == other.__class__
+        return False
```
lukasheinrich: maybe  we should make this a string comparison with strings that are not `__class__`? e.g. I'm anticipating setting backends in the `pyhf` CLI by name `pyhf cls --backend numpy` and there we might not want to be tied to the class names

kratsg: done

pyhf/tensor/backend.py
```diff
@@ -0,0 +1,20 @@
...
+            return self.__class__ == other.__class__
+        elif inspect.isclass(other):
+            # if this is comparing a Backend class
+            return self.__class__ == other
+        elif isinstance(other, string_types):
+            # if this is comparing to a string like 'numpy'
+            return self.name == other
+        else:
+            return False
+        return False # always return false
```
lukasheinrich: do we need this base class.. as far as I can see, we only ever do use `__eq__` to compare `self.tensorlib != tensorlib` which could just be done via `self.tensorlib.name ! = tensorlib.name` 

we might very well need a base class in the future, but perhaps for now it's premature?

kratsg: We can copy/paste this code into all backends -- but it'll be a pain to maintain?

pyhf/interpolate.py
```diff
@@ -20,24 +20,57 @@ def _slow_hfinterp_looper(histogramssets, alphasets, func):
...
+        if alphasets_shape_change or tensor_type_change:
+            self.mask_on  = tensorlib.ones(alphasets_shape)
+            self.mask_off = tensorlib.zeros(alphasets_shape)
+
+        # update after recomputation if needed
+        self.tensorlib = tensorlib
+        self.alphasets_shape = alphasets_shape
+
+        return True
+
```
lukasheinrich: nit: remove return value, as it holds no semantic meaning

pyhf/tensor/backend.py
lukasheinrich: this can be removed now that we have `.name`

pyhf/interpolate.py
```diff
@@ -52,25 +85,60 @@ def summand(down, nom, up, alpha):
...
+        self.alphasets_shape = None
+        self.histogramssets = histogramssets
+        # initial shape will be (nsysts, 1)
+        self._precompute(alphasets_shape=(histogramssets.shape[0], 1))
+
+    def _precompute(self, alphasets_shape):
+        tensorlib, _ = get_backend()
+
+        # did things change that we need to recompute?
+        tensor_type_change = tensorlib != self.tensorlib
```
lukasheinrich: change to .name comparison

pyhf/interpolate.py
```diff
@@ -20,24 +20,57 @@ def _slow_hfinterp_looper(histogramssets, alphasets, func):
...
```
lukasheinrich: change to .name comparison

pyhf/tensor/backend.py
kratsg: Do you not want to do somthing like `tensorlib == 'numpy'` or you want to do `tensorlib.name == 'numpy'` only?

lukasheinrich: yeah I think attribute access is sufficient, so we don't really need the `__eq__`

kratsg:
@lukasheinrich all resolved

pyhf/interpolate.py
```diff
@@ -1,7 +1,7 @@
 import logging
 log = logging.getLogger(__name__)
 
-from . import get_backend
+from . import get_backend, default_backend
```
lukasheinrich: should we give this another name? we could at some point change the default to TF but we would still not want those tensors in __init__ in TF.. it's like a  "prep backend" / "dark backend" or sth

kratsg: our tests rely on resetting the backend. I suppose if we switch it to tensorflow, then we need to explicitly change it to use numpy... I can't imagine the default backend would ever change to TF though, especially since like mxnet depends on numpy -- and this is all set up based on how our backends get imported (always require numpy).

lukasheinrich: ok let's stay with `default_backend` if we at some point dont' want to use our default for the prep stuff we can revisit later

pyhf/tensor/tensorflow_backend.py
```diff
@@ -44,8 +44,11 @@ def clip(self, tensor_in, min, max):
         return tf.clip_by_value(tensor_in, min, max)
 
     def tolist(self,tensor_in):
-        tensor_in = self.astensor(tensor_in)
-        return self.session.run(tensor_in).tolist()
+        try:
+            return self.session.run(tensor_in).tolist()
+        except AttributeError:
```
lukasheinrich: if there is no session we should raise with a message that tells us what's going on

```
if not self.session: raise RuntimeError('evaluation of tensor requested via .tolist() but no session defined')
```

or sth

PR #291, opened 2018-10-03
Update optimizers to allow for custom iterations for constrained fits

kratsg:
# Description

This PR is to help get ready for #285 by making it more manageable. Update optimizers to allow for constrained fits with a configurable number of iterations and handle exceptions better/more appropriately.

This is a much smaller PR.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

Update optimizers to allow for constrained fits with a configurable number of iterations.

### DIFF
--- a/pyhf/optimize/opt_tflow.py
+++ b/pyhf/optimize/opt_tflow.py
@@ -1,8 +1,16 @@
+import logging
 import numpy as np
 import tensorflow as tf
+from tensorflow.errors import InvalidArgumentError
+
+log = logging.getLogger(__name__)
+
 class tflow_optimizer(object):
     def __init__(self, tensorlib):
         self.tb = tensorlib
+        self.relax = 0.1
+        self.maxit = 1000
+        self.eps = 1e-4
 
     def unconstrained_bestfit(self, objective, data, pdf, init_pars, par_bounds):
         #the graph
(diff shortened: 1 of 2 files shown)


pyhf/optimize/opt_tflow.py
```diff
@@ -35,18 +53,35 @@ def constrained_bestfit(self, objective, constrained_mu, data, pdf, init_pars, p
         nuis_cat = self.tb.concatenate(nuis_pars)
         pars = self.tb.concatenate([nuis_cat[:0],poi_par,nuis_cat[0:]])
         objective = objective(pars,data,pdf)
-        hessian   = tf.hessians(objective, nuis_cat)[0]
+        hessian   = tf.hessians(objective, nuis_cat)[0]+1e-10
```
lukasheinrich: I'm not sure we really want to to this "dark" modification to the hessian that nobody knows is happening. I'd rather try to understand why the fit fails (if we can't figure it out, we should at least print a warning once that we are modifying the hessian). For now, I'd say we should remove this. It was just part of my testing

pyhf/optimize/opt_tflow.py
```diff
@@ -11,18 +19,28 @@ def unconstrained_bestfit(self, objective, data, pdf, init_pars, par_bounds):
...
+                if np.abs(np.max(up)) < self.eps:
+                    break
+            except InvalidArgumentError:
+                o,p,g,h = self.tb.session.run([
+                    objective,
+                    pars,
+                    gradient,
+                    hessian,
+                ], feed_dict={best_fit: best_fit})
+                raise
```
lukasheinrich: this will probably decrease coverage since we are adding lots of lines that we are not hitting in the tests. not sure how to cover it we would need to perhaps have a `validate_hessian` function that we could test stand-alone?

pyhf/optimize/opt_tflow.py
```diff
@@ -35,18 +53,35 @@ def constrained_bestfit(self, objective, constrained_mu, data, pdf, init_pars, p
...
+                    hessian,
+                ], feed_dict={nuis_cat: best_fit_nuis})
+                log.error('----- Constrained Fit -----\nIteration: {}\nObjective: {}\nPars: {}\nGradient: {}\nHessias was: {}'.format(
+                    i,
+                    self.tb.tolist(o),
+                    self.tb.tolist(p),
+                    self.tb.tolist(g),
+                    self.tb.tolist(h),
+                ))
+                raise
```
lukasheinrich: dito, here

lukasheinrich:
thanks @kratsg for continuing to port this stuff over, I added some comments.. if we are not sure how to deal with them, I think it's fine if this PR just adds self.maxiter etc to the optimizers and leaves out the e.g. exception handling and hessian modifications

kratsg:
> thanks @kratsg for continuing to port this stuff over, I added some comments.. if we are not sure how to deal with them, I think it's fine if this PR just adds self.maxiter etc to the optimizers and leaves out the e.g. exception handling and hessian modifications

I've removed the exception handling and hessian modifications as I agree that it's probably not needed for this, but was part of the local testing for right now. Until we can figure out why it happens so often in the optimization, etc...

PR #298, opened 2018-10-06
combined constraints

lukasheinrich:
# Description

Another PR to split up #285. This factors out the constraint pdf computation from the modifiers

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyhf/constraints.py
+++ b/pyhf/constraints.py
@@ -13,7 +13,6 @@ class gaussian_constraint_combined(object):
         self.tensorlib_name = None
         self._precompute()
         tensorlib, _ = get_backend()
-        self.tensorlib_name = tensorlib.name
 
     def _precompute(self):
         tensorlib, _ = get_backend()
@@ -21,6 +20,7 @@ class gaussian_constraint_combined(object):
         tensor_type_change = tensorlib.name != self.tensorlib_name
         if not tensor_type_change:
             return
+        self.tensorlib_name = tensorlib.name
         start_index = 0        
         normal_constraint_data = []
         normal_constraint_mean_indices = []
@@ -78,7 +78,6 @@ class poisson_constraint_combined(object):
         self.tensorlib_name = None
         self._precompute()
         tensorlib, _ = get_backend()
-        self.tensorlib_name = tensorlib.name
 
     def _precompute(self):
         tensorlib, _ = get_backend()
@@ -86,6 +85,7 @@ class poisson_constraint_combined(object):
         tensor_type_change = tensorlib.name != self.tensorlib_name
         if not tensor_type_change:
             return
+        self.tensorlib_name = tensorlib.name
         
         start_index = 0
         poisson_constraint_data = []
--- a/pyhf/constraints.py
+++ b/pyhf/constraints.py
@@ -82,7 +82,7 @@ class poisson_constraint_combined(object):
             poisson_rate_fac  = tensorlib.concatenate(map(lambda x: tensorlib.astensor(x,dtype = 'float'), poisson_constraint_rate_factors))
             poisson_data      = tensorlib.concatenate(map(lambda x: tensorlib.astensor(x,dtype = 'int'), poisson_constraint_data))
         else:
-            poisson_rate_idc, poisson_data = None, None
+            poisson_rate_idc, poisson_data, poisson_rate_fac = None, None, None
         self.prepped = (poisson_data,poisson_rate_idc,poisson_rate_fac)
 
     def logpdf(self,auxdata,pars):
--- a/pyhf/constraints.py
+++ b/pyhf/constraints.py
@@ -42,15 +42,16 @@ class gaussian_constraint_combined(object):
             normal_constraint_mean_indices.append(self.par_indices[modslice])
         
         if normal_constraint_mean_indices:
-            normal_mean_idc  = tensorlib.concatenate(list(map(lambda x: tensorlib.astensor(x,dtype = 'int'),normal_constraint_mean_indices)))
-            normal_sigmas    = tensorlib.concatenate(list(map(tensorlib.astensor,normal_constraint_sigmas)))
-            normal_data      = tensorlib.concatenate(list(map(lambda x: tensorlib.astensor(x,dtype = 'int'),normal_constraint_data)))
-        else:
-            normal_data, normal_sigmas, normal_mean_idc = None, None, None
+            normal_mean_idc  = default_backend.concatenate(list(map(lambda x: default_backend.astensor(x,dtype = 'int'),normal_constraint_mean_indices)))
+            normal_sigmas    = default_backend.concatenate(list(map(default_backend.astensor,normal_constraint_sigmas)))
+            normal_data      = default_backend.concatenate(list(map(lambda x: default_backend.astensor(x,dtype = 'int'),normal_constraint_data)))
 
-        self.normal_data = normal_data
-        self.normal_sigmas = normal_sigmas
-        self.normal_mean_idc = normal_mean_idc
+            self.normal_data = tensorlib.astensor(default_backend.tolist(normal_data),dtype = 'int')
+            self.normal_sigmas = tensorlib.astensor(default_backend.tolist(normal_sigmas))
+            self.normal_mean_idc = tensorlib.astensor(default_backend.tolist(normal_mean_idc),dtype = 'int')
+        else:
+            self.normal_data, self.normal_sigmas, self.normal_mean_idc = None, None, None
+
         self.prepped = True
 
     def logpdf(self,auxdata,pars):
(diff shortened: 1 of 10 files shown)


pyhf/constraints.py
```diff
@@ -0,0 +1,87 @@
...
+            normal_constraint_mean_indices.append(par_indices[modslice])
+        
+        if normal_constraint_mean_indices:
+            normal_mean_idc  = tensorlib.concatenate(map(lambda x: tensorlib.astensor(x,dtype = 'int'),normal_constraint_mean_indices))
+            normal_sigmas    = tensorlib.concatenate(map(tensorlib.astensor,normal_constraint_sigmas))
+            normal_data      = tensorlib.concatenate(map(lambda x: tensorlib.astensor(x,dtype = 'int'),normal_constraint_data))
+        else:
+            normal_data, normal_sigmas, normal_mean_idc = None, None, None
+
+        self.prepped =  (normal_data,normal_sigmas,normal_mean_idc)
```
kratsg: replace with `self.normal_data, self.normal_sigmas, self.normal_mean_idc` here. `self.prepped` is not very useful on its own and vague. Should also be called more like `normal_data_idx`, `normal_sigma_idx`, and `normal_mean_idx` or something.

pyhf/constraints.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def logpdf(self,auxdata,pars):
+        if self.prepped[0] is None:
+            return 0
+        tensorlib, _ = get_backend()
+        normal_data   = tensorlib.gather(auxdata,self.prepped[0])
+        normal_means  = tensorlib.gather(pars,self.prepped[2])
+        normal_sigmas = self.prepped[1]
+        normal = tensorlib.normal_logpdf(normal_data,normal_means,normal_sigmas)
+        return tensorlib.sum(normal)
```
kratsg: `return normal` instead of `tensorlib.sum(normal)`? You do same for poisson, so it makes more sense to do `tensorlib.sum(normal+poisson)` instead perhaps...

pyhf/pdf.py
```diff
@@ -131,6 +132,9 @@ def __init__(self, spec, **config_kwargs):
         # build up our representation of the specification
         self.config = _ModelConfig.from_spec(self.spec,**config_kwargs)
 
+        self.prepped_constraints_gaussian = gaussian_constraint_combined(self)
```
kratsg: i don't like passing in the `pdf` here, but maybe it's fine. It just makes the call look awkward of `self.... = constraint_combined(self)`.

pyhf/pdf.py
```diff
@@ -379,20 +383,9 @@ def expected_data(self, pars, include_auxdata=True):
...
-            modalphas = modifier.alphas(pars[modslice])
-            end_index = start_index + int(modalphas.shape[0])
-            thisauxdata = auxdata[start_index:end_index]
-            start_index = end_index
-            constraint_term = tensorlib.log(modifier.pdf(thisauxdata, modalphas))
-            summands = constraint_term if summands is None else tensorlib.concatenate([summands,constraint_term])
-        return tensorlib.sum(summands) if summands is not None else 0
+        normal  = self.prepped_constraints_gaussian.logpdf(auxdata,pars)
+        poisson = self.prepped_constraints_poisson.logpdf(auxdata,pars)
+        return normal + poisson
```
kratsg: is it really just a sum? I thought it was multiplied together.

lukasheinrich: it's logpdf, so sum is correct

pyhf/constraints.py
lukasheinrich: good point

pyhf/constraints.py
lukasheinrich: yeah we might save an op, but then logpdf is not a correct term since it is not a scalar anymore

pyhf/pdf.py
lukasheinrich: I think we only depend on pdf config.. so we could change that to `gaussian_constraint_combined(self.config)`

pyhf/constraints.py
```diff
@@ -0,0 +1,131 @@
...
+        self.mod_and_slice = [
+            (pdfconfig.modifier(cname),pdfconfig.par_slice(cname))
+            for cname in pdfconfig.auxdata_order
+        ]
+
+    def _precompute(self):
+        tensorlib, _ = get_backend()
+        # did things change that we need to recompute?
+        tensor_type_change = tensorlib.name != self.tensorlib_name
+        if (not tensor_type_change) and self.prepped:
```
kratsg: This logic is kinda odd to me. You're precomputing if the tensor type changes or if not prepped. You can get rid of prepped if you initialize the tensor type to "None" and call `_precompute()` in the `__init__()` which catches a tensor type change. Technically, it doesn't matter if you're prepped or not if the tensor type changes.

pyhf/constraints.py
```diff
@@ -0,0 +1,131 @@
...
+        normal_constraint_sigmas = []
+        for modifier,modslice in self.mod_and_slice:
+            end_index = start_index + modifier.n_parameters
+            thisauxdata = self.data_indices[start_index:end_index]
+            start_index = end_index
+            if not modifier.pdf_type == 'normal': continue
+
+            # print('combined gauss: {}/{}/{}'.format(
+            #     thisauxdata,par_indices[modslice],modifier.sigmas)
+            # )
```
kratsg: extra comments here to be removed.

pyhf/constraints.py
```diff
@@ -0,0 +1,131 @@
...
+            try:
```
kratsg: need a comment to explain why this try/except is needed

pyhf/constraints.py
```diff
@@ -0,0 +1,131 @@
...
+        for modifier,modslice in self.mod_and_slice:
+            end_index = start_index + modifier.n_parameters
+            thisauxdata = self.data_indices[start_index:end_index]
+            start_index = end_index
+            if not modifier.pdf_type == 'poisson': continue
+
+            poisson_constraint_data.append(thisauxdata)
+            poisson_constraint_rate_indices.append(self.par_indices[modslice])
+
+            try:
```
kratsg: need a comment here for this.

pyhf/modifiers/histosys.py
```diff
@@ -42,6 +42,8 @@ def expected_data(self, pars):
 
     def pdf(self, a, alpha):
         tensorlib, _ = get_backend()
+        # for d,m in zip(a,alpha):
+        #     print('slow hst data: {} mean: {} sigma: {}'.format(d,m,1))
```
kratsg: extra commented lines here

pyhf/modifiers/normsys.py
```diff
@@ -31,6 +31,8 @@ def expected_data(self, pars):
 
     def pdf(self, a, alpha):
         tensorlib, _ = get_backend()
+        # for d,m in zip(a,alpha):
+        #     print('slow norm data: {} mean: {} sigma: {}'.format(d,m,1))
```
kratsg: extra commented lines here

pyhf/modifiers/shapesys.py
```diff
@@ -26,6 +26,8 @@ def alphas(self, pars):
 
     def pdf(self, a, alpha):
         tensorlib, _ = get_backend()
+        # for d,r in zip(a,alpha):
+        #     print('slow shp data: {} rate: {}'.format(d,r))
```
kratsg: extra commented lines here

pyhf/modifiers/staterror.py
```diff
@@ -27,6 +36,9 @@ def pdf(self, a, alpha):
         inquad = tensorlib.sqrt(tensorlib.sum(tensorlib.power(self.uncertainties,2), axis=0))
         totals = tensorlib.sum(self.nominal_counts,axis=0)
         uncrts = tensorlib.divide(inquad,totals)
+        assert self.sigmas.tolist() == uncrts.tolist()
+        # for d,m,s in zip(a,alpha,uncrts):
+        #     print('slow stat data: {} mean: {} sigma: {}'.format(d,m,s))
```
kratsg: extra commented lines here

pyhf/modifiers/staterror.py
```diff
@@ -27,6 +36,9 @@ def pdf(self, a, alpha):
         inquad = tensorlib.sqrt(tensorlib.sum(tensorlib.power(self.uncertainties,2), axis=0))
         totals = tensorlib.sum(self.nominal_counts,axis=0)
         uncrts = tensorlib.divide(inquad,totals)
+        assert self.sigmas.tolist() == uncrts.tolist()
```
kratsg: this assert should be moved to a test statement, no?

pyhf/pdf.py
```diff
@@ -131,6 +132,17 @@ def __init__(self, spec, **config_kwargs):
         # build up our representation of the specification
         self.config = _ModelConfig.from_spec(self.spec,**config_kwargs)
 
+
+        for m in self.config.modifiers:
```
kratsg: probably require a `def finalize(self): pass` on all other modifiers to be honest.

kratsg:
Overall, looks a lot better. I'm still not sure if we should call them "constraints_combined". Are there non-combined versions in our code? If not, probably just remove the word "combined".

pyhf/pdf.py
lukasheinrich: hm not sure, so far we only have a single one that needs finalization. I'm not sure we will see more

lukasheinrich:
the `combined` refers to the fact that all gaussian constraints from a single modifier are calculated in one go, but yeah we don't specifically point this out in the interpolators either (where we compute multiple systs in one go, so maybe we can remove combined

pyhf/constraints.py
```diff
@@ -12,7 +10,10 @@ def __init__(self,pdfconfig):
             (pdfconfig.modifier(cname),pdfconfig.par_slice(cname))
             for cname in pdfconfig.auxdata_order
         ]
+        self.tensorlib_name = None
         self._precompute()
+        tensorlib, _ = get_backend()
+        self.tensorlib_name = tensorlib.name
```
kratsg: these two lines go into `_precompute()`. If the tensor type changes, you need to precompute, and then set the `self.tensorlib_name` again.

pyhf/constraints.py
```diff
@@ -76,7 +75,10 @@ def __init__(self,pdfconfig):
...
```
kratsg: same comment as above.

pyhf/pdf.py
kratsg: that's probably a good point. Can't the `finalize` in staterror just be part of the initialization and it provides a `self.sigmas` that you need to extract?

kratsg:
Everything looks good to me, except for maybe the name `combined` which we can deal w/ later and whether staterror needs a finalize or if it can't just go into the `init` or something.

PR #300, opened 2018-10-08
begin separation of constraints and modifiers

lukasheinrich:
# Description

This PR will separate modifier and contraint codes since some modifiers can share the same constraint (e.g. the qualify_names issue)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyhf/constraints.py
+++ b/pyhf/constraints.py
@@ -2,8 +2,12 @@ from . import get_backend, default_backend
 
 
 class standard_gaussian_constraint(object):
-    def __init__(self):
+    def __init__(self, n_parameters, inits, bounds, auxdata):
         self.pdf_type = 'normal'
+        self.n_parameters = n_parameters
+        self.suggested_init = inits
+        self.suggested_bounds = bounds
+        self.auxdata = auxdata
 
     def alphas(self, pars):
         return pars  # the nuisance parameters correspond directly to the alpha
@@ -12,9 +16,14 @@ class standard_gaussian_constraint(object):
         return self.alphas(pars)
 
 class factor_poisson_constraint(object):
-    def __init__(self, factors):
+    def __init__(self, n_parameters, inits, bounds, auxdata, factors):
         self.factors = factors
         self.pdf_type = 'poisson'
+        self.n_parameters = n_parameters
+        self.suggested_init = inits
+        self.suggested_bounds = bounds
+        self.auxdata = auxdata
+
     def alphas(self, pars):
         tensorlib, _ = get_backend()
         return tensorlib.product(tensorlib.stack([pars, tensorlib.astensor(self.factors)]), axis=0)
@@ -127,7 +136,7 @@ class poisson_constraint_combined(object):
             # with tau*b). If such a scale factor is not defined we just
             # take a factor of one
             try:
-                poisson_constraint_rate_factors.append(modifier.bkg_over_db_squared)
+                poisson_constraint_rate_factors.append(modifier.constraint.factors)
             except AttributeError:
                 poisson_constraint_rate_factors.append(default_backend.shape(self.par_indices[modslice]))
 
--- a/pyhf/constraints.py
+++ b/pyhf/constraints.py
@@ -17,12 +17,12 @@ class standard_gaussian_constraint(object):
 
 class factor_poisson_constraint(object):
     def __init__(self, n_parameters, inits, bounds, auxdata, factors):
-        self.factors = factors
         self.pdf_type = 'poisson'
         self.n_parameters = n_parameters
         self.suggested_init = inits
         self.suggested_bounds = bounds
         self.auxdata = auxdata
+        self.factors = factors
 
     def alphas(self, pars):
         tensorlib, _ = get_backend()
--- a/pyhf/constraints.py
+++ b/pyhf/constraints.py
@@ -17,12 +17,12 @@ class standard_gaussian_constraint(object):
 
 class factor_poisson_constraint(object):
     def __init__(self, n_parameters, inits, bounds, auxdata, factors):
+        self.factors = factors
         self.pdf_type = 'poisson'
         self.n_parameters = n_parameters
         self.suggested_init = inits
         self.suggested_bounds = bounds
         self.auxdata = auxdata
-        self.factors = factors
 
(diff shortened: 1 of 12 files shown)


kratsg:
I think I understand what's going on here -- since you're separating out the initialization/alphas of the constraints themselves from the modifiers. So each modifier has a constraint object that it refers to. Rather than duping code across modifiers?

A naive question is wondering why not go more OOP and creating different types of constraints that modifiers inherit from instead of setting it as the constraint object on the modifier itself? E.G. `class histosys(standard_gaussian_constraint)`

kratsg:
> @kratsg we do composition instead of inheritance precisely to address the qualify_names issue. Multiple modifier instances will point to the same constraint instance. Inheritance would keep the 1<->1 mapping

Only if you don't do meta-classing there to register things. But how do you ensure the same constraint instance?

kratsg:
I still don't see how you have a 1<->1 mapping since histosys would never have the poisson constraint, for example - based on the way this is refactored.

lukasheinrich:
this is not done yet. Right now each modifier does create its own instance (keeping 1<-> 1) .. in the end, we will inject the constraint object into the modifier on creation-time (where there will be some checks to assure that the constraint object is an instance of the constraint type that the modifier expects and that the dimensions are correct)

kratsg:
> this is not done yet. Right now each modifier does create its own instance (keeping 1<-> 1) .. in the end, we will inject the constraint object into the modifier on creation-time (where there will be some checks to assure that the constraint object is an instance of the constraint type that the modifier expects and that the dimensions are correct)

I mean, it's still much easier, I think, to handle this with singletons instead to some extent, rather than trying to inject the function into the code from outside of the modifier code itself. Otherwise, this will start to get messy.

pyhf/constraints.py
```diff
@@ -1,5 +1,36 @@
 from . import get_backend, default_backend
 
+
+class standard_gaussian_constraint(object):
```
kratsg: call this "gaussian_constraint"

pyhf/constraints.py
```diff
@@ -1,5 +1,36 @@
...
+    def __init__(self, n_parameters, inits, bounds, auxdata):
+        self.pdf_type = 'normal'
```
kratsg: do we want to change `pdf_type` from `normal` to `gaussian`?

pyhf/constraints.py
```diff
@@ -1,5 +1,36 @@
...
+        self.n_parameters = n_parameters
+        self.suggested_init = inits
+        self.suggested_bounds = bounds
+        self.auxdata = auxdata
+
+    def alphas(self, pars):
+        return pars  # the nuisance parameters correspond directly to the alpha
```
kratsg: change this to the function docstring

pyhf/constraints.py
```diff
@@ -1,5 +1,36 @@
...
+
+    def expected_data(self, pars):
+        return self.alphas(pars)
+
+class factor_poisson_constraint(object):
```
kratsg: call this `poisson_constraint`

pyhf/constraints.py
```diff
@@ -1,5 +1,36 @@
...
+    def __init__(self, n_parameters, inits, bounds, auxdata, factors):
+        self.factors = factors
+        self.pdf_type = 'poisson'
```
kratsg: self.pdf_type should be first (to mirror `gaussian_constraint`)

pyhf/constraints.py
```diff
@@ -1,5 +1,36 @@
...
+        self.n_parameters = n_parameters
+        self.suggested_init = inits
+        self.suggested_bounds = bounds
+        self.auxdata = auxdata
+
+    def alphas(self, pars):
+        tensorlib, _ = get_backend()
+        return tensorlib.product(tensorlib.stack([pars, tensorlib.astensor(self.factors)]), axis=0)
+    def expected_data(self, pars):
```
kratsg: missing a newline before this

pyhf/constraints.py
```diff
@@ -1,5 +1,36 @@
...
+class factor_poisson_constraint(object):
+    def __init__(self, n_parameters, inits, bounds, auxdata, factors):
+        self.factors = factors
+        self.pdf_type = 'poisson'
+        self.n_parameters = n_parameters
+        self.suggested_init = inits
+        self.suggested_bounds = bounds
+        self.auxdata = auxdata
+
+    def alphas(self, pars):
```
kratsg: need a comment on why you're doing the product/stack thing (maybe explain the math here)

pyhf/paramsets.py
```diff
@@ -0,0 +1,38 @@
+from . import get_backend
+class param_set(object):
```
kratsg: call this `paramset`

pyhf/paramsets.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self, n_parameters, inits, bounds):
+        self.n_parameters = n_parameters
+        self.suggested_init = inits
+        self.suggested_bounds = bounds
+
+class normal_constrained_set(param_set):
```
kratsg: call this `constrained_normal_paramset`

pyhf/paramsets.py
```diff
@@ -0,0 +1,38 @@
...
+        self.auxdata = auxdata
+
+    def alphas(self, pars):
+        '''the nuisance parameters correspond directly to the alpha'''
+        return pars  
+
+    def expected_data(self, pars):
+        return self.alphas(pars)
+
+class poisson_constrained_set(param_set):
```
kratsg: call this `constrained_poisson_paramset`

pyhf/paramsets.py
```diff
@@ -0,0 +1,38 @@
...
+        tensorlib, _ = get_backend()
+        return tensorlib.product(
+                tensorlib.stack([pars, tensorlib.astensor(self.factors)]
+            ),
+            axis=0
+        )
+
+    def expected_data(self, pars):
+        return self.alphas(pars)
+
```
kratsg: add a `class unconstrained_paramset(paramset)` and use that instead of the `paramset` class to help with naming.

PR #304, opened 2018-10-10
Add version info through __version__

matthewfeickert:
# Description

Add version information to the code in the form of `__version__` so that 

```
pyhf.__version__
```

returns information. So for example

**Edit below:** `0.0.16.dev` to `0.0.15`
```
>>> import pyhf
>>> pyhf.__version__
'0.0.15'
```
~~(implicit here is that as `0.0.15` is the current PyPI release that what is currently in `master` is the next release's development version. There is probably a more industry standard or sane way to do this then what I just proposed here.)~~

~~The [`version.py` file added here is copied from TensorFlow Probability](https://github.com/tensorflow/probability/blob/f8bd785031340f2231ed0e0f01c378c03ac3ce52/tensorflow_probability/python/version.py).~~

[`bumpversion`](https://github.com/peritus/bumpversion) controls the actual version number in all of the code to simplify the process.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -1,6 +1,5 @@
 from . import tensor, optimize
 from .version import __version__
-__all__ = ['__version__']  # Used to satisfy pyflakes
 tensorlib = tensor.numpy_backend()
 default_backend = tensorlib
 optimizer = optimize.scipy_optimizer()
@@ -59,4 +58,4 @@ def set_backend(backend, custom_optimizer=None):
 
 
 from .pdf import Model
-__all__ = ["Model", "utils", "modifiers"]
+__all__ = ['Model', 'utils', 'modifiers', '__version__']
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -1,6 +1,6 @@
 from . import tensor, optimize
 from .version import __version__
-assert __version__  # Used to satisfy pyflakes
+__all__ = ['__version__']  # Used to satisfy pyflakes
 tensorlib = tensor.numpy_backend()
 default_backend = tensorlib
 optimizer = optimize.scipy_optimizer()
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -1,4 +1,5 @@
 from . import tensor, optimize
+from .version import __version__
 tensorlib = tensor.numpy_backend()
 default_backend = tensorlib
 optimizer = optimize.scipy_optimizer()
--- a/pyhf/__init__.py
+++ b/pyhf/__init__.py
@@ -1,5 +1,6 @@
 from . import tensor, optimize
 from .version import __version__
+assert __version__  # Used to satisfy pyflakes
 tensorlib = tensor.numpy_backend()
 default_backend = tensorlib
 optimizer = optimize.scipy_optimizer()
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,14 @@
 #!/usr/bin/env python
 
+import os
+import sys
 from setuptools import setup, find_packages
-from pyhf.version import __version__
+
+# Import directly from version.py by adding to sys.path
+# Otherwise setup won't know what the module as pyhf isn't installed yet
+version_path = os.path.join(os.path.dirname(__file__), 'pyhf')
+sys.path.append(version_path)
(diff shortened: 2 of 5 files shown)


matthewfeickert:
@lukasheinrich, I assume that this will break bumpversion with how it is currently setup. I haven't used it before in the past, so is there a way to point `.bumpversion.cfg` to target `pyhf/version.py`? If not, I guess the information that is in `pyhf/version.py` can be moved to `.bumpversion.cfg` (or maybe can just be replaced by it?)

matthewfeickert:
Ah, `pyflakes` is currently breaking because of the following

```
pyhf/__init__.py:2: '.version.__version__' imported but unused
```

which I'll fix later (maybe though with just an ignore statement to `pyflakes`).

lukasheinrich:
@matthewfeickert you can give bumpversion a list of files it modifies on each bump. I think just adding a line

```
[bumpversion:file:pyhf/version.py]
```

can suffice. The unused error can be circumvented by just `assert`'ing the module or value

setup.py
```diff
@@ -1,7 +1,14 @@
...
 
+import os
+import sys
 from setuptools import setup, find_packages
 
+# Import directly from version.py by adding to sys.path
+version_path = os.path.join(os.path.dirname(__file__), 'pyhf')
+sys.path.append(version_path)
+from version import __version__
+
```
lukasheinrich: do we need this? shouldn't bump version just ensure for us that the various files are set to the same value?

kratsg: This shouldn't be needed. You should be able to do `from pyhf.version import __version__`

matthewfeickert:
> you can give bumpversion a list of files it modifies on each bump

Okay, so given that then we really don't need all of this chunck

https://github.com/diana-hep/pyhf/blob/8aab036f1608faf96e48afb6f0fe57309d6e736a/pyhf/version.py#L3-L18

we can just have this all replaced with

```
__version__ = whatever_the_version_number_will_be
```

and then `whatever_the_version_number_will_be` will just be taken care of by `bumpversion`, correct?

setup.py
matthewfeickert: I think this is needed, as in a clean install it doesn't know about `pyhf` yet. [This can be seen in the Travis builds with errors like this one](https://travis-ci.org/diana-hep/pyhf/jobs/439806439#L488-L489):

```
$ pip install --ignore-installed -U -q -e .[complete]
Command "python setup.py egg_info" failed with error code 1 in /home/travis/build/diana-hep/pyhf/
```

kratsg: Python, by default, will treat any folder with `__init__.py` as a module, even if it has not been installed. So I'm not sure why that won't work.

matthewfeickert: > shouldn't bumpversion just ensure for us that the various files are set to the same value?

Oh yeah. This would be better.

pyhf/__init__.py
```diff
@@ -1,4 +1,6 @@
 from . import tensor, optimize
+from .version import __version__
+assert __version__  # Used to satisfy pyflakes
```
kratsg: remove the assert and add `__all__ = ['__version__']` which should also make pyflakes happy, but is more correct.

pyhf/__init__.py
```diff
@@ -1,4 +1,6 @@
 from . import tensor, optimize
+from .version import __version__
+__all__ = ['__version__']  # Used to satisfy pyflakes
```
kratsg: there's already an `__all__` in `__init__.py` -- you should add it in there. This will just get overridden.

matthewfeickert:
Proposed PR commit log:

>- Add pyhf/version.py to add `pyhf.__version__`
>- Add pyhf/version.py to .bumpversion.cfg to continue to have bumpversion control the advancement of version numbers
>- Add `pyhf.__version__` to requested environment information for bug reports

pyhf/__init__.py
matthewfeickert: Yes. Sorry. Was rushing.

lukasheinrich:
@matthewfeickert maybe as a follow up PR, we could try adding a `--version` flag to pyhf?

PR #305, opened 2018-10-10
Add additional pdf information to aid with refactoring

kratsg:
# Description

This is to aid #285. Add additional information on the PDF config to aid with refactoring.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -21,35 +21,32 @@ class _ModelConfig(object):
         self.channels = []
         self.samples = []
         self.modifiers = []
-        self.modifiers_type = {}
         self.channel_nbins = {}
         # hacky, need to keep track in which order we added the constraints
         # so that we can generate correctly-ordered data
         for channel in spec['channels']:
-            channels.append(channel['name'])
-            channel_nbins[channel['name']] = len(channel['samples'][0]['data'])
+            self.channels.append(channel['name'])
+            self.channel_nbins[channel['name']] = len(channel['samples'][0]['data'])
             for sample in channel['samples']:
-                samples.append(sample['name'])
+                self.samples.append(sample['name'])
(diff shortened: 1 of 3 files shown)


pyhf/pdf.py
```diff
@@ -35,10 +38,13 @@ def from_spec(cls,spec,poiname = 'mu', qualify_names = False):
...
                     modifier.add_sample(channel, sample, modifier_def)
                     modifiers.append(modifier_def['name'])
+                    modifiers_type[modifier_def['name']] = modifier_def['type']
                     sample['modifiers_by_type'].setdefault(modifier_def['type'],[]).append(modifier_def['name'])
         instance.channels = list(set(channels))
         instance.samples = list(set(samples))
         instance.modifiers = list(set(modifiers))
+        instance.modifiers_type = modifiers_type
+        instance.channel_nbins = channel_nbins
         instance.set_poi(poiname)
```
lukasheinrich: I wonder if we should take the opportunity to move all the attribute-assigning things into the `__init__` of `Model` ?

kratsg: I'm not fully convinced that it belongs in `Model` because it's a private member variable. There's no reason for anyone to publicly access this -- which is why I think we agreed to set it on the underlying model config.

lukasheinrich: sorry I did not mean `Model` but config.. e.g. I was thinking we would pass these into init

```
instance = cls(channels samples, modifiers, poiname)
```

etc

kratsg:
This is a compromise to getting rid of qualify_names -- we will add `pdf.config.modifiers` and `pdf.config.parameters` -- the former which is qualified-named if needed, but the latter is never. This means that paramsets can refer to `pdf.config.parameters` while we slowly get rid of the `modifiers`.

PR #311, opened 2018-10-14
Add a way to subscribe/trigger events globally

kratsg:
# Description

This is to help with #285 which needs an easy way to trigger recomputation in the case where the tensorlib changes. Since this needs to be somewhat globally-triggering with callbacks, it's easier to go for some sort of sub/pub model. I've implemented a bare-bones version that should handle all we need to do.

The idea is to do something like

```python
def normsys_combined(self):
    def __init__(self):
        pass

    @pyhf.events.subscribe('tensorlib_change')
    def _precompute(self):
        # if this is called, the backend was changed so do something
```

and then `pyhf/__init__.py#set_backend` can simply call `pyhf.events.trigger('tensorlib_change')()` to trigger the change. There's zero checks on the arguments passed through to function calls, but it is flexible and can allow for passing through arguments from the trigger to the callbacks as long as the callbacks can handle the arguments.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyhf/events.py
+++ b/pyhf/events.py
@@ -1,6 +1,3 @@
-import logging
-log = logging.getLogger(__name__)
-
 __events = {}
 __disabled_events = set([])
 class Events(list):
--- a/pyhf/events.py
+++ b/pyhf/events.py
@@ -1,12 +1,12 @@
 __events = {}
 __disabled_events = set([])
-class Events(list):
+class Callables(list):
     def __call__(self, *args, **kwargs):
         for f in self:
             f(*args, **kwargs)
 
     def __repr__(self):
-        return "Events(%s)" % list.__repr__(self)
+        return "Callables(%s)" % list.__repr__(self)
 
 """
 
@@ -22,7 +22,7 @@ class Events(list):
 def subscribe(event):
     global __events
     def __decorator(func):
-        __events.setdefault(event, Events()).append(func)
+        __events.setdefault(event, Callables()).append(func)
         return func
     return __decorator
 
--- a/pyhf/events.py
+++ b/pyhf/events.py
@@ -1,0 +1,55 @@
+import sys
+
+# this is meant to wrap the Wrapper object
+class Wrapper(object):
+    __events = {}
+    class Events(list):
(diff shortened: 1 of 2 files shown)


kratsg:
NB: you'll probably wonder why the hell @kratsg wrote the `pyhf/events.py` module that way -- refer to this mail archive: https://mail.python.org/pipermail/python-ideas/2012-May/014969.html.

pyhf/events.py
```diff
@@ -0,0 +1,56 @@
...
+    class Events(list):
+        def __call__(self, *args, **kwargs):
+            for f in self:
+                f(*args, **kwargs)
+
+        def __repr__(self):
+            return "Events(%s)" % list.__repr__(self)
+
+    def __init__(self, name, filepath):
+        import logging
```
lukasheinrich: move import up

kratsg:
Example full-use case with this PR

```python
>>> import pyhf.events
>>> @pyhf.events.subscribe('test')
... def hello():
...   print 'hello world'
... 
>>> hello
<function hello at 0x1283d79b0>
>>> pyhf.events
<module 'pyhf.events' from 'pyhf/events.py'>
>>> pyhf.events.trigger('hello')
<function _trigger at 0x1283d7a28>
>>> pyhf.events.trigger('hello')()
>>> pyhf.events.trigger('test')
<function _trigger at 0x1283d7aa0>
>>> pyhf.events.trigger('test')()
hello world
>>> pyhf.events.disable('test')
>>> pyhf.events.trigger('test')
<function noop at 0x1283d7938>
>>> pyhf.events.trigger('test')()
>>> pyhf.events.enable('test')
>>> pyhf.events.trigger('test')
<function _trigger at 0x1283d7b18>
>>> pyhf.events.trigger('test')()
hello world
```

kratsg:
PR summary

```
- provide a global pub/sub with pyhf.events.subscribe / pyhf.events.trigger
- additionally allow for enabling/disabling specific events
- pyhf.events.noop for the cases where you need a noop function call
```

PR #316, opened 2018-10-16
[WIP] fix validation tests

### LINKED ISSUE #312: add more intermediate tests of the likelihood computation for more insight where validation fails (if it fails) or numerical precision plays a role
# Description

as part of the refactoring #285 we are running into some issues where we at some places we fail the existing tests due to numerical issues. As the numbers we compare are mostly the overall results of the CLs computation (which includes minimization etc) it would be good to add more tests for intermediate stages in order to be able to pin down where the sensitivity is introduced..

possible tests are: `pdf.expected_auxdata(..)`, `pdf.expected_actualdata(pars)` `pdf.mainlogpdf`, `pdf.constraint_logpdf` etc

kratsg:
# Description

This (will) resolves #312. Using `git bisect`, it was determined that by changing the continuous approximation to the poisson to using gamma (0eddbf9956d6a991cf22fb9eec9e2e1a7b3df618) caused the validation test to diverge more than expected. This is because the expected results were now invalid (made with a different continuous approximation) and needed to be updated.

This only came to light in #285 because the `abs()` of the difference wasn't taken causing the tests to pass when they really shouldn't have (negative numbers are always less than any positive tolerance one can define). Now one needs to figure out why the tests are not currently passing in master with `abs()` additionally added, even after updating the expected results.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -49,7 +49,7 @@ def spec_1bin_shapesys(source=source_1bin_example1()):
 def expected_result_1bin_shapesys(mu=1.):
     if mu == 1:
         expected_result = {
-            'obs': 0.4541865416107029,
+           'obs': 0.4541865416107029,
             'exp': [
                 0.06371799398864626,
                 0.15096503398048894,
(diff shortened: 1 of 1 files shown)


kratsg:
closed in favor of #317.

PR #317, opened 2018-10-16
fix validation data and tests

lukasheinrich:
# Description

This fixes the validation Data. In #45 a better poisson approximation was used and in #47 the histosys sharing was fixed but the validation data was from runs before these. but we also forgot checking the absolute difference in the tests and thus missed failing tests since then

The source of the validation data is the commit of #47 i.e. 0eddbf9 with this diff applied https://gist.github.com/lukasheinrich/03323eb9adcac1ef709e4b23ba496ee1

- [x] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -639,7 +639,7 @@ def setup_2bin_2channel_coupledshapefactor(
     }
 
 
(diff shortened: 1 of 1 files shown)


lukasheinrich:
we further improved the validation data by incorporating the change of #153 which has a 1e-5 effect on CLs in some tests

Issue #264, opened 2018-09-16
Suggestion: expose model library on command line? 'pyhf model --template'


lukasheinrich:
# Description

We keep a small (ok, it's one entry) library of example models in `pyhf.simplemodels`, which makes it easy to get started (see the nice example we have in `README`)

It might be interesting to expose this on the command line e.g. a `pyhf model` command that dumps a fully specified pyhf spec from a simpler input JSON (which we don't really validate, it's not another spec, but just a short-cut, we should keep this dead-simple key-value JSON without any nesting)

I'm not sure how I feel about this, so curious about inputs

```
pyhf model --template hepdata_like --signal 5.0 --bkg 50. --bkg_uncrt 7. | pyhf cls
```

```
cat << EOF| pyhf model --template hepdata_like |pyhf cls
{ 
"signal": 50.,
"bkg": 50.,
"bkg_uncrt": 7. 
}
EOF
```

matthewfeickert:
I was thinking about something similar (though not as well formed like this) today while writing some slides. So I would be in favor of it. :+1:

lukasheinrich:
@kratsg ?

lukasheinrich:
My thinking was that they'll should all take the same data format (JSON, I guess), but not necessarily the same contents. The implementation should always be just a single function like this one

https://github.com/diana-hep/pyhf/blob/master/pyhf/simplemodels.py#L3

which would document what the expected input is (could dump that help string via e.g. `pyhf model --template <name> --help`)

if this grows beyond a small handful of models (like the counting experiments you mention) it should probably be factored out (similar discussion as w/ visualization etc)

PR #321, opened 2018-10-18
store tests in a jsonnet library

lukasheinrich:
# Description

A nice way to store spec templates is to use jsonnet  https://github.com/google/jsonnet

I added this here as a demo for the tests, but actually I was thinking maybe we could use this in the CLI as well

this works

```
cat << EOF| jsonnet - --ext-code-file templates=validation/templates.libsonnet|pyhf cls
std.extVar("templates").simple {
    source:: { bindata: {
        sig: [10], bkg: [50], data: [51]
    }}
}
EOF
```

e.g. the tempates in `pyhf.simplemodels` could acutally be jsonnet templates

related to #264 

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,7 @@ extras_require = {
         'nbsphinx',
         'm2r',
         'jsonpatch',
+        'jsonnet',
         'ipython<7',  # jupyter_console and ipython clash in dependency requirement -- downgrade ipython for now
     ]
 }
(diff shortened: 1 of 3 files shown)


kratsg:
Are we saving a lot of extra code with the addition of another library here? It almost looks 1-1. It's not clear to me if we benefit strongly from this now, or if we switch to this now because it will pay off in the future.

kratsg:
I still want to see something like this happen, to be honest.

PR #323, opened 2018-10-19
Refactor/megachannel bookkeep paramsets

kratsg:
experimental development here


pyhf/modifiers/combined_mods.py
```diff
@@ -241,6 +240,8 @@ def __init__(self,shapefactor_mods,pdfconfig,mega_mods):
                 for s in pdfconfig.samples
             ] for m in shapefactor_mods
         ]
+        channel_bin_index_map = [j for c in pdfconfig.channels for j in range(pdfconfig.channel_nbins[c])]
```
kratsg: this tells us how to map the given parameter indices to the corresponding bins. For example..

```
[
 [c1b1,c1b2,c2b1c2b2,c2b3]  << signal
 [c1b1,c1b2,c2b1c2b2,c2b3]  << bkg1
 [c1b1,c1b2,c2b1c2b2,c2b3]  << background
]
```

ROOT will make nuispars: `gamma_shapefac_bin_1,gamma_shapefac_bin_2,gamma_shapefac_bin_3`. The 2 bins of the one channel will be affected by: `gamma_shapefac_bin_1,gamma_shapefac_bin_2` and the three bins of the other by: `gamma_shapefac_bin_1,gamma_shapefac_bin_2,gamma_shapefac_bin_3`

We need to count the number of bins per channel in the right order using `pdfconfig.channels` and generate indices for this using `range(num_bins_in_channel)`.

pyhf/modifiers/combined_mods.py
```diff
@@ -241,6 +240,8 @@ def __init__(self,shapefactor_mods,pdfconfig,mega_mods):
...
+        self._shapefactor_indices = [default_backend.tolist(default_backend.gather(default_backend.astensor(shapefactor_mod_parindices, dtype='int'), channel_bin_index_map)) for shapefactor_mod_parindices in self._shapefactor_indices]
```
kratsg: And then update shapefactor_indices by re-gathering with the corresponding parameter indices using the channel-bin index mapping.

pyhf/modifiers/combined_mods.py
```diff
@@ -258,7 +259,11 @@ def apply(self,pars):
         if not tensorlib.shape(shapefactor_indices)[0]:
             return
         shapefactors = tensorlib.gather(pars,shapefactor_indices)
-        results_shapefactor = shapefactor_mask * tensorlib.reshape(shapefactors,tensorlib.shape(shapefactors) + (1,1))
+
+        # needs to multiply across samples/mask value so (a,b) -> (a, 1, 1, b)
+        new_shape = list(tensorlib.shape(shapefactors))
+        new_shape[1:1] = [1,1]
+        results_shapefactor = shapefactor_mask * tensorlib.reshape(shapefactors, new_shape)
```
kratsg: this took some playing around, but for N shapefactors applied to M bins, we need to go from a shape of `shapefactor_mask.shape = (N, M)` to a `shapefactor_mask.shape = (N, 1, 1, M)` in order to multiply across correctly.

pyhf/modifiers/combined_mods.py
lukasheinrich: this is not a look-up table, so probably another name than `map` might be good.. maybe `global_bin_numbers`, `global_bin_index`

pyhf/modifiers/combined_mods.py
lukasheinrich: I don't understand this one.. why are we indexing through the global channels using indices from the parameters?  I don't think this will work in the general case.. e.g. say we have

```
channel1: [bin1,bin2]
channel2: [bin1,bin2]
channel_bin_index_map = [0,1,2,3] (2 2-bin channels concatenated)
```
let's say we have 3 samples

```
[s1g0,s1g1,s1g2,s1g3]
[s2g0,s2g1,s2g2,s2g3]
[s3g0,s3g1,s3g2,s3g3]
```
let's say we have one 2-bin shapefactor [nu1,nu2] and sample1 participates in channel1, sample2 in both and sample 3 in channel2.. the mask will be .. let's say we have many other parsets such that the global parsets ihas length 10 and 

```
                                nu1 nu2
                                  v v
self._parindices = [0,1,2,3,4,5,6,7,8,9]
```

assume [nu1,nu2] are located at [7,8] such that 

```
self._shapefactor_indices = [[7,8]]
```

```
[T,T,F,F]
[T,T,T,T]
[F,F,T,T]
```
then we want self.shapefactor_indices to be 

```
[ 7, 8,-1,-1]
[ 7, 8, 7, 8]
[-1,-1, 7, 8]
```

where the assumption is that the global parset is appended with a dummy value (at -1)

I guess my question is: shapefactor_mod_parindices is here [7,8] but it's not clear to me why we can index `channel_bin_index_map` with it

lukasheinrich: hm.. I think I understand a bit better now.. still looking at it. I realize 

it's not 
channel_bin_index_map = [0,1,2,3] (2 2-bin channels concatenated)

but
channel_bin_index_map = [0,1,0,1] (2 2-bin channels concatenated)

pyhf/modifiers/combined_mods.py
lukasheinrich: just making this concrete to help me understand.. this will result in e.g.

```
[
 [0,1,0,1,2]  << signal
 [0,1,0,1,2] << bkg1
 [0,1,0,1,2]  << background
]

```

pyhf/modifiers/combined_mods.py
lukasheinrich: ok I got confused by the argument order in `gather`.. I think I understand now

PR #329, opened 2018-10-20
Insert the POI parameter correctly using POI index in opt_tflow

kratsg:
# Description

`opt_tflow` doesn't insert the POI correctly. Assumed that it was the first one in the list of parameters, but might not always be the case. Need to respect the ordering.

Thanks to @dantrim for providing an example that found the bug.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyhf/optimize/opt_tflow.py
+++ b/pyhf/optimize/opt_tflow.py
@@ -41,7 +41,7 @@ class tflow_optimizer(object):
         poi_par   = self.tb.astensor([constrained_mu])
 
         nuis_cat = self.tb.concatenate(nuis_pars)
-        pars = self.tb.concatenate([nuis_cat[:0],poi_par,nuis_cat[0:]])
+        pars = self.tb.concatenate([nuis_cat[:pdf.config.poi_index],poi_par,nuis_cat[pdf.config.poi_index:]])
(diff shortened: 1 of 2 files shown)


lukasheinrich:
we should add a test that would make this fail, .. 


```
    spec = {
        'channels': [
            {
                'name': 'singlechannel',
                'samples': [
                    {
                        'name': 'signal',
                        'data': source['bindata']['sig'],
                        'modifiers': [
                            {
                                'name': 'mu',
                                'type': 'normfactor',
                                'data': None
                            }
                        ]
                    },
                    {
                        'name': 'background',
                        'data': source['bindata']['bkg'],
                        'modifiers': [
                            {
                                'name': 'bkg_norm',
                                'type': 'normsys',
                                'data': {'lo': 0.90, 'hi': 1.10}
                            }
                        ]
                    }
                ]
            }
        ]
    }
```

if we change the order of normsys and normfactor it should be covering that case

lukasheinrich:
I added a test which I verified would fail if @kratsg's fix is not applied.

lukasheinrich:
thanks @dantrim for the example!

matthewfeickert:
Excellent that a test has been added too! :)

PR #333, opened 2018-10-21
fix up indexing of summed_mask in shapesys, staterror

kratsg:
# Description

The main difficulty with shapesys/staterror as it affects multiple (non-consecutive) bins is the way we do masking. For situations where we need to figure out whether to enable the mask or not, it was filling in `-1` when the parameter index was `0` which is not good. Additionally, `tensorflow.gather()` is not happy with `-1`.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyhf/modifiers/staterror.py
+++ b/pyhf/modifiers/staterror.py
@@ -56,10 +56,6 @@ class staterror_combined(object):
                 # then apply the mask
                 summed_mask[positive_mask] = inds
                 summed_mask[zero_mask] = 0
-                # nb: old code above was
-                #     summed_mask[summed_mask > 0] = inds
-                #     summed_mask[summed_mask == 0] = -1
-                # This code broke when the `inds` included a '0' because it would replace that value with -1.
                 access_rows.append(summed_mask.tolist())
             self._factor_access_indices = default_backend.tolist(default_backend.stack(access_rows))
             self.finalize(pdfconfig)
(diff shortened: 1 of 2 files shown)


pyhf/modifiers/staterror.py
```diff
@@ -50,8 +50,16 @@ def __init__(self,staterr_mods,pdfconfig,mega_mods):
...
                 assert default_backend.shape(summed_mask[summed_mask >  0]) == default_backend.shape(default_backend.astensor(inds))
-                summed_mask[summed_mask >  0] = inds
-                summed_mask[summed_mask == 0] = -1
+                # make masks of > 0 and == 0
+                positive_mask = summed_mask > 0
+                zero_mask = summed_mask == 0
+                # then apply the mask
+                summed_mask[positive_mask] = inds
+                summed_mask[zero_mask] = 0
+                # nb: old code above was
```
matthewfeickert: These **nb** comments (below) probably don't need to be in the code, right? They could just exist in the PR, but we don't need them in the code base.

kratsg:
@matthewfeickert done ✅

lukasheinrich:
Hi, the -1 was the right index in order to pick up the default value which was appended at the end of the pars in apply, do i‘m not sure this is the right change

kratsg:
> Hi, the -1 was the right index in order to pick up the default value which was appended at the end of the pars in apply, do i‘m not sure this is the right change

This broke in tensorflow. `gather(....)` kept crashing since it can't access the index. Do you want to change `-1` to `len(inds) - 1`?

kratsg:
We also additionally changed shapesys in `refactor/megachannel` the same way to fix it so I just applied the same thing there to here.

PR #335, opened 2018-10-21
Fix indexing for real this time (instead of 0, use len(...) -1)

kratsg:
# Description

See #333. This is fixing it according to @lukasheinrich 's comment.

- [x] Tests are passing

# Before Merging

For the PR Assignees:


lukasheinrich:
Sorry for not catching this earlier

kratsg:
> Tbh I think the mask will enforce the rift value just before returning in apply

I thought the same thing too, but I'm not fully sure... That's what I initially set it to 0 thinking it'll just be masked out anyway...

PR #337, opened 2018-10-22
Add new interpolator codes 2 (two) and 4 (four)

kratsg:
# Description

This PR does two things: 
- restructure interpolators to be a sub-module of `pyhf` and additionally cleaning up the relevant API 
- add new interpolator codes (2=quadratic interpolator and linear extrapolation; 4=polynomial interpolation and exponential extrapolation)

The API for accessing interpolators takes on two forms:

```python
import pyhf.interpolators
interpolator = pyhf.interpolators.code0(histoset)
# or
interpolator = pyhf.interpolators.get(0)(histoset)
```

Additionally, the slow interpolator looper codes have been promoted to classes to provide the same API as their faster, tensorized counterpart. This makes the testing a lot more transparent and straightforward. Additionally, the function signatures are identical and should return identical types of data (e.g. tensor-types).

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- restructure interpolators to be a sub-module of `pyhf` and additionally cleaning up the relevant API 
- add new interpolator codes (2=quadratic interpolator and linear extrapolation; 4=polynomial interpolation and exponential extrapolation)
- slow interpolators are now promoted to classes to provide the same API to make testing more streamlined
- tests explicitly check interpolator structure as well as the subscription tensor backend changing to ensure that precomputation is being used
```


kratsg:
@matthewfeickert @lukasheinrich this is ready to be reviewed for sure.

lukasheinrich:
@matthewfeickert is there a way to view the docs from a PR?. I'll try to take a look. Clearly the code is self-consistent between the slow and the fast version, so that's great. But I need to double check one (probably the slow) to see if it's correct. @kratsg were you planning to add a learn nb?

kratsg:
> @kratsg were you planning to add a learn nb?

For now, no. I'm filing an issue to add it later. The somewhat elaborate documentation exists for that.

> @matthewfeickert is there a way to view the docs from a PR?.

Maybe we can define deployments for documentation generation.

kratsg:
NB: code2 and code4 need validation tests. Code4 in particular needs to be tested at 0 < alpha0 < 1.

matthewfeickert:
@lukasheinrich @kratsg 

> is there a way to view the docs from a PR?.

I'm not sure. I have just been checking out the branch locally and then building the docs to look at.

> Maybe we can define deployments for documentation generation.

This sounds good. Where would we be deploying to though?

PR #341, opened 2018-10-23
Change API from utils.runOnePoint to utils.hypotest

### LINKED ISSUE #220: clean up name/signature of `runOnePoint`
# Description

The signauture of `runOnePoint` is too complicated for newcomers. mostly people are only interested in CLs obs/exp. 

We could use the appraoch used in scikit of a signature that is *iteratively discoverable*

sucht that 

```
clsobs = pyhf.utils.cls(...)
clsobs, clsexp = pyhf.utils.cls(..., return_exp = True) #returns only two numbers
clsobs, clsexpset = pyhf.utils.cls(..., return_expset = True) #returns full 'brazil-band'
clsobs, clsexpset, qvals = pyhf.utils.cls(..., return_expset = True, return_teststatistics = True) #returns full 'brazil-band'  + test statistic values  
```

matthewfeickert:
# Description

Resolves #220 

**N.B.:** This is API breaking as both the name of the function and the return signature change.

- [x] Decide as a dev team on the new return signature
- [x] Add example of all `utils.hypotest` return options to Hello World example notebook

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- Change API from utils.runOnePoint to utils.hypotest and change the return signature of the function
- Use kwargs to have the signature of utils.hypotest be iteratively discoverable (in a similar vein to scikitlearn)
- Apply API change throughout pyhf to tests and notebooks
- Add utils.hypotest return structure test to test_utils.py
- Add example of all utils.hypotest return options to hello-world.ipynb notebook
- Improve documentation on the statistical interpretation of the return values of utils.hypotest
```


matthewfeickert:
@lukasheinrich @kratsg This is obviously not ready for review (and the docstring needs major revision) but given that this will change the API I figured the earlier feedback I could get from you on the new API the better.

pyhf/utils.py
```diff
@@ -127,27 +127,34 @@ def pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v):
     return CLsb, CLb, CLs
 
 
-def runOnePoint(muTest, data, pdf, init_pars=None, par_bounds=None):
+def cls(mu_test, data, pdf, init_pars=None, par_bounds=None, **kwargs):
```
lukasheinrich: I'd maybe call this `pyhf.utils.hypotest`

pyhf/utils.py
```diff
@@ -127,27 +127,34 @@ def pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v):
...
     r"""
-    Computes test statistics (and expected statistics) for a single value
-    of the parameter of interest
+    Computes test statistics (and expected statistics) for a single value of the parameter of interest
 
     Args:
-        muTest (Number or Tensor): The value of the parameter of interest (POI)
+        mu_test (Number or Tensor): The value of the parameter of interest (POI)
```
lukasheinrich: maybe `poi_test`

pyhf/utils.py
```diff
@@ -157,19 +164,50 @@ def runOnePoint(muTest, data, pdf, init_pars=None, par_bounds=None):
...
-        CLs_exp.append(pvals_from_teststat(sqrtqmu_v_sigma, sqrtqmuA_v)[-1])
-    CLs_exp = tensorlib.astensor(CLs_exp)
-    return qmu_v, qmuA_v, CLsb, CLb, CLs, CLs_exp
+    _returns = CLs
+    if kwargs:
+        _returns = [CLs]
+        if 'observed_set' in kwargs:
+            observed_set = kwargs['observed_set']
+        else:
+            observed_set = False
```
lukasheinrich: could just be `kwargs.get('observed_set')`. Not sure we need to assign to local vars..

```
if kwargs.get('test_statistics'): _returns.append([qmu_v, qmuA_v])
```

pyhf/utils.py
```diff
@@ -157,19 +164,50 @@ def runOnePoint(muTest, data, pdf, init_pars=None, par_bounds=None):
...
     CLsb, CLb, CLs = pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v)
-
-    CLs_exp = []
-    for nsigma in [-2, -1, 0, 1, 2]:
-        sqrtqmu_v_sigma = sqrtqmuA_v - nsigma
-        CLs_exp.append(pvals_from_teststat(sqrtqmu_v_sigma, sqrtqmuA_v)[-1])
-    CLs_exp = tensorlib.astensor(CLs_exp)
-    return qmu_v, qmuA_v, CLsb, CLb, CLs, CLs_exp
+    _returns = CLs
+    if kwargs:
```
lukasheinrich: I don't think this is needed.. if no kwargs are passed it should already be a empty dict

pyhf/utils.py
```diff
@@ -127,27 +127,34 @@ def pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v):
...
-        init_pars (Array or Tensor): the initial parameter values to be used for minimization
-        par_bounds (Array or Tensor): the parameter value bounds to be used for minimization
+        init_pars (Array or Tensor): The initial parameter values to be used for minimization
+        par_bounds (Array or Tensor): The parameter value bounds to be used for minimization
+
+    Keyword Args:
+        observed_set (bool): Bool for returning :math:`CL_{s+b}` and :math:`CL_{b}`
+        expected (bool): Bool for returning :math:`CL_{\textrm{exp}}`
+        expected_set (bool): Bool for returning the :math:`(-2,-1,0,1,2)\sigma` :math:`CL_{\textrm{exp}}` --- the "Brazil band"
+        test_statistics (bool): Bool for returning :math:`q_{\mu}` and :math:`q_{\mu,A}`
```
lukasheinrich: should these be called `return_X` ? similar to sklearn?

lukasheinrich:
Thanks @matthewfeickert. I left some mostly cosmetic comments

pyhf/utils.py
matthewfeickert: Good point

pyhf/utils.py
matthewfeickert: There might be a better way of doing this that will become apparent once I've had coffee, but as if there is `kwargs` then

```
_returns = tuple(_returns)
```

so in order to try to have a consistent return signature for the CLs_obs without `kwargs` (a Tensor and not a `tuple`) I added the default and the check for the empty dict.

pyhf/utils.py
matthewfeickert: Yeah, I had originally considered that as unnecessarily long, but The Zen of Python does tell us that

> Explicit is better than implicit.

so might as well follow sklearn's standard.

pyhf/utils.py
matthewfeickert: Changed in https://github.com/diana-hep/pyhf/pull/341/commits/490b22da1b8d2c6849523dcbfe340a211ac4dfd5

pyhf/utils.py
matthewfeickert: Changed in https://github.com/diana-hep/pyhf/pull/341/commits/490b22da1b8d2c6849523dcbfe340a211ac4dfd5

pyhf/utils.py
matthewfeickert: Changed in https://github.com/diana-hep/pyhf/pull/341/commits/490b22da1b8d2c6849523dcbfe340a211ac4dfd5

pyhf/utils.py
matthewfeickert: Changed in https://github.com/diana-hep/pyhf/pull/341/commits/c511057d206a6837d6a6bf5e00e64e2b28f4fa1a

pyhf/utils.py
matthewfeickert: Yes, of course. Sorry, I don't think I'm understanding what you're trying to show me. :? I don't see how that applies to what I'm saying RE: keeping the return signature of the CLs_obs the same regardless of `kwargs`.

```python
# Current: with guard
>>> import pyhf
>>> import pyhf.simplemodels
>>> import pyhf.utils
>>> 
>>> pdf = pyhf.simplemodels.hepdata_like(signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0])
>>> CLs_obs = pyhf.utils.hypotest(1.0, [51, 48] + pdf.config.auxdata, pdf)
>>> print(CLs_obs, type(CLs_obs))
[0.05290116] <class 'numpy.ndarray'>
>>> print('Observed: {}'.format(CLs_obs))
Observed: [0.05290116]
>>> 
>>> kwargs = {'return_expected': True}
>>> CLs_obs, CLs_exp = pyhf.utils.hypotest(1.0, [51, 48] + pdf.config.auxdata, pdf, **kwargs)
>>> print('Observed: {}, Expected: {}'.format(CLs_obs, CLs_exp))
Observed: [0.05290116], Expected: [0.06445521]
```

```python
# No guard: return signature is not consistent
>>> import pyhf
>>> import pyhf.simplemodels
>>> import pyhf.utils
>>> 
>>> pdf = pyhf.simplemodels.hepdata_like(signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0])
>>> CLs_obs = pyhf.utils.hypotest(1.0, [51, 48] + pdf.config.auxdata, pdf)
>>> print(CLs_obs, type(CLs_obs))
(array([0.05290116]),) <class 'tuple'>
>>> print('Observed: {}'.format(CLs_obs))
Observed: (array([0.05290116]),)
>>> 
>>> kwargs = {'return_expected': True}
>>> CLs_obs, CLs_exp = pyhf.utils.hypotest(1.0, [51, 48] + pdf.config.auxdata, pdf, **kwargs)
>>> print('Observed: {}, Expected: {}'.format(CLs_obs, CLs_exp))
Observed: [0.05290116], Expected: [0.06445521]
```

lukasheinrich: ah, sorry I misunderstood... so if I'm reading this right you only want to return a tuple in case there are two return values, other wise just the single object, or am I issing something?

we might just check at the very and the length of `_returns` and do `_returns if len(_returns) > 1 else _returns[0]`

pyhf/utils.py
lukasheinrich: thanks!

pyhf/utils.py
matthewfeickert: Yeah, I just want to have the returned `CLs_obs` be the same either way. Without some check then in the absence of `kwargs`

```python
CLs_obs = pyhf.utils.hypotest(...)
```
would have `CLs_obs` be of type `tuple`, where 

```python
CLs_obs, some_other_return = pyhf.utils.hypotest(..., **kwargs)
```

has `CLs_obs` being assigned an element of the returned tuple instead. The returned object is obviously the same, but the difference in what type of object the variable is assigned is not desirable.

Your suggested check of

```python
if len(_returns) > 1:
    _returns = tuple(_returns)
else:
    _returns = _returns[0]
```
is a fine alternative to checking for a non-empty dict. :+1:

lukasheinrich: ok .. sorry for the confusion!

pyhf/utils.py
```diff
@@ -127,27 +127,34 @@ def pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v):
...
-        muTest (Number or Tensor): The value of the parameter of interest (POI)
+        poi_test (Number or Tensor): The value of the parameter of interest (POI)
         data (Number or Tensor): The root of the calculated test statistic given the Asimov data, :math:`\sqrt{q_{\mu,A}}`
-        init_pars (Array or Tensor): the initial parameter values to be used for minimization
-        par_bounds (Array or Tensor): the parameter value bounds to be used for minimization
+        init_pars (Array or Tensor): The initial parameter values to be used for minimization
+        par_bounds (Array or Tensor): The parameter value bounds to be used for minimization
+
+    Keyword Args:
+        return_observed_set (bool): Bool for returning :math:`CL_{s+b}` and :math:`CL_{b}`
```
lukasheinrich: just commenting to remind ourselved to find a better name.. the name currently suggests some notion of analogy to the expected_set which should not be implied. But can't come up with one right now

pyhf/utils.py
matthewfeickert: No no, I think cleaner code would have avoided it in the first place. :)

lukasheinrich: out of curiosity if `return A if b else C` against black?

matthewfeickert: It is not! Changed in https://github.com/diana-hep/pyhf/pull/341/commits/4a8225ebf4adc31ad092146a8f2b077d9fc0b781

pyhf/utils.py
kratsg: Returning CLs set sounds clearer, since s+b and b is how you calculate CLs.

matthewfeickert: While I agree that just about anything is better than `return_observed_set` I think that `return_cls` might be confusing as by default a CLs is already returned. I don't have any useful suggestions at the moment though, so let me come back after I finish other work today with something to contribute.

lukasheinrich: yeah `cl_s` will be confusing. I was thinking perhaps `p_values` since this is exactly what they are

matthewfeickert: Changed in https://github.com/diana-hep/pyhf/pull/341/commits/51992a82c5422e9c8806db905d26491371ac41b5

pyhf/utils.py
```diff
@@ -157,19 +199,32 @@ def runOnePoint(muTest, data, pdf, init_pars=None, par_bounds=None):
...
 
     CLsb, CLb, CLs = pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v)
 
-    CLs_exp = []
-    for nsigma in [-2, -1, 0, 1, 2]:
-        sqrtqmu_v_sigma = sqrtqmuA_v - nsigma
-        CLs_exp.append(pvals_from_teststat(sqrtqmu_v_sigma, sqrtqmuA_v)[-1])
-    CLs_exp = tensorlib.astensor(CLs_exp)
-    return qmu_v, qmuA_v, CLsb, CLb, CLs, CLs_exp
+    _returns = [CLs]
```
matthewfeickert: Something I'd like you to think about in the review is if the order of returns (below) makes sense/is the most natural.

matthewfeickert:
Also, proposed PR commit log is above in the PR text body.

pyhf/utils.py
lukasheinrich: it's not :)

pyhf/utils.py
```diff
@@ -127,27 +126,70 @@ def pvals_from_teststat(sqrtqmu_v, sqrtqmuA_v):
...
 
 
-def runOnePoint(muTest, data, pdf, init_pars=None, par_bounds=None):
+def hypotest(poi_test, data, pdf, init_pars=None, par_bounds=None, **kwargs):
     r"""
-    Computes test statistics (and expected statistics) for a single value
-    of the parameter of interest
+    Computes :math:`p`-values and test statistics for a single value of the parameter of interest
 
     Args:
```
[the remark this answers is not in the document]
matthewfeickert: Fixed in https://github.com/diana-hep/pyhf/pull/341/commits/a456f387e78af355123349ce1e9aa93b96b1688c

pyhf/utils.py
matthewfeickert: Yeah. Do you have thoughts on what to do to improve this? Should we return to the tuple structure of `runOnePoint`'s return: `qmu_v, qmuA_v, CLsb, CLb, CLs, CLs_exp_set`?

matthewfeickert:
> Need to test the command line for all the available options and ensure that the various returns are returned as expected.

@kratsg As we currently only have [`cls` ](https://github.com/diana-hep/pyhf/blob/3f03d123f2cc81bf1693d0a4c55e3baa0e23c25d/pyhf/commandline.py#L66-L125) at the CLI that would interact with `hypotest` do you want a new CLI function to be added that directly exposes `hypotest`? Or just to add additional option flags to `cls` to interact with the `hypostest` `kwargs` (maybe something like `--return expected_set`)?

kratsg:
> @kratsg As we currently only have [`cls` ](https://github.com/diana-hep/pyhf/blob/3f03d123f2cc81bf1693d0a4c55e3baa0e23c25d/pyhf/commandline.py#L66-L125) at the CLI that would interact with `hypotest` do you want a new CLI function to be added that directly exposes `hypotest`? Or just to add additional option flags to `cls` to interact with the `hypostest` `kwargs` (maybe something like `--return expected_set`)?

Ahh, good point. Do we currently have existing tests to check the return of this function? Did I miss that change? I'll think more about what I want...

matthewfeickert:
> Do we currently have existing tests to check the return of this function? Did I miss that change?

No. In `tests/test_scripts.py` we have [`test_import_prepHistFactory_and_cls`](https://github.com/diana-hep/pyhf/blob/3f03d123f2cc81bf1693d0a4c55e3baa0e23c25d/tests/test_scripts.py#L46-L84) but that just [checks if it ran okay and then a few other things](https://github.com/diana-hep/pyhf/blob/3f03d123f2cc81bf1693d0a4c55e3baa0e23c25d/tests/test_scripts.py#L81-L84). We don't check the actual return values.

kratsg:
> If by "this function" you just mean `hypotest` though then yes. We currently do this in `test_validation.py` in [`validate_hypotest`](https://github.com/diana-hep/pyhf/blob/3f03d123f2cc81bf1693d0a4c55e3baa0e23c25d/tests/test_validation.py#L621-L631).

This should be in `test_utils.py` but doesn't validate results. Just checks that it returns the right structure as expected.

pyhf/utils.py
lukasheinrich: oh sorry I meant the order I chose did not make sense.. maybe it shoule be  `CLs, CLs_exp, CLsb, CLb, q_mu, qmu_A` .. is this what you have?

matthewfeickert: Ah. Yeah, it is the following:

```python
CLs_obs = pyhf.utils.hypotest(...)
CLs_obs, [CLsb, CLb] = pyhf.utils.hypotest(..., return_p_values=True)
CLs_obs, CLs_exp = pyhf.utils.hypotest(..., return_expected=True)
CLs_obs, [CLs_exp_band] = pyhf.utils.hypotest(..., return_expected_set=True)
CLs_obs, CLs_exp, [CLs_exp_band] = pyhf.utils.hypotest(
    ..., return_expected=True, return_expected_set=True
)
CLs_obs, [CLsb, CLb], CLs_exp, [CLs_exp_band] = pyhf.utils.hypotest(
    ..., return_p_values=True, return_expected=True, return_expected_set=True
)
CLs_obs, [CLsb, CLb], CLs_exp, [CLs_exp_band], [q_mu, q_mu_A] = pyhf.utils.hypotest(
    ...,
    return_p_values=True,
    return_expected=True,
    return_expected_set=True,
    return_test_statistics=True,
)
```

So if there are suggestions on how to make it nicer for a user to keep in their head I'm happy to change it.

matthewfeickert: Whatever we decide, I should probably add all the options to the Hello World notebook as well.

tests/test_utils.py
```diff
@@ -20,3 +23,66 @@ def test_load_custom_schema(tmpdir):
     temp = tmpdir.join("custom_schema.json")
     temp.write('{"foo": "bar"}')
     assert pyhf.utils.load_schema(temp.strpath)
+
+
+def test_hypotest(tmpdir):
```
matthewfeickert: @kratsg Suggestions for writing a more efficient testing scheme are welcome.

pyhf/utils.py
matthewfeickert: @kratsg Suggestions on revisions if you have them is welcomed.

If there are no suggested revisions on the return signature from you two then I will go ahead and cement it with some additional examples to the Hello World notebook (unless people think that is overkill that will clutter up a nice short example).

tests/test_utils.py
kratsg: This looks good enough to me for now. Perhaps splitting up the test into multiple test functions make more sense.. e.g. `test_hypotest_return_p_values` and so on.

pyhf/utils.py
```diff
@@ -162,15 +162,15 @@ def hypotest(poi_test, data, pdf, init_pars=None, par_bounds=None, **kwargs):
 
             to protect against excluding signal models in which there is little sensitivity. In the case that :math:`\textrm{CL}_{s} \leq \alpha` the given signal model is excluded.
 
-            - :math:`\left[\textrm{CL}_{s+b}, \textrm{CL}_{b}\right]`: The signal + background and background only :math:`p`-values as defined in Equations (75) and (76) of `arXiv:1007.1727`_
+            - :math:`\left[\textrm{CL}_{s+b}, \textrm{CL}_{b}\right]`: The signal + background :math:`p`-value and 1 minus the background only :math:`p`-value as defined in Equations (75) and (76) of `arXiv:1007.1727`_
```
matthewfeickert: @lukasheinrich @kratsg I noticed I had a typo in the documentation RE: the CL_b and the p-values. This makes me think though, is the naming choice `return_p_values` a good one if we return `p_{s+b}` (a p-value) and `1-p_b` (not a p-value)? I'm not sure that using `return_probabilities` is better though, as I don't think it gets across the same information. Thoughts?

lukasheinrich: I mean we could just return p_b and not CLb

matthewfeickert:
@lukasheinrich @kratsg I'm about to leave to drive to FNAL so I'll be offline for awhile. I think this is ready for review, so let me know if you have any additional thoughts or changes.

pyhf/utils.py
kratsg: > That doesn't make it a p-value though, right? The p-value is determined by what is your Null hypothesis, and so the direction of the integration shouldn't be the defining factor(?). Maybe I'm confused.

Direction of integration is somewhat important. Isn't the main reason for having the CLs method to handle the situations in when signal < background and signal > background correctly -- and the corresponding statistical fluctuations?

foo

"... background (blue) and signal+background (red) hypotheses for 10 observed events (black line). In this example, p(10 | b) = 0.018 and p(10 | s+b) = 0.125. The observed data is more likely under the s+b hypothesis than background-only."

kratsg:
LGTM

pyhf/utils.py
lukasheinrich: the reason these plots above may be confusing (which is why I would not recommend including them in the docs) is that in profile likelihood (unlike in the examples above) we would use different test statistics to test the mu=0 and mu=1 hypothesis.. there, the null hypothesis always follows a ~chi square (wilks theorem) and we always integrate to the right

lukasheinrich:
LGTM for me as well @matthewfeickert this is still marked WIP, should we merge?

kratsg:
> this is still marked WIP, should we merge?

It's not marked WIP -- but for some reason, the "WIP" is buggy.

matthewfeickert:
> It's not marked WIP -- but for some reason, the "WIP" is buggy.

Yeah, it shouldn't be. Let me force push and see if that fixes it.

pyhf/utils.py
matthewfeickert: Hm. Thanks for this discussion. I think I need to write a learn/lecture notebook (mostly for myself) on this now.

matthewfeickert:
Yeah, `git commit --amend --no-edit && git push --force-with-lease origin master revise/change-runOnePoint-API` fixed the WIP bot bug.

pyhf/utils.py
kratsg: I think I'll benefit from that too :)

PR #346, opened 2018-10-25
Get pyhf to import faster

### LINKED ISSUE #308: minimize unnecessary module loading
# Description

we want `pyhf` to be *fast*. A minmal click app can print in <0.1 seconds

```
import click
@click.command()
@click.option('--test')
def ok(test):
    pass

if __name__ == '__main__':
    ok()
```

```
time python what.py --help
Usage: what.py [OPTIONS]

Options:
  --test TEXT
  --help       Show this message and exit.
python what.py --help  0.05s user 0.03s system 81% cpu 0.091 total
```

while a simple `pyhf --version` takes 2 seconds

```
time  pyhf --version
pyhf, version 0.0.15
pyhf --version  1.45s user 0.48s system 95% cpu 2.027 total
```

which is probably due to excessive loading of modules (I think for example we load all backends into `__all__` irrespective of which one is used. For e.g. the MBJ fit (17s) this is 10% of the runtime, so it'd be nice to make this faster

kratsg:
# Description

This resolves #308 and similar issues. Part of the problem with the command line scripts is that it can take some time to do even simple tasks. Part of the reason this occurs is because we import a lot of things at the start, instead of deferring the imports until we actually need them. This PR will fix these issues in a way that works on both python2 and python3.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- Speed up pyhf import by lazy-loading the tensor backends
```


kratsg:
The current change to disguise the `tensor` module as a `_BackendRetriever` object gives this nice speed up for `$ time python -c "import pyhf"`

```
real	0m1.765s
user	0m1.725s
sys	0m0.320s
```

compared to master

```
real	0m6.550s
user	0m6.359s
sys	0m0.459s
```

without losing any functionality.

kratsg:
After lazy-loading the optimizers, I'm down to less than a second:

```
$ time pyhf --version
pyhf, version 0.0.15

real	0m0.765s
user	0m0.834s
sys	0m0.215s
```

pyhf/optimize/__init__.py
```diff
@@ -1,24 +1,39 @@
-from .opt_scipy import scipy_optimizer
+class _OptimizerRetriever(object):
+    def __getattr__(self, name):
+        if name == 'scipy_optimizer':
+            from .opt_scipy import scipy_optimizer
 
-assert scipy_optimizer
+            assert scipy_optimizer
+            self.scipy_optimizer = scipy_optimizer
```
lukasheinrich: what's the assignment for?

kratsg: When assigned, the custom `__getattr__` won't be called if the attribute already exists. The `__getattr__()` is only called when the attribute doesn't exist. This is the lazy-loading part... so once it's loaded, we just add it in and we're done.

lukasheinrich: shouldn't the import be lazy anyways?

```
import foo
import foo
```

should not really import twice, no?

kratsg: I think you're getting confused. In any other piece of code, we do something like `import pyhf.tensor`. When this is called once, all it does is load the object defined in that module. It doesn't have any backends loaded. When `pyhf.tensor.numpy_backend` is called, it looks through the existing attributes of the object, sees that it doesn't exist, then tries to import it. When we can import it, we then set that attribute on the object to bypass the `__getattr__` check which only occurs if the object didn't have the attribute in the first place.

lukasheinrich: I guess I am, I'd think that 

```
def lazyload():
    from bla import X
    return X

#nothing loaded until now

X = lazyload()
X = lazyload()
```
would achieve the same thing, without assignment

kratsg: This isn't about the import loading. The `__getattr__` is only run if it cannot find the attribute on the tensor object we're making. Since the object should just represent what we import, we should set the attribute on it so that when python goes through all attributes to match the call, it short-circuits instead of going to our `__getattr__` which should be fast since it doesn't actually import again.

We're not trying to speed up the import here, but the `getattr` check.

lukasheinrich: ok but clearly the bottleneck is in the import and not in the getattr check no? Maybe I'm overlooking something, how much are we saving?

kratsg: I've added comments to clarify.

PR #352, opened 2018-10-31
Add luminosity uncertainty calculations

kratsg:
# Description

This is the last piece needed for the MBJ validation. The luminosity is a modifier that affects some samples in some channels, but is a globally shared modifier (as such, the name is fixed to `lumi` in the spec). This is going to be similar to what normsys currently does, but globally shared.

- `readxml.py` is updated to check for this.
- `modifiers/lumi.py` is added with a gaussian constraint
- `tests/test_validation.py` is set-up to check for the calculation, but will need to run it in HistFactory to get the right calculation [e.g. use one of the validation XML files that already exist]

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- the luminosity uncertainty is a global multiplicative scale factor, constrained by a gaussian whose mu is the single nuisance parameter constrained as follows (based on HistFactory):
  - inits = [lumi]
  - auxdata = [lumi]
  - sigmas = [lumi * lumirelerror]
  - bounds = [[0, 10*lumi]]
- tests are added to validate the expected scaling behavior of the luminosity uncertainty
- the spec is updated to allow for sigmas to be configured for lumi and set for constrained_by_normal
- paramset reduction now handles the situation for paramset requirements that cannot have a default value (None) versus paramset requirements that should not be defined ('undefined')
```

### DIFF
--- a/pyhf/readxml.py
+++ b/pyhf/readxml.py
@@ -217,7 +217,7 @@ def process_measurements(toplvl):
             for param_name in param.text.split(' '):
                 # lumi will always be the first parameter
                 if param_name == 'Lumi':
-                    result['config']['parameters'][0].update(overall_param_object)
+                    result['config']['parameters'][0].update(overall_param_obj)
                 else:
                     param_obj = {'name': param_name}
                     param_obj.update(overall_param_obj)
--- a/pyhf/readxml.py
+++ b/pyhf/readxml.py
@@ -187,9 +187,23 @@ def process_channel(channelxml, rootdir, track_progress=False):
 def process_measurements(toplvl):
     results = []
     for x in toplvl.findall('Measurement'):
+        lumi = float(x.attrib['Lumi'])
+        lumierr = lumi * float(x.attrib['LumiRelErr'])
+
         result = {
             'name': x.attrib['Name'],
-            'config': {'poi': x.findall('POI')[0].text, 'parameters': []},
+            'config': {
+                'poi': x.findall('POI')[0].text,
+                'parameters': [
+                    {
+                        'name': 'lumi',
+                        'auxdata': [lumi],
+                        'bounds': [[0.0, 10.0 * lumi]],
+                        'inits': [lumi],
+                        'sigmas': [lumierr],
+                    }
+                ],
+            },
         }
         for param in x.findall('ParamSetting'):
             # determine what all parameters in the paramsetting have in common
@@ -201,20 +215,13 @@ def process_measurements(toplvl):
 
             # might be specifying multiple parameters in the same ParamSetting
             for param_name in param.text.split(' '):
+                # lumi will always be the first parameter
                 if param_name == 'Lumi':
-                    lumi = float(x.attrib['Lumi'])
-                    lumierr = lumi * float(x.attrib['LumiRelErr'])
-                    param_obj = {
-                        'name': 'lumi',
-                        'auxdata': [lumi],
-                        'bounds': [[0.0, 10.0 * lumi]],
-                        'inits': [lumi],
-                        'sigmas': [lumierr],
-                    }
+                    result['config']['parameters'][0].update(overall_param_object)
                 else:
                     param_obj = {'name': param_name}
-                param_obj.update(overall_param_obj)
-                result['config']['parameters'].append(param_obj)
+                    param_obj.update(overall_param_obj)
+                    result['config']['parameters'].append(param_obj)
         results.append(result)
     return results
 
--- a/pyhf/readxml.py
+++ b/pyhf/readxml.py
@@ -189,10 +189,7 @@ def process_measurements(toplvl):
     for x in toplvl.findall('Measurement'):
         result = {
             'name': x.attrib['Name'],
-            'config': {
-                'poi': x.findall('POI')[0].text,
-                'parameters': [],
-            },
+            'config': {'poi': x.findall('POI')[0].text, 'parameters': []},
         }
         for param in x.findall('ParamSetting'):
             # determine what all parameters in the paramsetting have in common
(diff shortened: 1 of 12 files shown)


tests/test_validation.py
```diff
@@ -73,6 +73,65 @@ def setup_1bin_shapesys(
...
+                ],
+            }
+        ]
+    }
+    return spec
+
+
+@pytest.fixture(scope='module')
+def expected_result_1bin_lumi(mu=1.0):
+    ## TO DO - FIX THIS
```
lukasheinrich: what's to fix here?

tests/test_validation.py
```diff
@@ -73,6 +73,65 @@ def setup_1bin_shapesys(
...
+                        'name': 'signal',
+                        'data': source['bindata']['sig'],
+                        'modifiers': [
+                            {'name': 'mu', 'type': 'normfactor', 'data': None}
+                        ],
+                    },
+                    {
+                        'name': 'background',
+                        'data': source['bindata']['bkg'],
+                        'modifiers': [{'name': 'lumi', 'type': 'lumi', 'data': None}],
```
lukasheinrich: don't we need to associate some data with the modifier? like the scale of the uncertainty?

pyhf/readxml.py
```diff
@@ -59,6 +59,9 @@ def process_sample(
     data, err = import_root_histogram(rootdir, inputfile, histopath, histoname)
 
     modifiers = []
+    # first check if we need to add lumi modifier for this sample
+    if sample.attrib.get("NormalizeByTheory", "False") == 'True':
+        modifiers.append({'name': 'lumi', 'type': 'lumi', 'data': None})
```
lukasheinrich: we need to somehow pass the lumi uncertainty in.. this is a bit tricky since it's not defined by the tag in the sample, but rather the toplevel measurement object.. so maybe we store that value when we read the toplevel object and then inject it into `data` here... or we set the lumi uncertainty from somewhere

tests/test_validation.py
kratsg: I need to run this with HistFactory to get actual expected results.

tests/test_validation.py
kratsg: Yep, I guess this goes into the auxdata -- but I'm still not sure where it goes yet. `readxml` doesn't have an easy way to access it and copy it across all modifiers.

pyhf/readxml.py
kratsg: Yep, that's the part I'm slightly stuck on.

kratsg:
@lukasheinrich , we now have something added to the `toplvl` as part of the measurements for lumi. This at least resolves being able to sorta inject things -- but the difficulty is that with multiple POIs, there can be multiple lumi uncerts -- so I'm not sure how to inject the lumi uncertainty correctly into what I have now. Any ideas? The below is what `readxml` provides now when parsing `validation/xmlimport_input/config/example.xml`:

```
{'channels': [{'name': 'channel1',
               'samples': [{'data': [20.0, 10.0],
                            'modifiers': [{'data': {'hi': 1.05, 'lo': 0.95},
                                           'name': 'syst1',
                                           'type': 'normsys'},
                                          {'data': None,
                                           'name': 'SigXsecOverSM',
                                           'type': 'normfactor'}],
                            'name': 'signal'},
                           {'data': [100.0, 0.0],
                            'modifiers': [{'data': None,
                                           'name': 'lumi',
                                           'type': 'lumi'},
                                          {'data': [5.000000074505806, 0.0],
                                           'name': 'staterror_channel1',
                                           'type': 'staterror'},
                                          {'data': {'hi': 1.05, 'lo': 0.95},
                                           'name': 'syst2',
                                           'type': 'normsys'}],
                            'name': 'background1'},
                           {'data': [0.0, 100.0],
                            'modifiers': [{'data': None,
                                           'name': 'lumi',
                                           'type': 'lumi'},
                                          {'data': [0.0, 10.0],
                                           'name': 'staterror_channel1',
                                           'type': 'staterror'},
                                          {'data': {'hi': 1.05, 'lo': 0.95},
                                           'name': 'syst3',
                                           'type': 'normsys'}],
                            'name': 'background2'}]}],
 'data': {'channel1': [122.0, 112.0]},
 'toplvl': {'measurements': [{'config': {'lumi': 1.0,
                                         'lumirelerr': 0.1,
                                         'parameters': [{'fixed': True,
                                                         'name': 'Lumi'},
                                                        {'fixed': True,
                                                         'name': 'alpha_syst1'}],
                                         'poi': 'SigXsecOverSM'},
                              'name': 'GaussExample'},
                             {'config': {'lumi': 1.0,
                                         'lumirelerr': 0.1,
                                         'parameters': [{'fixed': True,
                                                         'name': 'Lumi'},
                                                        {'fixed': True,
                                                         'name': 'alpha_syst1'}],
                                         'poi': 'SigXsecOverSM'},
                              'name': 'GammaExample'},
                             {'config': {'lumi': 1.0,
                                         'lumirelerr': 0.1,
                                         'parameters': [{'fixed': True,
                                                         'name': 'Lumi'},
                                                        {'fixed': True,
                                                         'name': 'alpha_syst1'}],
                                         'poi': 'SigXsecOverSM'},
                              'name': 'LogNormExample'},
                             {'config': {'lumi': 1.0,
                                         'lumirelerr': 0.1,
                                         'parameters': [{'fixed': True,
                                                         'name': 'Lumi'},
                                                        {'fixed': True,
                                                         'name': 'alpha_syst1'}],
                                         'poi': 'SigXsecOverSM'},
                              'name': 'ConstExample'}],
            'resultprefix': './results/example'}}
```

lukasheinrich:
@kratsg you need to change the README :)

PR #354, opened 2018-11-01
Enable shared nuisance parameters over modifier types

lukasheinrich:
# Description

this resolves the case in #272 of shared parsets over multiple modifiers

```
roohf not shared
{
    "CLs_exp": [
        0.15719136360999494,
        0.28707182742358117,
        0.49040108398224225,
        0.7391079566241138,
        0.9260249020571463
    ],
    "CLs_obs": 0.6624276086776395
}


roohf shared
{
    "CLs_exp": [
        0.1824533173540692,
        0.31861297860704774,
        0.5224666964320326,
        0.7615731173535661,
        0.9344146382006049
    ],
    "CLs_obs": 0.6768036954464488
}

pyhf shared
{
    "CLs_exp": [
        0.18257145128940405,
        0.31875687511186174,
        0.5226093840643298,
        0.761670602736219,
        0.9344501334751033
    ],
    "CLs_obs": 0.6768690617184364
}
```

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* expand config.modifiers to include mapping to parameter set (modname,type,paramset)
* use type/modname key to identify modifier data in mega_mods
* use paramset name in combined modifier appliers to select parset
* enables shared nuisance parameters across modifier types
```

### DIFF
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -58,7 +58,13 @@ class _ModelConfig(object):
                             )
                         )
                         raise exceptions.InvalidModifier()
-                    self.modifiers.append((modifier_def['name'], modifier_def['type']))
+                    self.modifiers.append(
+                        (
+                            modifier_def['name'],  # mod name
+                            modifier_def['type'],  # mod type
+                            modifier_def['name'],
+                        )  # parset name
+                    )
 
                     # check the shareability (e.g. for shapesys for example)
                     is_shared = paramset_requirements['is_shared']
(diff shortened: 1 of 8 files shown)


lukasheinrich:
this is passing tests.. does someone want to review?

kratsg:
@lukasheinrich i need more explanation here,  because I'm not fully sure this is doing what I expect it to do...

`normsys` and `histosys` have different `op_code` so when we reduce them, we pop off the set of op codes which could be "addition" or "multiplication" depending on the ordering in the set (if histosys and normsys are shared) and we're setting both of them to the same op_code... which means that's wrong?

pyhf/pdf.py
```diff
@@ -108,7 +114,7 @@ def param_set(self, name):
         return self.par_map[name]['paramset']
 
     def set_poi(self, name):
-        if name not in [x for x, _ in self.modifiers]:
+        if name not in [x for x, _, _ in self.modifiers]:
```
kratsg: this line is incorrect. we want `if name not in [x for _, _, x in self.modifiers]` and really, it should just become `if not not in self.parameters`.

Issue #355, opened 2018-11-02
optimization: allow setting arbitrary parameters constant 


lukasheinrich:
# Description

right now we only know how to do a single POI.. adding multiple POI support is easy if we can juts expand the optimizer API to set random sub-slices of the full parameters to constant values.

lukasheinrich:
closing as this was done #846

kratsg:
not merged yet!

PR #356, opened 2018-11-04
Allow for parameter configuration from the spec

kratsg:
# Description

This expands the spec to allow for (optional) parameter configuration. This will not break existing specs, since if this isn't specified, we will defer to default configurations in all cases. The addition of a parameter configuration allows to override these defaults in the code.

This will get us closer to being able to do something like #355 and enables us to implement globally defined systematics such as luminosity uncertainty in #352 .

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- Expand the spec to allow for optional parameter configuration
- Remove op_code from the paramset requirements
- Refactor to allow for op_code to be less hard-coded in pyhf/pdf::_modifications
```

### DIFF
--- a/pyhf/paramsets.py
+++ b/pyhf/paramsets.py
@@ -75,11 +75,15 @@ def reduce_paramset_requirements(paramset_requirements, paramsets_user_configs):
                 if isinstance(v, tuple):
                     v = list(v)
                 # this implies user-configured, so check that it has the right number of elements
-                elif isinstance(v, list) and len(v) != len(default_v):
+                elif isinstance(v, list) and default_v and len(v) != len(default_v):
                     raise exceptions.InvalidModel(
                         'Incorrect number of values ({}) for {} were configured by you, expected {}.'.format(
                             len(v), k, len(default_v)
                         )
+                    )
+                elif v and not default_v:
+                    raise exceptions.InvalidModel(
+                        '{} does not use the {} attribute.'.format(param_name, k)
                     )
 
                 combined_param[k] = v
--- a/pyhf/paramsets.py
+++ b/pyhf/paramsets.py
@@ -58,12 +58,13 @@ def reduce_paramset_requirements(paramset_requirements):
         for param in params:
             for k in param_keys:
                 v = param.get(k)
-                # need to convert lists to tuples
-                if k in ['inits', 'auxdata', 'factors']:
-                    v = tuple(v)
-                # need to convert lists of lists to tuples
-                elif k in ['bounds']:
-                    v = tuple(map(tuple, v))
+                if v:
+                    # need to convert lists to tuples
+                    if k in ['inits', 'auxdata', 'factors']:
+                        v = tuple(v)
+                    # need to convert lists of lists to tuples
+                    elif k in ['bounds']:
+                        v = tuple(map(tuple, v))
                 combined_param.setdefault(k, set([])).add(v)
 
         for k in param_keys:
--- a/pyhf/paramsets.py
+++ b/pyhf/paramsets.py
@@ -49,6 +49,14 @@ def reduce_paramsets_requirements(paramsets_requirements, paramsets_user_configs
         'factors',
     ]
 
+    """
+    - process all defined paramsets
+    - determine the unique set of paramsets by param-name
+    - if the paramset is not unique, complain
+    - if the paramset is unique, build the paramset using the set() defined on its options
+      - if the value is a tuple, this came from default options so convert to a list and use it
+      - if the value is a list, this came from user-define options, so use it
+    """
     for paramset_name in list(paramsets_requirements.keys()):
         paramset_requirements = paramsets_requirements[paramset_name]
         paramset_user_configs = paramsets_user_configs.get(paramset_name, {})
--- a/pyhf/paramsets.py
+++ b/pyhf/paramsets.py
@@ -37,7 +37,7 @@ class constrained_by_poisson(paramset):
         )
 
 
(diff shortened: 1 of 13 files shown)


pyhf/pdf.py
```diff
@@ -50,7 +63,10 @@ def __init__(self, spec, poiname='mu', qualify_names=False):
                     try:
                         paramset_requirements = modifiers.registry[
                             modifier_def['type']
-                        ].required_parset(len(sample['data']))
+                        ].required_parset(
+                            len(sample['data']),
+                            config=self.parameter_configs.get(modifier_def['name'], {}),
```
lukasheinrich: we should probably start explicitly differentiating between the param name and the mod name by introducing a local variable.. which then should be used here.

also further down:

```
self.parameters.append(param_name)
```

```
self.modifiers.append(
                         (
                             modifier_def['name'],  # mod name
                             modifier_def['type'],  # mod type
                             param_name
                         )  # parset name
```

lukasheinrich:
thanks for the PR! an overall question: why are we passing the config into the `required_parset` method of the modifiers? instead one could just collect all declared preferences from the modifiers' `required_parset` and the specs `parameters` key and resolve them by specifying a priority in a separate function. This would centralize the logic instead of scattering it across all `required_parset` functions

kratsg:
> thanks for the PR! an overall question: why are we passing the config into the `required_parset` method of the modifiers? instead one could just collect all declared preferences from the modifiers' `required_parset` and the specs `parameters` key and resolve them by specifying a priority in a separate function. This would centralize the logic instead of scattering it across all `required_parset` functions

Because then we have to duplicate the logic of auxdata/factors out of the modifiers themselves as that function would need to pass things in. E.G. we would have to know for shapesys that we shouldn't set defaults for auxdata/factors -- but can do so for other systematics.

lukasheinrich:
hm not sure I follow, why would we need to factor out auxdata etc? I would just treat the spec requirements as an additional data points for the parset through which consensus is reached, can you give an example?

kratsg:
> hm not sure I follow, why would we need to factor out auxdata etc? I would just treat the spec requirements as an additional data points for the parset through which consensus is reached, can you give an example? e.g. in your ex, that logic (e.g. to allow/disallow certain overrides) can be part of the consensus alg, no?

You'll have to explain more, but we don't have a way to prevent custom overrides. E.G.

```
paramset_requirements.update(paramset_configurations.get(param_name, {}))
```

is what you're proposing instead, but there's no protection if the user provides something of the wrong shape or provides auxdata and we don't want them to.

FWIW, we want the lumi uncertainty to be provided as a factor.

lukasheinrich:
`def reduce_paramset_requirements(paramset_requirements)` can be made arbitrarily complicated depending on what policy we want to enforce

kratsg:
> `def reduce_paramset_requirements(paramset_requirements)` can be made arbitrarily complicated depending on what policy we want to enforce

Ok, I have updated things. I can remove `auxdata` from the spec to only allow for `inits, bounds, factors` for user-configurability. See the `tests/test_schema.py` to see what we expect to fail. Removing `auxdata` from the spec prevents it from being overridden later on. If we additionally want to add protection again in `reduce_paramsets` as well (to be redundant), let me know - but otherwise, I find the schema-validation to be sufficient.

pyhf/paramsets.py
```diff
@@ -53,11 +52,13 @@ def reduce_paramset_requirements(paramset_requirements):
 
     for param_name in list(paramset_requirements.keys()):
         params = paramset_requirements[param_name]
```
lukasheinrich: I realize I'm to blame for this, but this should probably be `reqs` or something similar

pyhf/paramsets.py
```diff
@@ -53,11 +52,13 @@ def reduce_paramset_requirements(paramset_requirements):
...
+        param_user_configs = paramsets_user_configs.get(param_name, {})
 
         combined_param = {}
         for param in params:
             for k in param_keys:
-                combined_param.setdefault(k, set([])).add(param.get(k))
```
lukasheinrich: seems like if we reorder the loops here, we can only have one loop over `param_keys`?

kratsg: We can't. The first loop is iterating over all requirements and merging them all into a single (combined) requirement. After this loop is done, we look at the combined paramset to see if things are merge-able.

pyhf/paramsets.py
```diff
@@ -37,41 +37,55 @@ def expected_data(self, pars):
...
-        'op_code',
         'inits',
         'bounds',
         'auxdata',
         'factors',
     ]
 
-    for param_name in list(paramset_requirements.keys()):
-        params = paramset_requirements[param_name]
+    for paramset_name in list(paramsets_requirements.keys()):
```
lukasheinrich: why are we looping over `paramsets_requirements.keys()` and not `paramset_keys` ?

pyhf/pdf.py
```diff
@@ -315,8 +330,16 @@ def expected_auxdata(self, pars):
...
+        factor_mods = [
+            modtype
+            for modtype, mod in modifiers.uncombined.items()
+            if mod.op_code == 'multiplication'
+        ]
+        delta_mods = [
+            modtype
+            for modtype, mod in modifiers.uncombined.items()
+            if mod.op_code == 'addition'
+        ]
```
lukasheinrich: this is probably cachable in __init__?

lukasheinrich:
I just hada a minor nit-pick but in general this looks ready to merge

PR #365, opened 2018-11-08
adding cache dictionary to open root files only once in readxml

nikoladze:
# Description
Hi,

The xml parser (re-)opens the histogram rootfiles for each histogram. This can slow down the process extremely when there are many histograms in one rootfile (as is the case if one creates the xml and histograms with HistFitter). This might be easily avoided by adding a dictionary somewhere that tracks which files have already been opened.
My current solution is probably not the cleanest way  to do this (no way clean up of the dictionary), but it shows that the approach is effective. I tried it for an example with ~100k Histograms in one rootfile which seemed to take forever without that fix (i aborted it after 2 hours). With the cache dictionary it took a few minutes to generate the json file.
Maybe somebody has a good idea on how to incorporate this.

(remark: HistFactory's hist2workspace has the same issue, but there it is probably more involved to fix it)

Cheers,
Nikolai

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyhf/readxml.py
+++ b/pyhf/readxml.py
@@ -10,6 +10,7 @@ log = logging.getLogger(__name__)
 
 __FILECACHE__ = {}
 
+
 def extract_error(h):
     """
     Determine the bin uncertainties for a histogram.
--- a/pyhf/readxml.py
+++ b/pyhf/readxml.py
@@ -26,13 +26,18 @@ def extract_error(h):
     return np.sqrt(err).tolist()
 
 
-def import_root_histogram(rootdir, filename, path, name):
+def import_root_histogram(rootdir, filename, path, name, filecache={}):
     import uproot
 
     # strip leading slashes as uproot doesn't use "/" for top-level
     path = path or ''
     path = path.strip('/')
-    f = uproot.open(os.path.join(rootdir, filename))
+    fullpath = os.path.join(rootdir, filename)
+    if not fullpath in filecache:
+        f = uproot.open(fullpath)
+        filecache[fullpath] = f
+    else:
+        f = filecache[fullpath]
     try:
         h = f[name]
     except KeyError:
--- a/pyhf/readxml.py
+++ b/pyhf/readxml.py
@@ -7,6 +7,7 @@ import tqdm
 
 log = logging.getLogger(__name__)
 
+__FILECACHE__ = {}
 
 def extract_error(h):
     """
@@ -26,7 +27,10 @@ def extract_error(h):
     return np.sqrt(err).tolist()
 
 
-def import_root_histogram(rootdir, filename, path, name, filecache={}):
+def import_root_histogram(rootdir, filename, path, name, filecache=None):
+    global __FILECACHE__
+    filecache = filecache or __FILECACHE__
+
     import uproot
 
     # strip leading slashes as uproot doesn't use "/" for top-level
@@ -204,3 +208,7 @@ def parse(configfile, rootdir, track_progress=False):
         'channels': [{'name': k, 'samples': v['samples']} for k, v in channels.items()],
         'data': {k: v['data'] for k, v in channels.items()},
     }
+
+
+def clear_filecache():
+    __FILECACHE__ = {}
--- a/pyhf/readxml.py
+++ b/pyhf/readxml.py
@@ -4,6 +4,7 @@ import os
 import xml.etree.ElementTree as ET
 import numpy as np
 import tqdm
+import uproot
 
 log = logging.getLogger(__name__)
 
@@ -30,8 +31,6 @@ def extract_error(h):
 def import_root_histogram(rootdir, filename, path, name, filecache=None):
     global __FILECACHE__
     filecache = filecache or __FILECACHE__
-
-    import uproot
 
     # strip leading slashes as uproot doesn't use "/" for top-level
     path = path or ''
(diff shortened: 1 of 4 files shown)


pyhf/readxml.py
```diff
@@ -26,13 +26,18 @@ def extract_error(h):
     return np.sqrt(err).tolist()
 
 
-def import_root_histogram(rootdir, filename, path, name):
+def import_root_histogram(rootdir, filename, path, name, filecache={}):
```
lukasheinrich: yes if I'm not mistaken python can be a bit annoying with having mutable objects as default arguments

```pyhton
__FILECACHE__ = []

def import_root_histogram(rootdir, filename, path, name, filecache=None):
    global __FILECACHE__
    filecache = filecache or __FILECACHE__
```

should do the trick

nikoladze: In this case the python behavior (dictionary is created when function is defined) was intended, so the function has a persistent cache during multiple executions. But i agree that a global file cache in the module is probably better. The question remains where/if to automatically delete the cache?

lukasheinrich: we could have a `readxml.clearcache()` method . but honestly I don't think we will run into this often.. I expect that most of the time the readxml module will be called from the command line via `pyhf xml2json` and the cache will be torn out on exit anyways.. we can add this later on if this becomes a problem

lukasheinrich: (unless you already ran into a situation where this takes up too much memory)

nikoladze: > but honestly I don't think we will run into this often.. 

.... true - if the the file was produced with HistFitter it is anyways just one root file

nikoladze:
@kratsg Do you have a particular test in mind? I could imagine testing that `uproot.open` is not called again after calling `import_root_histogram` a second time on the same file (but still correct output)

lukasheinrich:
one could `mv` the file in the test to a different location.. the file in memory should not be affected but the filename itself will not point anymore to the inode

lukasheinrich:
or this https://stackoverflow.com/questions/21982964/how-to-get-the-call-count-using-mock-patch

lukasheinrich:
just adding more links.. i don't think we use any mocking so far but pytest seems to have support for it

https://github.com/pytest-dev/pytest-mock/

nikoladze:
I played around a bit and made a test which checks that uproot.open is only called once (using pytest-mock). In addition, i check if the dictionary before and after calling readxml.parse the first time has the same entries.

lukasheinrich:
hm pyflakes seems unhappy. but clearly we are using the variable in function.. if all else fails an `assert __FILECACHE__ is not None` should circumvent it (though a bit hacky)

lukasheinrich:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=703653

nikoladze:
Lets see if it works like that - i forgot to declare __FILECACHE__ global in the clear_filecache function, so the cache wasn't properly reset (and so the test failed when other tests ran before)

tests/test_import.py
```diff
@@ -83,3 +95,27 @@ def test_import_histosys():
     }
 
     assert channels['channel2']['samples'][0]['modifiers'][0]['type'] == 'histosys'
+
+
+def test_import_filecache(mocker):
+
+    # uproot is dynamically imported in readxml, so to wrap it we need to set it now
+    pyhf.readxml.uproot = uproot
+
```
lukasheinrich: would it help if we moved the uproot import to the module-level? it's imported dynamically in order to only do the import incase someone explicitly wants xml2json (if they already have json there's no need to install uproot) but we could easily just do 

```
try:
    import readxml
except ImportError
    pass # I guess we're not importing
```

in commandline.py (or factor xml2json into a separate module)

kratsg:
> just adding more links.. i don't think we use any mocking so far but pytest seems to have support for it
> 
> https://github.com/pytest-dev/pytest-mock/

We use mocking: https://github.com/diana-hep/pyhf/blob/master/tests/test_events.py

lukasheinrich:
`black` is not happy in python 3

     # check if uproot.open was only called once with the expected root file
-    pyhf.readxml.uproot.open.assert_called_once_with(os.path.join("validation/xmlimport_input", "./data/example.root"))
+    pyhf.readxml.uproot.open.assert_called_once_with(
+        os.path.join("validation/xmlimport_input", "./data/example.root")
+    )

nikoladze:
It seems pyflakes doesn't like that i'm trying to import uproot but don't use it:
https://github.com/diana-hep/pyhf/blob/b4cd7aec7292819d773da7561a22a8a15e29b9ea/pyhf/commandline.py#L40-L48
It felt better to catch the ImportError on importing uproot (what the error message will be about) th
an on readxml. Is there an easy way out?

lukasheinrich:
you can do

```
try:
    import uproot
    assert uproot
except:
    pass
```

lukasheinrich:
@nikoladze thanks for the fix. we seem to have issues with an unrelated piece #368 but that fix probably is easy, but we might need to rebase this PR once that fix is in master

matthewfeickert:
> Except for #370 I think?

Well, I pinned the version of pytest to be below `v4.0` in PR #369 as I didn't want to change two things in one PR. So while that would be nice so that I don't have to fix its uses in tests later it isn't _required_ to move this forward (I think).

PR #379, opened 2018-12-29
Make extract_error compliant with uproot-methods v0.3.0 API

matthewfeickert:
# Description

`tests/test_import.py` give an error of

```
E       ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

in `extract_error` due to a change in the return structure of `uproot-methods` `TH1.variances` return structure in [`uproot_methods` PR 21](https://github.com/scikit-hep/uproot-methods/pull/21) where the returned value [changed from a list to a `numpy.array`](https://github.com/scikit-hep/uproot-methods/commit/c2a78e4e05eb710c74d5c6ec5d017f70af28566b#diff-67fbe39413c91feb6ad80acd70f76d47).

This change was introduced in [`uproot-methods` release `v0.3.0`](https://github.com/scikit-hep/uproot-methods/releases/tag/0.3.0)

This necessitated a change in the query of `h.variances` on the pyhf side.

This PR is also affecting PR #377 and PR #378 

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Use h.variances.any() to match the numpy.array return structure of the variances property of uproot-methods TH1 objects introduced in uproot-methods v0.3.0
* Require uproot v3.3.0 or newer, which requires uproot-methods v0.3.0 or newer
* Temporarily pin PyTorch to be below v1.0 for the following PR to fix
```

PR #378, opened 2018-12-29
Change virtual environment link in docs to PyPA page

matthewfeickert:
The PyPA page on [Installing Packages](https://packaging.python.org/tutorials/installing-packages/) has a much more generally helpful section on [Creating Virtual Environments](https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments) that gives info for Python 2 and Python 3, has relevant links to docs, and is followed up with pip information. This is a much more useful site to direct new users to then the [virtualenvwrapper docs](https://virtualenvwrapper.readthedocs.io/en/latest/).

# Description

Please first read [CONTRIBUTING.md](https://github.com/diana-hep/pyhf/tree/master/CONTRIBUTING.md).

Please describe the purpose of this pull request in some detail. Reference and link to any relevant issues or pull requests.

# Checklist Before Requesting Reviewer

- [ ] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
The PyPA page on Installing Packages has a much more generally helpful section on Creating Virtual Environments.
```

PR #377, opened 2018-12-21
feat: Add jax backend

lukasheinrich:
# Description

Resolves #416 

pdf evaluation works modulo this diff

```
diff --git a/pyhf/pdf.py b/pyhf/pdf.py
index 259bf44..4f299ca 100644
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -423,7 +423,8 @@ class Model(object):
         tensorlib, _ = get_backend()
         lambdas_data = self.expected_actualdata(pars)
         summands = tensorlib.poisson_logpdf(maindata, lambdas_data)
-        tosum = tensorlib.boolean_mask(summands, tensorlib.isfinite(summands))
+        # tosum = tensorlib.boolean_mask(summands, tensorlib.isfinite(summands))
+        tosum = summands
         mainpdf = tensorlib.sum(tosum)
         return mainpdf
```

since bool-based indexing  seems to not work

```
import pyhf
import numpy as onp

pyhf.set_backend(pyhf.tensor.jax_backend())

m = pyhf.simplemodels.hepdata_like(signal_data = [5.], bkg_data = [17.], bkg_uncerts = [2.])
testdata = m.expected_data(m.config.suggested_init())
testpars = m.config.suggested_init()
value = m.logpdf(testpars,testdata)

print(type(value))
print(onp.asarray(value))
```

locally this gives

```
/Users/lukas/Code/pyhfdev/venv/lib/python3.6/site-packages/jax/lib/xla_bridge.py:143: UserWarning: No GPU found, falling back to CPU.
  warnings.warn('No GPU found, falling back to CPU.')
<class 'jax.interpreters.xla.DeviceArray'>
[-5.528393]
```

https://github.com/google/jax/issues/166

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* add JAX backend
```

PR #380, opened 2018-12-31
Require PyTorch v1.0 or newer and fix static TLS load error

matthewfeickert:
# Description

- Together with PR #379 fixes the CI for 2018 going into 2019.
- This PR needs to be rebased against `master` after PR #379 is accepted.
- This PR is also affecting PR #377 and PR #378.
- One reoccurring problem seems to be(?) that the `v1.0` release of PyTorch is [causing issues with memory allocation in Travis](https://travis-ci.org/diana-hep/pyhf/jobs/473431783#L481-L483) resulting in errors or segfault 11.

In requiring the latest version of PyTorch (v1.0) the loading of it after other imports in `tests/conftest.py` causes there to be an [error during setup of the pytest fixtures](https://travis-ci.org/diana-hep/pyhf/jobs/473509231#L650-L655)

```
ImportError while loading conftest '/home/travis/build/diana-hep/pyhf/tests/conftest.py'.
tests/conftest.py:45: in <module>
pyhf/tensor/__init__.py:24: in __getattr__
    e,
E   pyhf.exceptions.ImportBackendError: ('There was a problem importing PyTorch. The pytorch backend cannot be used.', ImportError('dlopen: cannot load any more object with static TLS',))
```

TLS appears to stand for thread local storage

* https://stackoverflow.com/questions/19268293/matlab-error-cannot-open-with-static-tls

and seems to be a reoccurring issue with PyTorch. The advice for how to deal with it from PyTorch [Issue 2083](https://github.com/pytorch/pytorch/issues/2083) and [Issue 2575](https://github.com/pytorch/pytorch/issues/2575) seems to be to simply change the import order so that

`import torch`

is always the first import in the files that are causing the issue. The rational for this fix is not understood to me, other then it just makes sure that libraries that (presumably) require a large amount of static TLS are dealt with first.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Require PyTorch v1.0 or newer
* Make torch the first import in tests/conftest.py to solve the "cannot load any more object with static TLS" (thread local storage) error that is common when using PyTorch along with other large libraries. c.f. PyTorch Issue 2083.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ extras_require = {
         'numpy<=1.14.5,>=1.14.0',  # Lower of 1.14.0 instead of 1.13.3 to ensure doctest pass
         'setuptools<=39.1.0',
     ],
-    'torch': ['torch<1.0.0,>=0.4.0'],
+    'torch': ['torch>=1.0.0'],
     'mxnet': [
         'mxnet>=1.0.0',
         'requests<2.19.0,>=2.18.4',
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import torch  # Solves "cannot load any more object with static TLS" (thread local storage) error
 import pytest
 import pyhf
 import tensorflow as tf


matthewfeickert:
This PR should wait for PR #381 to be approved and merged and then should be rebased against `master`.

PR #381, opened 2019-01-03
Fix corrupted size vs. prev_size memory error in CI benchmark stage

matthewfeickert:
# Description

In the merge of PR #379 into `master` it was seen that the [benchmark CI stage failed with the memory error of](https://travis-ci.org/diana-hep/pyhf/jobs/474501916#L859-L861)

```
*** Error in `/home/travis/virtualenv/python3.6.3/bin/python': corrupted size vs. prev_size: 0x0000000000c2de00 ***
/home/travis/.travis/job_stages: line 104:  4633 Aborted                 (core dumped) pytest -r sx --benchmark-sort=mean tests/benchmarks/
The command "pytest -r sx --benchmark-sort=mean tests/benchmarks/" exited with 134.
```

This error had been [seen before in Issue #67](https://github.com/diana-hep/pyhf/issues/67#issuecomment-378529720) and @kratsg provided a [fix for a relevant similar memory issue in PR #296](https://github.com/diana-hep/pyhf/pull/296#issuecomment-427536326). Applying that same fix, of installing and exporting the path for `libtcmalloc-minimal4` during `before_install` has provided a [passing build with the benchmark CI stage in a test](https://travis-ci.org/diana-hep/pyhf/builds/474967137), so it seems that this should revolve the issue here as well.

# Checklist Before Requesting Reviewer 

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Install and export the path for libtcmalloc-minimal4 during before_install for the benchmark stage of Travis CI to avoid a corrupted size vs. prev_size memory error. c.f. PR 296 for additional reference.
```

PR #383, opened 2019-01-04
Use Binder build API, remove trigger_binder.py and selenium dependency, and add Binder CI stage

### LINKED ISSUE #382: Use Binder Build API for Builds in PRs
# Description

After a discussion with @minrk and @betatim on the jupyterhub/binder Gitter, it was made clear that the use of Selenium in [`binder/trigger_binder.py`](https://github.com/diana-hep/pyhf/blob/c81f6007309f4c13241f9efac187594337d0bd08/binder/trigger_binder.py) (and the script itself) is unnecessary. Instead a simple API call can be made just using Python's `webbrowser` with an [endpoint of the form `https://mybinder.org/build/gh/owner/repo/ref`](https://gitter.im/jupyterhub/binder?at=5c2f87038dafa715c73ff54f) as can be [seen in the Binder Hub demo](https://github.com/jupyterhub/binderhub/blob/9ca8fa68bb8b69c6a2736f2275279583073f314f/examples/binder-api.py#L28) (thanks Tim for the link).

So, for example

```
python -m webbrowser "https://mybinder.org/build/gh/diana-hep/pyhf/master"
```

So asking [WWKHTD](https://github.com/kelseyhightower/nocode), this means that `binder/trigger_binder.py` is unnecessary and should be removed and `.travis.yml` should be updated to use the API calls.

matthewfeickert:
# Description

Resolves #382 

Waiting on #380 and #381 to be merged first.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use the Binder build API endpoint with Python's built in webbrowser to trigger builds
   - https://gitter.im/jupyterhub/binder?at=5c2f87038dafa715c73ff54f
   - https://github.com/jupyterhub/binderhub/blob/9ca8fa68bb8b69c6a2736f2275279583073f314f/examples/binder-api.py#L28
* Remove selenium and chromium-chromedriver dependency installation from Travis CI docs stage
* Remove binder/trigger_binder.py as being able to access the API makes it unnecessary
* Remove redundant Binder build API calls
* Move Binder build API call to separate Binder stage that runs only on PR merges to master
```


.travis.yml
```diff
@@ -56,30 +56,22 @@ jobs:
...
     script:
       - python -m doctest README.md
       - cd docs && make html && cd -
       - touch docs/_build/html/.nojekyll
-      # Open Binder example in browser to trigger repo2docker to build image
-      - python binder/trigger_binder.py --url https://mybinder.org/v2/gh/diana-hep/pyhf/master
-      - python binder/trigger_binder.py --url https://mybinder.org/v2/gh/diana-hep/pyhf/master?filepath=docs%2Fexamples%2Fnotebooks%2Fbinderexample%2FStatisticalAnalysis.ipynb
+      # Use Binder build API to trigger repo2docker to build image
+      - python -m webbrowser "https://mybinder.org/build/gh/diana-hep/pyhf/master"
+      - python -m webbrowser "https://mybinder.org/build/gh/diana-hep/pyhf/master?filepath=docs%2Fexamples%2Fnotebooks%2Fbinderexample%2FStatisticalAnalysis.ipynb"
```
minrk: FWIW, filepath isn't part of the build, it is only consumed by the browser redirects after build completes. These two lines won't do different things.

Additionally, you might want to use travis [environment variables](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) to ensure that the PR is being tested. I *think* that `https://mybinder.org/build/gh/$TRAVIS_REPO_SLUG/$TRAVIS_COMMIT` will be right. You *might* need to use $TRAVIS_PULL_REQUEST_SLUG for PRs since I believe that PR commits are available as refs in the target repo even before it's merged. Note that binder builds are cached by commit sha (it resolves branches and tags to a commit before checking), so building `abc123` and then requesting `master` later will use the cached build as long as the commits match.

Alternately, if this isn't really a test and you only want to build master, moving it to the `deploy` stage might be appropriate, since that only runs on master, not PRs.

matthewfeickert: @minrk thanks for the feedback!

>  filepath isn't part of the build, it is only consumed by the browser redirects after build completes. These two lines won't do different things.

Ah, so if I understand correctly, you're saying that I have been misunderstanding how Binder works and if I have Binder build an image off of the `master` branch then that image will be used for all subsequent requests off of that branch _regardless_ of filepaths. This makes sense now that I think about it, but I think I must have had some confusion in the past that if I requested a URL with `filepath=` in it then this actually built a *different* image then if `filepath=` wasn't there.

> Alternately, if this isn't really a test and you only want to build master, moving it to the `deploy` stage might be appropriate, since that only runs on master, not PRs.

This is a good point, as this is really what we're doing.

matthewfeickert:
@lukasheinrich Let's actually hold off on this one for now

.travis.yml
minrk: > if I have Binder build an image off of the master branch then that image will be used for all subsequent requests off of that branch regardless of filepaths

Correct. Binder has two stages: build and launch. The inputs to the build are only the repo and ref, while the 'filepath' is an input to launch, indicating what URL to redirect users to when the server is ready.

matthewfeickert:
@lukasheinrich @kratsg Given what @minrk has pointed out about how I've been adding redundant Binder build calls and the point about having it get called every time in the docs is unnecessary if we just want it to be built for every PR I've added a `binder` stage that like `benchmark` is only run on PR merges to `master`. I added a new stage as `deploy` only gets run on tags. Sound okay to you?

kratsg:
> @lukasheinrich @kratsg Given what @minrk has pointed out about how I've been adding redundant Binder build calls and the point about having it get called every time in the docs is unnecessary if we just want it to be built for every PR I've added a `binder` stage that like `benchmark` is only run on PR merges to `master`. I added a new stage as `deploy` only gets run on tags. Sound okay to you?

This is fine with me if (unless I've misunderstood @minrk ) we're guaranteed that the binder builds won't ever fail when the rest of our tests pass correctly.

.travis.yml
```diff
@@ -88,6 +79,14 @@ jobs:
       local_dir: docs/_build/html
       on:
         branch: master
+  - stage: binder
+    python: '3.6'
+    before_install: skip
+    install: skip
```
kratsg: are we sure the script will get called if `install: skip` is used? We had this issue before where it might have skipped the rest of the stages...

matthewfeickert:
> we're guaranteed that the binder builds won't ever fail when the rest of our tests pass correctly.

Correct. As the `binder` stage is the last to run then for the build to fail `python -m webbrowser` would need to die or BinderHub would need to be down.

.travis.yml
matthewfeickert: Good question. Luckily the answer is yes. :+1:  I tested this in a test branch and you can [see the Travis build here](https://travis-ci.org/diana-hep/pyhf/builds/476949260). You can see from the [`binder` stage](https://travis-ci.org/diana-hep/pyhf/jobs/476949265) that the call get executed properly and from [the config](https://travis-ci.org/diana-hep/pyhf/jobs/476949265/config) you can see that it used the following

```
  - stage: binder
    python: '3.6'
    before_install: skip
    install: skip
    script:
      - python -m webbrowser "https://mybinder.org/build/gh/diana-hep/pyhf/master"
    after_success: skip
```

PR #403, opened 2019-02-14
Add configurable optimizer (and options) through pyhf cls CLI

kratsg:
# Description

This updates the other optimizers to accept `maxiter` as a keyword argument for configuration. We should just keep this pretty consistent. This makes it easier for us to configure from the command line now for CLs. This might not be the best option, since we might want to support something like a `dotenv` to load default backend/optimizer instead...? Unclear.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- raise an exception (InvalidOptimizer) if an optimizer is requested that does not exist
- allow for **kwargs for all optimizer initializations so we can extract out the additional options for configuration
- create click.ParamType (EqDelimStringParamType) to parse CLI options for configuring optimizer using equal-delimited key-value pairs, inspired by yadage
```

### DIFF
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -4,7 +4,7 @@ import click
 import json
 import os
 
-from .utils import hypotest
+from .utils import hypotest, EqDelimStringParamType
 from .pdf import Workspace
 from .version import __version__
 
@@ -222,6 +222,7 @@ def inspect(workspace, output_file, measurement):
 @click.option('--testpoi', default=1.0)
 @click.option('--teststat', type=click.Choice(['q', 'qtilde']), default='qtilde')
 @click.option('--optimizer')
+@click.option('--optconf', type=EqDelimStringParamType(), multiple=True)
 @click.option('-n', '--max-iterations', default=1000)
 def cls(
     workspace,
@@ -231,6 +232,7 @@ def cls(
     testpoi,
     teststat,
     optimizer,
+    optconf,
     max_iterations,
 ):
     with click.open_file(workspace, 'r') as specstream:
@@ -246,6 +248,8 @@ def cls(
         patches=patches,
         modifier_settings={'normsys': {'interpcode': 'code4'}},
     )
+
+    optconf = {k: v for item in optconf for k, v in item.items()}
 
     # set the new optimizer
     if optimizer:
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -7,6 +7,7 @@ import os
 from .utils import hypotest, EqDelimStringParamType
 from .pdf import Workspace
 from .version import __version__
+from . import tensorlib, set_backend, optimize
 
 logging.basicConfig()
 log = logging.getLogger(__name__)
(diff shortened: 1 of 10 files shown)


pyhf/commandline.py
```diff
@@ -83,7 +83,9 @@ def json2xml(workspace, xmlfile, specroot, dataroot):
 )
 @click.option('--measurement', default=None)
 @click.option('-p', '--patch', multiple=True)
-def cls(workspace, output_file, measurement, patch):
+@click.option('--optimizer')
+@click.option('-n', '--max-iterations', default=1000)
+def cls(workspace, output_file, measurement, patch, optimizer, max_iterations):
```
lukasheinrich: going to be tricky since different optimizers will have different opts. One appraoch I used in https://github.com/yadage/yadage is to have a mainswitch to select type and then one for setting optiosn

```
... --backend xyc --backendopt hello=world --backendopt foo=bar --backendopt baz=qux
```

kratsg: This is now done with the `--optconf` option.

pyhf/optimize/__init__.py
```diff
@@ -49,6 +49,8 @@ def __getattr__(self, name):
                     "There was a problem importing Minuit. The minuit optimizer cannot be used.",
                     e,
                 )
+        elif name == '__wrapped__':  # doctest
```
matthewfeickert: @kratsg Can you add a simple/short explanation of where `__wrapped__` pops up in doctest to the PR description or the commit summary?

kratsg: https://bugs.python.org/issue25998 and https://github.com/timofurrer/ramlient/issues/51 explain some of the issues well here. The comment in the code + the context of the comment in the code is rather explanatory. I've had to add an exception for `__wrapped__` because doctesting is checking attributes on module as part of doctesting.

matthewfeickert: https://bugs.python.org/issue25998 is helpful, thanks. Given the context of these two resources I now agree with you that the comment is enough. Without them though it wasn't clear what was happening that this was actually happening.

pyhf/commandline.py
```diff
@@ -234,10 +240,19 @@ def cls(workspace, output_file, measurement, patch, testpoi, teststat):
...
+    if optimizer:
+        new_optimizer = getattr(optimize, optimizer)
+        set_backend(tensorlib, new_optimizer(**optconf))
+
     result = hypotest(testpoi, w.data(p), p, qtilde=is_qtilde, return_expected_set=True)
     result = {'CLs_obs': result[0].tolist()[0], 'CLs_exp': result[-1].ravel().tolist()}
+
     if output_file is None:
-        click.echo(json.dumps(result, indent=4, sort_keys=True))
+        print(json.dumps(result, indent=4, sort_keys=True))
```
matthewfeickert: @kratsg Just curious why this was changed to `print`. Just better formatting?

kratsg: Oh, no. it should've been fixed...

PR #409, opened 2019-02-18
Add Bash script to guard max time return value for Binder build

matthewfeickert:
# Description

Safeguards Binder build section of the CI from taking too long and causing a timeout error by Travis CI. The trigger function sets a time limit on curl and accepts `--max-time` exit return code, `28`, as valid and then returns success.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* The trigger function sets a time limit on curl and accepts --max-time exit return code, 28, as valid and then returns success.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,7 @@ jobs:
     install: skip
     script:
       # Use Binder build API to trigger repo2docker to build image
-      - curl https://mybinder.org/build/gh/diana-hep/pyhf/master
+      - bash binder/trigger_binder.sh https://mybinder.org/build/gh/diana-hep/pyhf/"${TRAVIS_BRANCH}"
     after_success: skip
   - stage: deploy
     python: '3.6'
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,7 @@ jobs:
     install: skip
     script:
       # Use Binder build API to trigger repo2docker to build image
-      - bash binder/trigger_binder.sh https://mybinder.org/build/gh/diana-hep/pyhf/"${TRAVIS_BRANCH}"
+      - ./binder/trigger_binder.sh https://mybinder.org/build/gh/diana-hep/pyhf/"${TRAVIS_BRANCH}"
     after_success: skip
   - stage: deploy
     python: '3.6'
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,7 @@ jobs:
     install: skip
     script:
       # Use Binder build API to trigger repo2docker to build image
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -87,7 +87,7 @@ jobs:
     install: skip
     script:
       # Use Binder build API to trigger repo2docker to build image
-      - curl https://mybinder.org/build/gh/diana-hep/pyhf/master
+      - bash binder/trigger_binder.sh https://mybinder.org/build/gh/diana-hep/pyhf/"${TRAVIS_BRANCH}"
```
kratsg: i don't think the `bash` is needed here. Travis will assume it's an executable.

matthewfeickert: I think that it will actually fail, [given this Travis build on another project with the same code](https://travis-ci.com/matthewfeickert/R-in-Jupyter-with-Binder/jobs/178277329#L931-L933), but if you want I can just run this on another trial branch.

kratsg: It's mostly environmental -- there's no reason to explicitly list `bash` (and `/usr/bin/env bash` may not match `bash` in some environments) and you've already written it in your shebang. I prefer the former instead of the latter.

matthewfeickert: I'll change it if you prefer. Though I prefer the opposite take, with making it explicit to the user what is happening and having the shebang be the fall back, not the default.

matthewfeickert: @kratsg It seems that changing it caused issues. On the [`trial/see-if-bash-needed` branch that I setup to test this it is failing from `Permission denied`](https://travis-ci.org/diana-hep/pyhf/jobs/494733188#L440-L442) with

```
./binder/trigger_binder.sh
```

kratsg: Permission denied means it didn't have a `chmod +x` on it. But we can leave it as it is.

matthewfeickert: This is (obviously) true &mdash; I had done `chmod +x` on the wrong branch (I guess I was tired?).

Dumb question: So do you consider it best practice to make all shell scripts executables and then encourage users to execute with `./pathto/script.sh` to enforce the shebang?

kratsg: It varies. Given that this is only for the CI, I don't care as much.

matthewfeickert: Sure, this is pretty low stakes. But given that my install scripts for all my projects are almost all Bash, I'm curious on your input here.

kratsg: I tend to use `./` because most people don't usually care what language it is. E.G. a python executable could be written with the shebang and then just ran as `./` similar to the bash script. So if in the future, you change the script to a new language, users don't change how they run it.

PR #413, opened 2019-03-02
Enforce consistent tensor shape for scalars

matthewfeickert:
# Description

In `mxnet >= v1.4.0` the shape of a `NDArray` when a scalar has been passed to it is empty. As a result, this should be caught and forced to have a consistent shape by wrapping the scalar in a list. However, at this point it is probably worth enforcing a default shape on tensors made from scalars, as the different tensor libraries handle them differently and we should have a consistent form.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Ensure non-empty tensor shape for consistency
   - Do this by asking for forgiveness and attempting to access shape information, and if it fails modifying the tensor object such that it has the proper shape
* Remove generic_len function from backends as a consistent tensor shape makes it unnecessary
* Add error logging and raise KeyError if an unsupported dtype is used for the tensorlibs astensor method
* Add tests for tensor shape for scalars
* Add tests for bad tensor dtypes
```

### DIFF
--- a/pyhf/tensor/mxnet_backend.py
+++ b/pyhf/tensor/mxnet_backend.py
@@ -108,6 +108,7 @@ class mxnet_backend(object):
         dtype = dtypemap[dtype]
         try:
             tensor = nd.array(tensor_in, dtype=dtype)
+            # Ensure non-empty tensor shape for consistency
             try:
                 tensor.shape[0]
             except IndexError:
--- a/pyhf/tensor/mxnet_backend.py
+++ b/pyhf/tensor/mxnet_backend.py
@@ -106,15 +106,12 @@ class mxnet_backend(object):
         """
         dtypemap = {'float': 'float32', 'int': 'int32', 'bool': 'uint8'}
         dtype = dtypemap[dtype]
-        try:
-            tensor = nd.array(tensor_in, dtype=dtype)
-            # Ensure non-empty tensor shape for consistency with mxnet v1.4.0+
-            try:
-                tensor.shape[0]
-            except IndexError:
-                tensor = tensor.broadcast_to((1,))
-        except ValueError:
+
+        if isinstance(tensor_in, (int, float)):
+            # Ensure non-empty tensor shape for consistency
             tensor = nd.array([tensor_in], dtype=dtype)
+        else:
+            tensor = nd.array(tensor_in, dtype=dtype)
         return tensor
 
     def sum(self, tensor_in, axis=None):
--- a/pyhf/tensor/mxnet_backend.py
+++ b/pyhf/tensor/mxnet_backend.py
@@ -111,11 +111,15 @@ class mxnet_backend(object):
             log.error('Invalid dtype: dtype must be float, int, or bool.')
             raise
 
-        if isinstance(tensor_in, (int, float)):
-            # Ensure non-empty tensor shape for consistency
-            tensor = nd.array([tensor_in], dtype=dtype)
-        else:
+        try:
             tensor = nd.array(tensor_in, dtype=dtype)
+            # Ensure non-empty tensor shape for consistency
+            try:
+                tensor.shape[0]
+            except IndexError:
+                tensor = tensor.broadcast_to((1,))
+        except ValueError:
+            tensor = nd.array([tensor_in], dtype=dtype)
         return tensor
 
     def sum(self, tensor_in, axis=None):
--- a/pyhf/tensor/mxnet_backend.py
+++ b/pyhf/tensor/mxnet_backend.py
@@ -108,11 +108,11 @@ class mxnet_backend(object):
         dtype = dtypemap[dtype]
         try:
             tensor = nd.array(tensor_in, dtype=dtype)
-            # Ensure non-empty tensor shape for consistency
+            # Ensure non-empty tensor shape for consistency with mxnet v1.4.0+
             try:
                 tensor.shape[0]
             except IndexError:
-                tensor = nd.array([tensor_in], dtype=dtype)
+                tensor = tensor.broadcast_to((1,))
         except ValueError:
             tensor = nd.array([tensor_in], dtype=dtype)
         return tensor
--- a/pyhf/tensor/mxnet_backend.py
+++ b/pyhf/tensor/mxnet_backend.py
@@ -105,7 +105,11 @@ class mxnet_backend(object):
             MXNet NDArray: A multi-dimensional, fixed-size homogenous array.
         """
         dtypemap = {'float': 'float32', 'int': 'int32', 'bool': 'uint8'}
-        dtype = dtypemap[dtype]
+        try:
+            dtype = dtypemap[dtype]
+        except KeyError:
+            log.error('Invalid dtype: dtype must be float, int, or bool.')
+            raise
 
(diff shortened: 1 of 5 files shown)


pyhf/tensor/mxnet_backend.py
```diff
@@ -108,6 +108,10 @@ def astensor(self, tensor_in, dtype='float'):
         dtype = dtypemap[dtype]
         try:
             tensor = nd.array(tensor_in, dtype=dtype)
+            try:
+                tensor.shape[0]
+            except IndexError:
```
kratsg: add an extra comment here about this exception

matthewfeickert: I added a 1-liner. @kratsg  let me know if this should be more clear (my brain is gone at this point in the evening).

kratsg:
We're missing coverage tests where we force the shape checks for the exceptions.

matthewfeickert:
Actually, looking at this in more depth now I think that the MXNet's backend's `simple_broadcast` needs to be rewritten instead with something like the [`generic_len` of the TensorFlow and PyTorch backends](https://github.com/diana-hep/pyhf/blob/4fa45dda8d2314138549ac6b67ebdc9d9a506742/pyhf/tensor/tensorflow_backend.py#L219-L226).

matthewfeickert:
@kratsg @lukasheinrich Looking at this again, is there a specific reason that for our backends the `astensor` method doesn't automatically convert input scalars to lists, turning 

```
tb = pyhf.tensorlib
tb.astensor(1.)
```

into the equivalent of
```
tb = pyhf.tensorlib
tb.astensor([1.])
```

such that they both return
```
that_libs_version_of_tensor([1.])
```

? I can't remember if there is a reason that we don't force this, beyond trying to keep parity with the underlying libraries themselves as

```
>>> np.array(1.)
array(1.)
>>> np.array([1.])
array([1.])
```

lukasheinrich:
Hi @matthewfeickert -- I think you are right, the point of the tensorlib is to act as a shim that unifies the interfaces across the different backends (as annoyingly they have small differences) in that sense, we should expect all shapes to be the same for the a given input shape (as well as decide on specific casts) and implement them accordingly

pyhf/tensor/mxnet_backend.py
```diff
@@ -105,11 +105,17 @@ def astensor(self, tensor_in, dtype='float'):
             MXNet NDArray: A multi-dimensional, fixed-size homogenous array.
         """
         dtypemap = {'float': 'float32', 'int': 'int32', 'bool': 'uint8'}
-        dtype = dtypemap[dtype]
         try:
-            tensor = nd.array(tensor_in, dtype=dtype)
-        except ValueError:
+            dtype = dtypemap[dtype]
+        except KeyError:
+            log.error('Invalid dtype: dtype must be float, int, or bool.')
```
matthewfeickert: If there is a better way to raise the offending error with a custom message that gives more information let me know and I'll implement it.

kratsg: that's fine.  just add a test to ensure coverage.

pyhf/tensor/mxnet_backend.py
```diff
@@ -105,11 +105,17 @@ def astensor(self, tensor_in, dtype='float'):
...
+            raise
+
+        if isinstance(tensor_in, (int, float)):
```
kratsg: all these isinstance checks will slow down the code a little.. and `astensor` is something we call *a lot*

kratsg:
I think we shouldn't cast scalars to lists. There's nothing in our code that is returning scalars -- so it's a user problem if scalars are passed in...

pyhf/tensor/mxnet_backend.py
matthewfeickert: Okay, I can revert them as we have gotten by without them with no problems.

matthewfeickert: I guess this is something that we should discuss on your other comment as well, but it seemed that this would be a cleaner way to do it then my original method of

```python
try:
    tensor.shape[0]
except IndexError:
    tensor.function_to_expand_the_shape 
```

Do you think catching the `IndexError` is better, or do you think we shouldn't be trying at all?

matthewfeickert:
> I think we shouldn't cast scalars to lists. There's nothing in our code that is returning scalars

So do you just want the offending tests to get changed then? The CI currently breaks without this because we don't enforce a consistent return structure.

kratsg:
> So do you just want the offending tests to get changed then? The [CI currently breaks without this](https://travis-ci.org/diana-hep/pyhf/jobs/500639468) because we don't enforce a consistent return structure.

Maybe. I wonder what @lukasheinrich thinks. The problem is that I think for a given fit, we call `astensor` something like 10k times?

pyhf/tensor/mxnet_backend.py
matthewfeickert: Doesn't the `test_astensor` test I added do that?

matthewfeickert:
> The problem is that I think for a given fit, we call `astensor` something like 10k times

Yeah, this is a very valid point, though we already have this in the TensorFlow backend

https://github.com/diana-hep/pyhf/blob/4fa45dda8d2314138549ac6b67ebdc9d9a506742/pyhf/tensor/tensorflow_backend.py#L110-L118

so do we know how much it gets slowed down?

pyhf/tensor/mxnet_backend.py
kratsg: I think this is better at least... because then slowness is fixable by making sure we don't pass in scalars...

pyhf/tensor/mxnet_backend.py
kratsg: > Doesn't the `test_astensor` test I added do that?

You don't have full coverage -- e.g. checking for the KeyError message, and so on.

pyhf/tensor/mxnet_backend.py
matthewfeickert: Yes, EAFP is Pythonic and I'm definitely being all LBYL. I'll revert these to how I originally had them. :+1:

pyhf/tensor/mxnet_backend.py
matthewfeickert: I probably don't know how to properly test failure cases in pytest (will go read now) but why doesn't

```
assert tb.astensor([1, 2, 3], dtype='long')
```

in `test_astensor` trigger the `KeyError` case? Or am I totally missing the point here?

matthewfeickert: Ah, did you mean that I should be using [`mark.xfail`'s `raises` argument](https://docs.pytest.org/en/latest/skipping.html#raises-parameter)?

matthewfeickert: Yes, you're right that `pytest.raises` is more appropriate here as from the [pytest docs](https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions)

> Using `pytest.raises` is likely to be better for cases where you are testing exceptions your own code is deliberately raising, whereas using `@pytest.mark.xfail` with a check function is probably better for something like documenting unfixed bugs (where the test describes what "should" happen) or bugs in dependencies.

PR #426, opened 2019-04-08
Feature: json2xml (XML+ROOT exporting)

### LINKED ISSUE #170: XML ROOT Export
# Description

Given that we observe that some models still scale better in ROOT, it's reasonable to have some an companion tools to `readxml`, namely `writexml` that writes XML + ROOT files corresponding to a model given as JSON.

kratsg:
# Description

This pull request adds the functionality of exporting JSON specs into XML+ROOT for use with RooStats/HistFactory/HistFitter. This resolves #170.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- flesh out writexml.py to actually write out the XML+ROOT dump
- update tests to match the expected command line interface
- added unit tests to check exporting functionality per-element
- added integration tests to check full export
- added validation tests to check a round-trip xml->json->xml->json for the simplified XML import examples
```

### DIFF
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -61,15 +61,23 @@ def xml2json(entrypoint_xml, basedir, output_file, track_progress):
 
 @pyhf.command()
 @click.argument('workspace', default='-')
-@click.argument('xmlfile', default='-')
-@click.option('--specroot', default=click.Path(exists=True))
-@click.option('--dataroot', default=click.Path(exists=True))
-def json2xml(workspace, xmlfile, specroot, dataroot):
+@click.option('--output-dir', type=click.Path(exists=True), default='.')
+@click.option('--xmlfile', default='FitConfig.xml')
+@click.option('--specroot', default='config')
+@click.option('--dataroot', default='results')
+def json2xml(workspace, output_dir, xmlfile, specroot, dataroot):
+    os.makedirs(os.path.join(output_dir, specroot), exist_ok=True)
+    os.makedirs(os.path.join(output_dir, dataroot), exist_ok=True)
     with click.open_file(workspace, 'r') as specstream:
         d = json.load(specstream)
-        with click.open_file(xmlfile, 'w') as outstream:
+        with click.open_file(os.path.join(output_dir, xmlfile), 'w') as outstream:
             outstream.write(
-                writexml.writexml(d, specroot, dataroot, '').decode('utf-8')
+                writexml.writexml(
+                    d,
+                    os.path.join(output_dir, specroot),
+                    os.path.join(output_dir, dataroot),
+                    '',
+                ).decode('utf-8')
             )
     sys.exit(0)
 
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -62,10 +62,10 @@ def xml2json(entrypoint_xml, basedir, output_file, track_progress):
 @pyhf.command()
 @click.argument('workspace', default='-')
 @click.option('--output-dir', type=click.Path(exists=True), default='.')
-@click.option('--xmlfile', default='FitConfig.xml')
 @click.option('--specroot', default='config')
 @click.option('--dataroot', default='data')
-def json2xml(workspace, output_dir, xmlfile, specroot, dataroot):
+@click.option('--resultprefix', default='FitConfig')
+def json2xml(workspace, output_dir, specroot, dataroot, resultprefix):
     os.makedirs(output_dir, exist_ok=True)
     with click.open_file(workspace, 'r') as specstream:
         d = json.load(specstream)
@@ -74,9 +74,11 @@ def json2xml(workspace, output_dir, xmlfile, specroot, dataroot):
             os.chdir(output_dir)
             os.makedirs(specroot, exist_ok=True)
             os.makedirs(dataroot, exist_ok=True)
-            with click.open_file(xmlfile, 'w') as outstream:
+            with click.open_file('{0:s}.xml'.format(resultprefix), 'w') as outstream:
                 outstream.write(
-                    writexml.writexml(d, specroot, dataroot, '').decode('utf-8')
+                    writexml.writexml(d, specroot, dataroot, resultprefix).decode(
+                        'utf-8'
+                    )
                 )
         finally:
             os.chdir(CWD)
--- a/pyhf/commandline.py
+++ b/pyhf/commandline.py
@@ -14,6 +14,14 @@ from .version import __version__
 logging.basicConfig()
 log = logging.getLogger(__name__)
 
+# This is only needed for Python 2/3 compatibility
+def ensure_dirs(path):
+    try:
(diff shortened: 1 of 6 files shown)


pyhf/commandline.py
```diff
@@ -61,16 +61,28 @@ def xml2json(entrypoint_xml, basedir, output_file, track_progress):
...
+    os.makedirs(output_dir, exist_ok=True)
     with click.open_file(workspace, 'r') as specstream:
         d = json.load(specstream)
-        with click.open_file(xmlfile, 'w') as outstream:
-            outstream.write(
-                writexml.writexml(d, specroot, dataroot, '').decode('utf-8')
-            )
+        CWD = os.getcwd()
+        try:
+            os.chdir(output_dir)
```
lukasheinrich: can we get around this somehow?

pyhf/writexml.py
```diff
@@ -1,45 +1,207 @@
+import logging
+
 import os
 import xml.etree.cElementTree as ET
+import numpy as np
+import uproot
+from uproot_methods.classes import TH1
+
+_ROOT_DATA_FILE = {}
+_HISTNAME = "h{sample}{modifier}{highlow}_{channel}_obs_cuts"
```
lukasheinrich: the `obs_cuts` is a HistFitter-ism I don't think we need to reintroduce it here

pyhf/writexml.py
```diff
@@ -1,45 +1,207 @@
...
+    #    raise KeyError('Duplicate key {0} being written.'.format(histname))
+    _ROOT_DATA_FILE[histname] = h
 
 
-def measurement(lumi, lumierr, poi, param_settings=None, name='Meas1'):
-    param_settings = param_settings or []
+# https://stackoverflow.com/a/4590052
+def indent(elem, level=0):
+    i = "\n" + level * "  "
+    if len(elem):
```
matthewfeickert: Can we use the [CodeFactor suggestion](https://www.codefactor.io/repository/github/diana-hep/pyhf/pull/426) here RE: "Do not use `len(SEQUENCE)` to determine if a sequence is empty"?

tests/test_export.py
```diff
@@ -0,0 +1,256 @@
...
+    channelspec = spec['channels'][0]
+    channelname = channelspec['name']
+
+    mocker.patch('pyhf.writexml.build_data', return_value=ET.Element("Data"))
+    mocker.patch('pyhf.writexml.build_sample', return_value=ET.Element("Sample"))
+    mocker.patch('pyhf.writexml._ROOT_DATA_FILE')
+    channel = pyhf.writexml.build_channel(channelspec, {})
+    assert channel.attrib['Name'] == channelspec['name']
+    assert channel.attrib['InputFile']
+    assert pyhf.writexml.build_data.called == False
```
matthewfeickert: Can we use the [CodeFactor suggestion](https://www.codefactor.io/repository/github/diana-hep/pyhf/pull/426) here RE: "Comparison to False should be 'expr is False'"?

matthewfeickert:
This might get fixed or was a 1 off, so I'll wait for the CI to finish, but at the moment there is [a Python 2.7 error in `test_import_and_export`](https://travis-ci.org/diana-hep/pyhf/jobs/517261613#L683-L704)

```
______________________ test_import_and_export[inprocess] _______________________
tmpdir = local('/tmp/pytest-of-travis/pytest-0/test_import_and_export_inproce0')
script_runner = <ScriptRunner inprocess>
    def test_import_and_export(tmpdir, script_runner):
        temp = tmpdir.join("parsed_output.json")
        command = 'pyhf xml2json validation/xmlimport_input/config/example.xml --basedir validation/xmlimport_input/ --output-file {0:s}'.format(
            temp.strpath
        )
    
        command = 'pyhf json2xml {0:s} --output-dir {1:s}'.format(
            temp.strpath, tmpdir.mkdir('output').strpath
        )
>       assert ret.success
E       assert False
E        +  where False = <pytest_console_scripts.RunResult object at 0x7efd81828b10>.success
tests/test_scripts.py:113: AssertionError
```

pyhf/writexml.py
matthewfeickert: Thanks! :slightly_smiling_face:

pyhf/readxml.py
```diff
@@ -235,14 +235,15 @@ def process_measurements(toplvl):
...
-                    result['config']['parameters'][0].update(overall_param_obj)
-                else:
-                    param_obj = {'name': param_name}
-                    param_obj.update(overall_param_obj)
-                    result['config']['parameters'].append(param_obj)
+            if param.text:
+                for param_name in param.text.split(' '):
+                    # lumi will always be the first parameter
+                    if param_name == 'Lumi':
+                        result['config']['parameters'][0].update(overall_param_obj)
```
lukasheinrich: are we guaranteed that Lumi will always be at index 0?

lukasheinrich: maybe we can add an `assert result['config']['parameters'][0]['name']=="Lumi"`?

pyhf/writexml.py
```diff
@@ -1,45 +1,207 @@
+import logging
+
 import os
 import xml.etree.cElementTree as ET
+import numpy as np
+import uproot
+from uproot_methods.classes import TH1
+
+_ROOT_DATA_FILE = {}
+_HISTNAME = "hist_{channel}_{sample}_{modifier}{highlow}"
```
lukasheinrich: I feel like this should rather me a function

```
def make_hist_name(channel, sample,modifier='', suffix='')
```

pyhf/readxml.py
kratsg: This wasn't code that I added.

lukasheinrich: wasn't it added here? https://github.com/diana-hep/pyhf/pull/352 but granted not this PR

kratsg: Yeah, I guess it *was* code I added. I'll fix momentarily.

kratsg: This is not a problem. Lumi is always guaranteed to be at index 0 here.

pyhf/writexml.py
```diff
@@ -1,45 +1,210 @@
+import logging
+
 import os
 import xml.etree.cElementTree as ET
+import numpy as np
+import uproot
+from uproot_methods.classes import TH1
+
+_ROOT_DATA_FILE = {}
```
lukasheinrich: I feel like initializing this as `{}` is misleading since it will become a `TFileRecreate` (which admittedly is dict-like). Since its type will change from dict to TFileRecreate anyways, isn't it better to initialize this as `None`?

kratsg: I can fix this now. I had switched it because I was trying to mock it as a dictionary (since all we'd had ever used it for was dictionary-like access which worked well)

kratsg:
@lukasheinrich , @matthewfeickert  -- all comments are addressed.

PR #428, opened 2019-04-09
Use sorted() to maintain ordering of channel/samples/modifiers/parameters

### LINKED ISSUE #414: Order of pdf.confg.samples not guaranteed (due to use of set)
# Description

The order of samples in [`pdf.config.samples`](https://github.com/diana-hep/pyhf/blob/4fa45dda8d2314138549ac6b67ebdc9d9a506742/pyhf/pdf.py#L163) (more specifically, [`pdf._ModelConfig.samples`](https://github.com/diana-hep/pyhf/blob/5ddfc6aa21d1f2e33a2fe95bfefb489541ae77ff/pyhf/pdf.py#L35)) should not change between runs with the same data. However, it can, which is highly problematic when trying to do something like plotting as we do in the examples. If `pdf.config.samples` does not maintain order between runs, then something like [`get_mc_counts`](https://github.com/diana-hep/pyhf/blob/4fa45dda8d2314138549ac6b67ebdc9d9a506742/docs/examples/notebooks/binderexample/StatisticalAnalysis.ipynb) will give different results on different runs because `pdf._modifications(pars)` will be giving different results.

I think this is coming from the fact that the use of `set` in

https://github.com/diana-hep/pyhf/blob/5ddfc6aa21d1f2e33a2fe95bfefb489541ae77ff/pyhf/pdf.py#L93

[doesn't guarantee order](https://stackoverflow.com/questions/9792664/converting-a-list-to-a-set-changes-element-order).

I think what should be used instead is a sorted set

```python
self.samples = sorted(set(self.samples), key=self.samples.index)
```

Consider the following two scripts:

`bug_report.py`
```python
#!/usr/bin/env python3
import pyhf


def main():
    pdf = pyhf.simplemodels.hepdata_like(
    )
    print(pdf.config.samples)


if __name__ == '__main__':
    main()
```

and `bug_resport.sh`
```shell
#!/usr/bin/env bash

for i in {1..10}; do
   python3 bug_report.py 
done
```

Thanks to @stephensekula for uncovering some of this behavior.

# Expected Behavior

Running `bash bug_report.sh` should return

```
$ bash bug_report.sh 
['background', 'signal']
['background', 'signal']
['background', 'signal']
['background', 'signal']
['background', 'signal']
['background', 'signal']
['background', 'signal']
['background', 'signal']
['background', 'signal']
['background', 'signal']
```

# Actual Behavior

```
$ bash bug_report.sh 
['background', 'signal']
['signal', 'background']
['signal', 'background']
['background', 'signal']
['signal', 'background']
['signal', 'background']
['signal', 'background']
['signal', 'background']
['background', 'signal']
['signal', 'background']
```

# Steps to Reproduce

This Issue.

# Checklist

kratsg:
# Description

This resolves #414.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- we use list(set(...)) to generate the metadata of a given spec but set() does not guarantee order
```

### DIFF
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -89,10 +89,10 @@ class _ModelConfig(object):
                         paramset_requirements
                     )
 
-        self.channels = list(set(self.channels))
-        self.samples = list(set(self.samples))
-        self.parameters = list(set(self.parameters))
-        self.modifiers = list(set(self.modifiers))
+        self.channels = sorted(list(set(self.channels)))
+        self.samples = sorted(list(set(self.samples)))
+        self.parameters = sorted(list(set(self.parameters)))
+        self.modifiers = sorted(list(set(self.modifiers)))
         self.channel_nbins = self.channel_nbins
         self._create_and_register_paramsets(
             _paramsets_requirements, _paramsets_user_configs


matthewfeickert:
@kratsg I've approved this, but let me know if you think there is any merit in using [`OrderedDict`](https://github.com/diana-hep/pyhf/issues/414#issuecomment-469902132). I honestly don't know as I haven't looked at it.

kratsg:
> @kratsg I've approved this, but let me know if you think there is any merit in using [`OrderedDict`](https://github.com/diana-hep/pyhf/issues/414#issuecomment-469902132). I honestly don't know as I haven't looked at it.

There's no need. The metadata we use for ordering is simply for self-consistency within a single pyhf invocation. The addition of `sorted()` just adds consistency across multiple invocations as well.

PR #431, opened 2019-04-09
fix divide to only divide non-zero pieces.

kratsg:
# Description

In situations where, for example, sample-data is zero - this causes a runtime warning about divide (and returns a undefined instead of zero) when we need to convert from absolute uncertainties back to relative uncertainties. This is a slight drawback in the translation, but should not affect the end-results whether or not we choose to return undefined or zero if it's treated identically in both cases.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- only divide out sample data in cases where sample data is non-zero when exporting to XML
- this only affects modifiers that convert from absolute uncertainties back to relative uncertainties
```

### DIFF
--- a/pyhf/writexml.py
+++ b/pyhf/writexml.py
@@ -132,6 +132,7 @@ def build_modifier(modifierspec, channelname, samplename, sampledata):
                 sampledata,
                 out=np.zeros_like(sampledata),
                 where=np.asarray(sampledata) != 0,
+                dtype='float',
             ).tolist(),
         )
     elif modifierspec['type'] == 'shapesys':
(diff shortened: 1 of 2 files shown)


matthewfeickert:
@kratsg I had thought that you had wanted to add tests for this which is why it hadn't gotten merged in a while ago. Do you still want to do that?

kratsg:
> @kratsg I had thought that you had wanted to add tests for this which is why it hadn't gotten merged in a while ago. Do you still want to do that?

Yeah, i need to add tests and get the thing to pass.

matthewfeickert:
> Yeah, i need to add tests and get the thing to pass.

Cool. This all looks good already so just ping when you're happy with tests and we can get this in. :+1:

Issue #410, opened 2019-02-19
PyPI documentation not rendered properly


matthewfeickert:
# Description

Currently ([pyhf `v0.0.16`](https://pypi.org/project/pyhf/0.0.16/)) the docs on PyPI are not being rendered correctly. This shouldn't be the case as

```
$ python setup.py sdist
$ python setup.py bdist_wheel --universal
$ twine check dist/*
Checking distribution dist/pyhf-0.0.16-py2.py3-none-any.whl: Passed
Checking distribution dist/pyhf-0.0.16.tar.gz: Passed
```

for
```
$ pip freeze | grep 'twine\|readme-renderer'
readme-renderer==24.0
twine==1.13.0
```

If there was a problem with the rendering then [`twine check dist/*`](https://packaging.python.org/guides/making-a-pypi-friendly-readme/#making-a-pypi-friendly-readme) should have picked it up. Yet, it is being rendered wrong.

@lukasheinrich Can you tell us what versions of `twine` you're using when you run 

```
twine upload dist/*
```

? I think there may have been an [issue](https://github.com/pypa/twine/pull/421) with `twine` `v1.12.0` that got fixed recently in `v1.13.0`

# Checklist

- [x] Run `git fetch` to get the most up to date version of `master`
- [x] Searched through existing Issues to confirm this is not a duplicate issue
- [x] Filled out the Description, Expected Behavior, Actual Behavior, and Steps to Reproduce sections above or have edited/removed them in a way that fully describes the issue

webknjaz:
I strongly recommend you using `pyproject.toml` and building dists using `pep517.build` via a standardized interface :)

matthewfeickert:
One of the things that is missing for PyPI is the following in [`MANIFEST.in`](https://github.com/diana-hep/pyhf/blob/fbba4b22a43ceafd1a7202b492cd7a25e2ba9f3c/MANIFEST.in)

```
include docs/_static/img/*.png
```

matthewfeickert:
> I strongly recommend you using `pyproject.toml` and building dists using `pep517.build` via a standardized interface

@webknjaz I agree with you. I need to get around to implementing Issue #399

matthewfeickert:
Ah, this is tricky. As is [described here](https://github.com/pypa/warehouse/issues/5246#issuecomment-451142432)

> PyPI does not support rendering from the package path itself, as it is not an image host.

[...]

Issue #399, opened 2019-01-31
Enable install through pyproject.toml


matthewfeickert:
# Description

Given [PEP 518](https://www.python.org/dev/peps/pep-0518/) it would be nice to have the ability to install pyhf through the `pyproject.toml` (which already exists since PR #322) as an alternative (...or in the future a replacement) to using the `setup.py` with `pip`. This would also allow installation with dependency managers like [Poetry](https://poetry.eustace.io/).

matthewfeickert:
Should also consider moving the bulk of `setup.py` to `setup.cfg` to take advantage of [`setuptools`'s work](https://github.com/pypa/setuptools/pull/1675) to support [PEP 517](https://www.python.org/dev/peps/pep-0517/)

c.f. [Bernat Gabor's post on this](https://www.bernat.tech/pep-517-518/)

webknjaz:
Just drop in a `pyproject.toml` like this and that's it:
```toml
[build-system]
requires = [
  "setuptools >= 41.0.0",
  "setuptools_scm >= 1.15.0",
  "setuptools_scm_git_archive >= 1.0",
  "wheel",
]
build-backend = "setuptools.build_meta"
```

matthewfeickert:
Given PR #455 this [Twitter thread](https://twitter.com/btskinn/status/1121633712209137665) from @bskinn will probably be useful in the future for doing this properly.

webknjaz:
This mostly hit `tox` users (because it always uses the latest published pip): https://github.com/tox-dev/tox/issues/1276. But the maintainers are quite responsive and the fix should be out soon anyway :)

matthewfeickert:
> This mostly hit tox users 

Yeah, it happened to us as well (PR #455). I've been watching [the Issue](https://github.com/pypa/pip/issues/6434), which should help in the future.

webknjaz:
It's not the same: tox updates pip to the latest version and the workaround to this is quite non-trivial for tox specifically :)

bskinn:
Huh. It must be a particular style of tox usage, because my tox (using it for factorial testing of different dependency matrices) works fine for two different projects.

webknjaz:
You'll hit a bug only if you have a combo of `pyproject.toml` file existing in the project + editable mode.

webknjaz:
The workaround, for now, is smth like this: https://github.com/ansible/molecule/pull/2009

bskinn:
[...]

PR #436, opened 2019-04-11
Finish adding deploy testing with TestPyPI

### LINKED ISSUE #425: Publish to Test PyPI to ensure robust release pipeline
# Description

Following [this tweet](https://twitter.com/webKnjaZ/status/1114198195222003712)'s suggestion we should be publishing each commit to [Test PyPI](https://test.pypi.org/) to make sure that the publishing pipeline works. This should nicely compliment Issue #410 and keep issues like it from happening in the future.

### LINKED ISSUE #432: Test Deploy feature is broken
# Description

See https://github.com/FaradayRF/faradayio/issues/35 for a related issue - I think. Failing job on master here: https://travis-ci.org/diana-hep/pyhf/builds/517678508?utm_source=github_status&utm_medium=notification

# Expected Behavior

Expect it to pass.

# Actual Behavior

Observe it failing.

# Checklist

matthewfeickert:
# Description

Completes PR #429 (which was merged early by accident) and resolves #425 and resolves #432.

Inspired by Issue #410 and @webknjaz's tweet which inspired Issue #425 to have more robust deployment to PyPI this PR implements PR level deployments to [TestPyPI](https://test.pypi.org/) that are versioned with version structure of `{next_version}.dev{distance}` through use of [`setuptools-scm`](https://github.com/pypa/setuptools_scm/) and [`setuptools-scm-git-archive`](https://github.com/Changaco/setuptools_scm_git_archive/) and heavy copying of [`ansible-lint`'s approach](https://github.com/ansible/ansible-lint/blob/d27451783dbc4f408cef04096a139f6be61f7611/setup.py).

The level of deployment is set to push commits on `master` as we commit frequently enough that there isn't a need to pollute the area with a huge number of deployed versions.

As an example of the system working, by turning on deployment from all branches this PR [deployed](https://travis-ci.org/diana-hep/pyhf/jobs/518643360) [`pyhf 0.0.17.dev23`](https://test.pypi.org/project/pyhf/0.0.17.dev23/) on TestPyPI.

This PR also sets up the path to then complete Issues #410 and #399.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- Add deployment of push commits to master to TestPyPI for deploy testing. This uses a TESTPYPI_UPLOAD environment variable to signal to setuptools-scm through setup.py.
- Add setup_requires of setuptools-scm and setuptools-scm-git-archive through setup.cfg
- Add .git_archival.txt and .gitattributes as required for setuptools-scm-git-archive
- For TestPyPI deploy job use type=push to avoid triggering on unintended builds
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ stages:
   - name: binder
     if: (branch = master) AND (NOT (type IN (pull_request)))
   - name: deploy
-    if: (tag =~ v[0-9\.]*)
+    # if: (tag =~ v[0-9\.]*)
 
 env:
   global:
@@ -122,6 +122,7 @@ jobs:
         secure: CqHxAg2GWY+7ogMKJYCqH8Eynmc1Xhh+WocntEfsJXOgsqksmHF5q9dp6IxGwrrts1ZWbxOs80mici2LSAxtHnVBAXfiqWm5cZ6nEv6+hvAzidvGO7Q2OVs3CIVKU4jCDg2BJIBo84Whqwe6T5S+DKUDo6PYQUtIvHqnO4PmDyB2Y2rLquavMqMkh8wIfFaDj51J5zm2aNdXx4349LvlMcW2yn0/Mh1t6yP1LUCkirCPjdIAljmqmrkxscEdCepR3sXjgYggpW/RLTXtMaUc/6Pz1Ax5uU3Y/mtuXT19lC+9PiFpciuLBvM6h0RjRz87tVE4US7DfbjZIIZFdrd97eog9Jqh0cDsD+lymgzTpeEaisXtDwBlI+osUtjB5d1rWCtsIu39fBRi368SLgBH3yrbYEIuNBzczeQ1wxMmJlWflQ/6nbPEaRxKpMZVjMRngdd8rccNj0vAJxsMlzqJZ99RbUjyMQ5n1JO1GU/9mnu1NH/+gZrvoGCLBaqKQDewrTkVKAMTmm/lti9hIEpgXr+U5KyAM0dtvU3ncLCEhIV6H9KujHx8S52hz/C1OJAXwh4otC2rnuiS21XhFfADJbVtXE9lBX0ANgc/pWs11Yu38ZbfoXnx5GkALuUnw2jZ2ZeEiugu5/dqlPTmleBgkxcRsVNx8/fft3raIscIYjM=
       distributions: "sdist bdist_wheel"
   - stage: deploy
+    if: (tag =~ v[0-9\.]*)
     python: '3.7'
     install: skip
     script: skip
--- a/.travis.yml
+++ b/.travis.yml
@@ -107,7 +107,7 @@ jobs:
   - stage: deploy
     # Tests deployment to PyPI for all commits to master
     name: "TestPyPI deploy"
-    if: (branch = master) AND (NOT (type IN (pull_request)))
+    if: (branch = master) AND type = push
     env:
       # Signal that setup.py should use use_scm_version with local_scheme
       TESTPYPI_UPLOAD: true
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -107,8 +104,13 @@ jobs:
       # Use Binder build API to trigger repo2docker to build image
       - bash binder/trigger_binder.sh https://mybinder.org/build/gh/diana-hep/pyhf/"${TRAVIS_BRANCH}"
     after_success: skip
-  - stage: test_deploy
+  - stage: deploy
+    # Tests deployment to PyPI for all commits to master
     name: "TestPyPI deploy"
+    if: (branch = master) AND (NOT (type IN (pull_request)))
```
webknjaz: Better
```suggestion
    if: branch = master AND type = push
```

Just because otherwise it'll be triggered on periodic cron builds and api (or manual via Travis UI) builds which will conflict in pypi

setup.py
```diff
@@ -56,6 +56,26 @@
...
+    configuration parameter with argument "version". This can be replaced
+    with a lambda as the desired version structure is {next_version}.dev{distance}
+    c.f. https://github.com/pypa/setuptools_scm/#importing-in-setuppy
+    """
+    from os import getenv
+
+    return (
+        {'local_scheme': lambda version: ''}
+        if getenv('TESTPYPI_UPLOAD') == 'true'
+        else False
```
webknjaz: This will cause a fallback to the version hardcoded in python code on line 81.

matthewfeickert: Yes, this is the intended behavior. We only want scm versioning for TestPyPI.

webknjaz: Ah, I see :)
It's an interesting approach which I haven't met before...

webknjaz: But be aware that this may cause version mismatches between tag and dist because of human errors

matthewfeickert: Yeah, the main reason is that we currently control everything through [bumpversion](https://github.com/peritus/bumpversion), and I wanted to keep this PR as atomic as possible. We could consider changing things in the future.

webknjaz: I'm currently working on a GitHub Action / GitHub App adding a Deploy button to Checks pages of each commit. Such an approach would allow triggering the whole flow with an actual click on GitHub UI ;)

matthewfeickert: We don't have GitHub actions on pyhf yet! But we are very much looking forward to using it when we get it, so we'll be following up with you! :)

webknjaz: You need to sign up first https://github.com/features/actions/signup

setup.cfg
```diff
@@ -3,3 +3,8 @@ universal = 1
 
 [metadata]
 license_file = LICENSE
+
+[options]
+setup_requires =
```
webknjaz: You know that with old setuptools this won't work, right?

matthewfeickert: Yup. But deployments happen with Python 3.7 and modern setuptools. We're in the slow process of starting the migration of getting ready to drop Python 2.7, but our field (high energy physics) is still using older everything so we haven't gone Python 3 and modern libs only yet.

setup.py
matthewfeickert: Indeed. I have it for my personal account, but our org doesn't yet.

setup.cfg
webknjaz: Well, as long as you ship binary wheels building them yourself it shouldn't matter.

setup.py
webknjaz: Oh, in this case, I invite you to complete this tutorial https://tutorial.octomachinery.dev :)

matthewfeickert: Thanks! That looks great. :)

matthewfeickert:
@webknjaz Thanks very much for being so generous with your time to give help and feedback! :+1:

webknjaz:
You're welcome :)

PR #440, opened 2019-04-12
Workspace specification created, specifications split up

kratsg:
# Description

This introduces minor, backwards-breaking changes to the JSON specs we know and love. `toplvl` and `resultsprefix` are now being dropped from our JSONs.
- `workspace.json` is defined to provide a specification for the workspace JSONs that are generated from `xml2json` and that are consumed by `json2xml`, as well as the new `Workspace` object
- `model.json` is defined to provide a specification for the JSONs that are consumed by the `Model` object
- `measurement.json` is defined to ensure that the `Workspace` provides an appropriate `measurement` when asking for a measurement (also usable for anyone to inject their own custom measurement as long as it follows the spec)
- `defs.json` is defined to contain the common objects / concepts in `pyhf` such as modifier and parameter

Additionally, now that the workspace exists, the commandline interface is being updated to take advantage of the new API to be more DRY.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- a Workspace object is created to improve the pyhf API for UI/UX
- the JSON schemas are split up and modularized to provide more robust checking against readxml/writexml
```


pyhf/pdf.py
```diff
@@ -450,3 +451,130 @@ def logpdf(self, pars, data):
...
+        self.measurement_names = []
+        for measurement in self.spec.get('measurements', []):
+            self.measurement_names.append(measurement['name'])
+
+    # NB: this is a wrapper function to validate the returned measurement object against the spec
+    def get_measurement(self, **config_kwargs):
+        """
+        Get (or create) a measurement object using the following logic:
+
+        1. if the poi name is given, create a measurement object for that poi
```
matthewfeickert: All these items in the `get_measurement` docstring are prefixed by `1.` Is this intentional, or were these supposed to be `1.`, `2.`, ...?

kratsg: I didn't check how the documentation rendered these. I generally assumed markdown rendering.

PR #454, opened 2019-04-25
Add case 4 of the piecewise interpolator

lukasheinrich:
# Description

This adds case 4 of the PiecewiseInterpolator of ROOT

https://root.cern.ch/doc/v614/PiecewiseInterpolation_8cxx_source.html

(line 225)

not to be confused of case 4 of the FlexibleInterp

as the naming becomes confusing i'm adding this as `codex` until we come up with better naming

- [x] Tests are passing

# Before Merging

For the PR Assignees:


matthewfeickert:
@lukasheinrich This commit I've added is just cherry-picked from PR #455 to get the CI to at least run through until it encounters the [`uproot-methods` Issue 53](https://github.com/scikit-hep/uproot-methods/issues/53) problem.

kratsg:
Let's call this `code4p` (p for piecewise)?

matthewfeickert:
@lukasheinrich if you have time to fix the `pyflakes` error tomorrow that would be great. I can get to it if you are busy though.

pyhf/interpolators/__init__.py
```diff
@@ -20,6 +20,7 @@ def _slow_interpolator_looper(histogramssets, alphasets, func):
 from .code1 import code1, _slow_code1
 from .code2 import code2, _slow_code2
 from .code4 import code4, _slow_code4
+from .codex import codex, _slow_codex
```
matthewfeickert: @lukasheinrich This all looks good, but what are your thoughts on @kratsg's proposed naming scheme?

lukasheinrich: i don't like it since it messes up the alignment :-p .. j/k no strong optinion either way

matthewfeickert: It isn't my most favorite to be honest, but I don't have a better name and it probably is better than `x`.

pyhf/interpolators/__init__.py
```diff
@@ -29,6 +30,7 @@ def get(interpcode, do_tensorized_calc=True):
         1: code1 if do_tensorized_calc else _slow_code1,
         2: code2 if do_tensorized_calc else _slow_code2,
         4: code4 if do_tensorized_calc else _slow_code4,
+        '4p': code4p if do_tensorized_calc else _slow_code4p,
```
matthewfeickert: @kratsg @lukasheinrich Is there any strong motivation to not just making things _very_ explicit and call this `4piecewise` or `4piece` to try to make things quite clear given the naming confusion?

kratsg: Notational short-hand in code is very different from the long-form we'll do in the documentation.

matthewfeickert: I think we've talked about this before, but I don't see the need to abbreviate much in names like this. But this isn't a hill I find worth dying on. :)

kratsg: if you can find two names that are non-abbreviated with the same number of characters ;)

PR #463, opened 2019-05-02
Remove editable flag for Binder to avoid PEP 517 issue with pip

matthewfeickert:
# Description

The `postBuild` file for Binder is breaking given the issues from pip v19.1 and PEP 517 that happened in PR #455. As Binder doesn't need to have an editable install of pyhf in it this can be resolved by not installing in editable mode, as there is no need for Binder.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- Follow up to PR #455
- Remove -e flag from pip install in postBuild to have Binder avoid issues with PEP 517 and pip
```

### DIFF
--- a/binder/postBuild
+++ b/binder/postBuild
@@ -1,1 +1,1 @@
-pip install --upgrade -e .[complete]
+pip install --upgrade .[complete]


matthewfeickert:
In what I can only describe as Travis CI being really weird and unreliable, [earlier the build for this PR was failing from memory errors](https://travis-ci.org/diana-hep/pyhf/builds/527914856) like that of PR #468. Then, without changing any code, I force pushed to restart a build and everything passes. :? I think moving to GitHub Actions will hopefully be a welcome relief in the future.

PR #452, opened 2019-04-23
Add /schemas to documentation

kratsg:
# Description

This adds the schemas in `pyhf` to the documentation under `/schemas`. Therefore paths as specified in the JSON specs we've written should work as expected, e.g.

```json
{
    "$schema": "http://json-schema.org/draft-06/schema#",
    "$id": "https://diana-hep.org/pyhf/schemas/defs.json",
    "definitions": {
...
```

# Checklist Before Requesting Reviewer

- [ ] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR
```
- move pyhf/data/ to pyhf/schemas/
- add pyhf/schemas/ to documentation root under schemas/
```

PR #472, opened 2019-05-07
Add semantic-versioning to JSON schemas

### LINKED ISSUE #451: add versioning to schema
# Description

For future-proofing we need to add versioning to the schemas. I would propose that we have a simple string-based version

```
version: "1.0"
.. remaining keys possibly version specific ..
```

and the following strategy

* when loading the model, it is validated against schemas of the declared version. 

* within the code, a given in-memory representation i.e. the `Workspace` class can be 
instantiated using specs of a subrange of versions `[v_min,v_max]`. possible the ctor will forward/backward evolve to a common format ( I suspect usually forward, i.e. accept a v1 representation but convert it into a v2 representations internally

* a given `pyhf` version  can read a subset of pyhf schema versions. The goal here should be to be as backward and forwards compatible as possible
  * forward compatible: validating the version number itself should be optional (i.e. a v2 spec might be structurally compatible with v1, e.g. if we add new optional fields) so it should still validate as a v1 model even though it has "version: v2" in the spec.
  * backward compatible: future pyhf versions should carry all older schemas as well so older models can be validated. if needed we will add converters for forward evolution to a more recent version once a model is validated in a older version
  * probably would be nice to have `pyhf` be able to output what version range is supported
  * to the extend that pyhf has translators from one version to another, a cli to do this conversion might be nice in order to normalize let's say multiple worksapces defined in different versions for a combination 

* if backward compatibility is not possible, this means a older version of the spec can only be read in a older version of pyhf. This is fine per se, this necessarily will require a new major release, but we should really try to avoid it.

kratsg:
# Description

This moves the JSON schemas into a versioned location and adds a version attribute to `workspace.json` that should match the `$id` as well.

This resolves #451.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- JSON schemas are now versioned
- workspace.json explicitly requires a version attribute as this is the JSON document that gets preserved/passed around (model.json and measurement.json do not explicitly require it)
- utils.validate and utils.load_schema are updated to handle versioning as well
```

### DIFF
--- a/pyhf/pdf.py
+++ b/pyhf/pdf.py
@@ -176,9 +176,10 @@ class Model(object):
     def __init__(self, spec, **config_kwargs):
         self.spec = copy.deepcopy(spec)  # may get modified by config
         self.schema = config_kwargs.pop('schema', 'model.json')
+        self.version = config_kwargs.pop('version', None)
         # run jsonschema validation of input specification against the (provided) schema
         log.info("Validating spec against schema: {0:s}".format(self.schema))
-        utils.validate(self.spec, self.schema)
+        utils.validate(self.spec, self.schema, version=self.version)
         # build up our representation of the specification
         self.config = _ModelConfig(self.spec, **config_kwargs)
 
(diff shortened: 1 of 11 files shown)


matthewfeickert:
@kratsg Does the result of PR #452 need to get updated for this to work?

kratsg:
> @kratsg Does the result of PR #452 need to get updated for this to work?

Nope. That runs `rsync -r ../pyhf/schemas/ $(BUILDDIR)/html/schemas` which means as long as everything is under `pyhf/schemas` -- it'll be copied over correctly -- for all versions. I wrote #452 forseeing that we'd move to versioning later.

matthewfeickert:
> Nope. That runs `rsync -r ../pyhf/schemas/ $(BUILDDIR)/html/schemas` which means as long as everything is under `pyhf/schemas` -- it'll be copied over correctly -- for all versions. 

@kratsg Ah yeah. Sorry, I was looking at PR #452 and had forgotten that this wasn't implemented until it was fixed in PR #458.

PR #488, opened 2019-06-25
Update workspace specification to use 'observations' instead of 'data'

### LINKED ISSUE #474: data specification for workspace is not specific enough
# Description

Related to #106.  https://github.com/diana-hep/pyhf/blob/master/pyhf/schemas/1.0.0/workspace.json#L8 currently describes data as an object, but that could be anything. We probably want to specify it as an object with keys for example... or perhaps as @lukasheinrich stated before, an array of items.

kratsg:
# Description

This is the last thing needed for the v1.0.0 of the JSON schema -- which is to switch out how we define the observed `data` of the workspace. I was initially lazy and defined it as a regular `object` which could've been anything... now we tighten up that definition ahead of release v1.0.0 of the schema.

This resolves #474.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- workspace 'data' renamed to 'observations'
- 'observation' specification is defined requiring a 'name' (channel) and 'data' (observations for that channel)
- code and tests are updated to reflect the change in the schema
- resolves #474
- additional test added to check that workspace data is retrieved correctly, and to handle situations in when the observed data requested for a channel does not exist
```

### DIFF
--- a/docs/examples/notebooks/binderexample/StatisticalAnalysis.ipynb
+++ b/docs/examples/notebooks/binderexample/StatisticalAnalysis.ipynb
@@ -92,7 +92,8 @@
    "outputs": [],
    "source": [
     "parsed = pyhf.readxml.parse('meas.xml',os.getcwd())\n",
-    "obs_data = next(obs for obs in parsed['observations'] if obs['name'] == 'channel1')['data']"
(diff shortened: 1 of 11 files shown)


lukasheinrich:
LGTM.

PR #490, opened 2019-06-25
Fix toy example of spec in docs and render parts of schema in docs from file

### LINKED ISSUE #489: Toy example in docs doesn't pass pyhf inspect
# Description

The [toy example in the docs](https://github.com/diana-hep/pyhf/blob/4869a15b8715e6d446428489f1fa33de96579e6c/docs/likelihood.rst#toy-example) needs to be updated as it doesn't pass `pyhf inspect`. It needs to use the new (as of PR #488) "observations" instead of "data" and is missing a version.

# Expected Behavior

```
$ pyhf inspect toy_example.json
          Summary       
    ------------------  
       channels  1
        samples  2
     parameters  2
      modifiers  2

       channels  nbins
     ----------  -----
  singlechannel    2  

        samples
     ----------
     background
         signal

     parameters  constraint              modifiers
     ----------  ----------              ----------
             mu  unconstrained           normfactor
uncorr_bkguncrt  constrained_by_poisson  shapesys

    measurement           poi            parameters
     ----------        ----------        ----------
(*) Measurement            mu            (none)

```

# Actual Behavior

```
$ pyhf inspect toy_example.json
  File "/home/mcf/Code/GitHub/pyhf/pyhf/utils.py", line 53, in validate
    return validator.validate(spec)
  File "/home/mcf/venvs/pyhf-dev/lib/python3.6/site-packages/jsonschema/validators.py", line 348, in validate
    raise error
jsonschema.exceptions.ValidationError: Additional properties are not allowed ('data' was unexpected)

Failed validating 'additionalProperties' in schema:
    {'$id': 'https://diana-hep.org/pyhf/schemas/1.0.0/workspace.json',
     '$schema': 'http://json-schema.org/draft-06/schema#',
     'additionalProperties': False,
     'properties': {'channels': {'items': {'$ref': 'defs.json#/definitions/channel'},
                                 'type': 'array'},
                    'measurements': {'items': {'$ref': 'defs.json#/definitions/measurement'},
                                     'type': 'array'},
                    'observations': {'items': {'$ref': 'defs.json#/definitions/observation'},
                                     'type': 'array'},
                    'version': {'const': '1.0.0'}},
     'required': ['channels', 'measurements', 'observations', 'version'],
     'type': 'object'}

On instance:
    {'channels': [{'name': 'singlechannel',
                   'samples': [{'data': [5.0, 10.0],
                                'modifiers': [{'data': None,
                                               'name': 'mu',
                                               'type': 'normfactor'}],
                                'name': 'signal'},
                               {'data': [50.0, 60.0],
                                'modifiers': [{'data': [5.0, 12.0],
                                               'name': 'uncorr_bkguncrt',
                                               'type': 'shapesys'}],
                                'name': 'background'}]}],
     'data': {'singlechannel': [50, 60]},
     'measurements': [{'config': {'parameters': [], 'poi': 'mu'},
                       'name': 'Measurement'}]}

  File "/home/mcf/venvs/pyhf-dev/bin/pyhf", line 11, in <module>
    load_entry_point('pyhf', 'console_scripts', 'pyhf')()
  File "/home/mcf/venvs/pyhf-dev/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/mcf/venvs/pyhf-dev/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/mcf/venvs/pyhf-dev/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/mcf/venvs/pyhf-dev/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/mcf/venvs/pyhf-dev/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/mcf/Code/GitHub/pyhf/pyhf/commandline.py", line 116, in inspect
    w = Workspace(wspec)
  File "/home/mcf/Code/GitHub/pyhf/pyhf/pdf.py", line 490, in __init__
    utils.validate(self.spec, self.schema, version=self.version)
  File "/home/mcf/Code/GitHub/pyhf/pyhf/utils.py", line 55, in validate
    raise InvalidSpecification(err)
pyhf.exceptions.InvalidSpecification: Additional properties are not allowed ('data' was unexpected).
	Path: 
	Instance: {'channels': [{'name': 'singlechannel', 'samples': [{'name': 'signal', 'data': [5.0, 10.0], 'modifiers': [{'name': 'mu', 'type': 'normfactor', 'data': None}]}, {'name': 'background', 'data': [50.0, 60.0], 'modifiers': [{'name': 'uncorr_bkguncrt', 'type': 'shapesys', 'data': [5.0, 12.0]}]}]}], 'data': {'singlechannel': [50, 60]}, 'measurements': [{'name': 'Measurement', 'config': {'poi': 'mu', 'parameters': []}}]}
```

# Checklist

matthewfeickert:
# Description

Resolves #489 

Uses the updated (as of #488) spec syntax of "observations" instead of "data" and add a spec "version". Additionally remove some missed (presumably copy and paste) unnecessary pieces of code.

https://pyhf-staging.readthedocs.io/en/docs-fix-observation-examples/

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Correct the toy example of the likelihood spec in the docs so that is passes pyhf inspect
   - Use "observations" syntax and add "version"
* Add tests for toy examples
* Render toy example in the docs from toy example JSON file
* Render workspace schema in the docs from the workspace JSON file
* Remove some unnecessary pieces of code
```


lukasheinrich:
should we try to ship some toy examples as part of the MANIFEST.in ?

lukasheinrich:
this way we could also test this and catch regressions

matthewfeickert:
> should we try to ship some toy examples as part of the `MANIFEST.in`? This way we could also test this and catch regressions.

Does it need to be in the manifest? Or should we just add a `test_toy_examples.py` to the tests? Either way I think that testing is a very good idea here.

matthewfeickert:
@kratsg this branch isn't showing up in https://readthedocs.org/projects/pyhf-staging/versions/ Do you know how to force ReadTheDocs to update itself?

lukasheinrich:
one way I'm looking at this is that (similar to pyhf.simplemodels) shipping
some demo workspaces might help with onboarding, but yeah they might not
need to be in a release but could instead be in the docs. (I just know that
a *lot* of people are happily using the HistFitter demo scripts as a
baseline which they adapt.

matthewfeickert:
@lukasheinrich I need to go, so I just threw some things in here to revise as needed later. Let me know if this is along the lines of what you were thinking.

kratsg:
> @kratsg this branch isn't showing up in https://readthedocs.org/projects/pyhf-staging/versions/ Do you know how to force ReadTheDocs to update itself?

@lukasheinrich , can you give us access to the settings so we can add integrations, etc? This is going to be a problem a lot (especially for releasing). 

@matthewfeickert  - go to https://readthedocs.org/projects/pyhf-staging/builds/ -- build the latest version which forces RTD to refresh the branch cache, then the new branch should show up, edit it, mark it as active, and RTD will build it.

kratsg:
> one way I'm looking at this is that (similar to pyhf.simplemodels) shipping some demo workspaces might help with onboarding, but yeah they might not need to be in a release but could instead be in the docs. (I just know that a *lot* of people are happily using the HistFitter demo scripts as a baseline which they adapt.

Put these examples directly in the docs. If we provide them as files, etc... this encourages bad habits. We want people to be _actively_ copying, instead of passively copying.

tests/test_workspace.py
```diff
@@ -101,12 +101,6 @@ def test_get_workspace_model_default(workspace_factory):
     assert m
 
 
-def test_get_workspace_model_default(workspace_factory):
```
kratsg: why was this removed? we actually need to test that we can access the default model.

tests/test_toy_examples.py
```diff
@@ -0,0 +1,7 @@
+import shlex
```
kratsg: this could be a doctest instead -- rather than an actual `tests/`

tests/test_workspace.py
matthewfeickert: @kratsg Look at what is above it. It is the exact same test as the one above it. If you meant for them to be different tests then I think you might have accidentally copied and pasted the same one twice.

tests/test_toy_examples.py
matthewfeickert: Won't this required additional code that doctest runs to evaluate the example? I would think that we would just want to show the example JSON and not additional stuff, which is why I created the separate file and the pytest test in the PR.

tests/test_workspace.py
kratsg: oh weird. I wonder how _that_ happened...

tests/test_toy_examples.py
kratsg: Yeah, I think I understand now. We should just include the JSON file from the `examples/` we already have -- but it'll only show latest schema version I suppose.

matthewfeickert: > but it'll only show latest schema version I suppose

Yeah. I don't think this is _too_ big of a problem until/if we start releasing docs for different releases.

tests/test_examples.py
```diff
@@ -0,0 +1,9 @@
+import shlex
+
+
+def test_2bin_singlechannel(tmpdir, script_runner):
+    command = 'pyhf inspect {0:s}'.format(
+        'docs/examples/JSON/2-bin_single-channel.json'
```
matthewfeickert: I initially felt a bit unsure if it was a good idea to have a file that is going to be tested as part of the code live under `docs/` instead of `pyhf/`, but doing so keeps it out of the manifest and I then also remembered that we test the Jupyter notebooks live under `docs/` so this shouldn't be any different.

PR #501, opened 2019-07-08
[WIP] toys

lukasheinrich:
# Description

this adds the possibility to draw toys (and some basic object oriented pdf classes).. API not finalized yet

```
import pyhf.simplemodels
m = pyhf.simplemodels.hepdata_like([0,0],[50,60],[1,1])
sim = m.pdfobj(m.config.suggested_init())
sim.sample((10,))


array([[  45,   65, 2518, 3568],
       [  68,   71, 2517, 3601],
       [  57,   57, 2454, 3545],
       [  41,   55, 2503, 3558],
       [  53,   64, 2512, 3639],
       [  62,   49, 2571, 3575],
       [  45,   43, 2418, 3405],
       [  46,   54, 2524, 3600],
       [  38,   57, 2506, 3587],
       [  52,   62, 2453, 3574]])
```

screenshot


lukasheinrich:
screenshot

lukasheinrich:
works for all three but somehow TF is much slower than torch & numpy .. thoughts? @kratsg @matthewfeickert

kratsg:
I'm gonna need more comments because I'm totally not understanding most of the changes.
- logpdf -> pdf?
- what's pyhf.prob (and pyhf.prob.Simultaneous?)
- why dataview?

pyhf/pdf.py
```diff
@@ -435,33 +436,48 @@ def expected_data(self, pars, include_auxdata=True):
...
-        poisson = self.constraints_poisson.logpdf(auxdata, pars)
-        return normal + poisson
-
-    def mainlogpdf(self, maindata, pars):
+    def constraint_pdf(self, pars):
+        tensorlib, _ = get_backend()
+        gview = self.constraints_gaussian.dataview()
+        pview = self.constraints_poisson.dataview()
+
+        factors = []
```
kratsg: what's this block of code starting here for? Why wasn't it needed before?

pyhf/pdf.py
```diff
@@ -435,33 +436,48 @@ def expected_data(self, pars, include_auxdata=True):
...
+            projections.append(gview)
+        if pview is not None:
+            ppdf = self.constraints_poisson.pdf(pars)
+            factors.append(ppdf)
+            projections.append(pview)
+
+        sim = Simultaneous(factors, projections)
+        return sim
+
+    def mainpdf(self, pars):
```
kratsg: I'm realizing I don't understand what mainpdf (or mainlogpdf) was for. E.G. why only Poisson?

pyhf/prob/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+        for fac_index,fac in enumerate(self.factors):
+            projected = self.project(value,fac_index)
+            projected = fac.log_prob(projected)
+            log_sum.append(tensorlib.reshape(projected,(value.shape[0],-1)))
+        result = tensorlib.reshape(tensorlib.sum(tensorlib.concatenate(log_sum,axis=-1),axis=-1), (value.shape[0],))
+        if flat:
+            return result[0]
+        return result
+
+    def project(self,data,factor_index):
```
kratsg: ???

pyhf/prob/__init__.py
```diff
@@ -0,0 +1,46 @@
+from .. import get_backend
+import numpy as np
+
+class Simultaneous(object):
+    def __init__(self, factors, projections):
+        self.factors = factors
+        self.projections = projections
+        self.batch_shape = (sum([x.batch_shape[0] for x in self.factors]),)
+    
+    def sample(self, sample_shape = None):
```
kratsg: ???

pyhf/prob/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+        tensorlib, _ = get_backend()
+        pieces = []
+        for fac in self.factors:
+            pieces.append(fac.sample(sample_shape if sample_shape is not None else (1,)))
+        result = tensorlib.concatenate(pieces,axis = -1) 
+        return result if sample_shape is not None else result[0]
+    
+    def log_prob(self, value):
```
kratsg: I guess the logpdf is calculated here now instead? There's a lot of changes/additions that I don't understand.

pyhf/tensor/pytorch_backend.py
```diff
@@ -297,3 +297,26 @@ def normal_cdf(self, x, mu=[0.0], sigma=[1.0]):
...
+
+    class Poisson(object):
+        def __init__(self, rate):
+            self.pdf =  torch.distributions.Poisson(rate = rate)
+            self.batch_shape = self.pdf.batch_shape
+        def sample(self, *args, **kwargs):
+            return self.pdf.sample(*args, **kwargs)
+        
+        def log_prob(self, data):
+            d = self.pdf.log_prob(data)
```
matthewfeickert: In 

`d = self.pdf.log_prob(data)`

this `d` never gets used. That doesn't seem right.

pyhf/tensor/tensorflow_backend.py
```diff
@@ -414,3 +414,26 @@ def normal_cdf(self, x, mu=0, sigma=1):
...
+
+    class Poisson(object):
+        def __init__(self, rate):
+            self.pdf =  tfp.distributions.Poisson(rate = rate)
+            self.batch_shape = self.pdf.batch_shape
+        def sample(self, *args, **kwargs):
+            return self.pdf.sample(*args, **kwargs)
+        
+        def log_prob(self, data):
+            d = self.pdf.log_prob(data)
```
matthewfeickert: In 

`d = self.pdf.log_prob(data)`

this `d` never gets used. That doesn't seem right.

matthewfeickert:
> works for all three but somehow TF is much slower than torch & numpy .. thoughts?

> I'm gonna need more comments because I'm totally not understanding most of the changes.

Yeah, @lukasheinrich I think if you can address some of @kratsg's questions that would clear things up for me as well.

pyhf/tensor/pytorch_backend.py
matthewfeickert: Thanks. Though I guess it didn't really matter, as it was getting bypassed.

lukasheinrich:
so the idea is here to delay the actualy computation of the pdf value (given parameters and data) and instead bubble up the computation wrapped in pdf objects. 

the reason is that we want to throw toys using a `model.sample()` call and thus are exapnding the set of operations we want to do from merely just evaluating the density to something more, so we need it as an object (lest we pepper everything with their own `sample()` 

it's for good reason we shied away from this for a while, while we were optimizing the computation but I think now we have a good grasp on how things ought to work.

the main composition of the pdf is

```
                                                full constraint
                                     _____ pdf.prob.Simultaneous ___  / Poisson Constraint
full pdf (pdf.prob.Simultaneous) ---/                                 \ Gaussian Constraint
                                    \___ poisson pdf


```

PR #506, opened 2019-07-11
Use math.pow to improve exponentiation in interpolators

matthewfeickert:
# Description

Given the discussion in the blog post "[Performance of Various Python Exponentiation Methods](https://chrissardegna.com/blog/posts/python-expontentiation-performance/)" and [this answer](https://stackoverflow.com/a/20970087/8931942) to the Stack Overflow question "Exponentials in python `x.**y` vs `math.pow(x, y)`" it is seen that anytime you're dealing with floats for either `x` and `y` it is more efficient to use `math.pow` instead of the builtin `**` operator.

Given this, [interpolator code 4](https://github.com/diana-hep/pyhf/blob/a019ba68f61bb56595aa4aeb98538c8f0c17ffcb/pyhf/interpolators/code4.py) is updated to use `math.pow`. The effect will be very small, as this happens in `init`, so this is more of a following best practices change. The slow implementation of the code for validation is also updated (as it is in code 1 as well) to have consistent style across the files.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use math.pow instead of the ** operator for exponentiation in the interpolator codes
```


pyhf/interpolators/code1.py
```diff
@@ -98,9 +99,9 @@ def product(self, down, nom, up, alpha):
         delta_up = up / nom
         delta_down = down / nom
         if alpha > 0:
-            delta = delta_up ** alpha
+            delta = math.pow(delta_up, alpha)
         else:
-            delta = delta_down ** (-alpha)
+            delta = math.pow(delta_down, (-alpha))
```
lukasheinrich: this is not code that's actually used in pyhf proper, but just reference code we check our production code against. I'm fine with changing it, but it'll not affect e.g. fit performance

lukasheinrich: (same is true also for the other _slow_x commands

matthewfeickert: I know. My reasoning above is:
> The slow implementation of the code for validation is also updated (as it is in code 1 as well) to have consistent style across the files.

lukasheinrich: hm, ok need to re-read (sorry) I didn't see anything in the PR that affected the non-slow impls (because I would have thought those all go through `tensorlib` anyways)

matthewfeickert: This commit is the one that changes the non-slow portion of code 4: https://github.com/diana-hep/pyhf/pull/506/commits/8844dbe869c4415a46f9c08c15c451d752790741

lukasheinrich: ah I see, thanks. I think I missed this because it's in the `__init__`, i.e. a one-time computation.

matthewfeickert: Yeah, this is a _very_ minor improvement PR.

matthewfeickert:
As the entire second commit is style based, let me know if you agree or if you want me to drop it.

PR #515, opened 2019-07-18
Apply recommended changes for CodeFactor issues

matthewfeickert:
# Description

This is simply a cleaning PR that goes through and addresses most of the "issues" that CodeFactor finds (though it leaves all the cylic import issues because those [can't be dealt with as long as we still have Python 2](https://github.com/diana-hep/pyhf/issues/469#issuecomment-511637392)).

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Apply recommended changes from CodeFactor's pylint linter
   - Removal of unnecessary pass
   - Removal of unnecessary return (Python returns None if there is no return statement)
   - Removal of len to check if list or dict is empty (empty arrays and empty dict evaluate to False in Python)
   - Use 'is' for singleton comparison (c.f. https://www.python.org/dev/peps/pep-0008/#programming-recommendations)
```

### DIFF
--- a/src/pyhf/exceptions/__init__.py
+++ b/src/pyhf/exceptions/__init__.py
@@ -41,8 +41,6 @@ class InvalidModel(Exception):
     This can occur, for example, when the provided parameter of interest to fit against does not get declared in the specification provided.
     """
 
-    pass
-
 
 class InvalidModifier(Exception):
     """
@@ -53,23 +51,17 @@ class InvalidModifier(Exception):
 
     """
 
-    pass
-
 
 class InvalidInterpCode(Exception):
     """
     InvalidInterpCode is raised when an invalid/unimplemented interpolation code is requested.
     """
 
-    pass
-
 
 class ImportBackendError(Exception):
     """
     MissingLibraries is raised when something is imported by sustained an import error due to missing additional, non-default libraries.
     """
-
-    pass
 
 
 class InvalidOptimizer(Exception):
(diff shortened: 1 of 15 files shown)


matthewfeickert:
As these are essentially all just style changes that were suggested by a linter I am more then happy to drop any of them.

matthewfeickert:
Additionally, if you feel there should be more explanation in the commit message I'm happy to extend it.

lukasheinrich:
this looks great. I went over this but I hope there are also some guarantees in CF that make sure the suggestions don't change functionality. But looking it over it seems find. Thanks!

matthewfeickert:
>...I hope there are also some guarantees in CF that make sure the suggestions don't change functionality.

I don't think there is any guarantee, but for the sake of completeness I'll break down the major changes that are implemented here.

- Removal of unnecessary `pass`
- Removal of unnecessary `return` (Python returns `None` if there is no `return` statement)
- Removal of `len` to check if `list` or `dict` is empty (empty arrays and empty dict evaluate to False in Python)
```
>>> bool([])
False
>>> bool({})
False
>>> not []
True
>>> not {}
True
```
- Use `is` for singleton comparison (c.f. https://www.python.org/dev/peps/pep-0008/#programming-recommendations)

Given this, we shouldn't see any change in evaluation behavior. As the CI is passing I believe all changes to be safe enough.

PR #519, opened 2019-07-23
Add ParamViewer class 

lukasheinrich:
# Description

adds a `ParamViewer` class to be used by the combined modifiers. Allows to splice out the correct subset of parameter values in batched and unbatched cases (see tests) based on the par_map

The philosophy of this PR is as followsL

It does three things

* introduce paramviewer (see below)
* adapt combined modifiers to use paramviewer
* adapt constraints to use paramviewer

for those things some minor changes are done elsewhere

* add tile method to tensor backends (PR #538)
* adapt pdf.py to constrain API homogenization (scalar return values)
* adapt paramset.py to work with batched constraints

Finally to make sure we do things correctly I added
* a lot of new tests

---------

In order to be fast these constraint codes used to pre-compute indices into the parameter vector in order to quickly grab all the parameter values needed to compute the constraint term

e.g.

```
par vector [ . . . . . x x x x . . . x . . . x x x x . . . . . x x ]
```

here the `x` denote parameters which are constrained by a normal (or poisson), the param_set `slice` attribute denoted exactly where these are in the order defined in the pdf config.

An index vector, one could pass to `tensorlib.gather` in this case would be

[ 6 7 8 9 13 17 18 19 20 26 27 ]

to get 

```
[ x x x x x x x x x x x ]
```

When we moving to batch evaluation we suddenly do not have a single `(N,)` param vector but a parameter "field"

```
par field
[ 
  [ . . . . . x x x x . . . x . . . x x x x . . . . . x x ],
  [ . . . . . y y y y . . . y . . . y y y y . . . . . y y ],
  [ . . . . . z z z z . . . z . . . z z z z . . . . . z z ],
]
```

so it's a bit harder to grab all these correctly. This is eventually done by the `acess_field` in the constraints, but the indices are computed (as before based on the param_set.slice attribute) using the `ParamViewer` helper class

----

(added by @kratsg with help from @lukasheinrich )

A real world situation of what the batching is trying to achieve is to imagine a simple model. Let's say you have a single channel, two sample (signal + background) model with a constrained modifier, an uncorrelated bkg uncertainty (`bkg_shapesysN`), and an normalisation factor on the signal (parameter of interest, `mu_sig`). Imagine the channel has 4 bins, so we have:

- one channel (4 bins)
- two samples
- 2 modifiers
- 2 parameter sets (`parsets`)
- 5 parameters (`mu_sig, bkg_shapesys1, bkg_shapesys2, bkg_shapesys3, bkg_shapesys4`)

The data space is 4 bin counts (plus 4 aux data for the `bkg_shapesysN`), so when evaluating

```
# (8,) + (5,) -> (1,)
model.logpdf(pars, data)
```

The `data` is `(8,)` tensor and `pars`  is a `(5, )` tensor. This PR allows us to *batch in parameter sets* which means evaluating multiple instances of parameter sets:

```
# (8,) + (B, 5) -> (B, 1)
model.logpdf(pars, data)
```

What this PR does is to enable moving to more parameter sets (`B = batch_size`) but we do not enable batching just yet. The groundwork is laid in place for a future PR.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add batched computation of constraints and modifiers 
* Add paramviewer class to handle selection of parameter subsets in batched regime
* Add tests for paramviewer, batched combined modifiers, and batched combined constraints
```


pytest.ini
```diff
@@ -1,2 +1,2 @@
 [pytest]
-addopts = --ignore=setup.py --ignore=validation/ --ignore=binder/ --cov=pyhf --cov-report=term-missing --cov-config=.coveragerc --cov-report html --cov-report html --doctest-modules --doctest-glob='*.rst'
```
kratsg: this needs to be uncommented

src/pyhf/constraints.py
```diff
@@ -1,25 +1,41 @@
...
             (pdfconfig.param_set(cname), pdfconfig.par_slice(cname))
             for cname in pdfconfig.auxdata_order
         ]
-        self._precompute()
-        events.subscribe('tensorlib_changed')(self._precompute)
 
-    def _precompute(self):
-        tensorlib, _ = get_backend()
+        pnames = [
+            cname
```
kratsg: what is cname here compared to pname? We should label these better...

src/pyhf/constraints.py
```diff
@@ -1,25 +1,41 @@
 from . import get_backend, default_backend
 from . import events
+from .paramview import ParamViewer
 
 
 class gaussian_constraint_combined(object):
-    def __init__(self, pdfconfig):
+    def __init__(self, pdfconfig, batch_size=None):
```
kratsg: set default batch size to 1

src/pyhf/constraints.py
```diff
@@ -1,25 +1,41 @@
...
 
-    def _precompute(self):
-        tensorlib, _ = get_backend()
+        pnames = [
+            cname
+            for cname in pdfconfig.auxdata_order
+            if pdfconfig.param_set(cname).pdf_type == 'normal'
+        ]
+
+        parfield_shape = (self.batch_size or 1, len(pdfconfig.suggested_init()))
```
kratsg: remove the `or 1` as default size should just be 1

src/pyhf/constraints.py
kratsg: we will need a check for batch size = 0 (or negative) perhaps.

src/pyhf/paramview.py
```diff
@@ -0,0 +1,69 @@
+from . import get_backend, default_backend, events
```
kratsg: `pnames` is not used for anything but inside `ParamViewer`. Given that this is the case, we should move the logic of building `pnames` into `ParamViewer`... in fact, thinking about this, ParamViewer relies on the `pdfconfig` and the `batch_size` only. One could imagine actually moving some of this out of the constraints entirely and just building a `ParamViewerFactory` that only takes in the `pdfconfig` and then you can configure it with the batch size after the fact, since it should be pretty deterministic.

src/pyhf/constraints.py
```diff
@@ -1,25 +1,41 @@
...
+        self.parameter_helper = ParamViewer(parfield_shape, pdfconfig.par_map, pnames)
+
+        # self.parameter_helper.index_selection  is [ (batch, parslice) ] ]
```
kratsg: what's going on in this block of code?

src/pyhf/constraints.py
```diff
@@ -39,57 +55,70 @@ def _precompute(self):
...
-            )
-            normal_data = default_backend.concatenate(
-                list(
-                    map(
-                        lambda x: default_backend.astensor(x, dtype='int'),
-                        normal_constraint_data,
-                    )
-                )
+
+        if self.parameter_helper.index_selection:
```
kratsg: when is this false-y?

src/pyhf/constraints.py
```diff
@@ -39,57 +55,70 @@ def _precompute(self):
...
         else:
-            self.normal_data, self.normal_sigmas, self.normal_mean_idc = (
-                None,
-                None,
-                None,
-            )
+            self._normal_data, self._normal_sigmas = (None, None)
+
+        sigmas = default_backend.reshape(self._normal_sigmas, (1, -1))  # (1, normals)
+        self._batched_sigmas = default_backend.tile(sigmas, (self.batch_size or 1, 1))
```
kratsg: remove the `or 1`

src/pyhf/constraints.py
```diff
@@ -39,57 +55,70 @@ def _precompute(self):
...
-            self.normal_mean_idc = tensorlib.astensor(
-                default_backend.tolist(normal_mean_idc), dtype='int'
-            )
         else:
-            self.normal_data, self.normal_sigmas, self.normal_mean_idc = (
-                None,
-                None,
-                None,
-            )
+            self._normal_data, self._normal_sigmas = (None, None)
```
kratsg: self.normal_mean_idc is gone?

src/pyhf/constraints.py
kratsg: why tile?

src/pyhf/constraints.py
```diff
@@ -39,57 +55,70 @@ def _precompute(self):
...
+        self.normal_data = tensorlib.astensor(self._normal_data, dtype='int')
+        self.normal_sigmas = tensorlib.astensor(self._normal_sigmas)
+        self.access_field = tensorlib.astensor(self._access_field, dtype='int')
 
     def logpdf(self, auxdata, pars):
-        if self.normal_data is None:
-            return 0
         tensorlib, _ = get_backend()
+        if not self.parameter_helper.index_selection:
+            print('return scalar!', self.batch_size, tensorlib.astensor(0.0))
```
kratsg: print statement

src/pyhf/constraints.py
```diff
@@ -129,58 +160,79 @@ def _precompute(self):
...
-                        lambda x: default_backend.astensor(x, dtype='int'),
-                        poisson_constraint_data,
-                    )
-                )
+            poisson_data = default_backend.astensor(
+                default_backend.concatenate(poisson_constraint_data), dtype='int'
             )
 
-            self.poisson_data = tensorlib.astensor(
+            self._poisson_data = default_backend.astensor(
```
kratsg: not sure what this line is doing as poisson_data is already in the tensor of default_backend from previous line

src/pyhf/constraints.py
```diff
@@ -129,58 +160,79 @@ def _precompute(self):
...
                 default_backend.tolist(poisson_data), dtype='int'
             )
-            self.poisson_rate_idc = tensorlib.astensor(
-                default_backend.tolist(poisson_rate_idc), dtype='int'
-            )
-            self.poisson_rate_fac = tensorlib.astensor(
+            self.poisson_rate_fac = default_backend.astensor(
```
kratsg: not sure what this line is doing as poisson_rate_fac is already in the tensor of default_backend from previous line

src/pyhf/tensor/numpy_backend.py
```diff
@@ -34,6 +34,9 @@ def clip(self, tensor_in, min_value, max_value):
         """
         return np.clip(tensor_in, min_value, max_value)
 
+    def tile(self, tensor_in, repeats):
```
kratsg: missing docs

src/pyhf/tensor/pytorch_backend.py
```diff
@@ -42,6 +42,9 @@ def tolist(self, tensor_in):
                 return tensor_in
             raise
 
+    def tile(self, tensor_in, repeats):
```
kratsg: missing docs

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -44,6 +44,9 @@ def clip(self, tensor_in, min_value, max_value):
             max_value = tf.reduce_max(tensor_in)
         return tf.clip_by_value(tensor_in, min_value, max_value)
 
+    def tile(self, tensor_in, repeats):
```
kratsg: missing docs

tests/test_init.py
```diff
@@ -19,14 +19,19 @@
             "tensorflow_backend",
             pytest.raises(pyhf.exceptions.ImportBackendError),
         ],
-        [
-            "mxnet",
-            "mxnet_backend",
-            "mxnet_backend",
-            pytest.raises(pyhf.exceptions.ImportBackendError),
-        ],
+        # [
```
kratsg: hrmm, commented out mxnet for now?

src/pyhf/constraints.py
lukasheinrich: cname is the name of a constrained parameter, pnames is the set of parameters constrained by 'normal'

src/pyhf/constraints.py
lukasheinrich: The idea is that as a pdf, there are 2 possibilities

* logpdf() returns a scalar ( shape `()`)  (batch_size = None)
* logpdf() returns a tensor ( shape `(batch_size,)` )

this is in preparation for moving to e.g. tensorflow_probability and pytorch.distributions. So this

src/pyhf/constraints.py
lukasheinrich: if we accept self.batch_size = None, this is necesary

src/pyhf/constraints.py
lukasheinrich: I removed this

src/pyhf/constraints.py
lukasheinrich: this sets up the `view` on the parameters needed by the constaint (when computing the constraint you do not need to access all parameters but just a subset of them, `ParamViewer` is a helper class that figures out the right indices to use

src/pyhf/tensor/numpy_backend.py
lukasheinrich: added

src/pyhf/constraints.py
lukasheinrich: when no constraints of this type are needed, and thus `pnames` is empty

src/pyhf/constraints.py
lukasheinrich: this is needed when self.batch_size = None is accepted

src/pyhf/constraints.py
lukasheinrich: yes this is handled by self.access_field

src/pyhf/constraints.py
lukasheinrich: you need to repeat the sigmas for each evaluation in the batch

lukasheinrich:
Hi @kratsg,

I addressed some of the easy comments. The philosophy of this PR is as follows

In order to be fast these constraint codes used to pre-compute indices into the parameter vector in order to quickly grab all the parameter values needed to compute the constraint term

e.g.

```
par vector [ . . . . . x x x x . . . x . . . x x x x . . . . . x x ]
```

here the `x` denote parameters which are constrained by a normal (or poisson), the param_set `slice` attribute denoted exactly where these are in the order defined in the pdf config.

An index field in this case would be

[ 6 7 8 9 13 17 18 19 20 26 27 ]

When we moving to batch evaluation we suddenly do not have a single `(N,)` param vector but a parameter "field"

```
par vector
[ 
  [ . . . . . x x x x . . . x . . . x x x x . . . . . x x ],
  [ . . . . . y y y y . . . y . . . y y y y . . . . . y y ],
  [ . . . . . z z z z . . . z . . . z z z z . . . . . z z ],
```

so it's a bit harder to grab all these correctly. This is eventually done by the `acess_field` in the constraints, but the indices are computed (as before based on the param_set.slice attribute) using the `ParamViewer` helper class

src/pyhf/constraints.py
kratsg: ok.

src/pyhf/constraints.py
kratsg: ok

kratsg:
> I'm not opposed to a common combined modifier class (if anything this will give a useful hook for future community-provided modifiers. We can point to it and: "implement this interface")

I agree.



> But I don't think ParamViewer belong in there. PV is used in a few places (constraints, soon pdf.py, modifiers).. It's really a thing that just deals with the parameters of the pdf. If anything this should be fused with param_sets.py in the future.

Yeah, on second thought, it should definitely be the interface to the parameters of a model. That is, a model builds/initializes a ParamViewer which then loads in the param sets, and given all the different modifiers built-in, can determine indices which can be grabbed by anything needing access to the parameter sets.

> I agree the tests become unwieldy, but i'm not sure how to resolve this, you need to test a ton of functionality. But I agree it is hard maintain, maybe we can refactor them better. At the same time I like explicitly coded out tests that don't rely on some other code, since you explicitly say what setup you are testing, otherwise the tests just become more logic to maintain

Yeah, only issue is that when a test is long and has more text, it's harder for new users to contribute (and we start getting into copy/pasting code...)

kratsg:
> e.g.
> 
> ```
> par vector [ . . . . . x x x x . . . x . . . x x x x . . . . . x x ]
> ```
> 
> here the `x` denote parameters which are constrained by a normal (or poisson), the param_set `slice` attribute denoted exactly where these are in the order defined in the pdf config.
> 
> An index vector, one could pass to `tensorlib.gather` in this case would be
> 
> [ 6 7 8 9 13 17 18 19 20 26 27 ]
> 
> to get
> 
> ```
> [ x x x x x x x x x x x ]
> ```

Can we add a test that ParamViewer handles such a simple use-case -- but this would also help clarify what it actually does. (Similar for the batched version).

lukasheinrich:
@kratsg I added tests to show this simple use case. PTAL

tests/test_tensor.py
```diff
@@ -155,6 +162,17 @@ def test_boolean_mask(backend):
     ) == [1, 2, 4]
 
 
+@pytest.mark.skip_mxnet
```
matthewfeickert: Would you have any objections to adding `tile` to the MxNet backend (which should be simple given [`mxnet.nd.tile`](https://mxnet.incubator.apache.org/api/python/ndarray/ndarray.html#mxnet.ndarray.NDArray.tile)) so that this test can have the `@pytest.mark.skip_mxnet` removed?

src/pyhf/tensor/numpy_backend.py
```diff
@@ -34,6 +34,27 @@ def clip(self, tensor_in, min_value, max_value):
...
+        """
+        Repeat tensor data along a specific dimension
+
+        Example:
```
matthewfeickert: To make `help()` and the docs as backend agnostic as possible can you also duplicate this example across the backends that you add `tile` to?

matthewfeickert:
@lukasheinrich Can you please rebase this against `master`? You're picking up the bug that PR #526 fixed which is keeping the tests from running.

matthewfeickert:
As this PR has the addition of `tile` to the tensorlib API can we split it out to be its own PR? I can also add it to PR #535 if you don't mind additional things getting batched in there.

src/pyhf/tensor/numpy_backend.py
matthewfeickert: Took care of this as I had some free cycles tonight.

lukasheinrich:
@matthewfeickert I prepped #538 to break out the `tile` method

lukasheinrich:
@matthewfeickert @kratsg rebased

src/pyhf/constraints.py
```diff
@@ -125,62 +151,82 @@ def _precompute(self):
             try:
                 poisson_constraint_rate_factors.append(parset.factors)
             except AttributeError:
-                poisson_constraint_rate_factors.append(
-                    default_backend.shape(self.par_indices[parslice])
-                )
+                # this seems to be dead code
+                # TODO: add coverage
```
matthewfeickert: @lukasheinrich can you comment more on the state of this? Or is this done and need to get removed?

lukasheinrich: The refactor to paramviewer does not touch this code so I chose not to fix it, but I noticed that it's dead code which we should review how that ended up in there and possibly remove it in a separate PR

matthewfeickert: Is there an issue for this? If not, can you make one and note the issue number in the comment?

tests/test_constraints.py
```diff
@@ -99,3 +127,208 @@ def fast(self, auxdata, pars):
     slow_result = pyhf.tensorlib.tolist(slow(m, auxd, pars))
     fast_result = pyhf.tensorlib.tolist(fast(m, auxd, pars))
     assert pytest.approx(slow_result) == fast_result
+
+
+def test_ok(backend):
```
matthewfeickert: Sorry, missed this earlier, but can we get a more descriptive test title? It is a few hundred lines, so having a more descriptive name will help people reading it in the future.

lukasheinrich: good catch. done!

tests/test_constraints.py
```diff
@@ -99,3 +127,208 @@ def fast(self, auxdata, pars):
...
+                    n_parameters=3,
+                    inits=[0] * 3,
+                    bounds=[[0, 10]] * 3,
+                    auxdata=[0, 0, 0],
+                ),
+                'slice': slice(5, 8),
+            },
+        },
+    )
+    sugg = [1.0] * 3 + [0.0] * 5  # 2 pois 5 norm
```
matthewfeickert: If I can make a suggestion, it might slightly simplify the tests here to create the following function

```python

    def _batched_constraint(config, init, constraint_callable, batch_size=None):
        constraint = constraint_callable(config, batch_size)
        return constraint.logpdf(config.auxdata, init)
```
which then allows this first test to be

```python
    suggested_init = [1.0] * 3 + [0.0] * 5  # 2 pois 5 norm
    result = default_backend.astensor(
        pyhf.tensorlib.tolist(
            _batched_constraint(c, suggested_init, poisson_constraint_combined)
        )
    )
```

and saves some common code being repeated, such that you can also later do

```python
    sugg = [1.1] * 3 + [0.0] * 5  # 2 pois 5 norm
    result = _batched_constraint(
        c, [sugg] * 10, poisson_constraint_combined, batch_size=10
    )
    assert result.shape == (10,)
```

If you like the idea I have a commit ready to go that implements this across the test.

tests/test_constraints.py
matthewfeickert: thanks!

tests/test_constraints.py
matthewfeickert: Sorry, read the PR better and now see those should have been `pars` and not `init`. Changed that in the example here.

lukasheinrich: generally I'm leaning towards having more verbose but explicit code in tests instead of logic, but I'm fine either way. If you think it's more readable and have a commit ready, go for it

src/pyhf/constraints.py
kratsg: add a comment or rename the variables to be clearer. `pname_constrained` and `pnames_normal` sound better to me -- than `pnames` and `cname`.

tests/test_constraints.py
matthewfeickert: If this is your preferred style for tests I'm okay with that. Though if you don't mind I will change this commit to be more explicit in the variable naming, as I found that I had misread what was actually happening the first time through.

matthewfeickert: > > I agree the tests become unwieldy, but i'm not sure how to resolve this, you need to test a ton of functionality. But I agree it is hard maintain, maybe we can refactor them better. At the same time I like explicitly coded out tests that don't rely on some other code, since you explicitly say what setup you are testing, otherwise the tests just become more logic to maintain
> 
> Yeah, only issue is that when a test is long and has more text, it's harder for new users to contribute (and we start getting into copy/pasting code...)

Ah, I see now that you and @kratsg already had some discussion on this. Giordon maybe you want to give some additional comments here as well before I push anything.

src/pyhf/constraints.py
```diff
@@ -1,33 +1,39 @@
...
-        tensorlib, _ = get_backend()
+        parfield_shape = (self.batch_size or 1, len(pdfconfig.suggested_init()))
+        self.parameter_helper = ParamViewer(parfield_shape, pdfconfig.par_map, pnames)
+
         start_index = 0
         normal_constraint_data = []
-        normal_constraint_mean_indices = []
         normal_constraint_sigmas = []
-        for parset, parslice in self.parset_and_slice:
+        for parset in self.parset_and_slice:
```
kratsg: `self.parset_and_slice` should just become `self.parsets`

src/pyhf/constraints.py
```diff
@@ -1,33 +1,39 @@
...
             end_index = start_index + parset.n_parameters
             thisauxdata = self.data_indices[start_index:end_index]
             start_index = end_index
             if not parset.pdf_type == 'normal':
                 continue
 
+            normal_constraint_data.append(thisauxdata)
```
kratsg: how are we making sure the ordering of `normal_constraint_data` is consistent?

src/pyhf/constraints.py
kratsg: please add a comment on that, as this is definitely not clear in the code

src/pyhf/constraints.py
```diff
@@ -38,85 +44,105 @@ def _precompute(self):
...
+                sigmas, (self.batch_size or 1, 1)
             )
-            normal_data = default_backend.concatenate(
-                list(
-                    map(
-                        lambda x: default_backend.astensor(x, dtype='int'),
-                        normal_constraint_data,
-                    )
+
+            # self.parameter_helper.index_selection  is [ (batch, parslice) ] ]
```
kratsg: i don't understand this comment

src/pyhf/constraints.py
```diff
@@ -38,85 +44,105 @@ def _precompute(self):
...
+            access_field = None
```
kratsg: this code should be refactored a bit more. This feels like something that could be done with a simply concatenate call against `axis=0` or something instead of an actual loop.

src/pyhf/constraints.py
```diff
@@ -38,85 +44,105 @@ def _precompute(self):
...
-            self.normal_data = tensorlib.astensor(
-                default_backend.tolist(normal_data), dtype='int'
-            )
-            self.normal_sigmas = tensorlib.astensor(
-                default_backend.tolist(normal_sigmas)
-            )
-            self.normal_mean_idc = tensorlib.astensor(
-                default_backend.tolist(normal_mean_idc), dtype='int'
-            )
         else:
```
kratsg: remove the else, set `self._normal_data, ... = None` before the `if` statement.

src/pyhf/constraints.py
```diff
@@ -38,85 +44,105 @@ def _precompute(self):
...
-            self.normal_data, self.normal_sigmas, self.normal_mean_idc = (
-                None,
-                None,
-                None,
-            )
+            self._normal_data = None
+            self._batched_sigmas = None
+            self._access_field = None
```
kratsg: set `self.batch_size = 0` here (instead of `None`) to remove the extra `if` check in `precompute`?

src/pyhf/constraints.py
```diff
@@ -38,85 +44,105 @@ def _precompute(self):
...
+        self.access_field = tensorlib.astensor(self._access_field, dtype='int')
 
     def logpdf(self, auxdata, pars):
-        if self.normal_data is None:
-            return 0
         tensorlib, _ = get_backend()
+        if not self.parameter_helper.index_selection:
+            return (
+                tensorlib.zeros(self.batch_size)
+                if self.batch_size is not None
```
kratsg: if `self. batch_size = 0`, then we don't need to have an `if` check in this case?

src/pyhf/constraints.py
```diff
@@ -38,85 +44,105 @@ def _precompute(self):
...
+                else tensorlib.astensor(0.0)[0]
+            )
+
+        pars = tensorlib.astensor(pars)
+        if self.batch_size == 1 or self.batch_size is None:
```
kratsg: how can `self.batch_size = None` at this point? Shouldn't it either be 1 or > 1?

src/pyhf/constraints.py
```diff
@@ -38,85 +44,105 @@ def _precompute(self):
...
+            batched_pars = tensorlib.reshape(
+                pars, (self.batch_size or 1,) + tensorlib.shape(pars)
```
kratsg: I don't understand this piece of code...

```
>>> import numpy as np
>>> pars = np.random.random(100)
>>> batch_size = 10
>>> np.reshape(pars, (batch_size or 1,) + np.shape(pars))
ValueError: cannot reshape array of size 100 into shape (10,100)
```

Shouldn't it be

```
tensorlib.reshape(pars, (self.batch_size or 1, -1))
```

src/pyhf/constraints.py
```diff
@@ -38,85 +44,105 @@ def _precompute(self):
...
+            )
+        else:
+            batched_pars = pars
+
+        flat_pars = tensorlib.reshape(batched_pars, (-1,))
+        normal_means = tensorlib.gather(flat_pars, self.access_field)
```
kratsg: why `means`? This variable just seems to be `normal_pars`.

src/pyhf/constraints.py
```diff
@@ -125,62 +151,82 @@ def _precompute(self):
             try:
```
kratsg: remove the dead code

src/pyhf/constraints.py
```diff
@@ -38,85 +44,105 @@ def _precompute(self):
...
-        tensorlib, _ = get_backend()
+        parfield_shape = (self.batch_size or 1, len(pdfconfig.suggested_init()))
+        self.parameter_helper = ParamViewer(parfield_shape, pdfconfig.par_map, pnames)
 
         start_index = 0
         poisson_constraint_data = []
-        poisson_constraint_rate_indices = []
         poisson_constraint_rate_factors = []
-        for parset, parslice in self.mod_and_slice:
+        for parset in self.parset_and_slice:
```
kratsg: change to `self.parsets`

src/pyhf/constraints.py
```diff
@@ -38,85 +44,105 @@ def _precompute(self):
...
             for cname in pdfconfig.auxdata_order
+            if pdfconfig.param_set(cname).pdf_type == 'poisson'
         ]
-        self._precompute()
-        events.subscribe('tensorlib_changed')(self._precompute)
 
-    def _precompute(self):
-        tensorlib, _ = get_backend()
+        parfield_shape = (self.batch_size or 1, len(pdfconfig.suggested_init()))
+        self.parameter_helper = ParamViewer(parfield_shape, pdfconfig.par_map, pnames)
```
kratsg: call it `self.paramviewer` instead of yet another name.

src/pyhf/constraints.py
```diff
@@ -125,62 +151,82 @@ def _precompute(self):
...
+                dtype='float',
             )
-            self.poisson_rate_idc = tensorlib.astensor(
-                default_backend.tolist(poisson_rate_idc), dtype='int'
-            )
-            self.poisson_rate_fac = tensorlib.astensor(
-                default_backend.tolist(poisson_rate_fac), dtype='float'
+            factors = default_backend.reshape(
+                _poisson_rate_fac, (1, -1)
+            )  # (1, normals)
```
kratsg: i don't understand the comment

src/pyhf/constraints.py
```diff
@@ -125,62 +151,82 @@ def _precompute(self):
...
+            self._batched_factors = default_backend.tile(
+                factors, (self.batch_size or 1, 1)
```
kratsg: Why is there a `1` in `(batch_size, 1)` and not `-1` or otherwise?

src/pyhf/constraints.py
```diff
@@ -125,62 +151,82 @@ def _precompute(self):
...
             )
+
+            # self.parameter_helper.index_selection  is [ (batch, parslice) ] ]
+            access_field = None
```
kratsg: same comment as previous

src/pyhf/constraints.py
```diff
@@ -125,62 +151,82 @@ def _precompute(self):
...
-        if self.poisson_data is None:
-            return 0
         tensorlib, _ = get_backend()
-        poisson_data = tensorlib.gather(auxdata, self.poisson_data)
-        poisson_rate_base = tensorlib.gather(pars, self.poisson_rate_idc)
-        poisson_factors = self.poisson_rate_fac
+        if not self.parameter_helper.index_selection:
+            return (
+                tensorlib.zeros(self.batch_size)
+                if self.batch_size is not None
```
kratsg: same comment as previous

src/pyhf/constraints.py
```diff
@@ -125,62 +151,82 @@ def _precompute(self):
...
+                else tensorlib.astensor(0.0)[0]
+            )
+        tensorlib, _ = get_backend()
 
-        poisson_rate = tensorlib.product(
-            tensorlib.stack([poisson_rate_base, poisson_factors]), axis=0
+        pars = tensorlib.astensor(pars)
+        if self.batch_size == 1 or self.batch_size is None:
+            batched_pars = tensorlib.reshape(
+                pars, (self.batch_size or 1,) + tensorlib.shape(pars)
```
kratsg: same comment as previous

src/pyhf/constraints.py
```diff
@@ -125,62 +151,82 @@ def _precompute(self):
...
+            )
+        else:
+            batched_pars = pars
+        flat_pars = tensorlib.reshape(batched_pars, (-1,))
+        nuispars = tensorlib.gather(flat_pars, self.access_field)
+
+        pois_rates = tensorlib.product(
```
kratsg: need a comment here for why the product/stack

src/pyhf/constraints.py
```diff
@@ -125,62 +151,82 @@ def _precompute(self):
...
+            tensorlib.stack([nuispars, self.batched_factors]), axis=0
         )
-        poisson = tensorlib.poisson_logpdf(poisson_data, poisson_rate)
-        return tensorlib.sum(poisson)
+
+        # pdf pars are done, now get data and compute
```
kratsg: describe the mathematical computation happening here in a comment briefly if possible

src/pyhf/modifiers/shapefactor.py
```diff
@@ -62,64 +63,71 @@ def __init__(self, shapefactor_mods, pdfconfig, mega_mods):
...
+        self.parameters_helper = ParamViewer(parfield_shape, pdfconfig.par_map, pnames)
+
         self._shapefactor_mask = [
             [[mega_mods[s][m]['data']['mask']] for s in pdfconfig.samples] for m in keys
         ]
+
         global_concatenated_bin_indices = [
-            j for c in pdfconfig.channels for j in range(pdfconfig.channel_nbins[c])
-        ]
-        # compute the max so that we can pad with 0s for the right shape
```
kratsg: these comments got deleted (and they were helpful)... the new change to this code is pretty opaque and I'm having a hard time actually understanding what's going on

src/pyhf/modifiers/shapefactor.py
```diff
@@ -62,64 +63,71 @@ def __init__(self, shapefactor_mods, pdfconfig, mega_mods):
...
+        if self.batch_size is None:
+            batched_pars = tensorlib.reshape(pars, (1,) + tensorlib.shape(pars))
+        else:
+            batched_pars = pars
+
+        flat_pars = tensorlib.reshape(batched_pars, (-1,))
+        shapefactors = tensorlib.gather(flat_pars, self.access_field)
         results_shapefactor = tensorlib.einsum(
-            's,a,mb->msab', self.sample_ones, self.alpha_ones, shapefactors
+            'yab,s->ysab', shapefactors, self.sample_ones
```
kratsg: `y`?

src/pyhf/modifiers/shapesys.py
```diff
@@ -58,46 +64,39 @@ def __init__(self, shapesys_mods, pdfconfig, mega_mods):
...
         self._precompute()
         events.subscribe('tensorlib_changed')(self._precompute)
 
     def _precompute(self):
         tensorlib, _ = get_backend()
+        if not self.parameters_helper.index_selection:
+            return
         self.shapesys_mask = tensorlib.astensor(self._shapesys_mask)
+        self.shapesys_mask = tensorlib.tile(
+            self.shapesys_mask, (1, 1, self.batch_size or 1, 1)
```
kratsg: why is the `batch_size` in that dimension?

src/pyhf/modifiers/staterror.py
```diff
@@ -50,47 +51,39 @@ def __init__(self, staterr_mods, pdfconfig, mega_mods):
...
 
-        if self._staterror_indices:
-            access_rows = []
-            staterror_mask = default_backend.astensor(self._staterror_mask)
-            for mask, inds in zip(staterror_mask, self._staterror_indices):
-                summed_mask = default_backend.sum(mask[:, 0, :], axis=0)
-                assert default_backend.shape(
-                    summed_mask[summed_mask > 0]
-                ) == default_backend.shape(default_backend.astensor(inds))
-                # make masks of > 0 and == 0
```
kratsg: how did we get rid of the masks? that's amazing...

src/pyhf/modifiers/staterror.py
```diff
@@ -123,19 +116,18 @@ def finalize(self, pdfconfig):
...
 
+        tensorlib, _ = get_backend()
+        if self.batch_size is None:
+            batched_pars = tensorlib.reshape(pars, (1,) + tensorlib.shape(pars))
+        else:
+            batched_pars = pars
+
+        flat_pars = tensorlib.reshape(batched_pars, (-1,))
+        statfactors = tensorlib.gather(flat_pars, self.access_field)
+        results_staterr = tensorlib.einsum('yab,s->ysab', statfactors, self.sample_ones)
```
kratsg: `y`?

src/pyhf/paramsets.py
```diff
@@ -35,6 +35,10 @@ def __init__(self, **kwargs):
 
     def expected_data(self, pars):
         tensorlib, _ = get_backend()
+        sh = tensorlib.shape(pars)
```
kratsg: this doesn't make much sense. Why are the `pars` multiplied by the `fc`?

src/pyhf/paramview.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def index_helper(name, tensor_shape, batch_shape, par_map):
```
kratsg: any reason for this to be outside of the ParamViewer class?

src/pyhf/constraints.py

src/pyhf/constraints.py
lukasheinrich: `self.parsets` is ordered via `pdfconfig.auxdata_order`

src/pyhf/constraints.py
lukasheinrich: was just temp comment recording the shapes of the tensors. they were outdated and I removed them completely

src/pyhf/modifiers/shapesys.py
lukasheinrich: because the alphasets shape is that dimension.

src/pyhf/modifiers/staterror.py
lukasheinrich: `yab`

* `y`: counter for instance of staterror
* `a`: batch direction
* `b`: global bin counter

src/pyhf/paramsets.py
lukasheinrich: why does it not make sense? this is just a generalization of the line below to the batched case.

src/pyhf/paramview.py
lukasheinrich: we can use anything except for the passed arguments. I tend to not include free functions in classes

src/pyhf/constraints.py
lukasheinrich: good catch.. replace with concatenation of `axis=1`

src/pyhf/constraints.py

src/pyhf/constraints.py
lukasheinrich: I don't think introducing a magic number `self.batch_size = 0` to signal an unbatched computation is a good idea. I'd rather just keep the additional `if`

src/pyhf/constraints.py
lukasheinrich: in your example batch size is "1"

```
import numpy as np
pars = np.random.random(100)
batch_size = 1
np.reshape(pars, (batch_size or 1,) + np.shape(pars))
```

src/pyhf/constraints.py
lukasheinrich: for this data member the unbatched computation reduces to a computation with batch_size = 1

src/pyhf/constraints.py
lukasheinrich: removed

src/pyhf/constraints.py

src/pyhf/constraints.py
lukasheinrich: yeah, the means of the normal constraint are set by the parameter directly but I think making this explicit is useful here

lukasheinrich: "means" here refers to the mean value of the gaussian constraint term

src/pyhf/modifiers/shapefactor.py
lukasheinrich: `yab`

* `y`: counter for instance of staterror
* `a`: batch direction
* `b`: global bin counter

same as below, if you prefer another letter lmk

src/pyhf/constraints.py

src/pyhf/constraints.py
lukasheinrich: below? it's just computing the `poisson_logpdf` no?

src/pyhf/constraints.py
lukasheinrich: see above re: my hesitation

src/pyhf/constraints.py
lukasheinrich: batch_size can be `None` or >=1.

src/pyhf/constraints.py
lukasheinrich: this code reshapes the parameters to be amenable to batched computation. If unbatched the (Npars,) tensor is reshaped into `(1,Npars)` .. in the batched case this is more or less a noop

src/pyhf/constraints.py
lukasheinrich: I'd leave this for #540

src/pyhf/constraints.py
lukasheinrich: see above for my hesitation

lukasheinrich: see above for my hesitation

src/pyhf/constraints.py
lukasheinrich: see answer above

src/pyhf/constraints.py

src/pyhf/modifiers/shapefactor.py
lukasheinrich: added extensive comment

src/pyhf/modifiers/staterror.py
lukasheinrich: using param_viewer we have better control over which indices we need and we can more explicitly construct the access field as commented on the code

src/pyhf/paramview.py
kratsg: Right, so my only concern is that it's used once -- and only once. Is it planned to be used elsewhere? I agree with you on free functions, but single-use functions are weird to me.

tests/test_constraints.py
kratsg: I prefer more verbose tests when one can't simplify it without losing clarity.

src/pyhf/constraints.py
kratsg: Ok.

src/pyhf/constraints.py
kratsg: Ok, I would like to have a comment in the code here along that effect

```suggestion
                # if batched, noop
                # if unbatched, reshape parameters to (1, N parameters)
```

src/pyhf/constraints.py
kratsg: Right, it happens that in this case, the `mu` of the gaussian is the parameter value itself... I would prefer to change to `normal_pars` with a comment above of `# NB: this is also the mu of the gaussian constraint term`

src/pyhf/constraints.py
kratsg: Not following. Is there a clear example of what you expect this calculation to do?

src/pyhf/constraints.py
kratsg: oh yes, you're right.

src/pyhf/modifiers/shapefactor.py
kratsg: why not `m` for modifier?

src/pyhf/modifiers/staterror.py
lukasheinrich: done

src/pyhf/paramsets.py
kratsg: I think this can be replaced with an einsum, unless I'm mistaken... What's the return shape of `expected_data` for `pars = (B, npars)`?

src/pyhf/paramview.py
```diff
@@ -0,0 +1,98 @@
...
+        self._index_selection = index_helper(
+            name, tensor_shape, self.batch_shape, self._par_map
+        )
+
+        if self._index_selection:
+            cat = default_backend.astensor(
+                default_backend.concatenate(self._index_selection, axis=-1), dtype='int'
+            )
+            if self.batch_shape:
+                self._indices_concatenated = default_backend.einsum('ij->ji', cat)
```
kratsg: not fully clear to me why a transpose is needed if there's batching vs no batching.

lukasheinrich: there is a switch in dimension order between `index_selection` and `indices_concatenated` (and thus the return type of `.get()` in order to remove superfluous ops in the modifier application.

Added a comment

src/pyhf/constraints.py
lukasheinrich: hm, I don't think that's much more cleaner. To me "normal_means" refers to the "means" or the normal distribution in this constraint computation. I'd lean towards keeping as is

src/pyhf/paramsets.py
lukasheinrich: i don't think `einsum` makes this more readable. The return shape is the same as `pars`

tests/test_constraints.py
matthewfeickert: It seems that you're in agreement so let's keep it as is.

src/pyhf/paramsets.py
kratsg:
```suggestion
        sh = tensorlib.shape(pars)
        # if batched, tile the factors as they only depend on background uncertainty
```

src/pyhf/modifiers/staterror.py
```diff
@@ -31,10 +31,11 @@ def __init__(self, staterr_mods, pdfconfig, mega_mods, batch_size=None):
         keys = ['{}/{}'.format(mtype, m) for m, mtype in staterr_mods]
         staterr_mods = [m for m, _ in staterr_mods]
```
kratsg: fix this up. use self._staterr_mods and remove the internal/temp variable probably?

lukasheinrich: done

src/pyhf/modifiers/.vscode/tags
```diff
@@ -0,0 +1,140 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
```
matthewfeickert: @lukasheinrich Please remove this. :)

matthewfeickert: Thank you!

matthewfeickert:
:tada: Congratulations @lukasheinrich on this heroic effort, and extra praise to @kratsg for a stellar review. :beers: are in order.

PR #520, opened 2019-07-28
feat: Add discovery p0 test statistic

nikoladze:
# Description

It would be nice to have the ability to calculate discovery p values. I started implementing the q_0 test statistic for discovery of a positive signal. To avoid code duplication i moved the calculation of the profile log likelihood ratio into a separate function.
This could potentially then be included as a separate flag (maybe similar to the `qtilde`) in `pyhf.utils.hypotest` but i still have to figure out what will be the best way to do this.
Also i don't have any tests for this yet.

ReadTheDocs build: https://pyhf.readthedocs.io/en/docs-dev-p0-trigger

- [ ] Tests are passing

# Before Merging

For the PR Assignees:


src/pyhf/utils.py
```diff
@@ -109,17 +147,48 @@ def qmu(mu, data, pdf, init_pars, par_bounds):
...
-    qmu = loglambdav(mubhathat, data, pdf) - loglambdav(muhatbhat, data, pdf)
+    mubhathat, muhatbhat, qmu = pll_and_estimators(mu, data, pdf, init_pars, par_bounds)
     qmu = tensorlib.where(muhatbhat[pdf.config.poi_index] > mu, [0], qmu)
     return qmu
 
 
+def q0(mu, data, pdf, init_pars, par_bounds):
+    r"""
+    The test statistic, :math:`q_{0}`, for discovery of a positive signal
+    as defiend in Equation (12) in `arXiv:1007.1727`_, for :math:`\mu=0`.
```
kratsg: defined

kratsg:
Had a brief look-over. Seems fine to me. We'll need to figure out a proper way to handle switching out test statistics for these things in the future (something a bit more generic -- perhaps akin to how we handle interpolation code configuration with strings->function mapping).

We will want tests for this however, with some expected outputs to assert that it's written correctly and computes the right answer.

nikoladze:
> Had a brief look-over. Seems fine to me. We'll need to figure out a proper way to handle switching out test statistics for these things in the future (something a bit more generic -- perhaps akin to how we handle interpolation code configuration with strings->function mapping).

That sounds reasonable. So one should have then for each test statistic one function that calculates the test statistic and one function that calculates the p-values? (Since this is different for e.g qmu vs qtilde). Should i go ahead with just another flag for q0 (called it `use_q0` now due to name clash with the function) and make another iteration for improving the mapping later or resolve this as well in this PR?

> We will want tests for this however, with some expected outputs to assert that it's written correctly and computes the right answer.

I would tabulate some values in `test_validation.py` similar to what is already there for CL_s and validate them against HistFactory?

kratsg:
> I would tabulate some values in `test_validation.py` similar to what is already there for CL_s and validate them against HistFactory?

That's roughly the way to go. We'd want to make sure it isn't too slow to calculate. It might be better to add a unit test around the function itself, rather than an integration test (which test_validation does).

lukasheinrich:
thanks @nikoladze for the PR. I agree this is much-needed functionality. I'll review this. At first glance it seems like it just touches the utils.py and not interfere with #519 (which will be quite a bit of a change in the pdf object side).

One thing that would be important is to add tests that check against ROOT, like we do in test_validation.py / the validation dir (here we extracted some hardcoded numbers from ROOT via run_singlepy we check against, at some point a more proper framework would be needed)

src/pyhf/utils.py
```diff
@@ -76,6 +76,44 @@ def loglambdav(pars, data, pdf):
...
+        data (Tensor): The data to be considered
+        pdf (|pyhf.pdf.Model|_): The HistFactory statistical model used in the likelihood ratio calculation
+        init_pars (Tensor): The initial parameters :math:`\pmb{\theta}`
+        par_bounds(Tensor): The bounds on the paramter values
+
+    .. |pyhf.pdf.Model| replace:: ``pyhf.pdf.Model``
+    .. _pyhf.pdf.Model: https://diana-hep.org/pyhf/_generated/pyhf.pdf.Model.html
+
+    Returns:
+        Tuple of (Tensor, Tensor, Float): The parameters :math:`\hat{\hat{\pmb{\theta}}}` that maximise the likelihood for the given value of :math:`\mu`, the parameters :math:`\hat{\pmb{\theta}}` that maximise the likelihood unconditionally (including :math:`\hat{\mu}`) and the value of :math:`-2\ln\lambda(\mu)`
```
lukasheinrich: is the third argument really a float and not a tensor? I don't think `loglambdav` bails out of tensorland

src/pyhf/utils.py
```diff
@@ -274,18 +351,21 @@ def hypotest(
 
     init_pars = init_pars or pdf.config.suggested_init()
     par_bounds = par_bounds or pdf.config.suggested_bounds()
+    test_stat = qmu if not use_q0 else q0
+    if qtilde and use_q0:
```
lukasheinrich: probably this should become a better look up generally now that we expand the number of test statistics
```
test_stats = {
    'q_mu': ..
    'q_tilde_mu': ..
    'q_0': ..
}
```

src/pyhf/utils.py
```diff
@@ -109,17 +147,48 @@ def qmu(mu, data, pdf, init_pars, par_bounds):
...
-    muhatbhat = optimizer.unconstrained_bestfit(
-        loglambdav, data, pdf, init_pars, par_bounds
-    )
-    qmu = loglambdav(mubhathat, data, pdf) - loglambdav(muhatbhat, data, pdf)
+    mubhathat, muhatbhat, qmu = pll_and_estimators(mu, data, pdf, init_pars, par_bounds)
     qmu = tensorlib.where(muhatbhat[pdf.config.poi_index] > mu, [0], qmu)
     return qmu
 
 
+def q0(mu, data, pdf, init_pars, par_bounds):
```
lukasheinrich: why does this have a `mu` parameter? shouldn't that be fixed to 0?

nikoladze: mhh - probably there is no real use in having this as a free parameter. On the other hand - when we want to switch to selecting the test statistic from a mapping it might be simpler when they all have the same parameters.

src/pyhf/utils.py
nikoladze: This is a good idea, but there are probably more places where a distinction has to be made. For example the way how to calculate p-values from the test statistic (mainly different between q_mu and q_tilde) and if one wants cls or pvalues (cls doesn't really make sense for q_0)

nikoladze:
> One thing that would be important is to add tests that check against ROOT, like we do in test_validation.py / the validation dir (here we extracted some hardcoded numbers from ROOT via run_singlepy we check against, at some point a more proper framework would be needed)

I added one test for single-bin, one systematic now to have at least something working - but it involved quite some copy&pasting and some not so elegant if-switches for q_0. Also, at least for some numbers i get a quite high relative difference between pyhf and the values i tabulated from HF (had to set tolerance to 3e-4

src/pyhf/utils.py
lukasheinrich: I think a way to do this is to have a kind of test statistic factory which takes kwargs that could be 

```
def get_teststat(type, **opts):
    ..
  teststat = get_teststat(type = 'qmu', mu = mu)
  teststat(data, pdf)

or 

  teststat = get_teststat(type = 'q0')
  teststat(data, pdf)

```

i.e. `mu` is not part of the test stat API (which just takes the data and the pdf, but part of the API that gives you the teststat

src/pyhf/utils.py
nikoladze: Indeed it is a Tensor - fixed that docstring now. Thanks!

src/pyhf/utils.py
kratsg: I think the best way for now is to keep it as it is, and then clean it up after this PR.

kratsg:
> I added one test for single-bin, one systematic now to have at least something working - but it involved quite some copy&pasting and some not so elegant if-switches for q_0. Also, at least for some numbers i get a quite high relative difference between pyhf and the values i tabulated from HF (had to set tolerance to 3e-4)

Not good. We should make sure we understand the difference if there is meant to be one (probably not!)

lukasheinrich:
I'll try to check what the root of the discrepancy is

nikoladze:
> I'll try to check what the root of the discrepancy is

Pun intended? ;)
I also tried to dig around a bit. I can't really reproduce the steps ROOT does in the AsymptoticCalculator, but what the logs indicate, the largest part of the discrepancy comes from different numbers in the Asimov data set. ROOT gives `[76.9869, 53.0479]`, while i get `[76.99019042, 53.05121471]` from pyhf. When i propagate the ROOT asimov values through the rest of the pyhf calculation the matching gets a bit better (~up to 2e-5 relative difference).
Interestingly the fitted parameter value for the nuisance paramater in the constrained fit for getting asimov data is pretty close between ROOT and pyhf (~1.5e-5 relative difference).
Now i'm a bit stuck in following how [AsymptoticCalculator::GenerateAsimovData](https://root.cern/doc/master/AsymptoticCalculator_8cxx_source.html#l01130) works ...

lukasheinrich:
hi @nikoladze,

sorry for the long silence, we're in a pretty big refactoring now, but i'll come back to this asap

nikoladze:
@lukasheinrich No problem - i'm also  quite busy right now, but i will have a look again when i have time

matthewfeickert:
It would be nice to revive this PR (as @lukasheinrich @kratsg and I were just talking about discovery stats today). @nikoladze, I may go ahead and do some rebasing and cleaning on your branch (so please fetch if needed). Additionally, maybe it would be nicer if we could move all of these ROOT files use for validation into the [`scikit-hep-testdata` project](https://github.com/scikit-hep/scikit-hep-testdata).

nikoladze:
Hi @matthewfeickert ,
Simple rebase won't do since some things changed (i didn't touch this in a very long time). So i just merged discarding the changes except for the tests and put everything in the new places until the my test for p0 passed again.

nikoladze:
I tried to investigate a bit where the difference between this p0 calculation and the one i get from ROOT/Histfactory comes from (the current test tolerance is set to 3e-4 which is rather high).

When looking at the "on-off" problem with `"bkg": [50.0], "bkgerr": [7.0], "sig": [25.0]` and scanning the observed data up to high values where the p-values get small we can see systematic discrepancies between the methods to calculate the p0 value starting when the number of observed events becomes larger than 120. Especially for the expected p0 one can see ROOT/Histfactory is doing something strange:

image

Now, for the "'on-off" case the CCGV paper actually has an exact solution for the MLEs (and cowan has a note where he put these into a formula: http://www.pp.rhul.ac.uk/~cowan/stat/medsig/medsigNote.pdf) so we can actually compare what should come out (called "exact asymptotic" in the plot). I did this for the observed p0:

image

So it seems either something is going wrong in ROOT or the model i created for histfactory doesn't exactly correspond to the "on-off" model we are targetting here. I already tried to set the Lumi uncertainty in the xml file to 0 and switched `NormalizeByTheory="False"` for the bkg sample, but this didn't seem to have an effect. One thing i want to try later is checking the MLEs that i get from the Histfactory model.

Side note:

While investigating this i found a numerical issue in the pyhf implementation of asymptotic p-values:

https://github.com/scikit-hep/pyhf/blob/73312338ef563305813198e5abc019de46fbbbc2/src/pyhf/infer/calculators.py#L74

This leads to problems when we have very low p-values since we are subtracting almost 1 from 1 to get something very small. For the numpy backend this might be solved by using `scipy.stats.norm.sf`, for the others i don't know atm. But maybe we should discuss this in a separate issue.

matthewfeickert:
> I tried to investigate a bit where the difference between this p0 calculation and the one i get from ROOT/Histfactory comes from (the current test tolerance is set to 3e-4 which is rather high).

This is all really nice to see &mdash; thanks! Can you share with us (maybe in a Gist) the code that you're using to do this check and make the plots? Would be nice to look at as well.

> While investigating this i found a numerical issue in the pyhf implementation of asymptotic p-values. This leads to problems when we have very low p-values since we are subtracting almost 1 from 1 to get something very small. But maybe we should discuss this in a separate issue.

Yeah, if you can open up another issue with an example of when this becomes problematic that would be great. Thanks!

nikoladze:
>  Can you share with us (maybe in a Gist) the code that you're using to do this check and make the plots? Would be nice to look at as well.

I put it in a separate git for now. Here is a script with wrapper functions (e.g run the Histfactory chain in a tempdir):
https://github.com/nikoladze/debug_pyhf/blob/master/run_everything.py
And here is the corresponding notebook that creates the plots i posted earlier (also containing the formula for the "exact" case):
https://github.com/nikoladze/debug_pyhf/blob/master/study_diff_hf_pyhf.ipynb

nikoladze:
I think i know more now: the strange drop in p-values for ROOT/HistFactory in this plot
image
comes from the default parameter range in HistFactory for gamma parameters (from 0 to 1 + 5 sigma). When i manually set it to the same range as in pyhf (from 1e-10 to 10) this drop goes away. In practice i think this shouldn't matter though since we will rarely run into situations where our observed data is so large that in the fit where mu is fixed to 1 the background is pulled up by more than 5 sigma ;)

So i would say that issue is resolved

However, the discrepancies in the validation test i set up for this PR persist (they concern a reasonable parameter range for the gamma parameter). But i think these can be attributed to differences between fitting with scipy or minuit. When i fit with minuit in pyhf i get results much closer to ROOT (1e-6 relative difference) . But - the "exact" asymptotic formula derived from the closed forms of the MLEs is closer to pyhf+scipy! The study can be found in [my  notebook](https://github.com/nikoladze/debug_pyhf/blob/c53948e4992e73cf39ccb93fa968991ceaaaa777/study_diff_hf_pyhf.ipynb)

matthewfeickert:
We should work on rebasing this. @nikoladze Do you have any problems with us just taking another branch at the state of `master` and then just using `git cherry-pick` and hand picking the relevant sections to get this up to date? There has been a lot of movement in the last year of the internals, so this will take a long time to rebase cleanly I think.

lukasheinrich:
@nikoladze i've been working on test statistic stuff in #1005 which hopefully should make it straight forward to add `q0` to

src/pyhf/infer/__init__.py
```diff
@@ -109,30 +119,38 @@ def hypotest(
     )
 
     _returns = [CLs]
+    if use_q0:
```
nikoladze, line 122: what follows after here is a little bit of a mess, now with even more if-else distinctions due to the `q0` - one might consider disentangling the calculation of pvalues and CLs?

src/pyhf/infer/calculators.py
```diff
@@ -156,23 +161,30 @@ def teststatistic(self, poi_test):
...
+            teststat_func = q0
+            asimov_mu = 1.0
+        else:
 
-        qmu_v = teststat_func(
-            poi_test, self.data, self.pdf, self.init_pars, self.par_bounds
-        )
-        sqrtqmu_v = tensorlib.sqrt(qmu_v)
+            def teststat_func(*args, **kwargs):
+                "make function signature without poi_test to be consistent with q0"
```
nikoladze, line 170: The reason for making a function signature without `poi_test` is that `q0` doesn't take a `poi_test` argument since it is always 0. The alternative would be to change `q0` to take a `poi_test` argument, regardless if it only makes sense to be 0.

src/pyhf/infer/test_statistics.py
```diff
@@ -226,3 +226,56 @@ def tmu_tilde(mu, data, pdf, init_pars, par_bounds):
             + 'Use the tmu test statistic (pyhf.infer.test_statistics.tmu) instead.'
         )
     return _tmu_like(mu, data, pdf, init_pars, par_bounds)
+
+
+def q0(data, pdf, init_pars, par_bounds):
```
nikoladze, line 231: `q0` still needs a separate function since it needs to clip for `muhat < 0` instead of `muhat > mu` as done in `_qmu_like`

nikoladze:
@matthewfeickert, @lukasheinrich i would be fine with starting over bit-by-bit from master, but actually it's only a few places where i had to resolve minor conflicts after merging. I added a few inline comments.

src/pyhf/infer/__init__.py
```diff
@@ -109,30 +119,38 @@ def hypotest(
...
+        _returns = [CLsb]
+
     if kwargs.get('return_tail_probs'):
         _returns.append([CLsb, CLb])
-    if kwargs.get('return_expected_set'):
+    if kwargs.get('return_expected_set') or kwargs.get('return_expected'):
+        n_sigma_list = [2, 1, 0, -1, -2]
+        if not kwargs.get('return_expected_set'):
+            n_sigma_list = [0]
```
nikoladze, line 130: added this to avoid having to type the code for calculating p0 and CLs (and doing the if-else) a second time for the case that only one expected value should be returned

matthewfeickert:
I'm going to start working on this again. For the time being I'm going to start working on a new branch from `origin/master` and then just copy what is needed from here. We can then decide if it is easier for me to just force push that branch here (preferable in my mind as it keeps the discussion), or if we just close out this PR and then add @nikoladze as co-author on a PR from my branch.

matthewfeickert:
@nikoladze We've closed this PR as we migrated it over to PR #1232. Thanks for your contribution here, which was a big help! :rocket:

Issue #508, opened 2019-07-12
Automate deployment to PyPI


matthewfeickert:
# Description

According to @lukasheinrich, the current workflow for deploying to PyPI is:

```
git checkout master
git pull
bumpversion patch
git commit
git push origin master --tags
```

This is a bit annoyingly manual and ideally should be done automatically.

Luckily, there is an [official PyPA GitHub action](https://discuss.python.org/t/official-github-action-for-publishing-to-pypi/1061) to do this:

https://github.com/pypa/gh-action-pypi-publish

However, we need GitHub actions for pyhf, so we have to wait.

matthewfeickert:
I'm happy to take this one once we get GitHub actions.

matthewfeickert:
An example of how this is actually hard without automation is that when [`v0.1.2`](https://github.com/diana-hep/pyhf/releases/tag/v0.1.2) was released to PyPI the commit got into the tags, but not the refs/heads, but still got [picked up by the CI, built and deployed](https://travis-ci.org/diana-hep/pyhf/builds/557986549). It was [later pushed to `master`](https://github.com/diana-hep/pyhf/commit/3e1e383c775a4407277dad49c810f5b16a56ef3b), but for about an hour there was a mismatch where PyPI was ahead of GitHub.

Having a CI that can properly do releases for us would solve this sort of confusing thing.

kratsg:
Why don't we have GitHub Actions yet?

matthewfeickert:
> Why don't we have GitHub Actions yet?

GitHub Actions is still in beta. 

![beta_version](https://user-images.githubusercontent.com/5142394/61195613-c0b27b00-a68e-11e9-96a2-091228004c2f.png)

They have tightly restricted the number of people who can enter the beta, and they denied my request to bump Lukas and diana-hep up in the queue.

matthewfeickert:
We have GitHub Actions now (:tada:), and @webknjaz has done an awesome job getting [`pypi-publish`](https://github.com/pypa/gh-action-pypi-publish) up [on the GitHub Marketplace](https://github.com/marketplace/actions/pypi-publish) and ready to go. Time to use it!

matthewfeickert:
[...]

PR #527, opened 2019-08-20
Add CI through GitHub actions

matthewfeickert:
# Description

- Add CI tests
- Add daily cron job of tests and benchmarks

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add CI tests with GitHub Actions
* Add daily cron job of tests and benchmarks with GitHub Actions
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,38 @@
+name: CI Python 3 on Linux
+
+on: [push, pull_request]
+
+jobs:
+  test:
+
+    strategy:
+      max-parallel: 4
+      matrix:
+        platform: [ubuntu-latest]
+        python-version: [3.6, 3.7]
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        pip install --ignore-installed -U -q --no-cache-dir -e .[complete]
+        pip list
+    - name: Lint with Pyflakes
+      run: |
+        pyflakes src
+        check-manifest
+    - name: Lint with Black
+      run: |
+        black --check --diff --verbose .
+    - name: Check MANIFEST
+      run: |
+        check-manifest
+    - name: Test with pytest
+      run: |
+        python -m pytest -r sx --ignore tests/benchmarks/
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -1,0 +1,44 @@
+name: Daily cron test
+
+on:
+  # Run daily at 0:01 UTC
+  schedule:
+  - cron:  '1 0 * * *'
+
+jobs:
+  test:
+
+    strategy:
+      max-parallel: 4
+      matrix:
+        platform: [ubuntu-latest]
+        python-version: [3.6, 3.7]
+    runs-on: ${{ matrix.platform }}
+
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        pip install --ignore-installed -U -q --no-cache-dir -e .[complete]
+        pip list
+    - name: Lint with Pyflakes
+      run: |
+        pyflakes src
+        check-manifest
+    - name: Lint with Black
+      run: |
+        black --check --diff --verbose .
+    - name: Check MANIFEST
+      run: |
+        check-manifest
+    - name: Test with pytest
+      run: |
+        python -m pytest -r sx --ignore tests/benchmarks/
+    - name: Run benchmarks
+      run: |
+        python -m pytest -r sx tests/benchmarks/test_benchmark.py


matthewfeickert:
@kratsg @lukasheinrich I don't have docs testing in here yet, but if you have comments now it is always great to get feedback early.

matthewfeickert:
We should also follow up on Issue #508

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
+name: CI Python 3 on Linux
+
+on: [push, pull_request]
+
+jobs:
+  test:
+
+    strategy:
+      max-parallel: 4
```
kratsg: why max parallel?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+      matrix:
+        platform: [ubuntu-latest]
```
kratsg: what other platforms can we test on?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+        python-version: [3.6, 3.7]
```
kratsg: are we dropping python 2 at this point?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        pip install --ignore-installed -U -q --no-cache-dir -e .[complete]
```
kratsg: have we checked if `--ignore-installed --no-cache-dir` are needed?

.github/workflows/cron.yml
```diff
@@ -0,0 +1,44 @@
+name: Daily cron test
```
kratsg: do we need to copy/paste? There's no way to do inheritance / includes?

.github/workflows/ci.yml
matthewfeickert: Artifact from when I had tried to run across 3 OS's. Can be removed, though it has no effect here as the matrix only expands to 4 jobs.

.github/workflows/ci.yml
matthewfeickert: I tried using

```yaml
 platform: [ubuntu-latest, macos-latest, windows-latest]
```

however, `macos-latest` doesn't support Python 3.7 at the moment, and `windows-latest` has issues with PyTorch. So given our current requirements we can only use the Linux virtual environments to test.

.github/workflows/ci.yml
matthewfeickert: Why should we include it here? It is already in Travis and since we're going to be dropping Python 2.7 at the end of the year we can just start fresh on a new CI platform and let Travis deal with 2.7 until Python 3 freedom day.

.github/workflows/ci.yml
matthewfeickert: `--ignore-installed` definitely isn't given the container used, but I just copied it from our docs to be consistent. `--no-cache-dir` is used as I don't think there is caching in anyway (so why bother even creating a cache dir?). If I'm wrong that would be great.

.github/workflows/cron.yml
matthewfeickert: Not that I know of. If I'm wrong, that would be great.

.github/workflows/ci.yml
matthewfeickert: Though it would be worth looking at trying to fix the Windows build.

matthewfeickert:
@webknjaz I know you've got plenty to do already, but from previous discussions I know you already have some [nice GitHub Actions workflows setup](https://github.com/sanitizers/octomachinery/blob/087b099198dfab2f1813545e96cde949bf75778f/.github/main.workflow) and have also looked at using the [new CI/CD for testing](https://github.com/sanitizers/octomachinery/blob/a4c28b5d431b7c1225b7d8f5ad37d7bc186f01e7/.github/workflows/python-tox.yml) so I thought I'd ask if you have suggestions on if there is a clean way to use the GitHub Actions CI/CD with reusable sections of workflows given @kratsg's [comment](https://github.com/diana-hep/pyhf/pull/527#discussion_r315723076).

webknjaz:
While it looks very similar to Azure Pipelines, I haven't found any native way to template workflows. I think that the smallest reusable component here is a GitHub Action (which can be parametrized with some input args).
I've also found other limitations, like [inability to generate env var based job matrix](https://github.community/t5/GitHub-API-Development-and/GitHub-Actions-new-BUG-Unable-to-create-environment-variables/m-p/29338/highlight/true#M2583).
I'm currently working on an Action for uploading dists to PyPI: https://github.com/pypa/gh-action-pypi-publish/pull/7 (old Action is being converted into the new format).
Currently, that's all I know. I also have a few links to docs collected @ https://github.com/pypa/gh-action-pypi-publish/issues/5.

matthewfeickert:
Thanks very much @webknjaz &mdash; that's very helpful information!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
+name: CI Python 3 on Linux
+
+on: [push, pull_request]
```
webknjaz: Try:
```suggestion
on:
  push:
  pull_request:
  # Run daily at 0:01 UTC
  schedule:
  - cron:  '1 0 * * *'
```
(I'm not sure about the push/pr events, you may need to transform them to smth like `push: true` or `push: {}` so please do some experiments with this)

webknjaz: @matthewfeickert ^

webknjaz: Then use `if: github.event_name == 'scheduled'` on that heavy task.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+        check-manifest
+    - name: Lint with Black
+      run: |
+        black --check --diff --verbose .
+    - name: Check MANIFEST
+      run: |
+        check-manifest
+    - name: Test with pytest
+      run: |
+        python -m pytest -r sx --ignore tests/benchmarks/
```
webknjaz:
```suggestion
        python -m pytest -r sx --ignore tests/benchmarks/
    - name: Run benchmarks
      if: github.event_name == 'scheduled'
      # Refs:
      # https://help.github.com/en/articles/workflow-syntax-for-github-actions#example-using-contexts
      # https://help.github.com/en/articles/events-that-trigger-workflows#webhook-events
      run: |
        python -m pytest -r sx tests/benchmarks/test_benchmark.py
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Run benchmarks
+      if: github.event_name == 'scheduled' && matrix.python-version == '3.7'
```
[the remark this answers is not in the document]
webknjaz: In any unclear situation, complain on Twitter :)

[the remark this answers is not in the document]
webknjaz: On the serious note, in your matrix, you use a float literal `3.7` while here you're comparing it to a string. I bet what's happening is that `3.7 != '3.7'`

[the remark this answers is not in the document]
webknjaz:
```suggestion
      if: github.event_name == 'scheduled' && matrix.python-version == 3.7
```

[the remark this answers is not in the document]
webknjaz: Weird. Docs say that `matrix` is in the context..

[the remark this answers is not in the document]
webknjaz: Ouch... that's definitely a bug

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
+name: CI Python 3 on Linux
+
+on:
+  push:
+  pull_request:
```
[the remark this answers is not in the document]
webknjaz: It is. The value is `null` AFAIR.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+        check-manifest
+    - name: Test with pytest
+      run: |
+        python -m pytest -r sx --ignore tests/benchmarks/
+    - name: Run benchmarks
+      if: github.event_name == 'schedule' && matrix.python-version == 3.7
+      run: |
+        python -m pytest -r sx tests/benchmarks/test_benchmark.py
+
+        docs:
```
webknjaz:
```suggestion
  docs:
```

PR #534, opened 2019-08-30
feat: Add backend option to pyhf CLI

matthewfeickert:
# Description

Add ability to change the computational backend used by `pyhf` from the command line with a `--backend` flag.

**Example:**

```
$ curl -sL https://git.io/fjxXE | pyhf cls
{
    "CLs_exp": [
        0.07807427911686152,
        0.17472571775474582,
        0.35998495263681274,
        0.6343568235898907,
        0.8809947004472013
    ],
    "CLs_obs": 0.3599845631401913
}
$ curl -sL https://git.io/fjxXE | pyhf cls --backend pytorch
{
    "CLs_exp": [
        0.07808291912078857,
        0.1747395098209381,
        0.36000293493270874,
        0.6343730092048645,
        0.8810024857521057
    ],
    "CLs_obs": 0.3600029945373535
}
$ curl -sL https://git.io/fjxXE | pyhf cls --backend tensorflow
{
    "CLs_exp": [
        0.0780724585056305,
        0.17472276091575623,
        0.35998106002807617,
        0.6343533396720886,
        0.8809930086135864
    ],
    "CLs_obs": 0.35998106002807617
}
```

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add backend option for the pyhf cls command line API
* Add tests to test_scripts.py to cover the pyhf cls backend option
* Answer the FAQ on setting the backend at the CLI
```

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -9,10 +9,10 @@ Questions
 Is it possible to set the backend from the CLI?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Not at the moment. `Pull Requests <https://github.com/scikit-hep/pyhf/compare>`__ are welcome.
-
-See also:
-  - :issue:`266`
+Yes.
+Use the :code:`--backend` option for :code:`pyhf cls` to specify a tensor backend.
+The default backend is NumPy.
+For more information see :code:`pyhf cls --help`.
 
 Troubleshooting
 ---------------
--- a/src/pyhf/cli/stats.py
+++ b/src/pyhf/cli/stats.py
@@ -5,7 +5,7 @@ import json
 
 from ..utils import hypotest, EqDelimStringParamType
 from ..workspace import Workspace
-from .. import tensorlib, set_backend, optimize
+from . import tensor, get_backend, set_backend, optimize
 
 logging.basicConfig()
 log = logging.getLogger(__name__)
@@ -27,10 +27,24 @@ def cli():
 @click.option('-p', '--patch', multiple=True)
 @click.option('--testpoi', default=1.0)
 @click.option('--teststat', type=click.Choice(['q', 'qtilde']), default='qtilde')
+@click.option(
+    '--backend',
+    type=click.Choice(['numpy', 'pytorch', 'tensorflow', 'np', 'torch', 'tf']),
+    help='The tensor backend used for the calculation.',
+    default='numpy',
+)
 @click.option('--optimizer')
 @click.option('--optconf', type=EqDelimStringParamType(), multiple=True)
 def cls(
-    workspace, output_file, measurement, patch, testpoi, teststat, optimizer, optconf
+    workspace,
+    output_file,
+    measurement,
+    patch,
+    testpoi,
+    teststat,
+    backend,
+    optimizer,
+    optconf,
 ):
     with click.open_file(workspace, 'r') as specstream:
         spec = json.load(specstream)
@@ -49,6 +63,15 @@ def cls(
         },
     )
 
+    # set the backend if not NumPy
+    if backend in ['pytorch', 'torch']:
+        set_backend(tensor.pytorch_backend())
+    elif backend in ['tensorflow', 'tf']:
+        from tensorflow.compat.v1 import Session
+
+        set_backend(tensor.tensorflow_backend(session=Session()))
+    tensorlib, _ = get_backend()
+
     optconf = {k: v for item in optconf for k, v in item.items()}
 
     # set the new optimizer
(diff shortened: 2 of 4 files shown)


matthewfeickert:
This still has some work to do on it, as TensorFlow is now complaining that

```python-traceback
...
  File "/home/mcf/Code/GitHub/pyhf/src/pyhf/utils.py", line 162, in pvals_from_teststat
    if sqrtqmu_v < sqrtqmuA_v:
  File "/home/mcf/venvs/pyhf-dev/lib/python3.6/site-packages/tensorflow/python/framework/ops.py", line 690, in __bool__
    raise TypeError("Using a `tf.Tensor` as a Python `bool` is not allowed. "
```

though I'm not sure why it hasn't picked this up before. To fix that requires something along the lines of replacing

```python
    if not qtilde:  # qmu
        nullval = sqrtqmu_v
        altval = -(sqrtqmuA_v - sqrtqmu_v)
    else:  # qtilde
        if sqrtqmu_v < sqrtqmuA_v:
            nullval = sqrtqmu_v
            altval = -(sqrtqmuA_v - sqrtqmu_v)
        else:
            qmu = tensorlib.power(sqrtqmu_v, 2)
            qmu_A = tensorlib.power(sqrtqmuA_v, 2)
            nullval = (qmu + qmu_A) / (2 * sqrtqmuA_v)
            altval = (qmu - qmu_A) / (2 * sqrtqmuA_v)
```

with

```python
    if not qtilde:  # qmu
        nullval = sqrtqmu_v
        altval = -(sqrtqmuA_v - sqrtqmu_v)
    else:  # qtilde
        def _sqrtqmu_v_case():
            nullval = sqrtqmu_v
            altval = -(sqrtqmuA_v - sqrtqmu_v)
            return nullval, altval
        
        def _sqrtqmuA_v_case():
            qmu = tensorlib.power(sqrtqmu_v, 2)
            qmu_A = tensorlib.power(sqrtqmuA_v, 2)
            nullval = (qmu + qmu_A) / (2 * sqrtqmuA_v)
            altval = (qmu - qmu_A) / (2 * sqrtqmuA_v)
            return nullval, altval
        
        import tensorflow as tf
        
        nullval, altval = tf.cond(
            tf.math.less(sqrtqmu_v[0], sqrtqmuA_v[0]), _sqrtqmu_v_case, _sqrtqmuA_v_case
        )
```

but to make that backend agnostic is going to take sometime (and I need some sleep).

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -44,6 +44,14 @@ def clip(self, tensor_in, min_value, max_value):
             max_value = tf.reduce_max(tensor_in)
         return tf.clip_by_value(tensor_in, min_value, max_value)
 
+    def conditional(self, bool_condition, true_callable, false_callable):
```
matthewfeickert: As this is introducing new functions to `tensorlib` I'll make this a separate PR, but I'll demo and protype them here before I do so.

matthewfeickert:
As I've answered our only FAQ posted at the moment I should also add another before this goes in.

matthewfeickert:
> I'm generally ok with these changes, but I don't think we should be supporting conditionals

Can you elaborate on why

> -- especially as the only place that is being used is for single-valued numbers

So you just want to have one line that is something like 

```
nullvl, atlval = true_callable() if predicate else false_callable()
```

> which should be done in `numpy` only

Why in numpy? Shouldn't this be backend independent?

> (and we're going to remove that code in the future). 

I've been defocused for the last months. What is getting removed and why?

matthewfeickert:
@kratsg The discussion RE: conditional above can probably get moved to PR #535 as I want to break this out into a few PRs.

matthewfeickert:
Given @kratsg's suggestion, this PR is going to change to promote the backend option to be a `pyhf` option instead of a `pyhf cls` option.

kratsg:
Does this need to stay as a draft?

matthewfeickert:
> Does this need to stay as a draft?

Yes, because as we agreed to not do PR #535 and PR #537 this is being blocked by PR #539. So it isn't ready for review and should stay as a draft.

lukasheinrich:
I think this would come after #583. Right now  I don't feel too confident in the non-numpy backend to successfuly do the fits tbh

lukasheinrich:
similar #535 I would vote to reopen this because we have a path to make the non-np backends fit robustly & reproduce the sbottom results

lukasheinrich:
@matthewfeickert do you still want to add this?

matthewfeickert:
> Do you still want to add this?

Sorry I missed this. Yes, I think this should still go in and not get closed, though we need to decide if we will accept it like it is (with modifications necessary for merge) or if we should do PR #541 first instead.

lukasheinrich:
i would do this now and not wait for the TF2 migration

kratsg:
> i would do this now and not wait for the TF2 migration

You and I have the same thought :)

PR #535, opened 2019-08-31
feat: Add 'conditional' method to tensorlib API

matthewfeickert:
# Description

In support of PR #534 add `conditional` method ~~, `less`, and `greater` methods~~ to tensorlib API. Also add tests for them.

**TODO:**
- [x] ~~Add `add` and `subtract` methods~~
- [x] Add tests for all added methods

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add 'conditional' method to the NumPy, TensorFlow, and PyTorch tensor backends
* Add tests for 'conditional'
* Add tests for overloaded operators on tensor operations that would be the equivalent of 'add', 'subtract', 'less', and 'greater'
```

### DIFF
--- a/src/pyhf/tensor/numpy_backend.py
+++ b/src/pyhf/tensor/numpy_backend.py
@@ -40,7 +40,7 @@ class numpy_backend(object):
 
     def add(self, tensor_in_1, tensor_in_2):
         """
-        Add two tensors element-wise
+        Add two tensors element-wise. Equivalent to :code:`sum([tensor_in_1, tensor_in_2], axis=0)`.
 
         Example:
 
--- a/src/pyhf/tensor/numpy_backend.py
+++ b/src/pyhf/tensor/numpy_backend.py
@@ -82,6 +82,80 @@ class numpy_backend(object):
         """
         return np.tile(tensor_in, repeats)
 
+    def conditional(self, predicate, true_callable, false_callable):
+        """
+        Runs a callable conditional on the boolean value of the evaulation of a predicate
+
+        Example:
+
+            >>> import pyhf
+            >>> import numpy as np
+            >>> pyhf.set_backend(pyhf.tensor.numpy_backend())
+            >>> a = pyhf.tensorlib.astensor([4])
+            >>> b = pyhf.tensorlib.astensor([5])
+            >>> pyhf.tensorlib.conditional(
+            ...     np.less(a, b)[0], lambda: np.add(a, b), lambda: np.subtract(a, b)
+            ... )
+            array([9.])
+
+        Args:
+            predicate (`scalar`): The logical condition that determines which callable to evaluate
+            true_callable (`callable`): The callable that is evaluated when the :code:`predicate` evalutes to :code:`true`
+            false_callable (`callable`): The callable that is evaluated when the :code:`predicate` evalutes to :code:`false`
+
+        Returns:
+            NumPy ndarray: The output of the callable that was evaluated
+        """
+        return true_callable() if predicate else false_callable()
+
+    def less(self, tensor_in_1, tensor_in_2):
+        """
+        The boolean value of :code:`(tensor_in_1 < tensor_in_2)` element-wise
+
+        Example:
+
+            >>> import pyhf
+            >>> pyhf.set_backend(pyhf.tensor.numpy_backend())
+            >>> a = pyhf.tensorlib.astensor([4])
+            >>> b = pyhf.tensorlib.astensor([5])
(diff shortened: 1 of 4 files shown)


matthewfeickert:
I might also add `add` and `subtract` methods to make the examples more uniform across `conditional`, `less`, and `greater`.

matthewfeickert:
A preview of the built docs is at RTD: https://pyhf-staging.readthedocs.io/en/feature-extend-tensorlib-api/

matthewfeickert:
@lukasheinrich @kratsg Following up on [the discussion](https://github.com/diana-hep/pyhf/pull/534#issuecomment-526857895) here, I want to add `conditional` to the API as TensorFlow will error with

```python-traceback
```

which disallows trying replace

https://github.com/diana-hep/pyhf/blob/f3b0680c115ef64cfb61a7631d9a550ec50c095f/src/pyhf/utils.py#L158-L169

with just

```python
    if not qtilde:  # qmu
        nullval = sqrtqmu_v
        altval = -(sqrtqmuA_v - sqrtqmu_v)
    else:  # qtilde
        if tensorlib.less(sqrtqmu_v, sqrtqmuA_v)[0]:
            nullval = sqrtqmu_v
            altval = -(sqrtqmuA_v - sqrtqmu_v)
        else:
            qmu = tensorlib.power(sqrtqmu_v, 2)
            qmu_A = tensorlib.power(sqrtqmuA_v, 2)
            nullval = (qmu + qmu_A) / (2 * sqrtqmuA_v)
            altval = (qmu - qmu_A) / (2 * sqrtqmuA_v)
    CLsb = 1 - tensorlib.normal_cdf(nullval)
    CLb = 1 - tensorlib.normal_cdf(altval)
    CLs = CLsb / CLb
    return CLsb, CLb, CLs
```

I don't see a way around this without adopting conditional control flow.

kratsg:
@matthewfeickert the problem is that you're getting into fixing `pyhf-stats` which is not where the focus of the package should be currently (since that falls under statisfactory technically). That's what i'm referring to when talking about this portion of the code. I'm not sure I understand the logic since the actual computation of the CLs does not require the piece of code with the conditional, so why can't we just typecast back to numpy arrays and do the conditional in that library is my question.

Numpy is the fundamental requirement for `pyhf`. And when `pyhf-stats` breaks away from `pyhf`, it won't have any connection to backend choice.

lukasheinrich:
Before we added the qtilde stuff, we did expand at some point our scope of tensor-ful computations to the test statistic, so I think just fixing the backend to be able to cover qtilde etc is fine. I need to review the TF APIs, since we might want to switch to eager execution entirely.

Even whenn we move to statisfactory, it will assume some kind of differentiable data model in the backend (what we currently view as a tensor backend) (i.e. what's the *type* of the pdf ?)

matthewfeickert:
> I need to review the TF APIs, since we might want to switch to eager execution entirely.

My understanding was that the default in TF 2.0 is eager execution, but that lazy execution will still be supported for those that want it.

matthewfeickert:
Closed in favor of PR #541

lukasheinrich:
hi @matthewfeickert,

I think we should revive this, since we now have a good path to using `scipy.optimize` for all backends and can reproduce the sbottom lhood result with tf,torch and numpy, which is very good to have and I expect the TF 2.0 migration to take a bit of time.

Can you rebase this see if you can get it passing? Why did we need greater and less? In my tests only adding `cond(pred,true_fn,false_fn)` to the backends was sufficient

matthewfeickert:
> Can you rebase this see if you can get it passing? 

Done

> Why did we need greater and less? In my tests only adding `cond(pred,true_fn,false_fn)` to the backends was sufficient

We don't need them explicitly. I think during this (maybe while writing the docstrings?) or PR #534 that they would be a common thing to want if the user wanted to write any sort of code that resembled

```python
>>> import pyhf
>>> tensorlib = pyhf.tensorlib
>>> tensorlib
<pyhf.tensor.pytorch_backend.pytorch_backend object at 0x7f6e0e99ce50>
>>> a = tensorlib.astensor([4])
>>> b = tensorlib.astensor([5])
>>> tensorlib.conditional(
...     tensorlib.less(a, b)[0],
...     lambda: tensorlib.add(a, b),
...     lambda: tensorlib.subtract(a, b)
... )
tensor([9.])
```

and that they seemed like an easy and simple extension to the API.

lukasheinrich:
I'd probably opt for just adding the minimal set since that increases our chances of adding new backends.. so maybe we go just for cond?

PR #537, opened 2019-09-02
Use conditional control flow in 'pvals_from_teststat'

matthewfeickert:
# Description

Conditional on PR #535 and in support of PR #534, use a conditional control flow in `pvals_from_teststat` to avoid a TensorFlow error of

```python-traceback
```

in the `if sqrtqmu_v < sqrtqmuA_v` of 

https://github.com/diana-hep/pyhf/blob/f3b0680c115ef64cfb61a7631d9a550ec50c095f/src/pyhf/utils.py#L158-L169

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use a conditional control flow in `pvals_from_teststat` to avoid a TensorFlow TypeError related to using a tf.Tensor as a bool
```

### DIFF
--- a/src/pyhf/tensor/mxnet_backend.py
+++ b/src/pyhf/tensor/mxnet_backend.py
@@ -42,7 +42,7 @@ class mxnet_backend(object):
 
     def subtract(self, tensor_in_1, tensor_in_2):
         """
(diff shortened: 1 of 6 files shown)


matthewfeickert:
Closed in favor of PR #541

PR #538, opened 2019-09-03
Add 'tile' method to tensorlib API

lukasheinrich:
# Description

broken out of #519 to add tile method

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add 'tile' method to the NumPy, TensorFlow, PyTorch, and MxNet tensor backends
* Add tests for 'tile' and 'normal_logpdf'
```

### DIFF
--- a/tests/test_tensor.py
+++ b/tests/test_tensor.py
@@ -60,6 +60,8 @@ def test_normal(backend):
     tb = pyhf.tensorlib
     assert tb.tolist(
         tb.normal_logpdf(tb.astensor([0]), tb.astensor([0]), tb.astensor([1]))
+    ) == pytest.approx(
+        [-0.9189385332046727], 1e-07
     )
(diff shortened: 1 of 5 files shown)


tests/test_tensor.py
```diff
@@ -56,6 +56,13 @@ def test_ones(backend):
     assert tb.tolist(tb.ones((4, 5))) == [[1.0] * 5] * 4
 
 
+def test_normal(backend):
```
lukasheinrich: this seems to have slipped in but I guess we can just add this test

matthewfeickert: What exactly is this testing though? It just is asserting the call, but not against any target. Is this what we want?

lukasheinrich: added an acfual value to test against

PR #546, opened 2019-09-05
Enable full batch evaluation

lukasheinrich:
# Description

this builds on PR #519 to enable full batch evaluation

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add ability to do batched evaluation
* pdf.logpdf(pars, data), where pars.shape = (B,N) with B==pdf.batch_size and N==len(pdf.config.suggested_config())
* Add test to expose bug in constrained_by_poisson.expected_data
* Add fix for bug in constrained_by_poisson.expected_data
* Relax tolerance in validation (suspect it's due to order of moving python arrays to tensors)
* Add arg in hepdata_like to return batched model
* Rename 'thenom' to 'nominal_rates'
```

### DIFF
--- a/src/pyhf/paramsets.py
+++ b/src/pyhf/paramsets.py
@@ -38,7 +38,9 @@ class constrained_by_poisson(paramset):
         sh = tensorlib.shape(pars)
         # if batched, tile the factors as they only depend on background uncertainty
         if len(sh) > 1:
-            fc = tensorlib.tile(self.factors, (sh[0], 1))
+            fc = tensorlib.tile(
+                tensorlib.reshape(tensorlib.astensor(self.factors), (1, -1)), (sh[0], 1)
+            )
             return pars * fc
         return tensorlib.product(
             tensorlib.stack([pars, tensorlib.astensor(self.factors)]), axis=0
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -478,8 +478,6 @@ class Model(object):
         tensorlib, _ = get_backend()
         lambdas_data = self.expected_actualdata(pars)
         summands = tensorlib.poisson_logpdf(maindata, lambdas_data)
-        tosum = tensorlib.boolean_mask(summands, tensorlib.isfinite(summands))
-        mainpdf = tensorlib.sum(tosum)
         if self.batch_size is None:
             return tensorlib.sum(summands, axis=0)
         return tensorlib.sum(summands, axis=1)
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -216,6 +216,12 @@ class Model(object):
             if mod.op_code == 'addition'
         ]
 
+        self.viewer_aux = ParamViewer(
+            (self.batch_size or 1, len(self.config.suggested_init())),
+            self.config.par_map,
+            self.config.auxdata_order,
+        )
+
     def _create_nominal_and_modifiers(self):
         default_data_makers = {
             'histosys': lambda: {
@@ -368,15 +374,10 @@ class Model(object):
     def expected_auxdata(self, pars):
         tensorlib, _ = get_backend()
         auxdata = None
-        viewer = ParamViewer(
-            (self.batch_size or 1, len(self.config.suggested_init())),
-            self.config.par_map,
-            self.config.auxdata_order,
-        )
-        if not viewer.index_selection:
+        if not self.viewer_aux.index_selection:
             return None
-        slice_data = viewer.get(pars)
-        for parname, sl in zip(self.config.auxdata_order, viewer.slices):
+        slice_data = self.viewer_aux.get(pars)
+        for parname, sl in zip(self.config.auxdata_order, self.viewer_aux.slices):
             # order matters! because we generated auxdata in a certain order
             thisaux = self.config.param_set(parname).expected_data(
                 tensorlib.einsum('ij->ji', slice_data[sl])
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -332,20 +332,20 @@ class Model(object):
 
         self.mega_mods = mega_mods
 
-        thenom = default_backend.astensor(
+        nominal_rates = default_backend.astensor(
             [mega_samples[s]['nom'] for s in self.config.samples]
         )
-        self._thenom = default_backend.reshape(
-            thenom,
+        self._nominal_rates = default_backend.reshape(
+            nominal_rates,
             (
-                1,  # modifier dimension.. thenom is the base
+                1,  # modifier dimension.. nominal_rates is the base
                 len(self.config.samples),
                 1,  # alphaset dimension
                 sum(list(self.config.channel_nbins.values())),
             ),
         )
-        self._thenom = default_backend.tile(
-            self._thenom, (1, 1, self.batch_size or 1, 1)
+        self._nominal_rates = default_backend.tile(
+            self._nominal_rates, (1, 1, self.batch_size or 1, 1)
         )
 
         self.modifiers_appliers = {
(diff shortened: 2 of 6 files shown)


src/pyhf/pdf.py
```diff
@@ -325,11 +332,10 @@ def _create_nominal_and_modifiers(self):
 
         self.mega_mods = mega_mods
 
-        tensorlib, _ = get_backend()
         thenom = default_backend.astensor(
             [mega_samples[s]['nom'] for s in self.config.samples]
         )
-        self.thenom = default_backend.reshape(
+        self._thenom = default_backend.reshape(
             thenom,
```
matthewfeickert: Can we replace all instances of `thenom` and `_thennom` with something like `nominal_rate`? Having much more explicit names here helps a lot in reading the code.

matthewfeickert: Thanks! Snappy service. :)

src/pyhf/pdf.py
```diff
@@ -325,40 +332,59 @@ def _create_nominal_and_modifiers(self):
...
+
+    def _precompute(self):
+        tensorlib, _ = get_backend()
+        self.nominal_rates = tensorlib.astensor(self._nominal_rates)
 
     def expected_auxdata(self, pars):
         tensorlib, _ = get_backend()
         auxdata = None
-        for parname in self.config.auxdata_order:
+        v = ParamViewer(
```
matthewfeickert: Would generally prefer to not have single letter variables.

```suggestion
        view = ParamViewer(
```

matthewfeickert:
This is pretty nice given that the tests were already added in PR #519

src/pyhf/pdf.py
```diff
@@ -432,21 +462,27 @@ def expected_data(self, pars, include_auxdata=True):
...
 
     def constraint_logpdf(self, auxdata, pars):
         tensorlib, _ = get_backend()
         normal = self.constraints_gaussian.logpdf(auxdata, pars)
         poisson = self.constraints_poisson.logpdf(auxdata, pars)
-        return tensorlib.sum(tensorlib.stack([normal, poisson]), axis=0)
+        if self.batch_size is None:
+            return normal + poisson
+        terms = tensorlib.stack([normal, poisson])
+        result = tensorlib.sum(terms, axis=0)
```
matthewfeickert: This is style not substance, but could follow the return structure style of above and do

```suggestion
        return tensorlib.sum(terms, axis=0)
```

with removal of `result` below.

tests/test_validation.py
```diff
@@ -710,7 +710,7 @@ def validate_hypotest(pdf, data, mu_test, expected_result, tolerance=1e-6):
     'setup_and_tolerance',
     [
         (setup_1bin_shapesys(), 1e-6),
-        (setup_1bin_lumi(), 1e-6),
+        (setup_1bin_lumi(), 4e-6),
```
matthewfeickert: Would be good to have a comment in the main body of the PR about why the tolerance needed to be changed here. I don't think we need a full study, just a comment about why it was changed.

tests/test_pdf.py
```diff
@@ -29,6 +29,38 @@ def test_pdf_inputs(backend):
...
+    }
+    pdf = pyhf.simplemodels.hepdata_like(
+        source['bindata']['sig'], source['bindata']['bkg'], source['bindata']['bkgerr']
+    )
+
+    pars = pdf.config.suggested_init()
+    data = source['bindata']['data'] + pdf.config.auxdata
+
+    tensorlib, _ = backend
+    assert tensorlib.tolist(pdf.pdf(pars, data))
```
matthewfeickert: Instead of just asserting the existence of the call, why not also assert against target values?

src/pyhf/pdf.py
lukasheinrich: this needs to be propagated to where we use `v` if we want to change the name

tests/test_validation.py
lukasheinrich: done

src/pyhf/pdf.py
matthewfeickert: Yes, would be a 

```
%s#v\.#view.#g
```

GitHub's suggestions only allow you to make a suggestion on a single line, not across multiple (I think).

tests/test_pdf.py
```diff
@@ -29,6 +29,38 @@ def test_pdf_inputs(backend):
...
+    pdf = pyhf.simplemodels.hepdata_like(
+        source['bindata']['sig'], source['bindata']['bkg'], source['bindata']['bkgerr']
+    )
+
+    pars = pdf.config.suggested_init()
+    data = source['bindata']['data'] + pdf.config.auxdata
+
+    tensorlib, _ = backend
+    assert tensorlib.tolist(pdf.pdf(pars, data)) == pytest.approx([0.002417118], 1e-05)
+    assert tensorlib.tolist(pdf.expected_data(pars))
```
matthewfeickert: In the same vein as the previous comment about comparing to values, this would be good to have across all of this test. I can add this in though.

src/pyhf/paramsets.py
```diff
@@ -38,7 +38,9 @@ def expected_data(self, pars):
         sh = tensorlib.shape(pars)
         # if batched, tile the factors as they only depend on background uncertainty
         if len(sh) > 1:
-            fc = tensorlib.tile(self.factors, (sh[0], 1))
+            fc = tensorlib.tile(
+                tensorlib.reshape(tensorlib.astensor(self.factors), (1, -1)), (sh[0], 1)
```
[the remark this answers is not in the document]
lukasheinrich: `factors` is `(N,)` and we want it to be something like `(B,N)`. So we first reshape as `(1,N)` and then tile by with `(B,1)` multiples to get `(B,N)`. We missed the reshape before, which wasn't covered by tests.

src/pyhf/pdf.py
```diff
@@ -325,40 +332,59 @@ def _create_nominal_and_modifiers(self):
...
                 mega_mods,
+                batch_size=self.batch_size,
                 **self.config.modifier_settings.get(k, {})
             )
             for k, c in modifiers.combined.items()
         }
+        self._precompute()
+        events.subscribe('tensorlib_changed')(self._precompute)
+
+    def _precompute(self):
```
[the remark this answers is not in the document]
lukasheinrich: this might create a new tensor everytime `nominal_rates` is accessed. So I'd rather to this once in `precompute`

src/pyhf/pdf.py
```diff
@@ -406,7 +432,9 @@ def expected_actualdata(self, pars):
 
         deltas, factors = self._modifications(pars)
 
-        allsum = tensorlib.concatenate(deltas + [tensorlib.astensor(self.thenom)])
+        allsum = tensorlib.concatenate(
+            deltas + [tensorlib.astensor(self.nominal_rates)]
```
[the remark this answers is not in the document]

src/pyhf/pdf.py
```diff
@@ -325,40 +332,59 @@ def _create_nominal_and_modifiers(self):
...
     def expected_auxdata(self, pars):
         tensorlib, _ = get_backend()
         auxdata = None
-        for parname in self.config.auxdata_order:
+        viewer = ParamViewer(
+            (self.batch_size or 1, len(self.config.suggested_init())),
+            self.config.par_map,
+            self.config.auxdata_order,
+        )
+        if not viewer.index_selection:
```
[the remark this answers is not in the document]
lukasheinrich: consider models without constraints

src/pyhf/pdf.py
```diff
@@ -173,7 +175,8 @@ def _create_and_register_paramsets(
 
 
 class Model(object):
-    def __init__(self, spec, **config_kwargs):
+    def __init__(self, spec, batch_size=None, **config_kwargs):
+        self.batch_size = batch_size
```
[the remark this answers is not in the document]
lukasheinrich: this is not necessarily a property of the pdf as a whole but rather of the constraint pdf. So I'd rather keep this very local to where we need it

lukasheinrich:
I think, right now, I consider the batch size to be a property of the model instance. This is also how scipy / tensorflow / pytorch do it. (you initialize with possibly batched parameters, and then can evaluate e.g. logpdf. This is anticipating some changes in actually making our Model a subclass of these types of classes.

PR #549, opened 2019-09-06
Rewrite Model into PDF composition of MainModel and ConstrainedModel

lukasheinrich:
# Description

NB: This is progress towards Statisfactory.

This is a cute refactoring that, only through moving some code blocks around, basically puts the "main model" of histfactory on equal footing with the constraint pdfs.

i.e. a histfactory model can be viewed as a "joint pdf" of three subpdfs

`p(data,pars) =  p_main(maindata, pars) * p_gaus(gausdata,pars) * p_pois(poisdata,pars)`

```
        self.main_model = MainModel(
            self.config,
            mega_mods=mega_mods,
            nominal_rates=_nominal_rates,
            batch_size=self.batch_size,
        )
     
        ... snip ...

        self.constraints_gaussian = gaussian_constraint_combined(
            self.config, batch_size=self.batch_size
        )
        self.constraints_poisson = poisson_constraint_combined(
            self.config, batch_size=self.batch_size
        )
```

each of which provides the `logpdf` interface, also it decouples exactly how the ingredients of the  main model get created. possibilities are

* from JSON
* from some other source, e.g. columnar analysis / differentiable upstream

this will connect with the early work on #501 (simultaneous.py)

Commit message

```
* introduce _MainModel and _ConstraintModel as sub pdf object
* prepares towards integration with prob prog ML libraries and statisfactory
* keeps around forwarding methods to be non-API breaking, expected to be remove when convenient
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -174,7 +174,25 @@ class _ModelConfig(object):
             self._register_paramset(param_name, paramset)
 
 
-class MainModel(object):
+class _ConstraintModel(object):
+    def __init__(self, gaussian, poisson, batch_size):
+        self.batch_size = batch_size
+        self.constraints_gaussian = gaussian
+        self.constraints_poisson = poisson
+        assert self.constraints_gaussian.batch_size == self.batch_size
+        assert self.constraints_poisson.batch_size == self.batch_size
+
+    def logpdf(self, auxdata, pars):
+        tensorlib, _ = get_backend()
+        normal = self.constraints_gaussian.logpdf(auxdata, pars)
+        poisson = self.constraints_poisson.logpdf(auxdata, pars)
+        if self.batch_size is None:
+            return normal + poisson
+        terms = tensorlib.stack([normal, poisson])
+        return tensorlib.sum(terms, axis=0)
+
+
+class _MainModel(object):
     def __init__(self, config, mega_mods, nominal_rates, batch_size):
         self._factor_mods = [
             modtype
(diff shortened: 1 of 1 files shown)


src/pyhf/pdf.py
```diff
@@ -205,24 +318,13 @@ def __init__(self, spec, batch_size=None, **config_kwargs):
...
-        ]
-
         self.viewer_aux = ParamViewer(
             (self.batch_size or 1, len(self.config.suggested_init())),
             self.config.par_map,
             self.config.auxdata_order,
         )
 
-    def _create_nominal_and_modifiers(self):
+    def _create_nominal_and_modifiers(self, config, spec):
```
kratsg: why did this get refactored to pass in config/spec?

lukasheinrich: I would like this to be at some point a "factory method" either as a `classmethod` or `staticmethod`..together with`_ModelConfig()`


if you look at the init we're almost at the point where

```
        self.config = _ModelConfig(self.spec, **config_kwargs)
        mega_mods, _nominal_rates = self._create_nominal_and_modifiers(
            self.config, self.spec
        )
        self.main_model = _MainModel(
            self.config,
            mega_mods=mega_mods,
            nominal_rates=_nominal_rates,
            batch_size=self.batch_size,
        )

       ... snip ...

        self.constraint_model = _ConstraintModel(
            config=self.config, batch_size=self.batch_size
        )
```

I'd like to be able to have multiple ctors .. one being from a JSON spec

lukasheinrich: note that `self` is not used at all in the body

lukasheinrich: screenshot

i.e. not how in`Model` all but `_create_nominal_and_modifiers` are either temporary forwarding functions to the  constituent models or functions that we consider to be part of the emerging statisfactory model  api

matthewfeickert:
@lukasheinrich We need a commit log summary

kratsg:
> @lukasheinrich We need a commit log summary

He added one. We're golden.

Issue #543, opened 2019-09-05
consolidate parameter handling


lukasheinrich:
# Description

we have `paramsets` and `paramviewer` and we will need some more metadata on parameters (which ones are constant, which ones variables) etc in the future. all of this could probably be consolidated in a set of related objects that replace the parameter tensor going into the pdf

PR #550, opened 2019-09-06
Refactor paramsets/paramview into parameters

kratsg:
# Description

It's a first step towards #543. This PR adds the `parameters` sub-package such that `pyhf.parameters` gives us the following:

- unconstrained
- constrained_by_normal
- constrained_by_poisson
- ParamViewer
- utils.reduce_paramsets_requirements

which cleans up our namespace a little bit and is a simple reorganization.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* cleans up paramsets/paramview namespace by putting it under parameters
```


src/pyhf/parameters/__init__.py
```diff
@@ -0,0 +1,21 @@
+import logging
+
+log = logging.getLogger(__name__)
```
lukasheinrich: what are we logging here?

src/pyhf/pdf.py
```diff
@@ -8,8 +8,8 @@
 from . import utils
 from . import events
 from .constraints import gaussian_constraint_combined, poisson_constraint_combined
-from .paramsets import reduce_paramsets_requirements
-from .paramview import ParamViewer
+from .parameters import ParamViewer
+from .parameters.utils import reduce_paramsets_requirements
```
lukasheinrich: why not add this to `__all__`?

src/pyhf/parameters/__init__.py
kratsg: removed.

src/pyhf/pdf.py
kratsg: i thought about this for a while. It occurred to me that this will probably be absorbed during another refactor, but I can put it under __all__.

kratsg:
> two minor questions / comments

resolved

PR #551, opened 2019-09-06
Add probability module

lukasheinrich:
# Description

another step towards statisfactory. this replaces `log_prob(data,pars)` calls directly to tensorlib with objects that mirror the API of `torch.distributions.Distribution`

see e.g. test (which should probably be added here)

https://gist.github.com/lukasheinrich/c69d89333d1c5bd6ea6a0a78eacc919b

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* adds pyhf.probability module
* pyhf.probability : stats classes of backends  :: tensorlib : array classes of backends
```


src/pyhf/probability.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def log_prob(self, value):
+        tensorlib, _ = get_backend()
+        return tensorlib.normal_logpdf(value, self.mu, self.sigma)
+
+
+class Independent(object):
+    '''
+    A probability density corresponding to the joint
+    likelihood of a batch of identically distributed random
```
matthewfeickert: Shouldn't this be joint distribution as this is frequentist stats, so the pdf and likelihoods are different (in their meaning)?

```suggestion
    distribution of a batch of identically distributed random
```

src/pyhf/probability.py
```diff
@@ -0,0 +1,48 @@
+from . import get_backend
+
+
+class Poisson(object):
```
matthewfeickert: I can add docstrings here, which are needed.

src/pyhf/probability.py
```diff
@@ -0,0 +1,48 @@
...
+    numbers.
+    '''
+
+    def __init__(self, batched_pdf, batch_size=None):
+        self.batch_size = batch_size
+        self._pdf = batched_pdf
+
+    def log_prob(self, value):
+        tensorlib, _ = get_backend()
+        result = self._pdf.log_prob(value)
```
matthewfeickert: Perhaps instead of reusing the same variable twice just change it to be more explicit here and below?

```suggestion
        _log_prob = self._pdf.log_prob(value)
        return tensorlib.sum(_log_prob, axis=-1)
```

src/pyhf/probability.py
```diff
@@ -0,0 +1,48 @@
...
+        result = tensorlib.sum(result, axis=-1)
+        return result
+
+
+def joint_logpdf(terms):
+    tensorlib, _ = get_backend()
+    terms = tensorlib.stack(terms)
+    result = tensorlib.sum(terms, axis=0)
```
matthewfeickert: As we don't use `result` maybe just `return` here?

```suggestion
    return tensorlib.sum(terms, axis=0)
```

src/pyhf/constraints.py
```diff
@@ -58,10 +59,13 @@ def __init__(self, pdfconfig, batch_size=None):
...
 
             _normal_sigmas = default_backend.concatenate(normal_constraint_sigmas)
-            sigmas = default_backend.reshape(_normal_sigmas, (1, -1))  # (1, normals)
-            self._batched_sigmas = default_backend.tile(
-                sigmas, (self.batch_size or 1, 1)
-            )
+            if self.batch_size:
+                sigmas = default_backend.reshape(
+                    _normal_sigmas, (1, -1)
+                )  # (1, normals)
```
matthewfeickert: IIRC from previous PRs, this is a Lukas to self comment that can get removed?

matthewfeickert:
I removed the docstrings I added and migrated them to PR #558

src/pyhf/probability.py
matthewfeickert: This will be done now in PR #558

matthewfeickert:
We also need a summary commit message for the PR

lukasheinrich:
agreed. tests added

src/pyhf/constraints.py

src/pyhf/constraints.py
```diff
@@ -58,10 +59,11 @@ def __init__(self, pdfconfig, batch_size=None):
             )
 
             _normal_sigmas = default_backend.concatenate(normal_constraint_sigmas)
-            sigmas = default_backend.reshape(_normal_sigmas, (1, -1))  # (1, normals)
-            self._batched_sigmas = default_backend.tile(
-                sigmas, (self.batch_size or 1, 1)
-            )
+            if self.batch_size:
+                sigmas = default_backend.reshape(_normal_sigmas, (1, -1))
+                self._sigmas = default_backend.tile(sigmas, (self.batch_size or 1, 1))
```
kratsg: should just be `self.batch_size` since you know it's not `None`

src/pyhf/constraints.py
```diff
@@ -75,10 +77,16 @@ def _precompute(self):
         if not self.param_viewer.index_selection:
             return
         tensorlib, _ = get_backend()
-        self.batched_sigmas = tensorlib.astensor(self._batched_sigmas)
+        self.sigmas = tensorlib.astensor(self._sigmas)
         self.normal_data = tensorlib.astensor(self._normal_data, dtype='int')
         self.access_field = tensorlib.astensor(self._access_field, dtype='int')
 
+    def _dataprojection(self, auxdata):
```
kratsg: we should have a comment that this returns only the auxiliary data associated with the modifier -- and you should be calling the signature `def _dataproject(self, data)` since you normally pass in all data, and extract the auxdata from it

src/pyhf/pdf.py
```diff
@@ -211,18 +212,21 @@ def expected_data(self, pars):
             return auxdata[0]
         return auxdata
 
+    def _dataprojection(self, data):
+        tensorlib, _ = get_backend()
+        cut = tensorlib.shape(data)[0] - len(self.config.auxdata)
```
kratsg: is this meant to only return the `auxdata`? I guess you might want to... add a comment on what this function does and test it separately.

src/pyhf/pdf.py
```diff
@@ -531,16 +538,20 @@ def logpdf(self, pars, data):
...
             pars, data = tensorlib.astensor(pars), tensorlib.astensor(data)
-            cut = tensorlib.shape(data)[0] - len(self.config.auxdata)
-            actual_data, aux_data = data[:cut], data[cut:]
 
-            mainpdf = self.mainlogpdf(actual_data, pars)
-            constraint = self.constraint_logpdf(aux_data, pars)
+            actual_data = self.main_model._dataprojection(data)
+            aux_data = self.constraint_model._dataprojection(data)
+
+            mainpdf = self.main_model.logpdf(actual_data, pars)
```
kratsg: thinking about whether `main_model.logpdf` should call `_dataprojection` or if `pyhf/pdf.py` should. You did originally call it as `_dataprojection` which implies you only ever expected it to be an internal call, but you're using it as a non-internal call here. It does certainly look cleaner to do

```
mainpdf = self.main_model.logpdf(data, pars)
constraintpdf = self.constraint_model.logpdf(data, pars)
```

Additionally, please call it `constraintpdf` and not `constraint`.

src/pyhf/pdf.py
```diff
@@ -531,16 +538,20 @@ def logpdf(self, pars, data):
...
+            constraint = self.constraint_model.logpdf(aux_data, pars)
+
+            result = prob.joint_logpdf([mainpdf, constraint])
```
kratsg: this almost gets us to the point of doing

```
jointpdf = prob.joint([self.main_model, self.constraint_model])(data, pars)
result = jointpdf.logpdf()
```

src/pyhf/pdf.py
```diff
@@ -531,16 +538,20 @@ def logpdf(self, pars, data):
...
 
-            result = tensorlib.sum(tensorlib.stack([mainpdf, constraint]), axis=0)
-            if not self.batch_size:
+            if (
```
kratsg:
```suggestion:-0+5
            return result if self.batch_size else tensorlib.reshape(result, (1,))
```

src/pyhf/probability.py
```diff
@@ -0,0 +1,46 @@
...
+        self.batch_size = batch_size
+        self._pdf = batched_pdf
+
+    def log_prob(self, value):
+        tensorlib, _ = get_backend()
+        _log_prob = self._pdf.log_prob(value)
+        return tensorlib.sum(_log_prob, axis=-1)
+
+
+def joint_logpdf(terms):
```
kratsg: Create a `Joint` class that handles `log_prob` correctly. This probably is the only thing missing.

src/pyhf/pdf.py
lukasheinrich: exactly. that's in the upcoming PRs

src/pyhf/pdf.py
lukasheinrich: this will get dropped anyways in the subsequent PRs

src/pyhf/probability.py
lukasheinrich: this is happening in subsequent PRs :)

src/pyhf/constraints.py
lukasheinrich: `_dataprojection` will be removed in subsequent PRs

src/pyhf/pdf.py
lukasheinrich: yes `_dataprojection` is meant to just extract the chunk of data relevant for this subpdf. in #553 this will be replaced with `TensorViewer.split`

src/pyhf/pdf.py

src/pyhf/constraints.py

PR #553, opened 2019-09-08
feat: Simultaneous pdf class

lukasheinrich:
# Description

a demo PR against `add_probability_module` #551 showing how we will arrive at a pyfitcore like API

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* extends pyhf.probability to have a joint pdf class pyhf.Simultaneous
* adapts pdf evaluation code to be thing layer around the pdf(pars).log_prob(data) pattern.
```

PR #556, opened 2019-09-08
feat: extend pdf API with expected_data method

lukasheinrich:
# Description

* patch against #553 that uses extends the basic pdf API to include expectation values and thus cleans up a lot of `expected_data` calls
* should still be passing tests and not break any APIs

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -45,8 +45,8 @@ Making Models from PDFs
    Model
    _ModelConfig
 
-Tensor Backends
----------------
+Backends
+--------
 
 The computational backends that :code:`pyhf` provides interfacing for the vector-based calculations.
 
@@ -61,7 +61,6 @@ The computational backends that :code:`pyhf` provides interfacing for the vector
    numpy_backend.numpy_backend
    pytorch_backend.pytorch_backend
    tensorflow_backend.tensorflow_backend
-   common.TensorViewer
 
 Optimizers
 ----------
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -45,8 +45,8 @@ Making Models from PDFs
    Model
    _ModelConfig
 
-Backends
---------
+Tensor Backends
+---------------
 
 The computational backends that :code:`pyhf` provides interfacing for the vector-based calculations.
 
@@ -61,6 +61,7 @@ The computational backends that :code:`pyhf` provides interfacing for the vector
    numpy_backend.numpy_backend
    pytorch_backend.pytorch_backend
    tensorflow_backend.tensorflow_backend
+   common.TensorViewer
 
 Optimizers
 ----------
--- a/src/pyhf/constraints.py
+++ b/src/pyhf/constraints.py
@@ -81,20 +81,7 @@ class gaussian_constraint_combined(object):
         self.normal_data = tensorlib.astensor(self._normal_data, dtype='int')
         self.access_field = tensorlib.astensor(self._access_field, dtype='int')
 
-    def _dataprojection(self, auxdata):
-        tensorlib, _ = get_backend()
-        auxdata = tensorlib.astensor(auxdata)
-        normal_data = tensorlib.gather(auxdata, self.normal_data)
-        return normal_data
-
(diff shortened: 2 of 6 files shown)


docs/api.rst
```diff
@@ -61,6 +61,7 @@ The computational backends that :code:`pyhf` provides interfacing for the vector
    numpy_backend.numpy_backend
    pytorch_backend.pytorch_backend
    tensorflow_backend.tensorflow_backend
+   common.TensorViewer
```
matthewfeickert: `TensorViewer` is in the public API and so should be documented as such. I actually think that it would be better to have this documentation be its own PR, but I thought I'd raise it here and if you agree then this can get dropped and it can get spun off into its own thing.

Alternatively, does `TensorViewer` need to be in the public API? If not, then can we just rename it to `_TensorViewer` and document it in the developer API later?

matthewfeickert:
This should unfortunately also wait to go in until the benchmarks in master are fixed and PR #570 goes in.

src/pyhf/constraints.py
```diff
@@ -116,10 +110,9 @@ def make_pdf(self, pars):
         if self.batch_size is None:
             normal_means = normal_means[0]
 
-        result = prob.Independent(
+        return prob.Independent(
             prob.Normal(normal_means, self.sigmas), batch_size=self.batch_size
         )
-        return result
 
     def logpdf(self, auxdata, pars):
```
kratsg: i didn't notice it before, but `logpdf` could be rewritten to just be `logpdf(self, pdf, auxdata)` since the only thing that `pars` is used for is to make the pdf `self.make_pdf(pars)`.

src/pyhf/pdf.py
```diff
@@ -549,7 +499,7 @@ def _create_nominal_and_modifiers(self, config, spec):
         return mega_mods, _nominal_rates
 
     def expected_auxdata(self, pars):
-        return self.constraint_model.expected_data(pars)
+        return self.make_pdf(pars).pdfobjs[1].expected_data()
```
kratsg: why did this get changed? `self.constraint_model` still works here, or have I missed something...

src/pyhf/pdf.py
```diff
@@ -559,30 +509,18 @@ def nominal_rates(self):
         return self.main_model.nominal_rates
 
     def expected_actualdata(self, pars):
-        return self.main_model.expected_data(pars)
+        return self.make_pdf(pars).pdfobjs[0].expected_data()
```
kratsg: why did this get changed here? `self.main_model` still works here, or have i missed something...

matthewfeickert:
> * TensorViewer needs to be out of the public API, so prefix with `_`

Okay, so we agree that `TensorViewer` should not be in the public API. So I'll go ahead and revert https://github.com/diana-hep/pyhf/pull/556/commits/6bc898446719b67d827fd801d1272498ede56734.

src/pyhf/pdf.py
lukasheinrich: I removed logpdf from self.main_model

src/pyhf/pdf.py
lukasheinrich: I removed logpdf from self.constraint_model

src/pyhf/constraints.py
lukasheinrich: yes, eventually this method would get dropped like the methods for _ConstraintModel and _MainModel. However I wanted this PR to be pure refactor without changing tests, where this signature is used

docs/api.rst
lukasheinrich: sounds good

matthewfeickert: Yeah, I renamed it to `_TensorViewer` in https://github.com/diana-hep/pyhf/pull/556/commits/77be0d307a6b1706009a1100efe4bcdc606381cb (as you've already seen).

matthewfeickert:
This PR is getting put on hold until we can fix the regression described in Issue #572

lukasheinrich:
closing in favor of #558

PR #558, opened 2019-09-09
feat: Add toy sampling

lukasheinrich:
# Description

toys

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add toy sampling capability
* Remove projection methods from constraint
* Add generalized tensorviewer (expect to merge w/ ParamViewer)
* Introduce make_pdf method to construct pdf with frozen parameters, evaluate with log_prob
* Add Poisson and Normal classes to Numpy backend with same API as PyTorch/Tensorflow
```


matthewfeickert:
I migrated docstrings from PR #551, but I'll come back to this later.

tests/test_toys.py
```diff
@@ -0,0 +1,14 @@
...
+def test_smoketest_toys(backend):
+    tb, _ = backend
+    m = pyhf.simplemodels.hepdata_like([6], [9], [3])
+    s = m.make_pdf(pyhf.tensorlib.astensor(m.config.suggested_init()))
+    assert np.asarray(tb.tolist(s.log_prob(s.sample((1000,))))).shape == (1000,)
+
+    tb, _ = backend
+    m = pyhf.simplemodels.hepdata_like([6, 6], [9, 9], [3, 3], batch_size=13)
+    s = m.make_pdf(pyhf.tensorlib.astensor(m.batch_size * [m.config.suggested_init()]))
+    np.asarray(tb.tolist(s.sample((10,)))).shape == (10, 13, 4)
```
kratsg: missing `assert` here (thanks CodeFactor!)

matthewfeickert: Fixed

src/pyhf/probability.py
```diff
@@ -1,40 +1,54 @@
...
+
+    def sample(self, sample_shape=()):
+        return self._pdf.sample(sample_shape)
+
+
+class Poisson(_SimpleDistributionMixin):
     def __init__(self, rate):
+        tensorlib, _ = get_backend()
         self.rate = rate
+        self._pdf = tensorlib.poisson_pdfcls(rate)
```
matthewfeickert: I realize this is going to be used to assign a private attribute, but I would suggest a more clear name than `poisson_pdfcls`. Maybe I'm missing something very obvious, but why the `cls`?

src/pyhf/probability.py
```diff
@@ -1,40 +1,54 @@
...
+        return self.rate
 
 
-class Normal(object):
+class Normal(_SimpleDistributionMixin):
     def __init__(self, loc, scale):
+        tensorlib, _ = get_backend()
         self.loc = loc
         self.scale = scale
+        self._pdf = tensorlib.normal_pdfcls(loc, scale)
```
matthewfeickert: Same comment as `poisson_pdfcls`.

src/pyhf/pdf.py
```diff
@@ -588,30 +578,20 @@ def nominal_rates(self):
...
-        )
-        if self.batch_size is None:
-            return tensorlib.concatenate(tocat, axis=0)
-        return tensorlib.concatenate(tocat, axis=1)
+            return self.make_pdf(pars).pdfobjs[0].expected_data()
+        return self.make_pdf(pars).expected_data()
 
     def constraint_logpdf(self, auxdata, pars):
-        return self.constraint_model.logpdf(auxdata, pars)
+        return self.make_pdf(pars).pdfobjs[1].log_prob(auxdata)
```
matthewfeickert: I think this signature is _really_ nice! :) Thanks for making this possible.

src/pyhf/probability.py
```diff
@@ -1,40 +1,54 @@
...
+        return self._pdf.log_prob(value)
+
+    def expected_data(self):
+        return self._pdf.expected_data()
+
+    def sample(self, sample_shape=()):
+        return self._pdf.sample(sample_shape)
+
+
+class Poisson(_SimpleDistributionMixin):
```
matthewfeickert: This is in the public API and so needs docstrings. I had already written these for this PR but it seems that they got removed in the the squash that was done. :( I'll add these again, but perhaps in the future this is another reason to avoid squashing PRs into 1 commit before merging.

src/pyhf/probability.py
```diff
@@ -1,40 +1,54 @@
...
     def __init__(self, rate):
+        tensorlib, _ = get_backend()
         self.rate = rate
+        self._pdf = tensorlib.poisson_pdfcls(rate)
 
-    def log_prob(self, value):
-        tensorlib, _ = get_backend()
-        return tensorlib.poisson_logpdf(value, self.rate)
+    def expected_data(self):
```
matthewfeickert: This is in the public API and so needs docstrings.

src/pyhf/probability.py
```diff
@@ -1,40 +1,54 @@
...
+        return self.rate
 
 
-class Normal(object):
+class Normal(_SimpleDistributionMixin):
```
matthewfeickert: All classes and methods in the public API needs docstrings. I'll redo these.

src/pyhf/probability.py
```diff
@@ -1,40 +1,54 @@
...
-    def log_prob(self, value):
-        tensorlib, _ = get_backend()
-        return tensorlib.normal_logpdf(value, self.loc, self.scale)
+    def expected_data(self):
+        return self.loc
 
 
-class Independent(object):
-    """
+class Independent(_SimpleDistributionMixin):
```
matthewfeickert: All classes and methods in the public API needs docstrings. I'll redo these.

src/pyhf/probability.py
```diff
@@ -1,40 +1,54 @@
...
     def log_prob(self, value):
         tensorlib, _ = get_backend()
-        _log_prob = self._pdf.log_prob(value)
-        return tensorlib.sum(_log_prob, axis=-1)
+        result = super(Independent, self).log_prob(value)
+        result = tensorlib.sum(result, axis=-1)
+        return result
 
 
 class Simultaneous(object):
```
matthewfeickert: All classes and methods in the public API needs docstrings. I'll redo these.

src/pyhf/tensor/numpy_backend.py
```diff
@@ -285,3 +311,9 @@ def normal_cdf(self, x, mu=0, sigma=1):
             NumPy float: The CDF
         """
         return norm.cdf(x, loc=mu, scale=sigma)
+
+    def poisson_pdfcls(self, rate):
```
matthewfeickert: All classes and methods in the public API needs docstrings. I will add these.

src/pyhf/tensor/pytorch_backend.py
```diff
@@ -320,3 +320,9 @@ def normal_cdf(self, x, mu=[0.0], sigma=[1.0]):
         sigma = self.astensor(sigma)
         normal = torch.distributions.Normal(mu, sigma)
         return normal.cdf(x)
+
+    def poisson_pdfcls(self, rate):
```
matthewfeickert: All classes and methods in the public API needs docstrings. I will add these.

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -441,3 +441,9 @@ def normal_cdf(self, x, mu=0, sigma=1):
         sigma = self.astensor(sigma)
         normal = tfp.distributions.Normal(mu, sigma)
         return normal.cdf(x)
+
+    def poisson_pdfcls(self, rate):
```
matthewfeickert: All classes and methods in the public API needs docstrings. I will add these.

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -102,12 +102,12 @@ def outer(self, tensor_in_1, tensor_in_2):
         tensor_in_2 = self.astensor(tensor_in_2)
         tensor_in_1 = (
             tensor_in_1
-            if tensor_in_1.dtype is not tf.bool
+            if tensor_in_1.dtype != tf.bool
```
matthewfeickert: Thanks for catching these.

matthewfeickert:
@lukasheinrich If you want to close PR #556 in favor of just doing everything here that's okay, but there are some changes that are in PR #556 that aren't propagated here. Notably that `TensorViewer` is removed from the public API. If you still want to do everything as one PR can you please do a check to make sure that everything got migrated and then close #556?

matthewfeickert:
View of the docs that are being built from this PR is here: https://pyhf-staging.readthedocs.io/en/add_probability_module_withjoint_withextended_pdf_api_withtoys/api.html#probability-distribution-functions-pdfs

src/pyhf/probability.py
lukasheinrich: we have:
* `poisson_logpdf`
* `poisson`
* `poisson_pdfcls`

so I was concerned `poisson_pdf` would be confused to be the non-log version of `poisson_logpdf` and return a number. Instead `poisson_pdfcls` returns a instance of a Poisson class

matthewfeickert: This is what I assumed, and I think it is reasonable to indicate that this is the pdf class. `cls` is okay, though I might be more inclined to just name the method `poisson_dist`, though I will concede that this no longer explicitly has `pdf` in the name.

lukasheinrich: I like `dist`

PR #564, opened 2019-09-16
fix: 1 bin lumi validation test

### LINKED ISSUE #554: Fix `1bin_lumi` test
# Description

This popped up as part of #553. The `1bin_lumi` validation model is buggy. The JSON spec is 2-bin, but the passed in data is 1-bin. We did not catch this before, but the test should be fixed and validation data updated.

See fix in 839ac67bd5aa4bbc2ef166191250f2dcba4dc872

# Expected Behavior

Describe what you thought should have happened

# Actual Behavior

Describe what did happen

# Steps to Reproduce

Detail your environment and `pyhf.__version__` and the commands that you are executing to generate the bug. If you have a large program please make a [Gist](https://gist.github.com/) and link it here. Additionally attach any screen shots as needed.

# Checklist

- [ ] Searched through existing Issues to confirm this is not a duplicate issue
- [ ] Filled out the Description, Expected Behavior, Actual Behavior, and Steps to Reproduce sections above or have edited/removed them in a way that fully describes the issue

matthewfeickert:
# Description

Resolves #554 

As `expected_result_1bin_lumi` is a 1 bin test it should not have 2 bins, which it erroneously had added in PR #352 .

This was caught by @lukasheinrich in PR #553, and is getting pulled out here.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* fix: Use 1 bin (instead of 2 bins) in the spec used in 'expected_result_1bin_lumi' to match the naming of the test
   - This revises an error in PR #352
* feat: Verify parameter and data shapes during call to pdf.logpdf
* feat: Add exception classes for invalid pdf inputs
* test: Add tests for the input shape of the pdf parameters and data
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -539,29 +539,22 @@ class Model(object):
             tensorlib, _ = get_backend()
             pars, data = tensorlib.astensor(pars), tensorlib.astensor(data)
             # Verify parameter and data shapes
-            try:
-                if pars.shape[-1] != len(self.config.suggested_init()):
-                    raise ValueError
-            except ValueError:
-                log.error(
+            if pars.shape[-1] != len(self.config.suggested_init()):
+                raise ValueError(
                     'eval failed as pars has len {} but {} was expected'.format(
                         pars.shape[-1], len(self.config.suggested_init())
                     )
                 )
-                raise
-            try:
-                if data.shape[-1] != self.nominal_rates.shape[-1] + len(
-                    self.config.auxdata
-                ):
-                    raise ValueError
-            except ValueError:
-                log.error(
+
+            if data.shape[-1] != self.nominal_rates.shape[-1] + len(
+                self.config.auxdata
+            ):
+                raise ValueError(
                     'eval failed as data has len {} but {} was expected'.format(
                         data.shape[-1],
                         self.nominal_rates.shape[-1] + len(self.config.auxdata),
                     )
                 )
-                raise
 
             actual_data = self.main_model._dataprojection(data)
             aux_data = self.constraint_model._dataprojection(data)
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -540,7 +540,7 @@ class Model(object):
             pars, data = tensorlib.astensor(pars), tensorlib.astensor(data)
             # Verify parameter and data shapes
             if pars.shape[-1] != len(self.config.suggested_init()):
-                raise ValueError(
+                raise exceptions.InvalidPdfParameters(
                     'eval failed as pars has len {} but {} was expected'.format(
                         pars.shape[-1], len(self.config.suggested_init())
                     )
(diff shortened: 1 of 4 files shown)


kratsg:
> Once this is merged it can be dropped from PR #553

Need to add assertions.

matthewfeickert:
> Need to add assertions.

What asserts do you want? `expected_result_1bin_lumi` is used in 

https://github.com/diana-hep/pyhf/blob/9b93f9d243ad71e9bde7f1428242df661be5c97d/tests/test_validation.py#L130-L136

which is used to parametrize `test_validation`

https://github.com/diana-hep/pyhf/blob/9b93f9d243ad71e9bde7f1428242df661be5c97d/tests/test_validation.py#L709-L732

where there are asserts.

tests/test_pdf.py
```diff
@@ -29,6 +29,60 @@ def test_pdf_inputs(backend):
     )
 
 
+def test_invalid_pdf_inputs():
```
matthewfeickert: A better, smaller test should be added. Just throwing this in for now. Suggestions welcome.

src/pyhf/pdf.py
```diff
@@ -538,6 +538,30 @@ def logpdf(self, pars, data):
         try:
             tensorlib, _ = get_backend()
             pars, data = tensorlib.astensor(pars), tensorlib.astensor(data)
+            # Verify parameter and data shapes
+            try:
+                if pars.shape[-1] != len(self.config.suggested_init()):
+                    raise ValueError
```
matthewfeickert: Using `ValueError` as a standin for now. Can implement a better pyhf exception later in the PR.

lukasheinrich: I would not catch the exception just rause 

```
raise ValueError('eval failed as pars has len {} but {} was expected'.format(...))

...

raise ValueError('eval failed as data has len {} but {} was expected'.format(...))
```

tests/test_pdf.py
lukasheinrich: can also use `pyhf.simplemodels`

tests/test_pdf.py
```diff
@@ -29,6 +29,60 @@ def test_pdf_inputs(backend):
...
+    if 'channels' in source:
+        data = []
+        for c in pdf.config.channels:
+            data += source['channels'][c]['bindata']['data']
+        data = data + pdf.config.auxdata
+    else:
+        data = source['bindata']['data'] + pdf.config.auxdata
+
+    with pytest.raises(ValueError):
+        pdf.logpdf(pars, data)
```
lukasheinrich: can also test the sub-cases (invalid data, invalid pars) by checking the message (if my suggestion above is implemented)

```
def test_recursion_depth():
    with pytest.raises(RuntimeError) as excinfo:

        def f():
            f()

        f()
    assert "maximum recursion" in str(excinfo.value)
```

kratsg: > can also test the sub-cases (invalid data, invalid pars) by checking the message (if my suggestion above is implemented)

Or we can just define multiple exceptions and make sure we can check them all instead of checking for messages (it's basically the same thing).

matthewfeickert: > Or we can just define multiple exceptions and make sure we can check them all instead of checking for messages (it's basically the same thing).

You had previously suggested adding custom exceptions, which I think is the better way to go. So yes, we should do this.

PR #574, opened 2019-09-20
fix: Correct warnings raised by LGTM scan

matthewfeickert:
# Description

The LGTM scans [reveal a few warnings](https://lgtm.com/projects/g/diana-hep/pyhf/alerts/?mode=list&severity=warning) that are easy to fix. This simply addresses them. 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* fix: Correct the warnings that are raised by the LGTM scans
```


matthewfeickert:
Revert is due to undoing the test to check if LGTM would pass any logic there. It does, which means it is getting confused as the logic is clearly the same.

Issue #572, opened 2019-09-19
Performance regression causes large slowdowns


matthewfeickert:
# Description

A performance regression was introduced in PR #553 that caused the benchmark tests with the Minuit optimizer to take too long for 200 bins and cause the CI systems to timeout and fail, which motivated PR #571.

The regression should be understood and fixed if possible and then the components of PR #571 that addressed this should be reverted.

# Expected Behavior

The speed of evaluation across all backends should remain unchanged from batching and simultaneous evaluation.

# Actual Behavior

The Minuit optimizer for the NumPy backend experiences significant enough slowdowns that it causes the CI systems to timeout.

# Steps to Reproduce

Run the benchmarks in the CI system.

# Checklist

- [x] Run `git fetch` to get the most up to date version of `master`
- [x] Searched through existing Issues to confirm this is not a duplicate issue
- [x] Filled out the Description, Expected Behavior, Actual Behavior, and Steps to Reproduce sections above or have edited/removed them in a way that fully describes the issue

lukasheinrich:
just to set the stage, there are a few commits that are relevant

```
86269a0 feat: Simultaneous pdf class (#553)
82a8bf0 Add probability module (#551)
657cd2f Rewrite Model into PDF composition of MainModel and ConstrainedModel (#549)
9b93f9d Enable full batch evaluation (#546)
673c1b6 Add ParamViewer class  (#519) 
c8a94f4 Add 'tile' method to tensorlib API (#538) << baseline 
```

the number of function calls according to cProfile to evaluate a pdf 100 times

```
import pyhf
import json
import cProfile
w = pyhf.Workspace(json.load(open('./s+b.json')))
m = w.model()
d = w.data(m)
cProfile.run('[m.logpdf(m.config.suggested_init(),d) for x in range(100)] ')
```

```
HEAD is now at 86269a0 feat: Simultaneous pdf class (#553)
         42304 function calls (42104 primitive calls) in 0.050 seconds
Previous HEAD position was 86269a0 feat: Simultaneous pdf class (#553)
HEAD is now at 82a8bf0 Add probability module (#551)
         34504 function calls in 0.035 seconds
Previous HEAD position was 82a8bf0 Add probability module (#551)
HEAD is now at 657cd2f Rewrite Model into PDF composition of MainModel and ConstrainedModel (#549)
         28704 function calls in 0.031 seconds
[...]

PR #576, opened 2019-09-22
fix: regression (better paramter handling in constraints and modifiers)

lukasheinrich:
# Description

this alleviates a part of the regression found in #572 

1) We know just from the spec whether we have gaussian or poissonian constraints (if at all). So we add a `has_pdf()` method to
    * `poisson_constraint_combined`
    * `gaussian_constraint_combined`
    * `_ConstraintModel`
    * `_MainModel` (here it's just a dummy returning True)

2) this allows us to pre-compute the `TensorViewer` objects before the parameter-full `make_pdf`

3) the `Simultaneous` class has been modified to 
    * receive a (precomputed) `TensorViewer` object directly
    * receive a `batch_size`, so that we can do some optimization for the batch-less case
    * `joint_logpdf` stack/sum is slower than simple sum in batch-less case

4) `TensorViewer` received `batch_size` so we can make optimizations
    * the axis shuffling with `einsum` is unnecessary in the batchless case

5)  removed a bunch of `astensor` calls. Surprisingly they're not free
6) computing `len(self.config.suggested_init())` on each call is non-negligible
    * I precompute ` len(self.config.suggested_init())` and ` len(self.config.auxdata)`

7) there was a lot or back-and-forth reshaping going on (which also is not free)
   * before all modifiers defaulted to a batched computation (reformulating the batch-less one to `batch_shape=1`)
   * now it's natively supporting the batchless case which simplifies things a bit
   * `Paramviewer.get` has been expanded by a optional argument so we can pass in a precomputed optimized index selection (used in `histosys` and `normsys`)

8) factored out `Workspace` to `workspace.py`

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
fix regression observed while moving to pyfitcore by being judicious about precomputation

* We know just from the spec whether we have gaussian or poissonian constraints (if at all). So we add a `has_pdf()` method to
* this allows us to pre-compute the `TensorViewer` objects before the parameter-full `make_pdf`* the `Simultaneous` class has been modified to 
* `TensorViewer` received `batch_size` so we can make optimizations
*  removed a bunch of `astensor` calls. Surprisingly they're not free
* computing `len(self.config.suggested_init())` on each call is non-negligible
* there was a lot or back-and-forth reshaping going on (which also is not free)
* factored out `Workspace` to `workspace.py`
```


tests/test_constraints.py
```diff
@@ -178,7 +178,12 @@ def test_batched_constraints(backend):
     suggested_pars = [1.0] * 3 + [0.0] * 5  # 2 pois 5 norm
```
matthewfeickert: Perhaps instead of wrapping `suggested_pars` multiple times in the tests

```
pyhf.tensorlib.astensor(suggested_pars)
```

instead could just do it once here?

```suggestion
    suggested_pars = pyhf.tensorlib.astensor([1.0] * 3 + [0.0] * 5)  # 2 pois 5 norm
```

src/pyhf/tensor/common.py
```diff
@@ -31,18 +33,23 @@ def stitch(self, data):
...
 
-        data = tensorlib.concatenate(data, axis=-1)
-        data = tensorlib.einsum('...j->j...', data)
-        stitched = tensorlib.gather(data, tensorlib.astensor(self.sorted_indices))
-        stitched = tensorlib.einsum('...j->j...', stitched)
-        return stitched
+        if self.batch_size is None:
+            data = tensorlib.concatenate(data, axis=-1)
+            stitched = tensorlib.gather(data, self.sorted_indices)
+            return stitched
```
matthewfeickert: Instead of multiple `returns` could just return the result of the `ifelse`
```suggestion
        else:
            data = tensorlib.concatenate(data, axis=-1)
            data = tensorlib.einsum('...j->j...', data)
            stitched = tensorlib.gather(data, self.sorted_indices)
            stitched = tensorlib.einsum('...j->j...', stitched)
        return stitched
```

src/pyhf/modifiers/lumi.py
```diff
@@ -56,17 +61,14 @@ def apply(self, pars):
         '''
         if not self.param_viewer.index_selection:
             return
+
         tensorlib, _ = get_backend()
         if self.batch_size is None:
-            batched_pars = tensorlib.reshape(pars, (1,) + tensorlib.shape(pars))
+            lumis = self.param_viewer.get(pars)
```
matthewfeickert: This is repeated in either case, so let's just move it before the `ifelse` block.

src/pyhf/modifiers/lumi.py
```diff
@@ -56,17 +61,14 @@ def apply(self, pars):
...
-
-        lumis = self.param_viewer.get(batched_pars)
-        # lumis is [(1,batch)]
-
-        # mask is (nsys, nsam, batch, globalbin)
-        results_lumi = tensorlib.einsum('msab,xa->msab', self.lumi_mask, lumis)
+            lumis = self.param_viewer.get(pars)
+            results_lumi = tensorlib.einsum('msab,xa->msab', self.lumi_mask, lumis)
 
         results_lumi = tensorlib.where(self.lumi_mask, results_lumi, self.lumi_default)
```
matthewfeickert: Just `return` here?

src/pyhf/constraints.py
```diff
@@ -83,10 +83,12 @@ def _precompute(self):
 
     def _dataprojection(self, auxdata):
         tensorlib, _ = get_backend()
-        auxdata = tensorlib.astensor(auxdata)
         normal_data = tensorlib.gather(auxdata, self.normal_data)
         return normal_data
 
+    def has_pdf(self):
```
matthewfeickert: Needs a docstring, but I can add this after we're done with all the benchmark testing so that I don't trigger unnecessarily long builds.

src/pyhf/constraints.py
```diff
@@ -219,10 +214,12 @@ def _precompute(self):
 
     def _dataprojection(self, auxdata):
         tensorlib, _ = get_backend()
-        auxdata = tensorlib.astensor(auxdata)
         poisson_data = tensorlib.gather(auxdata, self.poisson_data)
         return poisson_data
 
+    def has_pdf(self):
```
matthewfeickert: Needs a docstring, but I can add this after we're done with all the benchmark testing so that I don't trigger unnecessarily long builds.

src/pyhf/parameters/paramview.py
```diff
@@ -93,15 +93,19 @@ def slices(self):
         """
         return self._slices
 
-    def get(self, tensor):
+    def get(self, tensor, indices=None):
```
matthewfeickert: Would be good to docstring the arguments, but again, I can do this after the benchmark testing on this PR is done.

src/pyhf/pdf.py
```diff
@@ -302,6 +313,9 @@ def _precompute(self):
         tensorlib, _ = get_backend()
         self.nominal_rates = tensorlib.astensor(self._nominal_rates)
 
+    def has_pdf(self):
```
matthewfeickert: Needs a docstring, but I can add this after we're done with all the benchmark testing so that I don't trigger unnecessarily long builds.

src/pyhf/probability.py
```diff
@@ -42,17 +38,22 @@ def log_prob(self, value):
 
 
 class Simultaneous(object):
-    def __init__(self, pdfobjs, indices):
-        self.tv = TensorViewer(indices)
+    def __init__(self, pdfobjs, tv, batch_size):
```
matthewfeickert: Would prefer a more explicit argument name for the TensorViewer here than `tv`, but not the end of the world. Leaving `self.tv` is fine though, just for the class init would be nice.

tests/test_constraints.py
```diff
@@ -178,7 +178,12 @@ def test_batched_constraints(backend):
...
     constraint = poisson_constraint_combined(config)
     result = default_backend.astensor(
-        pyhf.tensorlib.tolist(constraint.logpdf(config.auxdata, suggested_pars))
+        pyhf.tensorlib.tolist(
+            constraint.logpdf(
+                pyhf.tensorlib.astensor(config.auxdata),
+                pyhf.tensorlib.astensor(suggested_pars),
```
matthewfeickert: This and the following would then get changed as

```suggestion
                suggested_pars,
```

Though maybe your wrapping like this is to have visual parity with `pyhf.tensorlib.astensor(config.auxdata)` and follow being super explicit about what is happening in tests?

matthewfeickert:
@lukasheinrich You have made things quite speedy indeed, as adding the 200 bin `numpy_minuit` now only adds on another ~7 minutes to the benchmarks. https://github.com/diana-hep/pyhf/pull/576/checks?check_run_id=233179398#step:9:419

src/pyhf/parameters/paramview.py
lukasheinrich: done

src/pyhf/constraints.py

src/pyhf/pdf.py
lukasheinrich: whops missed this

tests/test_constraints.py
lukasheinrich: I'd prefer to keep it explicit

tests/test_constraints.py
lukasheinrich: ^^

src/pyhf/pdf.py

src/pyhf/workspace.py
```diff
@@ -0,0 +1,152 @@
+import logging
```
matthewfeickert: @lukasheinrich Is the only reason that `Workspace` got its own file just to make CodeFactor happy? Or did you have plans to split it off already and this was a good opportunity? @kratsg had suggested moving `Workspace` and `Pdf` into a `pdf/` directory, but I guess I need some time to think on this.

lukasheinrich: I think moving it in `workspace.py` is fine. Yeah it was just to make CF happy, but it makes sense in general too

matthewfeickert: > but it makes sense in general too

I have to agree here.

matthewfeickert:
@lukasheinrich The PR body reads that

> this alleviates a part of the regression found in #572

This can now say

> Resolves #572

, correct?

lukasheinrich:
@matthewfeickert added a commit msg. now waiting for @kratsg to comment

src/pyhf/modifiers/histosys.py
```diff
@@ -35,7 +35,11 @@ def __init__(
         keys = ['{}/{}'.format(mtype, m) for m, mtype in histosys_mods]
         histosys_mods = [m for m, _ in histosys_mods]
 
-        parfield_shape = (self.batch_size or 1, len(pdfconfig.suggested_init()))
+        parfield_shape = (
+            (self.batch_size or 1, len(pdfconfig.suggested_init()))
```
kratsg: remove `or 1`

src/pyhf/modifiers/lumi.py
```diff
@@ -32,8 +32,13 @@ def __init__(self, lumi_mods, pdfconfig, mega_mods, batch_size=None):
         keys = ['{}/{}'.format(mtype, m) for m, mtype in lumi_mods]
         lumi_mods = [m for m, _ in lumi_mods]
 
-        parfield_shape = (self.batch_size or 1, len(pdfconfig.suggested_init()))
+        parfield_shape = (
+            (self.batch_size or 1, len(pdfconfig.suggested_init()))
```
kratsg: remove `or 1`

src/pyhf/modifiers/normfactor.py
```diff
@@ -29,7 +29,11 @@ def __init__(self, normfactor_mods, pdfconfig, mega_mods, batch_size=None):
         keys = ['{}/{}'.format(mtype, m) for m, mtype in normfactor_mods]
         normfactor_mods = [m for m, _ in normfactor_mods]
 
-        parfield_shape = (self.batch_size or 1, len(pdfconfig.suggested_init()))
+        parfield_shape = (
+            (self.batch_size or 1, len(pdfconfig.suggested_init()))
```
kratsg: remove `or 1`

src/pyhf/modifiers/normsys.py
```diff
@@ -36,7 +36,11 @@ def __init__(
 
         self.batch_size = batch_size
 
-        parfield_shape = (self.batch_size or 1, len(pdfconfig.suggested_init()))
+        parfield_shape = (
+            (self.batch_size or 1, len(pdfconfig.suggested_init()))
```
kratsg: remove `or 1`

src/pyhf/modifiers/histosys.py
kratsg: use `pdfconfig.npars`

src/pyhf/modifiers/lumi.py
kratsg: use `pdfconfig.npars`

src/pyhf/modifiers/lumi.py
```diff
@@ -32,8 +32,13 @@ def __init__(self, lumi_mods, pdfconfig, mega_mods, batch_size=None):
...
+            if self.batch_size
+            else (len(pdfconfig.suggested_init()),)
```
kratsg: use `pdfconfig.npars`

src/pyhf/modifiers/histosys.py
```diff
@@ -35,7 +35,11 @@ def __init__(
...
+            if self.batch_size
+            else (len(pdfconfig.suggested_init()),)
```
kratsg: use `pdfconfig.npars`

src/pyhf/modifiers/normfactor.py
kratsg: use `pdfconfig.npars`

src/pyhf/modifiers/normfactor.py
```diff
@@ -29,7 +29,11 @@ def __init__(self, normfactor_mods, pdfconfig, mega_mods, batch_size=None):
...
+            if self.batch_size
+            else (len(pdfconfig.suggested_init()),)
```
kratsg: use `pdfconfig.npars`

src/pyhf/modifiers/normsys.py
kratsg: use `pdfconfig.npars`

src/pyhf/modifiers/normsys.py
```diff
@@ -36,7 +36,11 @@ def __init__(
...
+            if self.batch_size
+            else (len(pdfconfig.suggested_init()),)
```
kratsg: use `pdfconfig.npars`

src/pyhf/probability.py
```diff
@@ -1,23 +1,19 @@
 from . import get_backend
-from .tensor.common import TensorViewer
 
 
 class Poisson(object):
     def __init__(self, rate):
-        tensorlib, _ = get_backend()
-        self.lam = tensorlib.astensor(rate)
+        self.lam = rate
```
kratsg: let's call it `self.lambda`

PR #578, opened 2019-09-22
fix: Correct error in invert_interval in Binder example notebook

### LINKED ISSUE #577: Type-O in Binder example notebook
# Description

The example Binned HEP Statistical Analysis in Pyton has a small type-O for the calculation of the obs upper limit.
https://diana-hep.org/pyhf/examples/notebooks/binderexample/StatisticalAnalysis.html#HistFactory

# Expected Behavior

In the plot for cell 12 it looks like the obs upper limit on mu should be close to 0.6. 

# Actual Behavior

The result printed in cell 13 says it is 0.96.
The issue is with the `invert_interal` function in cell 11: the line  `yvals = cls_obs` should be replaced with `y_vals = cls_obs` 

# Steps to Reproduce

Run the Binder example

# Checklist

N/a because has to do with the documentation

matthewfeickert:
# Description

Resolves #577 

* Pass in the original arguments to the interpolator instead of copying them to `y_vals`, as this makes it clearer what is happening and then also avoids the error of `yvals` (instead of `y_vals`) typo altogether.
* Apply Black to `invert_interval`
* Additionally split the cell that contains `plot_results` and `invert_interval`

Here are the relevent cells in the `nbdiff-web StatisticalAnalysis.ipynb`

nbdiff_1

nbdiff_2

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* fix: Revise 'invert_interval' function to fix error (from typo) and then avoid the error altogether in Binder example notebook (StatisticalAnalysis.ipynb)
* chore: Change the 'invert_interval' function to be the same in all public example notebooks 
```


matthewfeickert:
I think this should be an easy PR to review, as it is just changing the internals of a helper function to avoid an error.

matthewfeickert:
@lukasheinrich Can we get your review on this?

matthewfeickert:
Thanks for reviewing and getting this in, and also thanks to @nhartman94 for bringing this to our attention in the first place.

PR #582, opened 2019-09-25
perf: Effectively remove the use of self.astensor() in tensor backend calls

### LINKED ISSUE #295: Reduce the use of self.astensor() in tensor backend calls
# Description

Many of the calls in the backends wrap the `tensor_in` in `self.astensor` for convenience. This will probably result in some sloppy code and unnecessary conversions in the future. This issue is filed to remind us to clean this portion up.

matthewfeickert:
# Description

Resolves #295

Remove or reduce the guarding `self.astensor()` calls in the methods of the tensor backends. This makes it more explicit that tensors are required, move the responsibility to the user to format their data correctly, and should add a speedup by dramatically reducing the number of `self.astensor()` calls.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* perf: Remove all non-essential 'self.astensor' calls from the tensor backend methods
* Replace custom instances of 'where' with tensor library versions for TensorFlow, PyTorch, and MXNet backends
   - Uses: tf.compat.v2.where, torch.where, nd.where
* Make masks used in 'tensorlib.where' explicit bools
* docs: Update the docstring examples for the tensor backends
* test: Make all relevant arguments tensors in test_tensor.py
```

### DIFF
--- a/src/pyhf/interpolators/code0.py
+++ b/src/pyhf/interpolators/code0.py
@@ -57,7 +57,7 @@ class code0(object):
         tensorlib, _ = get_backend()
         self._precompute_alphasets(tensorlib.shape(alphasets))
         where_alphasets_positive = tensorlib.where(
-            tensorlib.astensor(alphasets > 0), self.mask_on, self.mask_off
+            alphasets > 0, self.mask_on, self.mask_off
         )
 
         # s: set under consideration (i.e. the modifier)
--- a/src/pyhf/interpolators/code0.py
+++ b/src/pyhf/interpolators/code0.py
@@ -57,7 +57,7 @@ class code0(object):
         tensorlib, _ = get_backend()
         self._precompute_alphasets(tensorlib.shape(alphasets))
         where_alphasets_positive = tensorlib.where(
-            alphasets > 0, self.mask_on, self.mask_off
+            tensorlib.astensor(alphasets > 0), self.mask_on, self.mask_off
         )
 
         # s: set under consideration (i.e. the modifier)
--- a/src/pyhf/interpolators/code0.py
+++ b/src/pyhf/interpolators/code0.py
@@ -71,8 +71,11 @@ class code0(object):
             'sa,shb->shab', alphasets, self.deltas_dn
         )
 
-        masks = tensorlib.einsum(
-            'sa,shb->shab', where_alphasets_positive, self.broadcast_helper
+        masks = tensorlib.astensor(
+            tensorlib.einsum(
+                'sa,shb->shab', where_alphasets_positive, self.broadcast_helper
+            ),
+            dtype="bool",
         )
 
         return tensorlib.where(masks, alphas_times_deltas_up, alphas_times_deltas_dn)
--- a/src/pyhf/modifiers/lumi.py
+++ b/src/pyhf/modifiers/lumi.py
@@ -69,4 +69,8 @@ class lumi_combined(object):
         else:
             results_lumi = tensorlib.einsum('msab,xa->msab', self.lumi_mask, lumis)
 
-        return tensorlib.where(self.lumi_mask, results_lumi, self.lumi_default)
+        return tensorlib.where(
+            tensorlib.astensor(self.lumi_mask, dtype="bool"),
+            results_lumi,
+            self.lumi_default,
+        )
--- a/src/pyhf/modifiers/lumi.py
+++ b/src/pyhf/modifiers/lumi.py
@@ -52,6 +52,7 @@ class lumi_combined(object):
         self.lumi_mask = tensorlib.tile(
             tensorlib.astensor(self._lumi_mask), (1, 1, self.batch_size or 1, 1)
         )
+        self.lumi_mask_bool = tensorlib.astensor(self.lumi_mask, dtype="bool")
         self.lumi_default = tensorlib.ones(self.lumi_mask.shape)
 
     def apply(self, pars):
@@ -69,8 +70,4 @@ class lumi_combined(object):
         else:
             results_lumi = tensorlib.einsum('msab,xa->msab', self.lumi_mask, lumis)
 
-        return tensorlib.where(
-            tensorlib.astensor(self.lumi_mask, dtype="bool"),
-            results_lumi,
-            self.lumi_default,
-        )
+        return tensorlib.where(self.lumi_mask_bool, results_lumi, self.lumi_default)
--- a/src/pyhf/modifiers/normfactor.py
+++ b/src/pyhf/modifiers/normfactor.py
@@ -51,6 +51,9 @@ class normfactor_combined(object):
         self.normfactor_mask = tensorlib.tile(
             tensorlib.astensor(self._normfactor_mask), (1, 1, self.batch_size or 1, 1)
         )
+        self.normfactor_mask_bool = tensorlib.astensor(
+            self.normfactor_mask, dtype="bool"
+        )
         self.normfactor_default = tensorlib.ones(self.normfactor_mask.shape)
 
     def apply(self, pars):
@@ -73,8 +76,6 @@ class normfactor_combined(object):
             )
 
         results_normfactor = tensorlib.where(
-            tensorlib.astensor(self.normfactor_mask, dtype="bool"),
-            results_normfactor,
-            self.normfactor_default,
+            self.normfactor_mask_bool, results_normfactor, self.normfactor_default
         )
         return results_normfactor
--- a/src/pyhf/modifiers/normfactor.py
+++ b/src/pyhf/modifiers/normfactor.py
@@ -48,9 +48,8 @@ class normfactor_combined(object):
         tensorlib, _ = get_backend()
         if not self.param_viewer.index_selection:
             return
-        self.normfactor_mask = tensorlib.astensor(self._normfactor_mask)
         self.normfactor_mask = tensorlib.tile(
-            self.normfactor_mask, (1, 1, self.batch_size or 1, 1)
+            tensorlib.astensor(self._normfactor_mask), (1, 1, self.batch_size or 1, 1)
         )
         self.normfactor_default = tensorlib.ones(self.normfactor_mask.shape)
 
@@ -74,6 +73,8 @@ class normfactor_combined(object):
             )
 
         results_normfactor = tensorlib.where(
-            self.normfactor_mask, results_normfactor, self.normfactor_default
+            tensorlib.astensor(self.normfactor_mask, dtype="bool"),
+            results_normfactor,
+            self.normfactor_default,
         )
         return results_normfactor
--- a/src/pyhf/modifiers/shapefactor.py
+++ b/src/pyhf/modifiers/shapefactor.py
@@ -118,9 +118,9 @@ class shapefactor_combined(object):
         if not self.param_viewer.index_selection:
             return
         tensorlib, _ = get_backend()
-        self.shapefactor_mask = tensorlib.astensor(self._shapefactor_mask, dtype="bool")
         self.shapefactor_mask = tensorlib.tile(
-            self.shapefactor_mask, (1, 1, self.batch_size or 1, 1)
+            tensorlib.astensor(self._shapefactor_mask, dtype="bool"),
+            (1, 1, self.batch_size or 1, 1),
         )
         self.access_field = tensorlib.astensor(self._access_field, dtype='int')
 
--- a/src/pyhf/modifiers/shapefactor.py
+++ b/src/pyhf/modifiers/shapefactor.py
@@ -118,7 +118,7 @@ class shapefactor_combined(object):
         if not self.param_viewer.index_selection:
             return
         tensorlib, _ = get_backend()
-        self.shapefactor_mask = tensorlib.astensor(self._shapefactor_mask)
+        self.shapefactor_mask = tensorlib.astensor(self._shapefactor_mask, dtype="bool")
         self.shapefactor_mask = tensorlib.tile(
             self.shapefactor_mask, (1, 1, self.batch_size or 1, 1)
         )
--- a/src/pyhf/tensor/common.py
+++ b/src/pyhf/tensor/common.py
@@ -17,10 +17,10 @@ class _TensorViewer(object):
         self.batch_size = batch_size
 
         self.partition_indices = indices
-        a = default_backend.astensor(
+        _concat_indices = default_backend.astensor(
             default_backend.concatenate(self.partition_indices), dtype='int'
         )
-        self._sorted_indices = default_backend.tolist(a.argsort())
+        self._sorted_indices = default_backend.tolist(_concat_indices.argsort())
 
         self._precompute()
         events.subscribe('tensorlib_changed')(self._precompute)
--- a/src/pyhf/tensor/common.py
+++ b/src/pyhf/tensor/common.py
@@ -45,9 +45,15 @@ class _TensorViewer(object):
     def split(self, data):
         tensorlib, _ = get_backend()
         if len(tensorlib.shape(data)) == 1:
-            return [tensorlib.gather(data, idx) for idx in self.partition_indices]
+            return [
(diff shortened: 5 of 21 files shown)


src/pyhf/interpolators/code0.py
```diff
@@ -57,7 +57,7 @@ def __call__(self, alphasets):
         tensorlib, _ = get_backend()
         self._precompute_alphasets(tensorlib.shape(alphasets))
         where_alphasets_positive = tensorlib.where(
-            alphasets > 0, self.mask_on, self.mask_off
+            tensorlib.astensor(alphasets > 0), self.mask_on, self.mask_off
```
lukasheinrich: why are we *adding* `astensor` calls here?

matthewfeickert: In short, because I removed the `mask = self.astensor(mask)` from the `where` method in the TensorFlow backend. As it requires that the `mask` argument is a tensor I decided to add `tensorlib.astensor` calls to a few places where the mask is a logical requirement and not already a tensor given that

```
$ git grep "tensorlib.where(" src/ | wc -l
26
```

is more then the number of calls that I added. So it seems that explicitly calling it as needed is preferable then having the backends call it everytime.

matthewfeickert: Alternatively, I could say that as this is a TensorFlow problem which might improve in TensorFlow 2.0 I could remove the calls to `tensorlib.astensor` that I added and just have TensorFlow do the redundant calls, until it might be possible to turn it off entirely.

lukasheinrich: could we skip this if we instead assured that alphasets is a tensor to start with?

matthewfeickert: > could we skip this if we instead assured that alphasets is a tensor to start with?

No. The issue is that `alphasets > 0` is being passed in. TensorFlow 1.0 needs to be told that this is a bool tensor with a `astensor` call.

matthewfeickert: The decision is if we're going to make a few calls across all backends (which we are currently already doing) in the interpolator codes and hypotest, or if we are going to isolate all of those calls to the TensorFlow backend and deal with them later in the TensorFlow 2.0 migration.

I think we can probably do the latter.

matthewfeickert:
The only `selt.astensor` that I think we don't need but I haven't figured out how to get is in

https://github.com/diana-hep/pyhf/blob/86269a013121d82d33662753caa3dfad7ed6dcd6/src/pyhf/tensor/pytorch_backend.py#L97-L99

[`torch.gather`](https://pytorch.org/docs/stable/torch.html#torch.gather) is confusing to me though, so I'm not sure how to properly get this.

src/pyhf/tensor/common.py
```diff
@@ -45,9 +45,15 @@ def stitch(self, data):
     def split(self, data):
         tensorlib, _ = get_backend()
         if len(tensorlib.shape(data)) == 1:
-            return [tensorlib.gather(data, idx) for idx in self.partition_indices]
+            return [
+                tensorlib.gather(data, tensorlib.astensor(idx, dtype='int'))
```
lukasheinrich: this can probaly go into `_precompute` since converting the indices into tensors is not dependent on `data`

src/pyhf/tensor/common.py
```diff
@@ -45,9 +45,15 @@ def stitch(self, data):
...
+                for idx in self.partition_indices
+            ]
         data = tensorlib.einsum('...j->j...', tensorlib.astensor(data))
         return [
-            tensorlib.einsum('j...->...j', tensorlib.gather(data, idx))
+            tensorlib.einsum(
+                'j...->...j',
+                tensorlib.gather(data, tensorlib.astensor(idx, dtype='int')),
+            )
```
lukasheinrich: dito

src/pyhf/tensor/common.py
matthewfeickert: Okay, I'll look into moving this. A note, this is _only_ being done to remove the `self.astensor` call in the PyTorch backend `gather` method (https://github.com/diana-hep/pyhf/pull/582/commits/2a163e71dc448494415960633faf72a6af1c974a). So this is a bit of a tradeoff, as this now requires extra ops by all other backends but it reduces the number of times that the PyTorch backend will have to call `astensor` overall.

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -195,26 +186,20 @@ def reshape(self, tensor, newshape):
...
 
     def stack(self, sequence, axis=0):
         return tf.stack(sequence, axis=axis)
 
     def where(self, mask, tensor_in_1, tensor_in_2):
-        mask = self.astensor(mask)
-        tensor_in_1 = self.astensor(tensor_in_1)
-        tensor_in_2 = self.astensor(tensor_in_2)
-        return mask * tensor_in_1 + (1 - mask) * tensor_in_2
+        mask = tf.cast(mask, tf.bool)
```
kratsg: why is there a cast here? The mask should be `bool` already.

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -195,26 +186,20 @@ def reshape(self, tensor, newshape):
...
+        return tf.compat.v2.where(mask, tensor_in_1, tensor_in_2)
```
kratsg: Can we always use latest, or does `tf` only version this method?

matthewfeickert: > Can we always use latest, or does `tf` only version this method?

@kratsg Sorry, I'm not quite sure what you mean by "latest" here, but if I guess a bit the reason I'm using `tf.compat.v2.where` is because it is [the alias](https://www.tensorflow.org/api_docs/python/tf/where) that TensorFlow recommends when you try to use `tf.where` while using the TF v1.0 API.

src/pyhf/tensor/tensorflow_backend.py
matthewfeickert: > why is there a cast here? The mask should be `bool` already.

I think that it is better to just cast it to make sure. There are times like the following where the mask seems like a valid `Bool`

https://github.com/diana-hep/pyhf/blob/c8a94f4d5a11edfdd5a28184153e1ef4b076e638/tests/test_tensor.py#L48-L50

but TensorFlow doesn't see it as such if the `tf.cast(mask, tf.bool)` is removed. So I think that it should just be used to ensure the right Tensor `dtype`.

kratsg: https://github.com/diana-hep/pyhf/blob/d9aeac8bd3b7bd384342ae2e0e7547b78714e66c/src/pyhf/tensor/tensorflow_backend.py#L124

That's because the default cast is to float. You should use `tb.astensor(tensor, dtype='bool')` for the mask input.

src/pyhf/tensor/pytorch_backend.py
```diff
@@ -131,35 +123,25 @@ def zeros(self, shape):
...
 
     def stack(self, sequence, axis=0):
         return torch.stack(sequence, dim=axis)
 
     def where(self, mask, tensor_in_1, tensor_in_2):
-        mask = self.astensor(mask).type(torch.FloatTensor)
-        tensor_in_1 = self.astensor(tensor_in_1)
-        tensor_in_2 = self.astensor(tensor_in_2)
-        return mask * tensor_in_1 + (1 - mask) * tensor_in_2
+        return torch.where(mask.type(torch.BoolTensor), tensor_in_1, tensor_in_2)
```
matthewfeickert: @kratsg Your comment got removed when I combined the two lines, so answering here again:

> why is there a cast here? The mask should be `bool` already.

I think that it is better to just cast it to make sure. There are times like the following where the mask seems like a valid `Bool`

https://github.com/diana-hep/pyhf/blob/c8a94f4d5a11edfdd5a28184153e1ef4b076e638/tests/test_tensor.py#L48-L50

but TensorFlow doesn't see it as such if the `tf.cast(mask, tf.bool)` is removed. So I think that it should just be used to ensure the right Tensor `dtype`.

src/pyhf/tensor/tensorflow_backend.py
kratsg: tl;dr - the test is wrong.

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -195,26 +186,19 @@ def reshape(self, tensor, newshape):
...
 
     def stack(self, sequence, axis=0):
         return tf.stack(sequence, axis=axis)
 
     def where(self, mask, tensor_in_1, tensor_in_2):
-        mask = self.astensor(mask)
-        tensor_in_1 = self.astensor(tensor_in_1)
-        tensor_in_2 = self.astensor(tensor_in_2)
-        return mask * tensor_in_1 + (1 - mask) * tensor_in_2
+        return tf.compat.v2.where(tf.cast(mask, tf.bool), tensor_in_1, tensor_in_2)
```
matthewfeickert: @kratsg Your comment got removed when I combined two lines, so answering here again.

> why is there a cast here? The mask should be `bool` already.

I think that it is better to just cast it to make sure. There are times like the following where the mask seems like a valid `Bool`

https://github.com/diana-hep/pyhf/blob/c8a94f4d5a11edfdd5a28184153e1ef4b076e638/tests/test_tensor.py#L48-L50

but TensorFlow doesn't see it as such if the `tf.cast(mask, tf.bool)` is removed. So I think that it should just be used to ensure the right Tensor `dtype`.

kratsg: https://github.com/diana-hep/pyhf/blob/d9aeac8bd3b7bd384342ae2e0e7547b78714e66c/src/pyhf/tensor/tensorflow_backend.py#L124

That's because the default cast is to float. You should use `tb.astensor(tensor, dtype='bool')` for the mask input.

tl;dr - the test is wrong.

kratsg:
```
>>> import pyhf
>>> pyhf.set_backend(pyhf.tensor.tensorflow_backend())
>>> from pyhf import tensorlib as tb
>>> tb.astensor([1,0,1])
array([1., 0., 1.])
>>> tb.astensor([1,0,1], dtype='float')
array([1., 0., 1.])
>>> tb.astensor([1,0,1], dtype='bool')
array([ True, False,  True])
```

src/pyhf/tensor/tensorflow_backend.py
matthewfeickert: > That's because the default cast is to float

Yes, I wasn't confused as to the _why_ of the behavior, but as to the reasonableness of it. I will change the test, but if you were a user it would seem to me that `tb.astensor([1, 0, 1])` could be a valid way to indicate a Bool mask. So I would opt to include that cast there, which should be faster then the `self.astensor()` call that we were previously including.

src/pyhf/tensor/tensorflow_backend.py
matthewfeickert: Not sure where my last comment went (somewhere into the void?) but yes, it is clear as to the _why_ of the behavior by not to the reasonableness. I have changed the test, but I think it is better to keep the cast there as a user could quite legitimately view the `tb.astensor([1,0,1])` as a valid form of Bool (as NumPy does). The cast should be faster then the `self.astensor` call that we have previously been doing, so it seems like keeping this cast in there should be reasonable and keep things safe.

kratsg: > as a valid form of Bool (as NumPy does)

It's not that numpy handles it as a valid form, it type-casts for you implicitly as part of the masking - but that might not be good behavior. We should have users be explicit about the type of tensor they're dealing with (bools, floats, ints) - because they all store memory in different ways and are allocated differently.

> The cast should be faster then the `self.astensor` call that we have previously been doing

Then we should change the tensorflow `astensor` to use the `cast` then... if it's faster.

src/pyhf/tensor/common.py
matthewfeickert: Should be resolved now.

src/pyhf/tensor/tensorflow_backend.py
matthewfeickert: > Then we should change the tensorflow `astensor` to use the `cast` then... if it's faster.

We do already

https://github.com/diana-hep/pyhf/blob/f952b8bca853e09259351b2ebc2a09c918c23e6f/src/pyhf/tensor/tensorflow_backend.py#L152-L153

doing this just skips the rest of the logic of `astensor`. I will still work on changing the calls in other parts, as needed, to explicitly request Bools.

src/pyhf/interpolators/code0.py
```diff
@@ -71,8 +71,11 @@ def __call__(self, alphasets):
             'sa,shb->shab', alphasets, self.deltas_dn
         )
 
-        masks = tensorlib.einsum(
-            'sa,shb->shab', where_alphasets_positive, self.broadcast_helper
+        masks = tensorlib.astensor(
```
lukasheinrich: doesn't einsum return a tensor anyways?

src/pyhf/modifiers/lumi.py
```diff
@@ -69,4 +69,8 @@ def apply(self, pars):
         else:
             results_lumi = tensorlib.einsum('msab,xa->msab', self.lumi_mask, lumis)
 
-        return tensorlib.where(self.lumi_mask, results_lumi, self.lumi_default)
+        return tensorlib.where(
+            tensorlib.astensor(self.lumi_mask, dtype="bool"),
```
lukasheinrich: this can probably go in the _precompute of the modifier where `self.lumi_mask` is set up

src/pyhf/modifiers/normfactor.py
```diff
@@ -74,6 +73,8 @@ def apply(self, pars):
             )
 
         results_normfactor = tensorlib.where(
-            self.normfactor_mask, results_normfactor, self.normfactor_default
+            tensorlib.astensor(self.normfactor_mask, dtype="bool"),
```
lukasheinrich: same here.. could just wrap the `tensorlib.tile` call in precompute to do the cast

src/pyhf/modifiers/shapefactor.py
```diff
@@ -118,7 +118,7 @@ def _precompute(self):
         if not self.param_viewer.index_selection:
             return
         tensorlib, _ = get_backend()
-        self.shapefactor_mask = tensorlib.astensor(self._shapefactor_mask)
+        self.shapefactor_mask = tensorlib.astensor(self._shapefactor_mask, dtype="bool")
```
lukasheinrich: same here.. could just wrap the tensorlib.tile call in precompute to do the cast

src/pyhf/interpolators/code0.py
matthewfeickert: @lukasheinrich It does. Dumb question: Do we support a way to cast the return Tensor from `tensorlib`?

src/pyhf/modifiers/lumi.py
matthewfeickert: @lukasheinrich I looked at this again and I don't think this will work. If `einsum` is given `self.lumi_mask` as a Bool tensor then it will crash as the backends don't support `Bool` there. Sorry, I ran into this last night but didn't mention it anywhere.

lukasheinrich: ok I guess we could have `self.lumi_mask_bool`, and `self.lumi_mask` and use each at their appropriate position, but probably doesn't make that much of a diff

src/pyhf/modifiers/shapefactor.py
matthewfeickert: @lukasheinrich I simplified this into one call instead of two, but this still needs to be done as `tile` requires that `self.shapefactor_mask` is a Tensor, so this needs to be done before `tile` gets evaluated.

src/pyhf/modifiers/lumi.py
matthewfeickert: I can do this.

PR #584, opened 2019-09-25
ci: Use Travis CI only for Python 2.7 testing and releases

matthewfeickert:
# Description

~~Do all testing in GitHub Actions and then have Travis CI be responsible for releases.~~ The setup action for Python 2.7 seems to be broken at the moment, so still use Travis for Python 2.7 testing. At the moment this still requires Travis CI to run one version of the test suite as Coveralls needs to be run.

~~Perhaps this is a good time to try to figure out how to get Coveralls to run in GitHub Actions as well.~~ Changed from Coveralls to Codecov in PR #598 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

~~* ci: Have GitHub Actions be responsible for all testing and benchmarks~~
~~* ci: Migrate Coveralls to GitHub Actions~~

```
* ci: Have GitHub Actions be responsible for all Python 3 testing (Travis CI only tests Python 2.7)
   - Travis CI is still responsible for releases
* ci: Add merged workflow to GitHub Actions
* ci: Migrate Binder build trigger from Travis CI to GitHub Actions
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,13 +18,12 @@ script:
 after_success: if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then coveralls; fi
 
 # test Python 2.7 on 'pr' builds and 'push' builds on master
-# test docs on 'pr' builds and 'push' builds on master and on docs branches
-# benchmark and deploy to PyPI only when merged into master (those mereges are 'push' builds)
+# test docs and deploy to PyPI only when merged into master (those mereges are 'push' builds)
 stages:
   - name: test
     if: (branch = master) AND (NOT (branch =~ /^docs\//))
   - name: docs
-    if: (branch = master) OR (branch =~ /^docs\//)
+    if: (branch = master) AND (NOT (type IN (pull_request)))
   - name: binder
     if: (branch = master) AND (NOT (type IN (pull_request)))
   - name: deploy
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ dist: xenial
 language: python
 cache: pip
 python:
-  - '2.7'
-  - '3.6'
   - '3.7'
 before_install:
   - sudo apt-get install libtcmalloc-minimal4
@@ -26,8 +24,6 @@ after_success: skip
 stages:
   - name: test
     if: NOT (branch =~ /^docs\//)
-  - name: benchmark
-    if: (branch = master) AND (NOT (type IN (pull_request)))
   - name: docs
     if: (branch = master) OR (branch =~ /^docs\//)
   - name: binder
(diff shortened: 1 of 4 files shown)


matthewfeickert:
Hm. Problems already. [GitHub Actions does not like Python 2.7](https://github.com/diana-hep/pyhf/commit/ed1157f2919e63b17b03a84306a28bb2e0bfe36d/checks#step:4:20)

matthewfeickert:
Strangely, I can get [Python 2.7 testing running solo fine on Travis](https://travis-ci.org/diana-hep/pyhf/builds/589669009), but on [GitHub Actions it fails to be able to properly install TensorFlow](https://github.com/diana-hep/pyhf/pull/584/checks?check_run_id=236446833#step:4:21). :?

matthewfeickert:
Quite confused as to what is happening in the Python 2.7 image as I can do the following

```
$ docker run --rm -it python:2.7 /bin/bash
# cd
# git clone https://github.com/diana-hep/pyhf.git
# cd pyhf/
# pip install -U -e .[complete]
```

and get a working install of pyhf.

matthewfeickert:
@damccorm, sorry to bug you, but as you seem to be [the primary developer](https://github.com/actions/setup-python/graphs/contributors) for the [`setup-python` action](https://github.com/actions/setup-python) I was wondering if you had seen anything like this before. I don't want to open an Issue on `setup-python` if this [inability to install with Python 2.7](https://github.com/diana-hep/pyhf/pull/584#issuecomment-535245889) isn't actually a real issue and is somehow related to other problems.

matthewfeickert:
As it seems that the Python 2.7 setup GitHub Action won't be fixed anytime soon I'm just going to move forward with this PR to have it just have Travis only test Python 2.7 on push builds. This should still significantly speed up the CI as we are always waiting on Travis CI to finish PR builds.

matthewfeickert:
Thanks to @lee-dohm for their [`post-release-notes` workflow](https://github.com/lee-dohm/atom-style-tweaks/blob/master/.github/workflows/post-release-notes.yml) which I used to add the merged workflow.

lee-dohm:
You're welcome @matthewfeickert. I'm glad it was helpful 🙇

.travis.yml
```diff
@@ -17,20 +15,14 @@ script:
   - pyflakes src
   - check-manifest || travis_terminate 1
   - python -m pytest -r sx --ignore tests/benchmarks/ --ignore tests/test_notebooks.py
-  - if [[ $TRAVIS_PYTHON_VERSION == '3.7' ]]; then black --check --diff --verbose .; fi
-after_success: if [[ $TRAVIS_PYTHON_VERSION == '3.7' ]]; then coveralls; fi
+after_success: coveralls
```
kratsg: do not use coveralls with python 2.7. This is a regression as we only ever did coveralls with python 3.7.

matthewfeickert: @kratsg Why does this matter? The coverage is the same, no?

kratsg: We've had issues in the past of coveralls reporting different or incorrect coverage because we call it multiple times during the CI (for different python versions). We should stick with 3.7 since we're quietly dropping 2.7 anyway.

matthewfeickert: > We've had issues in the past of coveralls reporting different or incorrect coverage because we call it multiple times during the CI (for different python versions). We should stick with 3.7 since we're quietly dropping 2.7 anyway.

We're not dropping 2.7 now though, and this new configuration only runs Travis once. Are we sure it was a Python runtime version thing, and not a push event vs. pull request event?

> We should stick with 3.7

I think it doesn't make much sense to continue to test Python 3.7 more slowly on Travis just for this. I was going to work on getting Coveralls running in GitHub Actions in Python 3.7 as a separate PR, but I can work on getting it setup in this PR if you'd prefer that.

kratsg: I'd prefer that. It was a problem with coveralls being run for both python 2.7 and python 3.7 -- as well as the python 2.7 incorrectly reporting coverage consistently.. so we switched to only python 3.7 as you can see in the changes you're making.

matthewfeickert: Okay, I'll work on getting Coveralls to GHA tomorrow.

matthewfeickert:
Hm, so [Coveralls is passing with GitHub Actions now](https://github.com/diana-hep/pyhf/pull/584/checks?check_run_id=245806463#step:9:1) and is correctly giving a HTTP response of 200, but it isn't commenting on the PR.

.travis.yml
matthewfeickert: Done now.

Issue #97, opened 2018-03-17
check docstrings via pydocstyle


lukasheinrich:
# Description

The pydocstyle package can check docstrings for compliance with style conventions such as numpy. 

https://github.com/PyCQA/pydocstyle

we should implement this as part of travis / documentation building

lukasheinrich:
we're doing this now (not complete coverage but working on it)

PR #588, opened 2019-09-29
ci: Use pydocstyle

lukasheinrich:
# Description

* adds `pydocstyle` testing (on pdf.py only for now) to CI
* makes pdf.py pass pydocstyle

Adresses #97 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add pydocstyle to "develop" environment requirements
* Add pydocstyle testing (on pdf.py only for now) to CI
* Add docstrings to all public classes and methods in 'pdf.py'
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,8 +98,8 @@ jobs:
         sudo apt-get -qq install pandoc
     - name: Check docstrings
       run: |
-        pydocstyle --select D1 src/pyhf/probability.py
-        pydocstyle --select D102 src/pyhf/pdf.py
+        # Group 1 is related to docstrings
+        pydocstyle --select D1 src/pyhf/pdf.py src/pyhf/probability.py
     - name: Test and build docs
       run: |
         python -m doctest README.md
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,8 +98,8 @@ jobs:
         sudo apt-get -qq install pandoc
     - name: Check docstrings
       run: |
-        # Group 1 is related to docstrings
         pydocstyle --select D1 src/pyhf/probability.py
+        pydocstyle --select D102 src/pyhf/pdf.py
     - name: Test and build docs
       run: |
         python -m doctest README.md
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -219,7 +219,7 @@ class _ConstraintModel(object):
 
     def make_pdf(self, pars):
         """
-        Construct a  pdf object for a given set of parameter values.
+        Construct a pdf object for a given set of parameter values.
 
         Args:
             pars (`tensor`): The model parameters
@@ -656,7 +656,7 @@ class Model(object):
 
     def make_pdf(self, pars):
         """
-        Construct a  pdf object for a given set of parameter values.
+        Construct a pdf object for a given set of parameter values.
 
         Args:
             pars (`tensor`): The model parameters
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -397,7 +397,7 @@ class Model(object):
         Construct a pyhf Model.
 
         Args:
-            spec (`jsonable`): The JSON specification
+            spec (`jsonable`): The HistFactory JSON specification
             batch_size (`None` or `int`): Number of simultaneous (batched) Models to compute.
             config_kwargs: Possible keyword arguments for the model configuration
 
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -628,7 +628,7 @@ class Model(object):
 
     def constraint_logpdf(self, auxdata, pars):
         """
-        Compute the log value of the constraint term.
+        Compute the log value of the constraint pdf.
 
         Args:
             auxdata (`tensor`): The auxiliary measurement data
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -136,7 +136,7 @@ class _ModelConfig(_ChannelSummaryMixin):
         self.poi_index = s.start
 
     def _register_paramset(self, param_name, paramset):
-        """Allocates n nuisance parameters and stores paramset > modifier map."""
+        """Allocates the nuisance parameters and stores paramset into a modifier map."""
         log.info(
             'adding modifier %s (%s new nuisance parameters)',
             param_name,
@@ -365,7 +365,7 @@ class _MainModel(object):
 
         So in the end we only make 3 calls to pdfs
 
-            1. The main pdf of data and modified rates
+            1. The pdf of data and modified rates
             2. All Gaussian constraint as one call
             3. All Poisson constraints as one call
 
@@ -583,7 +583,7 @@ class Model(object):
             pars (`tensor`): The parameter values
 
         Returns:
-            Tensor: The expected auxiliary data
+            Tensor: The expected data of the auxiliary pdf
 
         """
         return self.make_pdf(pars)[1].expected_data()
@@ -593,31 +593,31 @@ class Model(object):
 
     @property
     def nominal_rates(self):
-        """Nominal value of bin rates of the main measurment."""
+        """Nominal value of bin rates of the main model."""
         return self.main_model.nominal_rates
 
     def expected_actualdata(self, pars):
         """
-        Compute the expected value of the main measurements.
+        Compute the expected value of the main model.
 
         Args:
             pars (`tensor`): The parameter values
 
         Returns:
-            Tensor: The expected main data
+            Tensor: The expected data of the main model (no auxiliary data)
 
         """
         return self.make_pdf(pars)[0].expected_data()
 
     def expected_data(self, pars, include_auxdata=True):
         """
-        Compute the expected value of the main measurements.
+        Compute the expected value of the main model
 
         Args:
             pars (`tensor`): The parameter values
 
         Returns:
-            Tensor: The expected main data
+            Tensor: The expected data of the main and auxiliary model
 
         """
         tensorlib, _ = get_backend()
@@ -727,7 +727,7 @@ class Model(object):
 
     def pdf(self, pars, data):
         """
-        Compute the value of the full density.
+        Compute the density at a given observed point in data space of the full model.
 
         Args:
             pars (`tensor`): The parameter values
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -394,7 +394,7 @@ class Model(object):
 
     def __init__(self, spec, batch_size=None, **config_kwargs):
         """
-        Construct a pyhf Model.
+        Construct a HistFactory Model.
 
         Args:
             spec (`jsonable`): The HistFactory JSON specification
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -253,7 +253,7 @@ class _ConstraintModel(object):
             pars (`tensor`): The model parameters
 
         Returns:
-            log pdf value: the log of the pdf value
+            log pdf value: The log of the pdf value
 
         """
         simpdf = self.make_pdf(pars)
@@ -322,7 +322,7 @@ class _MainModel(object):
             pars (`tensor`): The model parameters
 
         Returns:
-            log pdf value: the log of the pdf value
+            log pdf value: The log of the pdf value
 
         """
         return self.make_pdf(pars).log_prob(maindata)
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -66,8 +66,12 @@ jobs:
         pip list
         sudo apt-get update
         sudo apt-get -qq install pandoc
+    - name: Check docstrings
+      run: |
+        pydocstyle --select D102 src/pyhf/pdf.py
```
lukasheinrich: D102 is "missing docstring" I propose we enable other error codes only once all src files pass D102

lukasheinrich:
yeah I didn't want to make us have to address all our backlog at once.. better to get started somehow than be blocked by not having time to do 500 docstrings. 

As mentioend in the comment, i'd first only test for "missing docstring" (step by step moving to testing all of `src`) and then enable other msgs. I would imagine a series of PRs a long the lines of "test docstrings for some/set/of/files/*/py"

matthewfeickert:
> yeah I didn't want to make us have to address all our backlog at once.. better to get started somehow than be blocked by not having time to do 500 docstrings.

Good. :)

> As mentioned in the comment, i'd first only test for "missing docstring" (step by step moving to testing all of `src`) and then enable other msgs. I would imagine a series of PRs a long the lines of "test docstrings for some/set/of/files/*/py"

Excellent. We can make a separate issue for expanding beyond pydocstyle code D102.

.github/workflows/ci.yml
```diff
@@ -68,8 +68,12 @@ jobs:
...
     - name: Test and build docs
       run: |
         python -m doctest README.md
         cd docs && make html && cd -
         touch docs/_build/html/.nojekyll
+
```
matthewfeickert: Can probably avoid blank lines like this

src/pyhf/pdf.py
```diff
@@ -176,9 +178,7 @@ def _create_and_register_paramsets(
 
 
 class _ConstraintModel(object):
-    """
-    Factory class to create pdfs for the constraint terms
-    """
+    """Factory class to create pdfs for the constraint terms."""
```
matthewfeickert: This PR is also a good motivation for us to formalize our docstring style. Just as one example here, do we want to say that as our style we do

```
"""
Description line of class or method
"""
```

or

```
"""Description line of class or method"""
```
?

src/pyhf/pdf.py
```diff
@@ -231,21 +231,27 @@ def _dataprojection(self, data):
...
 
     def has_pdf(self):
-        '''
+        """
+        Indicate whether the model has a constraint.
+
         Returns:
             flag (`bool`): Whether the model has a constraint term
-        '''
+
```
matthewfeickert: Also following up on docstring style, do we want

```
Returns:
    <information on the return>
"""
```

or with the extra space

```
Returns:
    <information on the return>

"""
```

?

src/pyhf/pdf.py
```diff
@@ -402,7 +418,21 @@ def expected_data(self, pars):
...
 
 class Model(object):
+    """The main pyhf model class."""
+
     def __init__(self, spec, batch_size=None, **config_kwargs):
+        """
+        Construct a pyhf Model.
+
+        Args:
+            spec (`jsonable`): a json specification
```
matthewfeickert: Probably have consistent capitalization on the docstring explanation. (Where I vote for using upper case not lower case as the standard)

src/pyhf/pdf.py
lukasheinrich: pydocstyle requires a newline

src/pyhf/pdf.py
matthewfeickert: Ah okay, that greatly simplifies things. Thanks!

src/pyhf/pdf.py
```diff
@@ -152,7 +154,7 @@ def set_poi(self, name):
         self.poi_index = s.start
 
     def _register_paramset(self, param_name, paramset):
-        """allocates n nuisance parameters and stores paramset > modifier map"""
+        """Allocates n nuisance parameters and stores paramset > modifier map."""
```
kratsg: this doesn't read well to me. What is `n`? What does `paramset > modifier map` mean?

src/pyhf/pdf.py
```diff
@@ -350,7 +363,9 @@ def _modifications(self, pars):
 
     def _expected_data(self, pars):
         """
-        For a single channel single sample, we compute
+        Compute the expected rates for given values of parameters.
+
+        For a single channel single sample, we compute:
 
             Pois(d | fac(pars) * (delta(pars) + nom) ) * Gaus(a | pars[is_gaus], sigmas) * Pois(a * cfac | pars[is_poi] * cfac)
```
kratsg: can we change this to the mathematical formula using the notation we introduced in the docs?

src/pyhf/pdf.py
```diff
@@ -392,7 +408,21 @@ def _expected_data(self, pars):
...
 
 class Model(object):
+    """The main pyhf model class."""
+
     def __init__(self, spec, batch_size=None, **config_kwargs):
+        """
+        Construct a pyhf Model.
+
+        Args:
+            spec (`jsonable`): The JSON specification
```
kratsg:
```suggestion
            spec (`jsonable`): The HistFactory JSON specification
```

src/pyhf/pdf.py
```diff
@@ -226,21 +226,27 @@ def expected_data(self, pars):
         return auxdata
 
     def has_pdf(self):
-        '''
+        """
+        Indicate whether the model has a constraint.
```
kratsg:
```suggestion
        Indicate whether this model has a constraint.
```

.github/workflows/ci.yml
```diff
@@ -68,6 +68,10 @@ jobs:
         pip list
         sudo apt-get update
         sudo apt-get -qq install pandoc
+    - name: Check docstrings
+      run: |
+        # Group 1 is related to docstrings
+        pydocstyle --select D1 src/pyhf/pdf.py
```
kratsg:
```suggestion
        pydocstyle --select D107 src/pyhf/pdf.py
```

src/pyhf/pdf.py
```diff
@@ -392,7 +408,21 @@ def _expected_data(self, pars):
 
 
 class Model(object):
+    """The main pyhf model class."""
+
     def __init__(self, spec, batch_size=None, **config_kwargs):
+        """
+        Construct a pyhf Model.
```
kratsg:
```suggestion
        Construct a HistFactory Model.
```

src/pyhf/pdf.py
```diff
@@ -567,38 +597,94 @@ def _create_nominal_and_modifiers(self, config, spec):
...
+
+        """
         return self.make_pdf(pars).pdfobjs[1].expected_data()
 
     def _modifications(self, pars):
         return self.main_model._modifications(pars)
 
     @property
     def nominal_rates(self):
+        """Nominal value of bin rates of the main measurment."""
```
kratsg:
```suggestion
        """Nominal event rates of the main measurement."""
```

kratsg: I don't know what `measurement` means here.

src/pyhf/pdf.py
```diff
@@ -567,38 +597,94 @@ def _create_nominal_and_modifiers(self, config, spec):
...
         return self.main_model.nominal_rates
 
     def expected_actualdata(self, pars):
+        """
+        Compute the expected value of the main measurements.
```
kratsg: Main measurements?

src/pyhf/pdf.py
```diff
@@ -567,38 +597,94 @@ def _create_nominal_and_modifiers(self, config, spec):
...
+
+        Returns:
+            Tensor: The expected main data
+
+        """
         return self.make_pdf(pars).pdfobjs[0].expected_data()
 
     def expected_data(self, pars, include_auxdata=True):
+        """
+        Compute the expected value of the main measurements.
```
kratsg: Main measurements?

src/pyhf/pdf.py
```diff
@@ -567,38 +597,94 @@ def _create_nominal_and_modifiers(self, config, spec):
...
 
     def expected_actualdata(self, pars):
+        """
+        Compute the expected value of the main measurements.
+
+        Args:
+            pars (`tensor`): The parameter values
+
+        Returns:
+            Tensor: The expected main data
```
kratsg: How is this different from `expected_data` since they both "return" the same thing?

src/pyhf/pdf.py
```diff
@@ -567,38 +597,94 @@ def _create_nominal_and_modifiers(self, config, spec):
...
 
     def expected_data(self, pars, include_auxdata=True):
+        """
+        Compute the expected value of the main measurements.
+
+        Args:
+            pars (`tensor`): The parameter values
+
+        Returns:
+            Tensor: The expected main data
```
kratsg: How is this different from `expected_actualdata` since they both "return" the same thing?

src/pyhf/pdf.py
```diff
@@ -567,38 +597,94 @@ def _create_nominal_and_modifiers(self, config, spec):
...
+        """
         tensorlib, _ = get_backend()
         pars = tensorlib.astensor(pars)
         if not include_auxdata:
             return self.make_pdf(pars).pdfobjs[0].expected_data()
         return self.make_pdf(pars).expected_data()
 
     def constraint_logpdf(self, auxdata, pars):
+        """
+        Compute the log value of the constraint term.
```
kratsg:
```suggestion
        Compute the log value of the constraint pdf.
```

src/pyhf/pdf.py
```diff
@@ -651,5 +748,16 @@ def logpdf(self, pars, data):
             raise
 
     def pdf(self, pars, data):
+        """
+        Compute the value of the full density.
```
kratsg: What is "value"? What is "full density"?

.github/workflows/ci.yml
matthewfeickert: @kratsg Why do you only want to run [107 instead of all of group 1](http://www.pydocstyle.org/en/latest/error_codes.html)?

kratsg: I thought you only wanted to fix the issue related to missing docstrings in `__init__`. If we want it to pass only on pdf.py -- then should make sure we have no missing docstrings there?

src/pyhf/pdf.py
lukasheinrich: could be singular, but then again  each bin (or channel) is more or less its own measurement

src/pyhf/pdf.py
lukasheinrich: we can do this, but I'd reserve this for another PR, this is mostly about the CI testing about docstring existence

src/pyhf/pdf.py
lukasheinrich: this returns the non-aux slice of the data

src/pyhf/pdf.py
lukasheinrich: the full model (constraint + main).. value is the density at a given observed point in data space

.github/workflows/ci.yml
matthewfeickert: > If we want it to pass only on pdf.py -- then should make sure we have no missing docstrings there?

Yes. `D1` checks all of the error codes in the 1 group. So as `pdf.py` passes `D100` through `D107` then why not run with that?

kratsg: Oh, I misunderstood.

src/pyhf/pdf.py
kratsg: Change the phrasing.

src/pyhf/pdf.py
kratsg: Change the phrasing.

src/pyhf/pdf.py
kratsg: Change the phrasing.

matthewfeickert:
@lukasheinrich As this was rebased to fix some issues can you please `reset --hard` against this before making the requested changes?

src/pyhf/pdf.py
lukasheinrich: to?

src/pyhf/pdf.py
lukasheinrich: to what?

src/pyhf/pdf.py
kratsg: "Compute the density at a given observed point in data space of the full model"

src/pyhf/pdf.py
kratsg: "Tensor: The expected data of the main pdf (no auxiliary data)"

PR #607, opened 2019-10-10
revert: Drop support for MXNet

### LINKED ISSUE #86: implement MXNet autograd based optimizer (newton's method)

### LINKED ISSUE #593: Drop MXNet Support
# Description

MXNet isn't going to be implementing statistical distributions anytime soon, so it seems like they will be diverging from the rest of our ML framework backends even more. Instead of keeping MXNet around as a vestigial dependency I would suggest we just drop support for it.

Thoughts @kratsg @lukasheinrich?

matthewfeickert:
# Description

Resolves #593 

As MXNet doesn't have full enough support and is currently a bit of overhead in the tests this drops support of MXNet for the time being. It may come back in the future, but there are no plans to do so unless MXNet improves drastically its support of statistical distributions.

Closes #86 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Drop support of MXNet indefinitely, but not necessarily permanently
```


matthewfeickert:
> sad to see you go mxnet!

Indeed. Though that's why I say

>  indefinitely, but not necessarily permanently

as I hope things will be different in the future.

PR #617, opened 2019-10-16
feat: pyhf spec CLI

### LINKED ISSUE #663: use sphinx.ext.napoleon, not sphinxcontrib.napoleon
# Description

As stated here: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/

> As of Sphinx 1.3, the napoleon extension will come packaged with Sphinx under sphinx.ext.napoleon. The sphinxcontrib.napoleon extension will continue to work with Sphinx <= 1.2.

I'll fix this in #617 .

kratsg:
# Description

- [x] `pyhf prune --channel ... --sample ... --modifier ... --modifier-type ... --measurement ... spec.json`
- [x] `pyhf rename --channel ... --sample ... --modifier ... --measurement ... spec.json`
- [x] `pyhf combine left.json right.json`

This requires:
- #657
- #656

Addresses (but does not resolve) an issue in #660 with `pyhf.Workspace` not showing up in our docs.

Resolves #663.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add specification operations to the workspace (prune, rename, combine)
* Add CLI entrypoints for prune, rename, combine
* Add API docs
* Update doc dependencies
* Fix docstrings for workspace.py
```

### DIFF
--- a/src/pyhf/cli/spec.py
+++ b/src/pyhf/cli/spec.py
@@ -122,3 +122,38 @@ def inspect(workspace, output_file, measurement):
         with open(output_file, 'w+') as out_file:
             json.dump(result, out_file, indent=4, sort_keys=True)
         log.debug("Written to {0:s}".format(output_file))
+
+
+@cli.command()
+@click.argument('workspace', default='-')
+@click.option(
+    '--output-file',
+    help='The location of the output json file. If not specified, prints to screen.',
+    default=None,
+)
+@click.option('-c', '--channel', default=[], multiple=True)
+@click.option('-s', '--sample', default=[], multiple=True)
+@click.option('-m', '--modifier', default=[], multiple=True)
+@click.option('-t', '--modifier-type', default=[], multiple=True)
+@click.option('-e', '--measurement', default=[], multiple=True)
+def prune(
+    workspace, output_file, channel, sample, modifier, modifier_type, measurement
+):
+    with click.open_file(workspace, 'r') as specstream:
+        spec = json.load(specstream)
+
+    ws = Workspace(spec)
+    pruned_ws = ws.prune(
+        channels=channel,
+        samples=sample,
+        modifiers=modifier,
+        modifier_types=modifier_type,
+        measurements=measurement,
+    )
+
+    if output_file is None:
+        click.echo(json.dumps(pruned_ws, indent=4, sort_keys=True))
+    else:
+        with open(output_file, 'w+') as out_file:
+            json.dump(pruned_ws, out_file, indent=4, sort_keys=True)
+        log.debug("Written to {0:s}".format(output_file))
--- a/src/pyhf/cli/spec.py
+++ b/src/pyhf/cli/spec.py
@@ -202,7 +202,6 @@ def prune(
     metavar='<PATTERN> <REPLACE>...',
 )
 @click.option(
-    '-e',
     '--measurement',
     default=[],
     multiple=True,
--- a/src/pyhf/cli/spec.py
+++ b/src/pyhf/cli/spec.py
@@ -243,7 +243,7 @@ def combine(workspace_one, workspace_two, output_file):
 
     ws_one = Workspace(spec_one)
     ws_two = Workspace(spec_two)
-    combined_ws = ws_one + ws_two
+    combined_ws = ws_one.combine(ws_two)
 
     if output_file is None:
         click.echo(json.dumps(combined_ws, indent=4, sort_keys=True))
--- a/src/pyhf/cli/spec.py
+++ b/src/pyhf/cli/spec.py
@@ -131,11 +131,19 @@ def inspect(workspace, output_file, measurement):
     help='The location of the output json file. If not specified, prints to screen.',
     default=None,
 )
-@click.option('-c', '--channel', default=[], multiple=True)
-@click.option('-s', '--sample', default=[], multiple=True)
-@click.option('-m', '--modifier', default=[], multiple=True)
-@click.option('-t', '--modifier-type', default=[], multiple=True)
-@click.option('-e', '--measurement', default=[], multiple=True)
+@click.option('-c', '--channel', default=[], multiple=True, metavar='<CHANNEL>...')
+@click.option('-s', '--sample', default=[], multiple=True, metavar='<SAMPLE>...')
+@click.option('-m', '--modifier', default=[], multiple=True, metavar='<MODIFIER>...')
+@click.option(
+    '-t',
+    '--modifier-type',
+    default=[],
+    multiple=True,
+    type=click.Choice(modifiers.uncombined.keys()),
+)
+@click.option(
+    '-e', '--measurement', default=[], multiple=True, metavar='<MEASUREMENT>...'
+)
 def prune(
     workspace, output_file, channel, sample, modifier, modifier_type, measurement
 ):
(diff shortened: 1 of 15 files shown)


src/pyhf/cli/rootio.py
```diff
@@ -0,0 +1,100 @@
...
+@click.option('-p', '--patch', multiple=True)
+def json2xml(workspace, output_dir, specroot, dataroot, resultprefix, patch):
+    try:
+        import uproot
+
+        assert uproot
+    except ImportError:
+        log.error(
+            "json2xml requires uproot, please install pyhf using the "
+            "xmlio extra: pip install pyhf[xmlio] or install uproot "
```
matthewfeickert: Maybe we shouldn't advocate for installing `uproot` by itself but only just through `pyhf[xmlio]`? That was we are assured that `uproot` will always be a version that is consistent with pyhf.

```suggestion
            "xmlio extra: pip install pyhf[xmlio]."
```

src/pyhf/cli/rootio.py
```diff
@@ -0,0 +1,100 @@
...
+            "manually: pip install uproot"
+        )
+    from .. import writexml
+
+    ensure_dirs(output_dir)
+    with click.open_file(workspace, 'r') as specstream:
+        d = json.load(specstream)
```
matthewfeickert: Why `d` instead of something more readable like `spec`?

```suggestion
        spec = json.load(specstream)
```

src/pyhf/schemas/1.0.0/model.json
```diff
@@ -3,7 +3,7 @@
     "$id": "https://scikit-hep.org/pyhf/schemas/1.0.0/model.json",
     "type": "object",
     "properties": {
-        "channels": { "type": "array", "items": {"$ref": "defs.json#/definitions/channel"} },
+        "channels": { "type": "array", "items": {"$ref": "defs.json#/definitions/channel"}, "minItems": 1 },
```
matthewfeickert: As this changes the schema can we bump it to `1.0.1`?

src/pyhf/cli/rootio.py
kratsg: this is a comment on something I never changed though. It was always `d` before.

src/pyhf/schemas/1.0.0/model.json
kratsg: This actually doesn't change the schema. Empty channels was always going to break our code. Unless people are dealing with empty channels a lot, this is fixing a problem on our side.

src/pyhf/cli/rootio.py
matthewfeickert: Ah yeah. I guess this all [got started back in PR #205](https://github.com/scikit-hep/pyhf/pull/205/files#diff-ae6f1fd62a5763e85e44acfd7c788873R36), which I didn't review and so never complained about then. :P Since you're moving things around which will create a change anyway would you be okay with changing this as well?

src/pyhf/workspace.py
```diff
@@ -161,3 +161,186 @@ def data(self, model, with_aux=True):
         if with_aux:
             observed_data += model.config.auxdata
         return observed_data
+
+    def _prune_and_rename(
```
lukasheinrich: could this maybe more generally be a visitor-like pattern in which prune returns `None`, and rename returns the renamed object?

kratsg: If you're thinking of traversal, I tried.. but if you have a better way to implement it... The `_prune_and_rename` isn't a public api though. If you prefer, i can just copy/paste that code into the two methods and get rid of this one.

matthewfeickert:
Here's the Read the Docs build for the webpage: https://pyhf-staging.readthedocs.io/en/feat-cli-pruning/

@kratsg There seems to be an issue with the rendering of the [exception highlighting for `pyhf.exceptions.InvalidMeasurement`](https://pyhf-staging.readthedocs.io/en/feat-cli-pruning/_generated/pyhf.workspace.Workspace.html#pyhf.workspace.Workspace.get_measurement). Also still need to add the CLI.

kratsg:
> @kratsg There seems to be an issue with the rendering of the [exception highlighting for `pyhf.exceptions.InvalidMeasurement`](https://pyhf-staging.readthedocs.io/en/feat-cli-pruning/_generated/pyhf.workspace.Workspace.html#pyhf.workspace.Workspace.get_measurement). Also still need to add the CLI.

That's an issue filed in #663 .

matthewfeickert:
> That's an issue filed in #663 .

Yup. I saw the issue after I posted. :)

src/pyhf/workspace.py
```diff
@@ -161,3 +163,186 @@ def data(self, model, with_aux=True):
...
+        prune_measurements=[],
+        rename_modifiers={},
+        rename_samples={},
+        rename_channels={},
+        rename_measurements={},
+    ):
+        """
+        Return a new, pruned, renamed workspace specification. This will not modify the original workspace.
+
+        The pruned, renamed workspace must also be a valid workspace.
```
lukasheinrich: I think we need to somewhere define what we mean by prunig, either here, or in the docs

kratsg: done.

src/pyhf/cli/spec.py
```diff
@@ -122,3 +122,132 @@ def inspect(workspace, output_file, measurement):
...
+@click.option('-m', '--modifier', default=[], multiple=True, metavar='<MODIFIER>...')
+@click.option(
+    '-t',
+    '--modifier-type',
+    default=[],
+    multiple=True,
+    type=click.Choice(modifiers.uncombined.keys()),
+)
+@click.option(
+    '-e', '--measurement', default=[], multiple=True, metavar='<MEASUREMENT>...'
```
matthewfeickert: why `e`? This doesn't seem super intuitive to me. I realize that `m` is already taken.

lukasheinrich:
could just have no short option name, since probably thiis sill not we used
all to often anyways

matthewfeickert:
@kratsg The cross listing between `spec` and `workspace` is super nice! :)

src/pyhf/workspace.py
```diff
@@ -161,3 +163,195 @@ def data(self, model, with_aux=True):
...
+        """
+        common_channels = set(left.channels).intersection(right.channels)
+        if common_channels:
+            raise exceptions.InvalidWorkspaceOperation(
+                "Workspaces cannot have any channels in common: {}".format(
+                    common_channels
+                )
+            )
+
+        common_measurements = set(left.measurement_names).intersection(
```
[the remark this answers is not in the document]
kratsg: the common POI is fine, but we didn't discuss how to handle merging parameter configurations under a given measurement.. and that's a bit complicated because combining workspaces might end up with configuring a given parameter two different ways...

lukasheinrich:
don't we have the paramset reduction for that to check compatibility?

src/pyhf/workspace.py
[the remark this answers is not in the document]
lukasheinrich: don't we have the paramset reduction for that to check compatibility?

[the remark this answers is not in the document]
kratsg: Good point. I've removed common measurements and just checked for incompatible POI.

Issue #469, opened 2019-05-07
Drop Python 2.7 support


matthewfeickert:
# Description

Given 

> DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.

can be viewed as something to look forward to (so we can be free to Travis CI issues among other things), pyhf should drop Python 2.7 support in the last week of December 2019.

Once this is done, a future Issue should be created to move all print statements to using f-strings (c.f. [PEP 498 -- Literal String Interpolation](https://www.python.org/dev/peps/pep-0498/)).

lukasheinrich:
hm, as long as we don't have a blocking reason to drop it I think i'd still keep it. Given that all CERN Centos 7 still has python 2.7 as its default. I agree in general with pushing for depreciation and perhaps we won't *support* 2.7 but we shouldn't intentionally break it either

matthewfeickert:
> Given that all CERN Centos 7 still has python 2.7 as its default. I agree in general with pushing for depreciation and perhaps we won't _support_ 2.7 but we shouldn't intentionally break it either

My main motivation for trying to force people to use Python 3 is related to this as I would guess (though maybe you know more than me?) that CERN will be keeping Python 2.7 alive for a long time to come, and it might be easier to make a clean break. While I don't think that having pyhf working on Python 2.7 in the future is a _bad_ thing, it seems almost inevitable that if we drop testing it and drop support then as we continue to develop we'll break something Python 2.7 related. If part of Python 2.7 is broken, and we aren't going to fix it, then why not go all in on Python 3? Or would you suggest that we just wait until we get our first report of Python 2.7 being broken?

kratsg:
We are already using dependencies that are sunsetting/dropping python 2.7 support. It's going to get _very_ difficult to start maintaining things if we keep it up...

lukasheinrich:
yeah I think if we use core components in numpy that will need py3 or similar i'm fine .. I guess what I'm saying is I might reconsider to break 2.7  if it's only due to f-strings

matthewfeickert:
>  I guess what I'm saying is I might reconsider to break 2.7 if it's only due to f-strings

Sure. Perhaps a compromise where we drop support for Python 2.7 but then keep testing it in the CI and then once we find that the CI breaks on it due to Python 2 vs. 3 issues/dependencies we can move to f-strings?

[...]

PR #623, opened 2019-10-25
test: Add regression test on ATLAS sbottom public likelihood

### LINKED ISSUE #518: add public sbottom likelihood as test
# Description

once public we would want to add a wide range of public likelihoods into the CI as integration tests. Not breaking these likelihoods will be a priority of pyhf for future pyhf releases and probably should be an arbiter on whether the pyhf version in the spec needs to be updated for breaking changes

matthewfeickert:
# Description

Resolves #518 

Add a regression test that uses the observed and expected CLs values of the [ATLAS sbottom published likelihood](https://www.hepdata.net/record/ins1748602)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add regression test on ATLAS sbottom public likelihood
   - c.f. https://www.hepdata.net/record/ins1748602
```

### DIFF
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -4,6 +4,7 @@ import tarfile
 import json
 import os
 import pyhf
+import numpy as np
 
 
 # curl -sL https://www.hepdata.net/record/resource/997020?view=true
@@ -68,13 +69,18 @@ def test_sbottom_regionA(regionA_bkgonly_json, regionA_signal_patch_json):
     CLs_obs = result[0].tolist()[0]
     CLs_exp = result[-1].ravel().tolist()
     assert CLs_obs == pytest.approx(0.24443635754482018, rel=1e-6)
-    assert CLs_exp == pytest.approx(
-        [
-            0.09022521939741368,
-            0.19378411715432514,
-            0.3843236961508878,
-            0.6557759457699649,
-            0.8910421945189615,
-        ],
-        rel=1e-6,
+    assert np.all(
+        np.isclose(
+            np.array(CLs_exp),
+            np.array(
+                [
+                    0.09022521939741368,
+                    0.19378411715432514,
+                    0.3843236961508878,
+                    0.6557759457699649,
+                    0.8910421945189615,
+                ]
+            ),
+            rtol=1e-6,
+        )
     )
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -1,9 +1,9 @@
-import pyhf
+import pytest
 import requests
 import tarfile
 import json
 import os
-import pytest
+import pyhf
 
 
 # def test_sbottom_download():
@@ -27,6 +27,7 @@ import pytest
 
 @pytest.fixture(scope='module')
 def sbottom_likelihoods_download():
+    """Download the sbottom likelihoods tarball from HEPData"""
     sbottom_HEPData_URL = "https://www.hepdata.net/record/resource/997020?view=true"
     targz_filename = "sbottom_workspaces.tar.gz"
     # Download the tar.gz of the likelihoods
@@ -44,6 +45,7 @@ def sbottom_likelihoods_download():
 
 @pytest.fixture(scope='module')
 def regionA_bkgonly_json(sbottom_likelihoods_download):
+    """Extract the background only model from sbottom Region A"""
     tarfile = sbottom_likelihoods_download
     bkgonly_json_data = (
         tarfile.extractfile(tarfile.getmember("RegionA/BkgOnly.json"))
@@ -55,6 +57,7 @@ def regionA_bkgonly_json(sbottom_likelihoods_download):
 
 @pytest.fixture(scope='module')
 def regionA_signal_patch_json(sbottom_likelihoods_download):
+    """Extract a signal model from sbottom Region A"""
     tarfile = sbottom_likelihoods_download
     signal_patch_json_data = (
         tarfile.extractfile(tarfile.getmember("RegionA/patch.sbottom_1300_205_60.json"))
@@ -64,7 +67,7 @@ def regionA_signal_patch_json(sbottom_likelihoods_download):
     return json.loads(signal_patch_json_data)
 
 
-def test_gzip_file(regionA_bkgonly_json, regionA_signal_patch_json):
+def test_sbottom_regionA(regionA_bkgonly_json, regionA_signal_patch_json):
     bkg_json = regionA_bkgonly_json
     signal_json = regionA_signal_patch_json
     workspace = pyhf.workspace.Workspace(bkg_json)
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -34,6 +34,30 @@ def regionA_bkgonly_json(sbottom_likelihoods_download):
 
 
 @pytest.fixture(scope='module')
+def regionB_bkgonly_json(sbottom_likelihoods_download):
+    """Extract the background only model from sbottom Region B"""
+    tarfile = sbottom_likelihoods_download
+    bkgonly_json = (
+        tarfile.extractfile(tarfile.getmember("RegionB/BkgOnly.json"))
+        .read()
+        .decode("utf8")
+    )
+    return json.loads(bkgonly_json)
+
+
+@pytest.fixture(scope='module')
+def regionC_bkgonly_json(sbottom_likelihoods_download):
+    """Extract the background only model from sbottom Region C"""
+    tarfile = sbottom_likelihoods_download
+    bkgonly_json = (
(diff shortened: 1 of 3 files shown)


matthewfeickert:
For the Region C test I had to raise the tolerance to `1e-4` as Python 2.7 on Travis CI was failing for one of the CLs_exp points with the tolerance at `1e-5`. However, all the Python 3 instances on GitHub actions passed this fine, so once we drop Python 2.7 in Issue #469 then we can set the tolerance back to `1e-5` again.

matthewfeickert:
If there are more intelligent or more interesting points to be testing let me know and I will happily add them.

kratsg:
> For the Region C test I had to raise the tolerance to `1e-4` as Python 2.7 on Travis CI was failing for one of the CLs_exp points with the tolerance at `1e-5`. However, all the Python 3 instances on GitHub actions passed this fine, so once we drop Python 2.7 in Issue #469 then we can set the tolerance back to `1e-5` again.

Just drop python 2.7. It's not even worth getting it to pass. We made all of the validations for the pub note on python3 where possible.

Where are you getting those numbers from btw? For RegionC, I have

```
$ cat data/outputs/results/pyhf/RegionC/pyhf_result_sbottom_1000_205_60.json 
{
    "CLs_exp": [
        0.8906470732412857,
        0.9280262743211622,
        0.9614288407343238,
        0.9857553063165135,
        0.9971958190844394
    ],
    "CLs_obs": 0.9424009499519606
}```

Additionally, with the way the tests are written, it's very hard to understand what results we're comparing because there's way too much modularization...

File: tests/test_regression.py [hunk 1]
```diff
@@ -0,0 +1,163 @@
...
+import os
+import pyhf
+import numpy as np
+
+
+@pytest.fixture(scope='module')
+def sbottom_likelihoods_download():
+    """Download the sbottom likelihoods tarball from HEPData"""
+    sbottom_HEPData_URL = "https://www.hepdata.net/record/resource/997020?view=true"
+    targz_filename = "sbottom_workspaces.tar.gz"
```
lukasheinrich: should we compare some download checksums here?

matthewfeickert:
> Just drop python 2.7. It's not even worth getting it to pass. We made all of the validations for the pub note on python3 where possible.

I'll let you and @lukasheinrich discuss this here, as in the past when this has gotten brought up we've agreed based on his concerns to drop Python 2.7 at the end of the year after having cut a final release so that we avoid breaking people who are Python 2.7 bound for as long as possible.
 
> Where are you getting those numbers from btw? For RegionC, I have
> 
> ```
> $ cat data/outputs/results/pyhf/RegionC/pyhf_result_sbottom_1000_205_60.json 
> {
>     "CLs_exp": [
>         0.8906470732412857,
>         0.9280262743211622,
>         0.9614288407343238,
>         0.9857553063165135,
>         0.9971958190844394
>     ],
>     "CLs_obs": 0.9424009499519606
> }
> ``` 

```
$ curl -sL https://www.hepdata.net/record/resource/997020?view=true \
   | tar -O -xzv RegionC/BkgOnly.json \
   | pyhf cls --patch \
    <(curl -sL https://www.hepdata.net/record/resource/997020?view=true \
   | tar -O -xzv RegionC/patch.sbottom_1000_205_60.json)
RegionC/BkgOnly.json
RegionC/patch.sbottom_1000_205_60.json
{
    "CLs_exp": [
        0.8906506884676416,
        0.9280287127442725,
        0.9614301796189283,
        0.9857558128338463,
        0.9971959212073871
    ],
    "CLs_obs": 0.9424021663134358
}
```

> Additionally, with the way the tests are written, it's very hard to understand what results we're comparing because there's way too much modularization...

How so? Beyond the fixtures, which I think are reasonable, there are only two functions, and given the boilerplate nature of them I don't see how having them as functions is a problem. Can  you suggest a better naming scheme if things are confusing?

tests/test_regression.py
matthewfeickert: This is a good idea.

kratsg:
> How so? Beyond the fixtures, which I think are reasonable, there are only two functions, and given the boilerplate nature of them I don't see how having them as functions is a problem. Can you suggest a better naming scheme if things are confusing?

When you run the test, it only states "RegionA" -- it doesn't list the mass point that was run over. That should be part of the test name.

You should use my numbers for the RegionC -- since that's what the pub note was published on (which means we're regressing to the exact version of pyhf, but had to delay adding the test).

matthewfeickert:
> When you run the test, it only states "RegionA" -- it doesn't list the mass point that was run over. That should be part of the test name.

That's fair. My thought was to keep the naming simple and then if multiple points were going to be tested change the name then, but you're right that it is better to be explicit up front now rather then have someone have to deal with name confusion or conflicts in the future. Let me know if you want changes to the modifications I've made.
 
> You should use my numbers for the RegionC -- since that's what the pub note was published on (which means we're regressing to the exact version of pyhf, but had to delay adding the test).

Sure. I've added them for Region C. Can you either give me these or add them for Region A and Region B?

tests/test_regression.py
matthewfeickert: @lukasheinrich Done.

lukasheinrich:
I think it woudl be sueful to add some points close to the contour e.g. 1400,1000 or so

kratsg:
```
Lord Stark:~/sbottom-lhood-pubnote-code (triple_overlay)$ cat data/outputs/results/pyhf/RegionA/*1300_205_60*
{
    "CLs_exp": [
        0.09022509053507759,
        0.1937839194960632,
        0.38432344933992,
        0.6557757334303531,
        0.8910420971601081
    ],
    "CLs_obs": 0.24443627759085326
}Lord Stark:~/sbottom-lhood-pubnote-code (triple_overlay)$ cat data/outputs/results/pyhf/RegionA/*1400_950_60*
{
    "CLs_exp": [
        0.002644707461012826,
        0.013976754489151644,
        0.06497313811425813,
        0.23644505123524753,
        0.5744843501873754
    ],
    "CLs_obs": 0.021373283911064852
}Lord Stark:~/sbottom-lhood-pubnote-code (triple_overlay)$ cat data/outputs/results/pyhf/RegionA/*1500_850_60*
{
    "CLs_exp": [
        0.0059847029077065295,
        0.026103516126601122,
        0.10093985752614597,
        0.3101988586187604,
        0.6553686728646031
    ],
    "CLs_obs": 0.04536774062150508
}Lord Stark:~/sbottom-lhood-pubnote-code (triple_overlay)$ cat data/outputs/results/pyhf/RegionB/*1500_850_60*
{
    "CLs_exp": [
        0.9999682931142808,
        0.9999796235899027,
        0.9999893398756928,
        0.9999961574993774,
        0.999999261851925
    ],
    "CLs_obs": 0.9999893250416583
}Lord Stark:~/sbottom-lhood-pubnote-code (triple_overlay)$ cat data/outputs/results/pyhf/RegionC/*1000_205_60*
{
    "CLs_exp": [
        0.8906470732412857,
        0.9280262743211622,
        0.9614288407343238,
        0.9857553063165135,
        0.9971958190844394
    ],
    "CLs_obs": 0.9424009499519606
}
```

matthewfeickert:
@kratsg Can you give me a different point for Region B than `1500_850_60`? Unless @lukasheinrich and you can agree on when to drop Python 2.7 support then the point you chose for Region B will require me to raise the tolerance from `1e-5` to `1e-4` for Python 2.7 on Travis CI.

kratsg:
> raise the tolerance from `1e-5` to `1e-4`

You mean lower, but ok... try this one?

```
$ cat data/outputs/results/pyhf/RegionB/pyhf_result_sbottom_1400_550_60.json 
{
    "CLs_exp": [
        0.9338879894557114,
        0.9569045303300702,
        0.9771296335437559,
        0.9916370124133669,
        0.9983701133999316
    ],
    "CLs_obs": 0.9744675266677597
}
```

matthewfeickert:
> You mean lower, but ok... try this one?

Yes. Sorry.

strike-that

Also that works now, so thanks.

matthewfeickert:
@kratsg Can I also get output for Region C point `1600_850_60`?

kratsg:
```
$ cat ~/sbottom-lhood-pubnote-code/data/outputs/results/pyhf/RegionC/*1600_850_60*
{
    "CLs_exp": [
        0.2955492909588046,
        0.4446885457298284,
        0.6371473864200973,
        0.8336149623750603,
        0.9585901381554178
    ],
    "CLs_obs": 0.711023707425625
}
```

matthewfeickert:
@kratsg @lukasheinrich I think this is ready for review again now.

matthewfeickert:
Thanks for the helpful reviews on this &mdash; it definitely needed them.

Issue #583, opened 2019-09-25
move to /add TF and PyTorch  BFGS optimizers


lukasheinrich:
# Description

our use of optimizers for TF and PyTorch is a bit rudimentary. Both TF and Pytorch include BGFS optimizers

* https://pytorch.org/docs/stable/_modules/torch/optim/lbfgs.html
* https://www.tensorflow.org/probability/api_docs/python/tfp/optimizer/bfgs_minimize

which is also the algorithm on which our scipy optimizer is based (SLSQP)

PR #625, opened 2019-11-02
feat: Add autodiff driven SciPy optimizer

lukasheinrich:
# Description

This adds a optimizer based on scipy. The objective and gradient are computed by the respective backend but the parameter stepping is done by `scipy.minimize` since #583 seems to be blocked by https://github.com/tensorflow/probability/issues/637 and additionally it's not clear we can implement parameter bounds (which in scipy need `SLSQP` method)

this optimizer is subclassable by any autodifferentiable backend.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Drop tensor-ful hand-written optimization for scipy.minimize
* Leverage autodiff capability in TensorFlow and PyTorch to pass exact gradients/jacobians
* Can reproduce sbottom contour w/ TF and PyTorch (PyTorch is faster than NumPy)
```


matthewfeickert:
@lukasheinrich Is this ready for review, or is there further WIP?

lukasheinrich:
yeah this is ready for review

lukasheinrich:
this is ready for review @kratsg @matthewfeickert

src/pyhf/optimize/opt_tflow.py
```diff
@@ -1,76 +1,55 @@
...
+        data = tf.placeholder(tf.float32, (m.config.nmaindata + m.config.nauxdata,))
+        nll = objective(pars, data, m)
+        nllgrad = tf.identity(tf.gradients(nll, pars)[0])
+        func = eval_func([nll, nllgrad], pars, data, d)
+        return func, init_pars, par_bounds
+
+    def setup_constrained(self, objective, poival, m, d, init_pars, par_bounds):
+        tensorlib, _ = get_backend()
+        data = tf.placeholder(tf.float32, (m.config.nmaindata + m.config.nauxdata,))
+        idx = default_backend.astensor(range(m.config.npars), dtype='int')
```
kratsg: probably should support `tensorlib.range`

src/pyhf/pdf.py
```diff
@@ -422,6 +422,7 @@ def __init__(self, spec, batch_size=None, **config_kwargs):
             if hasattr(parset, 'pdf_type'):  # is constrained
                 self.config.auxdata += parset.auxdata
                 self.config.auxdata_order.append(k)
+        self.config.nauxdata = len(self.config.auxdata)
```
kratsg: why does this get set later instead of at `init` time?

src/pyhf/optimize/opt_tflow.py
lukasheinrich: PRs welcome :)

src/pyhf/pdf.py
lukasheinrich: because the stuff that `appends` to auxdata is done here see comment above

```
        # this is tricky, must happen before constraint
        # terms try to access auxdata but after
        # combined mods have been created that
        # set the aux data
```

src/pyhf/optimize/autodiff.py
```diff
@@ -0,0 +1,40 @@
...
+        )
+        unconstr_pars = fitresult.x
+        tensorlib, _ = get_backend()
+        return tensorlib.astensor(unconstr_pars)
+
+    def constrained_bestfit(
+        self, objective, constrained_mu, data, pdf, init_pars, par_bounds
+    ):
+        poival = constrained_mu
+        for_constr = self.setup_constrained(
```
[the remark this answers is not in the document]
lukasheinrich: done

src/pyhf/optimize/autodiff.py
```diff
@@ -0,0 +1,40 @@
+import scipy
+from .. import default_backend
+from .. import get_backend
+
+
+class AutoDiffOptimizerMixin(object):
+    def unconstrained_bestfit(self, objective, data, pdf, init_pars, par_bounds):
+        for_unconstr = self.setup_unconstrained(
```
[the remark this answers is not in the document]
lukasheinrich: done

lukasheinrich:
pushed some more commits. PTAL

src/pyhf/pdf.py
kratsg: i see comment now. yes. this makes sense... not obvious initially.

src/pyhf/optimize/opt_tflow.py
kratsg: #636 is filed for now.

src/pyhf/optimize/opt_pytorch.py
```diff
@@ -1,58 +1,46 @@
...
+
+
+class pytorch_optimizer(AutoDiffOptimizerMixin):
+    def __init__(self, *args, **kargs):
+        pass
+
+    def setup_unconstrained(self, objective, pdf, data, init_pars, par_bounds):
+        def func(pars):
+            tensorlib, _ = get_backend()
+            pars = tensorlib.astensor(pars)
```
kratsg: why not `pars = tensorlib.astensor(pars, requires_grad=True)`?

src/pyhf/optimize/opt_pytorch.py
```diff
@@ -1,58 +1,46 @@
...
+            pars.requires_grad = True
+            r = objective(pars, data, pdf)
+            grad = torch.autograd.grad(r, pars)[0]
```
kratsg: what's the `[0]` for? do we throw away the rest? Why not

`grad, _ = torch.autograd.grad(r, pars)`

kratsg:
```suggestion
            grad, _ = torch.autograd.grad(r, pars)
```

src/pyhf/optimize/opt_pytorch.py
kratsg:
```suggestion
            pars = tensorlib.astensor(pars, requires_grad=True)
```

src/pyhf/optimize/opt_pytorch.py
```diff
@@ -1,58 +1,46 @@
...
+            return r.detach().numpy(), grad
```
kratsg: why detach?

src/pyhf/optimize/opt_pytorch.py
matthewfeickert: Just to check, the only reason that we need to explicitly cast as a tensor here is because these `pars` are being passed into `func` by the internals of

```
fitresult = scipy.optimize.minimize(func, init, method='SLSQP', jac=True, bounds=bounds)
```

which won't be using tensors, correct? We can't avoid having to do this cast?

src/pyhf/optimize/opt_pytorch.py
```diff
@@ -1,58 +1,46 @@
...
+
+        return func, init_pars, par_bounds
+
+    def setup_constrained(self, objective, poival, pdf, data, init_pars, par_bounds):
+        tensorlib, _ = get_backend()
+        idx = default_backend.astensor(range(pdf.config.npars), dtype='int')
+        init = default_backend.astensor(init_pars)
```
matthewfeickert: See question above RE: if the `astensor` calls that don't explicitly transform the `dtype` are needed as a result of use in `scipy.optimize.minimize`.

src/pyhf/optimize/opt_pytorch.py
```diff
@@ -1,58 +1,46 @@
...
+        nuisinit = default_backend.concatenate(
+            [init[: pdf.config.poi_index], init[pdf.config.poi_index + 1 :]]
+        ).tolist()
+        nuisidx = default_backend.concatenate(
+            [idx[: pdf.config.poi_index], idx[pdf.config.poi_index + 1 :]]
+        ).tolist()
+        nuisbounds = [par_bounds[i] for i in nuisidx]
+        tv = _TensorViewer([[pdf.config.poi_index], nuisidx])
+
+        data = tensorlib.astensor(data)
```
matthewfeickert: See question above RE: if the `astensor` calls that don't explicitly transform the `dtype` are needed as a result of use in `scipy.optimize.minimize`.

src/pyhf/optimize/opt_pytorch.py
```diff
@@ -1,58 +1,46 @@
...
+        poivals = tensorlib.astensor([poival], dtype='float')
+
+        def func(pars):
+            pars = tensorlib.astensor(pars)
```
matthewfeickert: See question above RE: if the `astensor` calls that don't explicitly transform the `dtype` are needed as a result of use in `scipy.optimize.minimize`.

src/pyhf/optimize/opt_tflow.py
```diff
@@ -1,76 +1,64 @@
...
+        data_placeholder = tf.placeholder(
+            tf.float32, (pdf.config.nmaindata + pdf.config.nauxdata,)
         )
-        objective = objective(pars, data, pdf)
-        hessian = tf.hessians(objective, nuis_cat)[0]
-        gradient = tf.gradients(objective, nuis_cat)[0]
-        invhess = tf.linalg.inv(hessian)
-        update = tf.transpose(tf.matmul(invhess, tf.transpose(tf.stack([gradient]))))[0]
+        idx = default_backend.astensor(range(pdf.config.npars), dtype='int')
+        init = default_backend.astensor(init_pars)
```
matthewfeickert: See question above RE: if the `astensor` calls that don't explicitly transform the `dtype` are needed as a result of use in `scipy.optimize.minimize`.

src/pyhf/optimize/opt_pytorch.py
lukasheinrich: because scipy doesn't know anything about tensors other than numpy, so we just use the backedns to compute the logpdf -> cast to numpy, and scipy.optimize will return a new parameter vector (as numpy array) at the location where it wants to evaluate a likelihood nexxt

src/pyhf/optimize/autodiff.py
```diff
@@ -0,0 +1,36 @@
+import scipy
+from .. import default_backend
+from .. import get_backend
+
+
+class AutoDiffOptimizerMixin(object):
+    def unconstrained_bestfit(self, objective, data, pdf, init_pars, par_bounds):
+        func, init, bounds = self.setup_unconstrained(
+            objective, pdf, data, init_pars, par_bounds
```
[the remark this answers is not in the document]
lukasheinrich: done

src/pyhf/optimize/opt_pytorch.py
lukasheinrich: it's not  tuple return, but a one-element list.. i'd chose `a,_` only for tuple returns tbh

src/pyhf/optimize/opt_pytorch.py
lukasheinrich: because it needs to be detached from the gradient calculation in  pytorch before it can be cast to numpy

src/pyhf/optimize/opt_tflow.py
```diff
@@ -1,76 +1,55 @@
...
-                nuis_cat[: pdf.config.poi_index],
-                poi_par,
-                nuis_cat[pdf.config.poi_index :],
-            ]
+def eval_func(op, argop, dataop, d):
+    def func(p):
+        tensorlib, _ = get_backend()
+        xx = tensorlib.tolist(p) if isinstance(p, tf.Tensor) else p
+        yy = tensorlib.tolist(d) if isinstance(d, tf.Tensor) else d
+        value = tensorlib.session.run(op, feed_dict={argop: xx, dataop: yy})
```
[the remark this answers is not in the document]
lukasheinrich: done

src/pyhf/optimize/opt_pytorch.py
```diff
@@ -1,58 +1,46 @@
...
+    def __init__(self, *args, **kargs):
+        pass
+
+    def setup_unconstrained(self, objective, data, pdf, init_pars, par_bounds):
+        def func(pars):
+            tensorlib, _ = get_backend()
+            pars = tensorlib.astensor(pars)
+            pars.requires_grad = True
+            r = objective(pars, data, pdf)
+            grad, _ = torch.autograd.grad(r, pars)
```
kratsg:
```suggestion
            grad = torch.autograd.grad(r, pars)[0]
```

PR #638, opened 2019-11-14
ci: Migrate publishing from Travis CI to GitHub Actions

### LINKED ISSUE #508: Automate deployment to PyPI

matthewfeickert:
# Description

Resolves #508

Migrate the publication of the distributions from Travis CI to GitHub Actions using the [PyPA's "PyPI publish GitHub Action"](https://github.com/pypa/gh-action-pypi-publish) using [PyPI's API token](https://pypi.org/help/#apitoken). This workflow publishes to TestPyPI on every commit to `master` and published to PyPI on every tag that is pushed to `master`. Thus, to publish, one simply just needs to run `bumpversion` on `master` and then do `git push origin master --tags`.

Many thanks go to the PyPA team for their tool and their [well written guide to it](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/), and special thanks go to @webknjaz for his long patience and helpful advice. This workflow is based off of the PyPA example workflow and @kratsg's [workflow for `stare`](https://github.com/kratsg/stare/blob/6d063d4f12ced9c5440328b9ff12350dc53b5020/.github/workflows/publish-package-to-pypi.yml). 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add publishing workflow to GitHub Actions for TestPyPI and PyPI using API tokens
* Remove publishing from Travis CI
* Add publishing section to developer docs
```

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -1,0 +1,34 @@
+name: Publish distributions
+on:
+  push:
+    branches:
+    - master
+    tags:
+    - v*
+
+jobs:
+  build-and-publish:
+    name: Build and publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install pep517
+      run: |
+        python -m pip install pep517 --user
+    - name: Build a binary wheel and a source tarball
+      run: |
+        python -m pep517.build --source --binary --out-dir dist/ .
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.testpypi_password }}
+        repository_url: https://test.pypi.org/legacy/
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        password: ${{ secrets.pypi_password }}
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -23,7 +23,7 @@ jobs:
       run: |
         python -m pep517.build --source --binary --out-dir dist/ .
     - name: Publish distribution 📦 to Test PyPI
-      uses: pypa/gh-action-pypi-publish@master
+      uses: pypa/gh-action-pypi-publish@v1.0.0a0
       env:
         IS_TESTPYPI: ${{ true }}
       with:
@@ -32,7 +32,7 @@ jobs:
         repository_url: https://test.pypi.org/legacy/
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
-      uses: pypa/gh-action-pypi-publish@master
+      uses: pypa/gh-action-pypi-publish@v1.0.0a0
       with:
         user: __token__
         password: ${{ secrets.pypi_password }}
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -27,10 +27,12 @@ jobs:
       env:
         IS_TESTPYPI: ${{ true }}
       with:
+        user: __token__
         password: ${{ secrets.test_pypi_password }}
         repository_url: https://test.pypi.org/legacy/
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.ref, 'refs/tags')
       uses: pypa/gh-action-pypi-publish@master
       with:
+        user: __token__
         password: ${{ secrets.pypi_password }}
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -1,4 +1,4 @@
-name: Publish distributions
+name: publish distributions
 on:
   push:
     branches:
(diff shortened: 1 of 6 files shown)


matthewfeickert:
For reference, this is what the badges will look like and do now:

[GitHub Actions Status: CI](https://github.com/diana-hep/pyhf/actions?query=workflow%3ACI%2FCD+branch%3Amaster)
[GitHub Actions Status: Publish](https://github.com/diana-hep/pyhf/actions?query=workflow%3A%22Publish+distributions%22+branch%3Amaster)

matthewfeickert:
Also, here is a dummy workflow that I setup on another branch to show that the environmental variable would get picked up correctly: https://github.com/diana-hep/pyhf/commit/3cbe77f298501f7dde5227b5295c2b6cf36671e9/checks?check_suite_id=311784905#step:5:8

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,38 @@
+name: publish distributions
+on:
+  push:
+    branches:
+    - master
```
lukasheinrich: on master as well? why not only on tags?

kratsg: if you tag on non-master, you don't want it to run.

matthewfeickert: This is to trigger the TestPyPI releases. The PyPI releases are still protected with

```yaml
      if: startsWith(github.ref, 'refs/tags')
```

@kratsg I'm heavily basing this off of `stare` so let me know if I'm missing any other reason. The reason I gave is why I kept it like you had it.

matthewfeickert:
As [`__token__` is the default username in the GitHub Action](https://github.com/pypa/gh-action-pypi-publish/commit/25d67dc29880c27f76b67eda7296231916ae4eaf#diff-04c6e90faac2675aa89e2176d2eec7d8R17) it doesn't need to be specified here, so no username is given. HT @kratsg

docs/development.rst
```diff
@@ -18,3 +18,21 @@ Then setup the Git pre-commit hook for `Black <https://github.com/psf/black>`__
...
+.. code-block:: console
+
+    bumpversion [major|minor|patch]
+
+to update the release version and get a tagged commit and then push the commit
+and tag to :code:`master` with
+
+.. code-block:: console
+
+    git push origin master --tags
```
webknjaz: @kratsg this way you both publish `master` branch and **all** tags which may be problematic.

You could push the specific tag rather than pushing all, I guess.
```suggestion
    git push origin master <specific_tag_here>
```

Moreover, you could adopt some wrapper like tox/nox/invoke to wrap a number of commands and pre-validation into a script that would also protect you from creating tags in unwanted places.

The point here is that this whole process should be triggered by a single "button click" rather than have any manual actions in between because humans are error-prone...

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Install pep517
+      run: |
+        python -m pip install pep517 --user
+    - name: Build a binary wheel and a source tarball
+      run: |
+        python -m pep517.build --source --binary --out-dir dist/ .
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@v1.0.0a0
+      env:
+        IS_TESTPYPI: ${{ true }}
```
webknjaz: Why use a boolean literal here? Env vars values are always strings so why not just use a string literal?
```suggestion
        IS_TESTPYPI: "true"
```

docs/development.rst
kratsg: I would love for a way to have three buttons to click for major/minor/patch and it automatically bumpversions and pushes the tag for us.

.github/workflows/publish-package.yml
kratsg: yep, this ended up getting replaced entirely by what I did in kratsg/stare

docs/development.rst
webknjaz: I actually have a DEMO of such a button in the octomachinery repo. But the button implementation requires a GitHub App.
Basically, my app publishes a "Deploy" page via Checks API on each commit on master. That page contains a clickable button (it's a native feature of Checks on GitHub) and that click triggers a "deployment" event.
So then there's a Workflow that's triggered by the `deployment` event and does the actual work.
You can add up to 3 buttons to a single Checks page. So you could integrate it with calls to `bumpversion` with 3 different options :)
The source is here: https://github.com/sanitizers/diactoros-github-app.
atm it's not configurable but could be tweaked to be :)

kratsg: This is nice! How do you handle permissions with this (so that not _anyone_ can trigger?)

webknjaz: The button for the last commit in octomachinery is here: https://github.com/sanitizers/octomachinery/runs/289377465. You don't see it but I do. The last time I checked it only showed up for users with `write` permission on the repo. I filed an FR to make it more flexible but haven't heard back. But if you want to make ACL even stricter, the App would need to be modified to check this. The button click is an event and is delivered to the App using webhooks just like any other events.
The current button click processing is fairly simple: https://github.com/sanitizers/diactoros-github-app/blob/6e68364/diactoros/__main__.py#L60-L86. It just verifies that the button has `deploy` ID and punches GitHub API to create a deployment request, also specifying that it doesn't matter if any of the other CI checks are green (by default, API'd return an error if at the time of the API request there'd be any non-green checks posted on that commit).

webknjaz: You can join me complaining about the ACL here: https://github.community/t5/GitHub-API-Development-and/FR-Make-possible-to-show-action-buttons-posted-via-Checks-API-to/m-p/21423/highlight/true#M1342

PR #640, opened 2019-11-15
fix: Correct reference check in GitHub Actions publish to TestPyPI step

kratsg:
# Description

This adds in changes to #639. GHA triggers pushes on PRs as well... so you can't tell if a push event comes from a PR or not. Also need to fix logic of whether to use `scm` or not depending on if we're tagged or not.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* More fixes to workflow in #639, only push to testpypi on pushes on master
* Only use scm if the current commit is not tagged
* Protect forks from publishing
```

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -30,13 +30,13 @@ jobs:
         python -m pep517.build --source --binary --out-dir dist/ .
     - name: Publish distribution 📦 to Test PyPI
       # every PR will trigger a push event on master, so check the push event is actually coming from master
-      if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+      if: success() && github.event_name == 'push' && github.ref == 'refs/heads/master' && github.repository == "diana-hep/pyhf"
       uses: pypa/gh-action-pypi-publish@v1.0.0a0
       with:
         password: ${{ secrets.test_pypi_password }}
         repository_url: https://test.pypi.org/legacy/
     - name: Publish distribution 📦 to PyPI
-      if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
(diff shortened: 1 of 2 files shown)


matthewfeickert:
For posterity, to document the logic that is being used here that @kratsg had to explain to me as I missed it: Given that

> GHA triggers pushes on PRs as well... so you can't tell if a push event comes from a PR or not. 

this means that

1. As every PR will trigger a `push` even on `master` then you need to check to make sure that the reference for the the push event is _actually_ coming from `master`

```yaml
if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
2. If it is, then you have to check if the commit is tagged (and so meant for PyPI). If it is, then skip using `scm`. If it is not

```python
if getenv('IS_COMMIT_TAGGED') == 'false'
```

then use `scm` and this commit will get pushed up to TestPyPI.
3. If the commit is tagged, then push to PyPI. You shouldn't try to check here as the only information that `github.ref` can give is the tag, not which branch the tag was coming from.

```yaml
if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```

.github/workflows/publish-package.yml
```diff
@@ -26,10 +26,11 @@ jobs:
       run: |
         python -m pep517.build --source --binary --out-dir dist/ .
     - name: Publish distribution 📦 to Test PyPI
-      if: github.event_name == 'push'
+      if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
matthewfeickert: As I think this is the most confusing part of the logic, above this can we have the comment?
```suggestion
      # every PR will trigger a push event on master, so check the push event is actually coming from master
      if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```

kratsg:
We're safe from malicious PRs: https://github.com/r-lib/ghactions/issues/262#issuecomment-540233688

PR #647, opened 2019-11-21
docs: Improve discussion of publication workflow

matthewfeickert:
# Description

Extend and improve the instructions for validating and then publishing a release using GitHub Actions.

This follows up on [some discussion](https://github.com/scikit-hep/pyhf/pull/638#discussion_r346845955) on PR #638 with @webknjaz and advocates for locally merging on the command line so as to avoid a merge commit (as the branch would be just ahead of `master` and so can easily fastforward).

ReadTheDocs build: https://pyhf.readthedocs.io/en/docs-improve-publishing-docs/development.html

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Update the instructions for validating and then publishing a release using GitHub Actions
```

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -46,29 +46,14 @@ Publishing
 ----------
 
 Publishing to `PyPI <https://pypi.org/project/pyhf/>`__ and `TestPyPI <https://test.pypi.org/project/pyhf/>`__
-is automated through the `PyPA's PyPI publish GitHub Action <https://github.com/pypa/gh-action-pypi-publish>`__.
-Publishing a release to PyPI begins with creating a new branch :code:`release/cut-<release tag>`
-and then from that branch running
-
-.. code-block:: console
-
-    bumpversion [major|minor|patch]
-
-to update the release version and get a tagged commit.
-Then, push only the branch to GitHub and open a PR to ensure that the code
-you are going to publish is indeed stable.
-Once the PR has passed all checks and has been approved, a maintainer should merge
-the branch from the command line
-
-.. code-block:: console
-
-    git checkout master
-    git merge release/cut-<release tag>
-
-and then push the commit and tag to :code:`master` with
-
-.. code-block:: console
-
-    git push origin master <release tag>
-
-which will both merge and close the PR and start the release publication workflow.
+is automated through the `PyPA's PyPI publish GitHub Action <https://github.com/pypa/gh-action-pypi-publish>`__
+and the ``pyhf`` `Tag Creator GitHub Actions workflow <https://github.com/scikit-hep/pyhf/blob/master/.github/workflows/tag.yml>`__.
+A release can be created from any PR created by a core developer by adding a
+``bumpversion`` tag to it that corresponds to the release type:
+`major <https://github.com/scikit-hep/pyhf/labels/bumpversion%2Fmajor>`__,
+`minor <https://github.com/scikit-hep/pyhf/labels/bumpversion%2Fminor>`__,
+`patch <https://github.com/scikit-hep/pyhf/labels/bumpversion%2Fpatch>`__.
+Once the PR is tagged with the label, the GitHub Actions bot will post a comment
+with information on the actions it will take once the PR is merged. When the PR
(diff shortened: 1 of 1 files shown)


matthewfeickert:
A [build of this branch on ReadTheDocs is here](https://pyhf-staging.readthedocs.io/en/docs-improve-publishing-docs/development.html#publishing).

webknjaz:
@matthewfeickert FYI there's a GitHub App for per-PR RTD previews: https://github.com/apps/rtd-helper

webknjaz:
I noticed
> publish distributions / Build and publish Python distro to (Test)PyPI (pull_request

in the list of checks. 

If it's not needed in PRs, I recently discovered that you can add `if:` clause on the job level. This may help.

docs/development.rst
```diff
@@ -24,15 +24,28 @@ Publishing
...
+    git checkout master
+    git merge release/cut-<release tag>
+
+and then push the commit and tag to :code:`master` with
+
+.. code-block:: console
+
+    git push origin master <release tag>
+
+which will both merge and close the PR and start the release publication workflow.
```
webknjaz: The tag push may generate a separate `push` event (additionally to branch's `push` and `pull_request`). Maybe it's worth mentioning that one can visually see more jobs than they expect.

docs/development.rst
```diff
@@ -24,15 +24,28 @@ Publishing
...
 
-    git push origin master --tags
+    git checkout master
+    git merge release/cut-<release tag>
+
+and then push the commit and tag to :code:`master` with
+
+.. code-block:: console
+
+    git push origin master <release tag>
```
webknjaz: The idea is that you push the tag first, before merging the PR and if the publishing happens properly, only then do the merge.

webknjaz:
> to avoid a merge commit

Actually, my point wasn't about avoiding the merge commit or doing the fast-forward merge. Personally I prefer explicit `--no-ff` mode but that's not the point.
It's about avoiding squash and rebase during merge.

matthewfeickert:
> I noticed
> 
> > publish distributions / Build and publish Python distro to (Test)PyPI (pull_request
> 
> in the list of checks.
> 
> If it's not needed in PRs, I recently discovered that you can add `if:` clause on the job level. This may help.

Yup. At the moment we have it in on every PR just to make sure that there is nothing that would break the build of the sdist or wheel (the idea being if so, then catch this _before_ we try to make a release).

docs/development.rst
matthewfeickert: Correct. As can be seen with release `v0.2.2`

```
git push origin master v0.2.2
```

[kicked off 5 actions](https://github.com/scikit-hep/pyhf/commit/2175c9ff9120791a628e2164a05107310894c5b0/checks): 2 publish actions (TestPyPI and PyPI), 2 `push` actions, and then our Binder trigger action. So good point on this.

Though I think that all of this is gong to get rewritten given PR #650 as @kratsg is changing it from Issue driven to PR driven.

docs/development.rst
matthewfeickert: Ah, sorry. Yes, this is a **_very_** good point, and apparently not clear to me late at night while working on this. Thank you for pointing this out!

webknjaz:
> just to make sure that there is nothing that would break the build of the sdist or wheel

You could also have a separate job to build dists that would use `upload-artifacts` action and then all other actions (including publish) would use `download-artifacts` and work with those. The idea is that in the end, you publish exactly what you tested and not something that you build separately from testing. With this approach, you'd be able to completely hide the upload job from PRs.

matthewfeickert:
> You could also have a separate job to build dists that would use `upload-artifacts` action and then all other actions (including publish) would use `download-artifacts` and work with those. 

@kratsg We might want to look into this.

webknjaz:
P.S. As a bonus, those artifacts are available for download on the job page meaning that you can easily inspect them locally should the need arise.

kratsg:
> @kratsg We might want to look into this.

File an issue for it.

PR #648, opened 2019-11-21
refactor: Add better segmented environments

matthewfeickert:
# Description

Add further segmented `test`, `docs`, and `develop` environments in `setup.py` to avoid having to install unneeded dependencies.

This makes very little difference in terms of core developer operations, but the finer segmentation makes it easier to understand what all the dependencies a core developer would install are meant for.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Redistribute the "develop" environment into separate "test", "docs", and "develop" environments
   - "docs" contains all of the backend environments to successfully build all the API docs
   - "develop" still combines "test", "docs", and the "develop" dependencies together
* Use more specific environments for testing in CI to make environment dependencies more clear
* In doing so, fix the API section of the docs by building the docs in an environment with all the backends
   - Fixes bug introduced in PR #527
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
-        python -m pip install --ignore-installed -U -q --no-cache-dir -e .[complete]
+        python -m pip install --ignore-installed -U -q --no-cache-dir -e .[tensorflow,torch,minuit,xmlio,test]
         python -m pip list
     - name: Lint with Pyflakes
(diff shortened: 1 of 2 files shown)


matthewfeickert:
A [build of this branch on ReadTheDocs is here](https://pyhf-staging.readthedocs.io/en/revise-better-segment-environments).

matthewfeickert:
Ah, yeah, I now see that the docs have been a bit broken in that they failed to show all the backends in the API since PR #527 as I used [`develop` which doesn't contain TensorFlow or PyTorch](https://github.com/scikit-hep/pyhf/blob/f3b0680c115ef64cfb61a7631d9a550ec50c095f/setup.py#L24-L53). So this PR also fixes that now.

PR #650, opened 2019-11-22
ci: workflow that runs bumpversion and pushes when issue filed

kratsg:
# Description

When a PR is labeled with one of the following:

- `bumpversion/major`
- `bumpversion/minor`
- `bumpversion/patch`

The workflow will trigger. This will run `bumpversion $part`, update the commit to refer to the issue number filed, and update the tag to summarize the commits since the previous tag to the new tag.

The changes will be commit messages that follow the semantic format `$semantic: $description`.

When the PR is merged in, then the commit and tag will be pushed to master as well (using a personal access token) which will trigger the publish release.

Successful action (PR from branch):
- [on labeling PR](https://github.com/kratsg/github-action-test/runs/315583006)
- [on merging labeled PR](https://github.com/kratsg/github-action-test/runs/315585704)

Failed action (PR from fork):
- [on labeling PR](https://github.com/kratsg/github-action-test/runs/315572210)
- [on merging labeled PR](https://github.com/kratsg/github-action-test/runs/315581175)

For the failing ones -- it fails on labeling PR because the branch name is from a fork, so cannot be checked out locally. It fails on merging labeled PR because the `secrets.GITHUB_TOKEN` does not have the necessary permissions (which is probably a little confusing to me - @webknjaz?).

In any case, I think we could probably impose the policy that any PR which gets tagged for a release should come from people who have write access to the repo. Since this is actually an improvement over just an `issue` - because a PR is also an issue - it means we can do a PR with an empty commit, discuss, label it - to ensure that it can succeed, and then merge the PR.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add tag workflow that relies on tagging a PR with a specific label
```

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,0 +1,52 @@
+name: Tag Creator
+on:
+  issues:
+    types: [opened]
+jobs:
+  tag_master:
+    if: >-
+      startsWith( github.event.issue.title, 'bumpversion:' ) &&
+      && (
+        contains( github.event.issue.title, 'major' ) ||
+        contains( github.event.issue.title, 'minor' ) ||
+        contains( github.event.issue.title, 'patch' )
+      )
+      && (
+        github.event.issue.user.login == 'lukasheinrich' ||
+        github.event.issue.user.login == 'matthewfeickert' ||
+        github.event.issue.user.login == 'kratsg'
+      )
+    env:
+      IS_MAJOR: >-
+          ${{ contains( github.event.issue.title, 'major' ) }}
+      IS_MINOR: >-
+          ${{ contains( github.event.issue.title, 'minor' ) }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: install bumpversion
+      run: |
+      python -m pip install --upgrade pip setuptools wheel
+      pip install bumpversion
+    - name: set up git user
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
+    - name: run bumpversion
+      run: |
+        if [ $IS_MAJOR == 'true' ]
+        then
+          bumpversion major
+        elif [ $IS_MINOR == 'true' ]
+        then
+          bumpversion minor
+        else
+          bumpversion patch
+        fi
+    - name: push changes
+      run: |
+        git rev-parse HEAD
+        git rev-parse master
+        git describe --tags
+        git rev-list -n 1 $(git describe --tags)
+        git status
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -94,7 +94,7 @@ jobs:
         github.event.action == 'closed' && github.event.pull_request.merged
       uses: ad-m/github-push-action@v0.5.0
       with:
-        github_token: ${{ secrets.GITHUB_TOKEN }}
+        github_token: ${{ secrets.GITHUB_PAT }}
     - name: Comment that something failed
       if: failure()
       uses: actions/github@v1.0.0
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -88,9 +88,8 @@ jobs:
       uses: actions/github-script@v0.3.0
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
-      script: |
-        echo "abc"
-        github.issues.createComment({...context.issue, body: "I've queued this up. When it gets merged, I'll create a ${{ env['BV_PART'] }} release from ${{ env['OLD_TAG'] }} → ${{ env['NEW_TAG'] }} which includes the following ${{ env['NUM_CHANGES'] }} change(s) [including this PR]:<br />${{ env['CHANGES'] }}<br />If you make any more changes, you probably want to re-trigger me again by removing the `bumpversion/${{ env['BV_PART'] }}` label and then adding it back again."})
+        script: |
+          github.issues.createComment({...context.issue, body: "I've queued this up. When it gets merged, I'll create a ${{ env['BV_PART'] }} release from ${{ env['OLD_TAG'] }} → ${{ env['NEW_TAG'] }} which includes the following ${{ env['NUM_CHANGES'] }} change(s) [including this PR]:<br />${{ env['CHANGES'] }}<br />If you make any more changes, you probably want to re-trigger me again by removing the `bumpversion/${{ env['BV_PART'] }}` label and then adding it back again."})
     - name: Push changes
       if: >-
         github.event.action == 'closed' && github.event.pull_request.merged
@@ -102,9 +101,9 @@ jobs:
       uses: actions/github-script@v0.3.0
       with:
         github-token: ${{ secrets.GITHUB_TOKEN }}
-      script: |
-        github.issues.createComment({...context.issue, body: ":cry: Something went wrong. I am not able to push. Check the [Actions pipeline](https://github.com/${{ github.repository }}/actions?query=workflow%3A%22Tag+Creator%22) to see what happened. If you make any more changes, you probably want to re-trigger me again by adding the `bumpversion/${{ env['BV_PART'] }}` label again."})
-        github.issues.removeLabels({...context.issue, "bumpversion/${{ env['BV_PART'] }}" })
+        script: |
+          github.issues.createComment({...context.issue, body: ":cry: Something went wrong. I am not able to push. Check the [Actions pipeline](https://github.com/${{ github.repository }}/actions?query=workflow%3A%22Tag+Creator%22) to see what happened. If you make any more changes, you probably want to re-trigger me again by adding the `bumpversion/${{ env['BV_PART'] }}` label again."})
+          github.issues.removeLabels({...context.issue, "bumpversion/${{ env['BV_PART'] }}" })
 
   always_job:
     name: Always run job
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -43,7 +43,7 @@ jobs:
           echo "::set-env name=BV_PART::patch"
         fi
     - name: Checkout repository
-      uses: actions/checkout@master
+      uses: actions/checkout@v1.2.0
     - name: Set up git user
       run: |
         git config --local user.email "action@github.com"
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -62,7 +62,7 @@ jobs:
     - name: Install bumpversion
       run: |
         python -m pip install --upgrade pip setuptools wheel
-        pip install bumpversion
+        python -m pip install bumpversion
     - name: Run bumpversion ${{ env['BV_PART'] }}
       run: |
         OLD_TAG=$(git describe --tags --abbrev=0)
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -20,6 +20,8 @@ jobs:
         contains( github.event.pull_request.labels.*.name, 'bumpversion/major' ) ||
         contains( github.event.pull_request.labels.*.name, 'bumpversion/minor' ) ||
         contains( github.event.pull_request.labels.*.name, 'bumpversion/patch' )
+      ) && (
+        contains( ['lukasheinrich', 'matthewfeickert', 'kratsg'], github.event.sender.login )
       )
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,52 +1,104 @@
+# NB: release notes inspired by https://blogs.sap.com/2018/06/22/generating-release-notes-from-git-commit-messages-using-basic-shell-commands-gitgrep/
 name: Tag Creator
 on:
   issues:
     types: [opened]
 jobs:
-  tag_master:
+  bumpversion:
     if: >-
-      startsWith( github.event.issue.title, 'bumpversion:' ) &&
-      && (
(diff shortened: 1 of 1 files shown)


.github/workflows/tag.yml
```diff
@@ -0,0 +1,104 @@
...
+        echo "::set-env name=NUM_CHANGES::$NUM_CHANGES"
+        git tag $NEW_TAG $NEW_TAG^{} -f -m "$(printf "This is a $BV_PART release from $OLD_TAG → $NEW_TAG.\n\nChanges:\n$CHANGES")"
+        git tag -n99 -l $NEW_TAG
+    - name: comment on issue
+      uses: actions/github@v1.0.0
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        args: |
+          comment "This is a ${{ env['BV_PART'] }} release from ${{ env['OLD_TAG'] }} → [${{ env['NUM_CHANGES'] }} change(s)] → ${{ env['NEW_TAG'] }}. I'm going to attempt to push now."
```
matthewfeickert:
```suggestion
          comment "This is a ${{ env['BV_PART'] }} release from ${{ env['OLD_TAG'] }} → ${{ env['NEW_TAG'] }} with ${{ env['NUM_CHANGES'] }} change(s). I'm going to attempt to push now."
```

I'd suggest having the number of changes not be part of the tag change chain. As we only ever want releases to be _just_ releases, then can we have it succeed only if the number of changes is 1?

.github/workflows/tag.yml
```diff
@@ -0,0 +1,104 @@
+# NB: release notes inspired by https://blogs.sap.com/2018/06/22/generating-release-notes-from-git-commit-messages-using-basic-shell-commands-gitgrep/
+name: Tag Creator
+on:
+  issues:
+    types: [opened]
```
matthewfeickert: I understand why this is `opened` and not `closed` but it feels like it would be nice to have some way to have group consensus in the public record before this gets kicked off.

.github/workflows/tag.yml
```diff
@@ -0,0 +1,104 @@
...
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        args: |
+          comment "This is a ${{ env['BV_PART'] }} release from ${{ env['OLD_TAG'] }} → [${{ env['NUM_CHANGES'] }} change(s)] → ${{ env['NEW_TAG'] }}. I'm going to attempt to push now."
+    - name: push changes
+      uses: ad-m/github-push-action@v0.5.0
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+    - name: comment that it failed
```
matthewfeickert:
```suggestion
    - name: Comment that the push failed
```

.github/workflows/tag.yml
```diff
@@ -0,0 +1,104 @@
...
+    env:
+      IS_MAJOR: >-
+          ${{ startsWith( github.event.issue.title, '[bumpversion:major]' ) }}
+      IS_MINOR: >-
+          ${{ startsWith( github.event.issue.title, '[bumpversion:minor]' ) }}
+      IS_PATCH: >-
+          ${{ startsWith( github.event.issue.title, '[bumpversion:patch]' ) }}
+    runs-on: ubuntu-latest
+    steps:
+    - name: determine part
```
matthewfeickert:
```suggestion
    - name: Determine part
```

Some of the step names are captialized, but not all, so go all start with caps?

.github/workflows/tag.yml
```diff
@@ -0,0 +1,104 @@
...
+        if [ $IS_MAJOR == 'true' ]
+        then
+          echo "::set-env name=BV_PART::major"
+        elif [ $IS_MINOR == 'true' ]
+        then
+          echo "::set-env name=BV_PART::minor"
+        else
+          echo "::set-env name=BV_PART::patch"
+        fi
+    - name: add bumpversion/${{ env['BV_PART'] }} label
```
matthewfeickert:
```suggestion
    - name: Add bumpversion/${{ env['BV_PART'] }} label
```

Some of the step names are captialized, but not all, so go all start with caps?

.github/workflows/tag.yml
```diff
@@ -0,0 +1,104 @@
...
+      with:
+        args: label bumpversion/${{ env['BV_PART'] }}
+    - uses: actions/checkout@master
+    - name: Prepare repository
+      run: git checkout "${GITHUB_REF:11}"
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: install bumpversion
```
matthewfeickert:
```suggestion
    - name: Install bumpversion
```

Some of the step names are capitalized, but not all, so go all start with caps?

.github/workflows/tag.yml
```diff
@@ -0,0 +1,104 @@
...
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        pip install bumpversion
+    - name: set up git user
```
matthewfeickert:
```suggestion
    - name: Set up Git user
```

Some of the step names are capitalized, but not all, so go all start with caps?

.github/workflows/tag.yml
```diff
@@ -0,0 +1,104 @@
...
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
+    - name: run bumpversion ${{ env['BV_PART'] }}
```
matthewfeickert:
```suggestion
    - name: Run bumpversion ${{ env['BV_PART'] }}
```

Some of the step names are capitalized, but not all, so go all start with caps?

.github/workflows/tag.yml
```diff
@@ -0,0 +1,104 @@
...
+        NEW_TAG=$(git describe --tags --abbrev=0)
+        echo "::set-env name=NEW_TAG::$NEW_TAG"
+        git tag -n99 -l $NEW_TAG
+
+        CHANGES="$(git log --pretty=format:'%s' $OLD_TAG..HEAD -i -E --grep='^([a-z]*?):' | sed 's/^/  - /')"
+        NUM_CHANGES=$(echo -n "$CHANGES" | grep -c '^')
+        echo "::set-env name=NUM_CHANGES::$NUM_CHANGES"
+        git tag $NEW_TAG $NEW_TAG^{} -f -m "$(printf "This is a $BV_PART release from $OLD_TAG → $NEW_TAG.\n\nChanges:\n$CHANGES")"
+        git tag -n99 -l $NEW_TAG
+    - name: comment on issue
```
matthewfeickert:
```suggestion
    - name: Comment on Issue
```

Some of the step names are capitalized, but not all, so go all start with caps?

.github/workflows/tag.yml
```diff
@@ -0,0 +1,104 @@
...
+      uses: actions/github@v1.0.0
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        args: |
+          comment "This is a ${{ env['BV_PART'] }} release from ${{ env['OLD_TAG'] }} → [${{ env['NUM_CHANGES'] }} change(s)] → ${{ env['NEW_TAG'] }}. I'm going to attempt to push now."
+    - name: push changes
```
matthewfeickert:
```suggestion
    - name: Push changes
```

Some of the step names are capitalized, but not all, so go all start with caps?

matthewfeickert:
As @webknjaz [pointed out](https://github.com/scikit-hep/pyhf/pull/647#discussion_r349316627) in PR #647, what we really want is 

> push the tag first, before merging the PR and if the publishing happens properly, only then do the merge.

If it is possible to manage this order, and ability to trigger before merge, in the PR version of this that would be important.

matthewfeickert:
> > If it is possible to manage this order, and ability to trigger before merge, in the PR version of this that would be important.
> 
> Not really... we can't... make it wait. The thing is that merging the PR is what literally triggers making the tag (and pushing it).

I think the way that you have it setup now in the PR mode is fine.

In general though:
- Can you link to some examples of this running as expected (maybe an example of it making a new release on `stare`)?
- Is this still WIP?

.github/workflows/tag.yml
```diff
@@ -0,0 +1,110 @@
+# NB: release notes inspired by https://blogs.sap.com/2018/06/22/generating-release-notes-from-git-commit-messages-using-basic-shell-commands-gitgrep/
```
webknjaz: Git commit message contents don't map to a good changelog... https://twitter.com/yetanothertw/status/1197634865988067329

Using Towncrier and forcing PR authors to include human-readable change fragments is a more popular practice that many Python projects use. I even wrote a GitHub App that checks the presence of such fragments in PRs...

.github/workflows/tag.yml
```diff
@@ -0,0 +1,110 @@
...
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        args: |
+          comment "I've queued this up. When it gets merged, I'll create a ${{ env['BV_PART'] }} release from ${{ env['OLD_TAG'] }} → ${{ env['NEW_TAG'] }} which includes the following ${{ env['NUM_CHANGES'] }} change(s) [including this PR]:<br />${{ env['CHANGES'] }}<br />If you make any more changes, you probably want to re-trigger me again by removing the `bumpversion/${{ env['BV_PART'] }}` label and then adding it back again."
+    - name: Push changes
+      if: >-
+        github.event.action == 'closed' && github.event.pull_request.merged
+      uses: ad-m/github-push-action@v0.5.0
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
```
webknjaz: When you use the Actions' token, it'll not trigger *any* events at all. Meaning that *none* of the workflows that listen to the `push` event, in particular, will start as a result of this.

.github/workflows/tag.yml
kratsg: It's a good point -- and I don't want the actual GitHub release to contain the commits -- just the tag annotated with the list. That way, when I draft a release, I can refer to the changes incorporated in the tag (to remind myself) and write a more appropriate release / Towncrier.

.github/workflows/tag.yml
kratsg: ... what, really? I don't understand how that works. It's a regular `git push` that should always trigger, no matter what? If it doesn't trigger, then... ... huh...

kratsg: Yeah, so I guess this is a documented limitation. Maybe I should make a developer token and use that instead (create a bot for the tag/release) and then I can avoid that limitation...

webknjaz: > Yeah, so I guess this is a documented limitation

This is protection from shooting yourself in a leg. Otherwise, it's easy to create a loop of recursively reacting to own events over and over again.
GitHub Apps don't have such limitation but then you're responsible for ignoring events caused by yourself.



> Maybe I should make a developer token

I strongly recommend using a GitHub App token. You'll get access to more APIs with it and the ACL is more granular (accessing everything your account can access vs. just one-repo-related stuff). You can put a fake `localhost` URL there if you don't want to actually deploy a web app.

For a full app deployment, though, it'd be useful for you to skim-through this tutorial: https://tutorial.octomachinery.dev/.

kratsg: Is there a solution that doesn't require using something like Heroku? The whole point here is to have a self-contained solution that doesn't require putting things _outside_ of GitHub Actions...

webknjaz: Heroku part is only for receiving events. If you want to just use a token, you can use a fake address like localhost.

lukasheinrich:
can we ensure that only us can attach these labels?

webknjaz:
Any user with write access to the repo or a GitHub App with write access to labels can create them. But each event payload has a `sender` object which represents the entity that caused it. It has a `type` field (that can be either `User` or `Bot`) and a `login` field (that will end with `[bot]` for GitHub Apps). This can be used for the ACL implementation.

.github/workflows/tag.yml
kratsg: FYI, I tried to use localhost for the GitHub App -- and it absolutely failed and I could never get a useable token from it. It seems like it provides a secret one uses to generate access tokens (and expires in an hour) and those access tokens aren't that useable as github secrets.

kratsg:
> can we ensure that only us can attach these labels?

well, at the moment, we're the only developers.

> Any user with write access to the repo or a GitHub App with write access to labels can create them. But each event payload has a `sender` object which represents the entity that caused it. It has a `type` field (that can be either `User` or `Bot`) and a `login` field (that will end with `[bot]` for GitHub Apps). This can be used for the ACL implementation.

oh, sweet. I'll look at the API and use the one that let's me specify our github usernames (are those logins, or is that the email that's logins?)

.github/workflows/tag.yml
kratsg: For now, I'll use a personal access token. Since I have ACL logic added, I think I'm ok with it.

kratsg:
This seems to be doing exactly what I expect when labeled. I've removed the label now, so that we aren't making a tag/release when this is merged in.

.github/workflows/tag.yml
webknjaz: @kratsg yes. Basically, you need to put a private key (PEM file that you get from the GitHub App page) in Secrets and then use it inside of an Aciton to fetch an access token (yes, it has a 1-hour lifetime) using it. It's almost the same as GitHub does when it injects its pre-created tokens — they also have 1-hour lifetime. Maybe even a few minutes less. Wrapper script example: https://github.com/ansible-community/collection_migration/blob/82ca900/gh.py.

PR #611, opened 2019-10-10
test: Add public API canary tests to CI and nightly tests for current PyPI release

matthewfeickert:
# Description

- Resolves #609 
- Resolves #605 

Add canary tests and GitHub Actions workflow to check if the current release of pyhf on PyPI is functional and able to work with the current API.

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Add GitHub Actions workflow to nightly check if the current release of pyhf on PyPI is valid for the public API
* Add canary tests of the public API
   - Runs on both GitHub in CI and in against the releases from PyPI
* Use tf.compat.v1.Session to avoid warning
```

PR #652, opened 2019-11-25
fix: Add missing pytest dependencies to Current Release tests

matthewfeickert:
# Description

Install missing `pytest` (`v3`) and `pytest-cov` to ["Current release" tests](https://github.com/scikit-hep/pyhf/blob/0e56c6c2e4470b5dc2ad8fb02d1afeeb4184ec18/.github/workflows/release_tests.yml). This was forgotten in PR #611

Please describe the purpose of this pull request in some detail. Reference and link to any relevant issues or pull requests.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add pytest dependency to install
   - pytest is regulated to be below v4.0 as pyhf isn't compliant with modern pytest
* Fixes bug introduced in PR #611
* Run "Current release" tests on PRs as well as on a daily schedule
```

### DIFF
--- a/.github/workflows/release_tests.yml
+++ b/.github/workflows/release_tests.yml
@@ -1,6 +1,7 @@
 name: Current Release
 
 on:
+  pull_request:
   # Run daily at 0:01 UTC
   schedule:
   - cron:  '1 0 * * *'
--- a/.github/workflows/release_tests.yml
+++ b/.github/workflows/release_tests.yml
@@ -25,6 +25,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install pyhf[tensorflow,torch,minuit,xmlio]
+        python -m pip install pytest
         python -m pip list
     - name: Canary test public API
       run: |
--- a/.github/workflows/release_tests.yml
+++ b/.github/workflows/release_tests.yml
@@ -25,7 +25,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install pyhf[tensorflow,torch,minuit,xmlio]
-        python -m pip install pytest pytest-cov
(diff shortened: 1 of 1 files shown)


matthewfeickert:
> Make sure we test `release_tests` on PRs as well as cron jobs. That way, we know if we ever break it.

Done.

.github/workflows/release_tests.yml
```diff
@@ -25,6 +26,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install pyhf[tensorflow,torch,minuit,xmlio]
+        python -m pip install 'pytest~=3.5' pytest-cov
```
lukasheinrich: why is `pytest-cov` not installed as part of the development extra?

matthewfeickert: As in why not do 

```
python -m pip install pyhf[tensorflow,torch,minuit,xmlio,develop]
```
?

We could do that. The main reason for not just doing `python -m pip install pyhf[complete]` is to be very explicit what is actually being tested in release tests.

kratsg: I prefer this way, because we need to check if everything (except `develop`) is necessary for performing the tests. We have a lot of extra stuff that doesn't need to be checked for user-facing API usage.

lukasheinrich: that's a good reason I guess

matthewfeickert: If https://github.com/scikit-hep/pyhf/pull/648 had gone in before release `v0.2.2` then I would say this line would just be

```
python -m pip install pyhf[test]
```

(maybe something to keep in mind for the next release). However, as it didn't, then I'm trying to be more explicit for the time being.

PR #656, opened 2019-11-27
refactor: Use a CLI subpackage

kratsg:
# Description

Refactor the [`commandline.py`](https://github.com/scikit-hep/pyhf/blob/e49733e261efc60d2fd1e92414676499c95d68ec/src/pyhf/commandline.py) into a separate CLI subpackage.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Refactor the commandline.py into a CLI subpackage
```


src/pyhf/cli/cli.py
```diff
@@ -0,0 +1,26 @@
...
+log = logging.getLogger(__name__)
+
+
+@click.group(context_settings=dict(help_option_names=['-h', '--help']))
+@click.version_option(version=__version__)
+def pyhf():
+    pass
+
+
+# pyhf.add_command(rootio.cli)
```
matthewfeickert: These comments can get deleted now?
```suggestion
```

src/pyhf/cli/cli.py
```diff
@@ -0,0 +1,26 @@
...
+pyhf.add_command(rootio.json2xml)
+pyhf.add_command(rootio.xml2json)
+
+# pyhf.add_command(spec.cli)
```
matthewfeickert: These comments can get deleted now?
```suggestion
```

src/pyhf/cli/cli.py
```diff
@@ -0,0 +1,26 @@
...
+pyhf.add_command(spec.inspect)
+
+# pyhf.add_command(stats.cli)
```
matthewfeickert: These comments can get deleted now?
```suggestion
```

src/pyhf/cli/rootio.py
```diff
@@ -0,0 +1,100 @@
...
+    """ Entrypoint XML: The top-level XML file for the PDF definition. """
+    try:
+        import uproot
+
+        assert uproot
+    except ImportError:
+        log.error(
+            "xml2json requires uproot, please install pyhf using the "
+            "xmlio extra: pip install pyhf[xmlio] or install uproot "
+            "manually: pip install uproot"
```
matthewfeickert: I would advocate that we don't tell people to install dependencies for extras manually, but let us do it.
```suggestion
```

So I'd suggest the last two lines just be
```
            "xmlio extra: pip install pyhf[xmlio]"
```

src/pyhf/cli/rootio.py
```diff
@@ -0,0 +1,100 @@
...
+def json2xml(workspace, output_dir, specroot, dataroot, resultprefix, patch):
+    try:
+        import uproot
+
+        assert uproot
+    except ImportError:
+        log.error(
+            "json2xml requires uproot, please install pyhf using the "
+            "xmlio extra: pip install pyhf[xmlio]."
+            "manually: pip install uproot"
```
matthewfeickert: I would advocate that we don't tell people to install dependencies for extras manually, but let us do it.
```suggestion
```

So I'd suggest the last two lines just be
```
            "xmlio extra: pip install pyhf[xmlio]"
```

src/pyhf/cli/stats.py
```diff
@@ -0,0 +1,67 @@
...
+):
+    with click.open_file(workspace, 'r') as specstream:
+        wspec = json.load(specstream)
+
+    w = Workspace(wspec)
+
+    is_qtilde = teststat == 'qtilde'
+
+    patches = [json.loads(click.open_file(pfile, 'r').read()) for pfile in patch]
+    p = w.model(
```
matthewfeickert: I'd advocate to in general avoid 1 letter variables 
```suggestion
    pdf = ws.model(
```

src/pyhf/cli/stats.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+    optconf = {k: v for item in optconf for k, v in item.items()}
+
+    # set the new optimizer
+    if optimizer:
+        new_optimizer = getattr(optimize, optimizer)
+        set_backend(tensorlib, new_optimizer(**optconf))
+
+    result = hypotest(testpoi, w.data(p), p, qtilde=is_qtilde, return_expected_set=True)
```
matthewfeickert: I'd advocate to in general avoid 1 letter variables 
```suggestion
    result = hypotest(testpoi, ws.data(pdf), pdf, qtilde=is_qtilde, return_expected_set=True)
```

src/pyhf/cli/stats.py
```diff
@@ -0,0 +1,67 @@
...
+@click.option('--teststat', type=click.Choice(['q', 'qtilde']), default='qtilde')
+@click.option('--optimizer')
+@click.option('--optconf', type=EqDelimStringParamType(), multiple=True)
+def cls(
+    workspace, output_file, measurement, patch, testpoi, teststat, optimizer, optconf
+):
+    with click.open_file(workspace, 'r') as specstream:
+        wspec = json.load(specstream)
+
+    w = Workspace(wspec)
```
matthewfeickert: Use consistent variables names with `inspect`
```suggestion
    ws = Workspace(wspec)
```

src/pyhf/cli/cli.py
kratsg: depends on what we want to do with the api. If we want `pyhf rootio xml2json` and `pyhf rootio json2xml` etc...

src/pyhf/cli/cli.py
kratsg: depends on what we want to do with the api. If we want `pyhf rootio xml2json` and `pyhf rootio json2xml` etc...

src/pyhf/cli/cli.py
kratsg: depends on what we want to do with the api. If we want `pyhf rootio xml2json` and `pyhf rootio json2xml` etc...

src/pyhf/cli/cli.py
matthewfeickert: You're saying that if we do want 

```
$ pyhf rootio xml2json
```

then we remove the comment and go with

```python
pyhf.add_command(rootio.json2xml)
pyhf.add_command(rootio.xml2json)
```

but if we would prefer a CLI like 

```
$ pyhf xml2json
```

then we only have

```python
pyhf.add_command(rootio.cli)
```

yes?

If so, then FWIW, I would put my vote behind `pyhf rootio xml2json` to keep the cli from returning a huge number of results when someone tab completes. The grouping should help with that, and I think it is cleaner.

kratsg: Other way around.

`pyhf.add_command(rootio.cli)` adds the `pyhf rootio` -- while `pyhf.add_command(rootio.json2xml)` adds the `pyhf json2xml`.

I haven't gotten autocomplete set up on my end ,but it would be nice to have with Click + pyhf...

matthewfeickert: Ah okay &mdash; sorry. Yeah, since you're preserving the current behavior with how you currently have it I see why you have those lines commented hinting at what would be to come.

Since moving to `pyhf.add_command(rootio.cli)` would be an API change maybe that should happen in a separate `feat` PR?

kratsg: Yep, which is why it's commented out for now (and the tests need to be updated to use the right command).

PR #657, opened 2019-11-27
feat: Add Channel summary mixin, JSON-serializable workspace object

kratsg:
# Description

This adds a channel summary mixin to be used by any object that needs a summary of the channel information in the spec they're using. This will apply to both `pyhf.Workspace` and `pyhf.pdf._ModelConfig`.

The end goal, if it's not obvious, is that unlike `pyhf.Model` which doesn't currently have a spec.. I want the `pyhf.Workspace` to be a "light" layer on top of the `workspace.json` spec we have. This means that we should do (directly)

```python
spec = pyhf.Workspace(json.load(open('spec.json')))
assert spec['version'] == 1.0.0
assert 'signal' in spec.samples
```

where it quacks like a python dictionary, but has some extra niceties such as being able to build a model from it, or listing summary information about itself. 

You might ask, well, why not just provide a `pyhf.Workspace::to_json()` method instead?

- this will break when doing nested `json.dump()` calls

Or some sort of `__json__` that gets called?

- all python json libraries do something different and the one we use doesn't call any magic function and instead expects that we pass in a custom encoder/decoder to the `json.dump()` calls...

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add Channel Summary mixin
* Make the pyhf.Workspace JSON-serializable
```

### DIFF
--- a/src/pyhf/workspace.py
+++ b/src/pyhf/workspace.py
@@ -30,6 +30,11 @@ class Workspace(_ChannelSummaryMixin, dict):
         for obs in self['observations']:
             self.observations[obs['name']] = obs['data']
 
+    def __eq__(self, other):
+        if not isinstance(other, Workspace):
+            return False
+        return dict(self) == dict(other)
+
     def __repr__(self):
         return object.__repr__(self)
 
--- a/src/pyhf/workspace.py
+++ b/src/pyhf/workspace.py
@@ -1,9 +1,6 @@
 import logging
 import jsonpatch
 from copy import deepcopy
-from collections import (
-    UserDict,
-)  # underlying dict is accessible as an attribute (self.data)
 from . import exceptions
 from . import utils
 from .pdf import Model
@@ -13,7 +10,7 @@ logging.basicConfig()
 log = logging.getLogger(__name__)
 
 
-class Workspace(_ChannelSummaryMixin, UserDict):
+class Workspace(_ChannelSummaryMixin, dict):
     """
     A JSON-serializable object that is built from an object that follows the `workspace.json` schema.
     """
@@ -24,7 +21,7 @@ class Workspace(_ChannelSummaryMixin, UserDict):
         self.version = config_kwargs.pop('version', None)
         # run jsonschema validation of input specification against the (provided) schema
         log.info("Validating spec against schema: {0:s}".format(self.schema))
-        utils.validate(self.data, self.schema, version=self.version)
+        utils.validate(self, self.schema, version=self.version)
 
         self.measurement_names = []
         for measurement in self.get('measurements', []):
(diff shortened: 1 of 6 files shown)


src/pyhf/workspace.py
```diff
@@ -3,33 +3,41 @@
...
-        for measurement in self.spec.get('measurements', []):
+        for measurement in self.get('measurements', []):
             self.measurement_names.append(measurement['name'])
 
         self.observations = {}
-        for obs in self.spec['observations']:
+        for obs in self['observations']:
             self.observations[obs['name']] = obs['data']
 
+    def __eq__(self, other):
```
lukasheinrich: do we eally want this high level logic?

src/pyhf/workspace.py
```diff
@@ -3,33 +3,41 @@
...
 from . import utils
 from .pdf import Model
+from .mixins import _ChannelSummaryMixin
 
 logging.basicConfig()
 log = logging.getLogger(__name__)
 
 
-class Workspace(object):
+class Workspace(_ChannelSummaryMixin, dict):
```
lukasheinrich: is inheriting from `dict` necessary only to make w.get and w[] working?

kratsg: It's both for that and to make `json.dumps(workspace)` possible :)

src/pyhf/workspace.py
kratsg: LGTM complains if I don't include it...

kratsg: See [the __eq__](https://lgtm.com/projects/g/scikit-hep/pyhf/rev/pr-710232ea6f84678b0578b28d44820bfb7ff5cbf1) warning and the subsequent [__ne__](https://lgtm.com/projects/g/scikit-hep/pyhf/rev/pr-033b34773cbd178d22e6e5a4308022eb37878f5c) warning.

lukasheinrich: ok

src/pyhf/workspace.py
lukasheinrich: cool!

kratsg: The end goal, if it's not obvious, is that unlike `pyhf.Model` which doesn't currently have a spec.. I want the `pyhf.Workspace` to be a "light" layer on top of the `workspace.json` spec we have. This means that we should do (directly)

```python
spec = pyhf.Workspace(json.load(open('spec.json')))
assert spec['version'] == 1.0.0
```

where it quacks like a python dictionary, but has some extra niceties such as being able to build a model from it, or listing summary information about itself.

kratsg: You might ask, well, why not just provide a `pyhf.Workspace::to_json()` method instead?

- this will break when doing nested `json.dump()` calls

Or some sort of `__json__` that gets called?

- all python json libraries do something different and the one we use doesn't call any magic function and instead expects that we pass in a custom encoder/decoder to the `json.dump()` calls...

PR #665, opened 2019-12-03
docs: Use Sphinx cross-referencing '~' syntax to dynamically link the codebase

matthewfeickert:
# Description

Use the [`~`](https://www.sphinx-doc.org/en/1.5/markup/inline.html#cross-referencing-syntax) cross-referencing syntax instead of a hyperlink to dynamically link to functions in the codebase in the docs (results in _Model_ instead of `pyhf.pdf.Model`).

Thanks to @kratsg for pointing this out in PR #617.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use ~ instead of a hyperlink to dynamically link to functions in the codebase in the docs
   - c.f. https://www.sphinx-doc.org/en/1.5/markup/inline.html#cross-referencing-syntax
```

### DIFF
--- a/src/pyhf/utils.py
+++ b/src/pyhf/utils.py
@@ -98,7 +98,7 @@ def qmu(mu, data, pdf, init_pars, par_bounds):
     Args:
         mu (Number or Tensor): The signal strength parameter
         data (Tensor): The data to be considered
-        pdf (:func:`pyhf.pdf.Model`): The HistFactory statistical model used in the likelihood ratio calculation
+        pdf (:obj:`pyhf.pdf.Model`): The HistFactory statistical model used in the likelihood ratio calculation
         init_pars (Tensor): The initial parameters
         par_bounds(Tensor): The bounds on the paramter values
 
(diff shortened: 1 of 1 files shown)


src/pyhf/utils.py
```diff
@@ -98,13 +98,10 @@ def qmu(mu, data, pdf, init_pars, par_bounds):
     Args:
         mu (Number or Tensor): The signal strength parameter
         data (Tensor): The data to be considered
-        pdf (|pyhf.pdf.Model|_): The HistFactory statistical model used in the likelihood ratio calculation
+        pdf (:func:`pyhf.pdf.Model`): The HistFactory statistical model used in the likelihood ratio calculation
```
matthewfeickert: @kratsg Do you know if it is possible for us to use this nicer API but still keep the particular styling where we get `pyhf.pdf.Model` (like we have now below) and not `pyhf.pdf.Model()`?

nice_look

kratsg: Just don't use the `:func:` for parameter types. You only use it when you need to explicitly force linking.

kratsg: Screenshot 2019-12-03 12 15 48

is done with

```
        Raises:
          ~pyhf.exceptions.InvalidWorkspaceOperation: The workspaces have common channel names, incompatible measurements, or incompatible schema versions.

        Args:

        Returns:
            ~pyhf.workspace.Workspace: A new combined workspace object
```

matthewfeickert: I used `:obj:` instead to maintain the entire namespace of `pyhf.pdf.Model`. Would you prefer it just be `Model` with `~pyhf.pdf.Model`?

kratsg: Yes, I prefer the `~pyhf.pdf.Model`. In documentation types like this, you only need the short name of the class rather than the full path. The full path will be on the class' doc page anyway.

PR #666, opened 2019-12-04
feat: normfactor import/export parameter configuration

### LINKED ISSUE #661: Parse normalization factor ranges from xml
# Description

The .xml workspace specification includes the ranges of normalization factors (NFs), such as 
```
<NormFactor Name="my_NF"  Val="1"  High="20"  Low="-20"  Const="False"   />
```
A simple example can be found [here](https://github.com/alexander-held/template_fit_workflows/blob/master/TRExFitter/minimal_example/RooStats/minimal_example_Signal_region.xml).
When the NF affects multiple samples or channels, the information is consistently specified in all instances.

Given that this information is specified in the xml, it should be picked up by `pyhf.readxml.parse()` and subsequently be available within pyhf.

The aim of this is to reduce small differences between the results of `xml workspace -> fit within ROOT` and `xml workspace -> translate to pyhf -> fit` that can be hard to track down.

I would like the ranges to be picked up automatically by pyhf.

Could alternatively specify ranges manually from within pyhf.

None that I am aware of.

# Additional context
None

kratsg:
# Description

This resolves #661. This allows for importing xml/exporting json of the NormFactor parameter configurations. 

A very strong assumption is made on import (checked via deduplication) that NormFactors (and the set of parameter configs grabbed from the channel specifications) do not conflict with one another. These parameter configurations are stored for all measurements.

A second assumption on import is that NormFactors defined `const` in ParamSetting override NormFactor settings.

A strong assumption is made on export that, for NormFactor, the first measurement should have all the same parameter configurations as the rest of the measurements (if there are more than one measurement). The Val/Low/High parameter configuration is only extracted from the first measurement and stored in the resulting XML.

This is done in such a way that our pyhf JSON spec does not need to change.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* normfactor parameter configurations (Val, High, Low) are now imported/exported
```

### DIFF
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -105,6 +105,7 @@ def process_sample(
             parameter_config = {
                 'name': modtag.attrib['Name'],
                 'bounds': [[float(modtag.attrib['Low']), float(modtag.attrib['High'])]],
+                'inits': [float(modtag.attrib['Val'])],
             }
             if modtag.attrib.get('Const'):
                 parameter_config['fixed'] = modtag.attrib['Const'] == 'True'
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -69,6 +69,7 @@ def process_sample(
 
     data, err = import_root_histogram(rootdir, inputfile, histopath, histoname)
 
+    parameter_configs = []
     modifiers = []
     # first check if we need to add lumi modifier for this sample
     if sample.attrib.get("NormalizeByTheory", "False") == 'True':
@@ -101,6 +102,14 @@ def process_sample(
             modifiers.append(
                 {'name': modtag.attrib['Name'], 'type': 'normfactor', 'data': None}
             )
+            parameter_config = {
+                'name': modtag.attrib['Name'],
+                'bounds': [[float(modtag.attrib['Low']), float(modtag.attrib['High'])]],
+            }
+            if modtag.attrib.get('Const'):
+                parameter_config['fixed'] = modtag.attrib['Const'] == 'True'
+
+            parameter_configs.append(parameter_config)
         elif modtag.tag == 'HistoSys':
             lo, _ = import_root_histogram(
                 rootdir,
@@ -165,7 +174,12 @@ def process_sample(
         else:
             log.warning('not considering modifier tag %s', modtag)
 
-    return {'name': sample.attrib['Name'], 'data': data, 'modifiers': modifiers}
+    return {
+        'name': sample.attrib['Name'],
+        'data': data,
+        'modifiers': modifiers,
+        'parameter_configs': parameter_configs,
+    }
 
 
 def process_data(sample, rootdir, inputfile, histopath):
@@ -197,17 +211,19 @@ def process_channel(channelxml, rootdir, track_progress=False):
     channelname = channel.attrib['Name']
 
     results = []
+    channel_parameter_configs = []
     for sample in samples:
         samples.set_description('  - sample {}'.format(sample.attrib.get('Name')))
         result = process_sample(
             sample, rootdir, inputfile, histopath, channelname, track_progress
         )
+        channel_parameter_configs.extend(result.pop('parameter_configs'))
         results.append(result)
 
-    return channelname, parsed_data, results
+    return channelname, parsed_data, results, channel_parameter_configs
 
 
-def process_measurements(toplvl):
+def process_measurements(toplvl, other_parameters=None):
     results = []
     for x in toplvl.findall('Measurement'):
         lumi = float(x.attrib['Lumi'])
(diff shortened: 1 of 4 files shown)


kratsg:
Honestly, I'm 100% against merging this. I strongly believe HistFactory XML is absolutely wrong in doing it the way they did it, because it makes it _very_ hard to actually fix things.

THIS PR BREAKS ROUND-TRIPPING.

XML: `<NormFactor Name=mu Val=1 High=10, Low=0 Const=true/>`
xml2json
JSON: `[{'name': 'mu', inits=[1.0], bounds=[[0.0, 10.0]], fixed=true}]`
json2xml
XML: `<NormFactor Name=mu Val=1 High=10, Low=0 />` and additionally `<ParamSetting Const=true>mu</ParamSetting>` is added
xml2json
JSON: (crashes due to duplicate `mu` parameter configuration settings from two different places -- which we can't even protect against in the first place!)

matthewfeickert:
> Honestly, I'm 100% against merging this. I strongly believe HistFactory XML is absolutely wrong in doing it the way they did it, because it makes it _very_ hard to actually fix things.
> 
> THIS PR BREAKS ROUND-TRIPPING.

Before reviewing, and assuming this is still the case today, then I would say that I'm not in favor of accepting the PR. Breaking the ability to do round-trip checks makes me quite uncomfortable.

kratsg:
> Before reviewing, and assuming this is still the case today, then I would say that I'm not in favor of accepting the PR. Breaking the ability to do round-trip checks makes me quite uncomfortable.

Updated my comment. It's all good now. Given assumptions in the PR body.

src/pyhf/writexml.py
```diff
@@ -115,9 +137,32 @@ def build_modifier(modifierspec, channelname, samplename, sampledata):
...
+        # However, since I strive for perfection, the "Const" attribute will
+        # never be set here, but at the per-measurement configuration instead
+        # like all other parameters. This is an acceptable compromise.
+        #
+        # Lastly, if a normfactor parameter configuration doesn't exist in the
+        # first measurement parameter configuration, then set defaults.
+        val = 1
+        low = 0
+        high = 10
+        # Whatever. Let's get on with the show, shall we?
```
matthewfeickert:
```suggestion
```

tests/test_import.py
```diff
@@ -16,6 +17,92 @@ def assert_equal_dictionary(d1, d2):
...
+    )
+    result = {
+        m['name']: {k['name']: k for k in m['config']['parameters']} for m in result
+    }
+    assert result
+    # make sure ParamSetting configs override NormFactor configs
+    assert result['NormalMeasurement']['mu_both']['fixed']
+    assert result['NormalMeasurement']['mu_both']['inits'] == [2.0]
+    assert result['NormalMeasurement']['mu_both']['bounds'] == [[1.0, 5.0]]
+    # make sure ParamSetting is doing the right thing
```
matthewfeickert: To make these `assert` blocks visually easier to read can you separate them with a new line before the relevant comment?

tests/test_import.py
```diff
@@ -26,9 +113,9 @@ def test_import_measurements():
     measurement_configs = measurements[0]['config']
 
     assert 'parameters' in measurement_configs
-    assert len(measurement_configs['parameters']) == 2
-    assert measurement_configs['parameters'][0]['name'] == 'lumi'
-    assert measurement_configs['parameters'][1]['name'] == 'alpha_syst1'
+    assert len(measurement_configs['parameters']) == 3
+    pnames = [p['name'] for p in measurement_configs['parameters']]
```
matthewfeickert: Suggestion
```suggestion
    parnames = [p['name'] for p in measurement_configs['parameters']]
    assert sorted(parnames) == sorted(['lumi', 'SigXsecOverSM', 'alpha_syst1'])
```

tests/test_import.py
```diff
@@ -16,6 +17,92 @@ def assert_equal_dictionary(d1, d2):
...
+        Name='NormalMeasurement',
+        Lumi=str(1.0),
+        LumiRelErr=str(0.017),
+        ExportOnly=str(True),
+    )
+    poiel = ET.Element('POI')
+    poiel.text = 'mu_SIG'
+    meas.append(poiel)
+
+    se = ET.Element('ParamSetting', Const='True')
```
matthewfeickert:
```suggestion
    setting = ET.Element('ParamSetting', Const='True')
    setting.text = ' '.join(['Lumi', 'mu_both', 'mu_paramSettingOnly'])
    meas.append(setting)
```

tests/test_import.py
```diff
@@ -16,6 +17,92 @@ def assert_equal_dictionary(d1, d2):
...
+    se.text = ' '.join(['Lumi', 'mu_both', 'mu_paramSettingOnly'])
+    meas.append(se)
+
+    se = ET.Element('ParamSetting', Val='2.0')
```
matthewfeickert:
```suggestion
    setting = ET.Element('ParamSetting', Val='2.0')
    setting.text = ' '.join(['mu_both'])
    meas.append(setting)
```

tests/test_import.py
```diff
@@ -16,6 +17,92 @@ def assert_equal_dictionary(d1, d2):
...
+        Name='ParallelMeasurement',
+        Lumi=str(1.0),
+        LumiRelErr=str(0.017),
+        ExportOnly=str(True),
+    )
+    poiel = ET.Element('POI')
+    poiel.text = 'mu_BKG'
+    meas.append(poiel)
+
+    se = ET.Element('ParamSetting', Val='3.0')
```
matthewfeickert:
```suggestion
    setting = ET.Element('ParamSetting', Val='3.0')
    setting.text = ' '.join(['mu_both'])
    meas.append(setting)
```

src/pyhf/readxml.py
```diff
@@ -243,13 +264,40 @@ def process_measurements(toplvl):
...
         results.append(result)
+
     return results
 
 
+def dedupe_parameters(parameters):
+    duplicates = {}
+    for p in parameters:
+        duplicates.setdefault(p['name'], []).append(p)
+    for pname in duplicates.keys():
```
matthewfeickert:
```suggestion
    for parname in duplicates.keys():
        parameter_list = duplicates[parname]
```

src/pyhf/readxml.py
```diff
@@ -243,13 +264,40 @@ def process_measurements(toplvl):
...
+        parameter_list = duplicates[pname]
+        if len(parameter_list) == 1:
+            continue
+        elif any(p != parameter_list[0] for p in parameter_list[1:]):
+            for p in parameter_list:
+                log.warning(p)
+            raise RuntimeError(
+                'cannot import workspace due to incompatible parameter configurations for {0:s}.'.format(
+                    pname
```
matthewfeickert:
```suggestion
                    parname
```

PR #672, opened 2019-12-06
feat: Add string alias option for setting tensorlib in set_backend

### LINKED ISSUE #671: Use strings of backend names for set_backend
# Description

At the moment we set the tensor backend by passing an object:

```python
import pyhf
pyhf.set_backend(pyhf.tensor.pytorch_backend())
```

however, it would be nicer and easier for new users to just set the backend through a string

```python
import pyhf
pyhf.set_backend("pytorch")
```

There is the obvious catch on the fact that as long as TensorFlow v1 is used then the session needs to be managed. This will go away when TensorFlow v2 is adopted, but a somewhat inelegant solution that shouldn't affect anything is to add a kwarg of `_session` that defaults to `None`

```python
import pyhf
import tensorflow as tf
sess = tf.compat.v1.Session()
pyhf.set_backend("tensorflow", _session=sess)
```

@kratsg @lukasheinrich What are your thoughts on this? I currently have a branch that is passing CI with all of this implemented, so if you are in favor I can open a PR on it and we can go from there.

I don't have a solution to how one would also incorporate something similar for the optimizer choices at the moment as they can have args associated with themselves, so that still looks ugly

```python
import pyhf
pyhf.set_backend("numpy_minuit", pyhf.optimize.minuit_optimizer(verbose=True))
```

matthewfeickert:
# Description

Resolves #671 

Adds the ability to set the tensor backend with either a string alias or by passing a pyhf tensor backend object to `pyhf.set_backend`. This modifies the signature of `pyhf.set_backend` to be:

```
set_backend(backend, custom_optimizer=None, _session=None)
```

where `backend` can be either `str` or a pyhf tesnor backend object. As a user is allowed to pass a custom backend they have developed, a check is also added to make sure that any non-supported backends used do not share name attributes with supported backends.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add ability to set tensorlib with a string alias
   - Maintain ability to set tensorlib with a pyhf tensor backend object
* Add '_session' kwarg for TensorFlow v1 session management
* Add InvalidBackend exception
```


matthewfeickert:
For reasons I don't fully understand both CodeFactor and Codecov aren't happy ([Codecov is very confusing here](https://codecov.io/gh/scikit-hep/pyhf/pull/672/diff)). I'll let you decide if you want these to be investigated more if you're okay with them not passing everything.

kratsg:
> For reasons I don't fully understand both CodeFactor and Codecov aren't happy ([Codecov is very confusing here](https://codecov.io/gh/scikit-hep/pyhf/pull/672/diff)). I'll let you decide if you want these to be investigated more if you're okay with them not passing everything.

I'll comment on the lines that are problematic with how to fix it.

tests/test_pdf.py
```diff
@@ -17,7 +17,7 @@ def test_pdf_inputs(backend):
     pars = pdf.config.suggested_init()
     data = source['bindata']['data'] + pdf.config.auxdata
 
-    tensorlib, _ = backend
+    tensorlib, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_pdf.py
```diff
@@ -72,7 +72,7 @@ def test_pdf_basicapi_tests(backend):
...
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_pdf.py
```diff
@@ -90,7 +90,7 @@ def test_pdf_basicapi_tests(backend):
     pars = [pdf.config.suggested_init()] * 2
     data = source['bindata']['data'] + pdf.config.auxdata
 
-    tensorlib, _ = backend
+    tensorlib, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_pdf.py
```diff
@@ -173,7 +173,7 @@ def test_pdf_integration_staterror(backend):
     pdf = pyhf.Model(spec)
     par = pdf.config.par_slice('stat_firstchannel')
     par_set = pdf.config.param_set('stat_firstchannel')
-    tensorlib, _ = backend
+    tensorlib, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_probability.py
```diff
@@ -1,9 +1,10 @@
+import pyhf
 from pyhf import probability
 import numpy as np
 
 
 def test_poisson(backend):
-    tb, _ = backend
+    tb, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_probability.py
```diff
@@ -22,7 +23,7 @@ def test_poisson(backend):
 
 
 def test_normal(backend):
-    tb, _ = backend
+    tb, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_probability.py
```diff
@@ -45,7 +46,7 @@ def test_normal(backend):
 
 
 def test_joint(backend):
-    tb, _ = backend
+    tb, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_probability.py
```diff
@@ -54,7 +55,7 @@ def test_joint(backend):
 
 
 def test_independent(backend):
-    tb, _ = backend
+    tb, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_public_api.py
```diff
@@ -34,15 +34,15 @@ def test_hypotest(backend, model_setup):
 
 
 def test_prob_models(backend):
-    tb, _ = backend
+    tb, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_public_api.py
```diff
@@ -34,15 +34,15 @@ def test_hypotest(backend, model_setup):
...
     pyhf.probability.Poisson(tb.astensor([10.0])).log_prob(tb.astensor(2.0))
     pyhf.probability.Normal(tb.astensor([10.0]), tb.astensor([1])).log_prob(
         tb.astensor(2.0)
     )
 
 
 def test_pdf_batched(backend):
-    tb, _ = backend
+    tb, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_top_level.py
```diff
@@ -0,0 +1,20 @@
+import pyhf
```
kratsg: this should be in the `test_api` instead of a separate python test here. "top level" is vague.

tests/test_toys.py
```diff
@@ -3,12 +3,12 @@
 
 
 def test_smoketest_toys(backend):
-    tb, _ = backend
+    tb, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

tests/test_toys.py
```diff
@@ -3,12 +3,12 @@
...
     m = pyhf.simplemodels.hepdata_like([6], [9], [3])
     s = m.make_pdf(pyhf.tensorlib.astensor(m.config.suggested_init()))
     assert np.asarray(tb.tolist(s.log_prob(s.sample((1000,))))).shape == (1000,)
 
-    tb, _ = backend
+    tb, _ = pyhf.get_backend()
```
kratsg: this should not have been changed. We have to inject the backend this way for testing purposes (if the tests are in weird orders, the `pyhf.get_backend()` isn't properly reset for tests).

src/pyhf/__init__.py
```diff
@@ -27,44 +27,77 @@ def get_backend():
...
+            backend = tensor.numpy_backend()
+        elif backend == "tensorflow":
+            backend = tensor.tensorflow_backend(session=_session)
+        elif backend == "pytorch":
+            backend = tensor.pytorch_backend()
+        elif backend == "numpy_minuit":
+            backend = tensor.numpy_backend(poisson_from_normal=True)
+    elif not isinstance(backend, supported_backend_types) and not custom_backend:
+        raise ValueError(
+            "'{}' is not a supported backend.\n             Select from one of the supported backends: numpy, tensorflow, pytorch".format(
```
kratsg: In order to fix CodeFactor... (well, technically, bandit):

```suggestion
            "'{0:s}' is not a supported backend.\n             Select from one of the supported backends: numpy, tensorflow, pytorch".format(
```

src/pyhf/__init__.py
```diff
@@ -27,44 +27,77 @@ def get_backend():
...
     global optimizer
 
+    supported_backend_types = (
+        tensor.numpy_backend,
+        tensor.tensorflow_backend,
+        tensor.pytorch_backend,
+    )
+    supported_backend_names = (backend().name for backend in supported_backend_types)
+
+    if isinstance(backend, str) and (
```
kratsg: This doesn't work on python3 -- you either need to use something like

```
try:
  basestring
except:
  basestring = str
```

or use `six`

src/pyhf/__init__.py
```diff
@@ -27,44 +27,77 @@ def get_backend():
...
+    .. |tf.compat.v1.Session| replace:: ``tf.compat.v1.Session``
+    .. _tf.compat.v1.Session: https://www.tensorflow.org/api_docs/python/tf/compat/v1/Session
 
     Returns:
         None
     """
     global tensorlib
     global optimizer
 
+    supported_backend_types = (
```
kratsg: you can't use this piece of code. it will break if someone doesn't have tensorflow/pytorch installed and raise a `pyhf.exceptions.ImportBackendError` error.

src/pyhf/__init__.py
```diff
@@ -27,44 +27,77 @@ def get_backend():
...
+        tensor.numpy_backend,
+        tensor.tensorflow_backend,
+        tensor.pytorch_backend,
+    )
+    supported_backend_names = (backend().name for backend in supported_backend_types)
+
+    if isinstance(backend, str) and (
+        backend in supported_backend_names or backend == "numpy_minuit"
+    ):
+        if backend == "numpy":
```
kratsg: CodeFactor is complaining about a complex if/else conditional here. This can be replaced with fewer lines:

```python
backend = getattr(pyhf.tensor, '{0:s}_backend'.format(backend))
if not backend:
  raise ValueError("'{0:s}' is not a supported backend....")
```

src/pyhf/__init__.py
```diff
@@ -27,44 +27,77 @@ def get_backend():
...
+                backend
+            )
+        )
+
     # need to determine if the tensorlib changed or the optimizer changed for events
     tensorlib_changed = bool(backend.name != tensorlib.name)
     optimizer_changed = False
 
-    if backend.name == 'tensorflow':
+    # Check on instance of type instead of name to prevent custom name conflicts
```
kratsg: why not just check for `backend.name == "numpy"` or similar?

matthewfeickert: If we are going to be allowing for the possibility of custom backends there is nothing keeping someone from giving any name. So if they also named their custom backend "numpy" this would improperly treat it as the supported numpy backend, no?

src/pyhf/__init__.py
matthewfeickert: For some reason I was thinking that it would be important to support "numpy_minuit". I don't know why I thought that as it isn't, not is the name really good. So yes, this is the better way to do things if I ditch it. Thanks! :)

src/pyhf/__init__.py
matthewfeickert: Yes, very good point. Thanks for catching this. As we discussed, we should also add tests that run in different CI environments to test a (lack of) dependency in the environments.

matthewfeickert: This is now https://github.com/scikit-hep/pyhf/issues/674

src/pyhf/__init__.py
matthewfeickert: This is going to get changed anyway, but sorry to be daft but can you explain what is the problem here? I either don't understand what line you're referring to or I'm very confused.

```python
>>> import pyhf
>>> backend = "numpy"
>>> isinstance(backend, str)
True
>>> backend = pyhf.tensor.numpy_backend()
>>> isinstance(backend, str)
False
>>> isinstance(backend, pyhf.tensor.numpy_backend)
True
```

kratsg: anything that relies on `str` can't (for python3 -- it's `isinstance(backend, (str, bytes))`

matthewfeickert:
The tests that are currently failing are expected as they are the current release tests and this PR changes the public API.

src/pyhf/__init__.py
```diff
@@ -27,24 +28,55 @@ def get_backend():
...
+            backend = backend.decode("utf-8")
+        # Needed while still using TF v1.0 API
+        if backend == "tensorflow":
+            backend = tensor.tensorflow_backend(session=_session)
+        else:
+            try:
+                backend = getattr(tensor, "{0:s}_backend".format(backend))()
+            except TypeError:
+                raise InvalidBackend(
+                    "\"{0:s}\" is not a supported backend. Select from one of the supported backends: numpy, tensorflow, pytorch".format(
```
kratsg:
```suggestion
                    "The backend provided is not supported: {0:s}. Select from one of the supported backends: numpy, tensorflow, pytorch".format(
```

src/pyhf/__init__.py
```diff
@@ -27,24 +28,55 @@ def get_backend():
...
+def set_backend(backend, custom_optimizer=None, _session=None):
     """
     Set the backend and the associated optimizer
 
     Example:
         >>> import pyhf
         >>> import tensorflow as tf
-        >>> pyhf.set_backend(pyhf.tensor.tensorflow_backend(session=tf.Session()))
+        >>> sess = tf.compat.v1.Session()
+        >>> pyhf.set_backend("tensorflow", _session=sess)
```
kratsg: should show both kinds of examples.. using objects or using strings.

src/pyhf/__init__.py
```diff
@@ -27,24 +28,55 @@ def get_backend():
...
+            try:
+                backend = getattr(tensor, "{0:s}_backend".format(backend))()
+            except TypeError:
+                raise InvalidBackend(
+                    "\"{0:s}\" is not a supported backend. Select from one of the supported backends: numpy, tensorflow, pytorch".format(
+                        backend
+                    )
+                )
+
+    _name_supported = getattr(tensor, "{0:s}_backend".format(backend.name))
```
kratsg: what's the difference btwn this and line 63 above?

src/pyhf/exceptions/__init__.py
```diff
@@ -68,6 +68,12 @@ class ImportBackendError(Exception):
     """
 
 
+class InvalidBackend(Exception):
+    """
+    InvalidBackend is raised when trying to set using a backend that does not exist.
```
kratsg: "trying to set using"

tests/test_public_api.py
```diff
@@ -14,6 +14,58 @@ def model_setup(backend):
     return model, data, init_pars
 
 
+def test_set_backend_by_name(backend):
+    tb, _ = pyhf.get_backend()
+    pyhf.set_backend(tb.name)
```
kratsg: this is kinda confusing, since `set_backend()` is called with the `backend` fixture

tests/test_public_api.py
```diff
@@ -14,6 +14,58 @@ def model_setup(backend):
...
+
+@pytest.mark.parametrize("backend_name", ["numpy", "tensorflow", "pytorch"])
+def test_set_backend_by_string(backend_name):
+    pyhf.set_backend(backend_name)
+    assert isinstance(
+        pyhf.tensorlib, getattr(pyhf.tensor, "{0:s}_backend".format(backend_name))
+    )
+
+
+@pytest.mark.parametrize("backend_name", [b"numpy", b"tensorflow", b"pytorch"])
```
kratsg: how does this work in python2?

src/pyhf/__init__.py
matthewfeickert: Line 63 is:

```python
getattr(tensor, "{0:s}_backend".format(backend))()
```

while this is
```python
getattr(tensor, "{0:s}_backend".format(backend.name))
```

In like 63 `backend` is a string, and in line 71 `backend` is a `pyhf.tensor` object.

src/pyhf/exceptions/__init__.py
matthewfeickert: ? Sorry I'm not sure what your comment is suggesting.

kratsg: this isn't a grammatically correct sentence, and it reads awkwardly.

tests/test_public_api.py
matthewfeickert: I'm not sure what part of this would be incompatible with Python 2.7. Can you elaborate?

But:

```
$ docker run --rm -it -v $PWD:/home -w /home python:2.7 /bin/bash
# pip install --upgrade pip setuptools wheel
# pip install .
# python
>>> import pyhf
>>> backend_name = b"numpy"
>>> pyhf.set_backend(backend_name)
>>> assert isinstance(
...         pyhf.tensorlib, getattr(pyhf.tensor, "{0:s}_backend".format(backend_name))
...     )
>>> pyhf.tensorlib
<pyhf.tensor.numpy_backend.numpy_backend object at 0x7f644d6425d0>
>>> getattr(pyhf.tensor, "{0:s}_backend".format(backend_name))
<class 'pyhf.tensor.numpy_backend.numpy_backend'>

```

kratsg: ok, this is fine. Just wasn't sure.

src/pyhf/__init__.py
matthewfeickert: Done.

src/pyhf/exceptions/__init__.py
matthewfeickert: I copied the [docstring you wrote for `InvalidOptimizer`](https://github.com/scikit-hep/pyhf/pull/403/files#diff-37200bfdeb3cc4323cb5f51ba7cd9670R77) and then just changed "optimizer" to "backend". Do you want to change both at the same time?

How about
```
InvalidX is raised when trying to set a X that does not exist.
```

?

kratsg: that sounds better

matthewfeickert:
Btw, I think I would prefer to do the bumpversion in a separate dedicated PR. The last time we did it inside the PR and then no `v0.3.0` went up on TestPyPI.

kratsg:
> Btw, I think I would prefer to do the bumpversion in a separate dedicated PR. The last time we did it inside the PR and then no `v0.3.0` went up on TestPyPI.

Can't. because then we push two different v0.3.0 tags which don't represent the same commit, but represent the same state of code (maybe)...

matthewfeickert:
> because then we push two different v0.3.0 tags which don't represent the same commit, but represent the same state of code

I don't follow. That didn't happen when we did PR #645 (Bump version: 0.2.1 → 0.2.2), which is what I'm proposing we do again. Where are the two different tags coming from?

kratsg:
> I don't follow. That didn't happen when we did PR #645 (Bump version: 0.2.1 → 0.2.2), which is what I'm proposing we do again. Where are the two different tags coming from?

Tagging in the PR for tag A and then tagging in master for tag A will have the same tag point to two different commits because of how the squashing is done.

matthewfeickert:
> Tagging in the PR for tag A and then tagging in master for tag A will have the same tag point to two different commits because of how the squashing is done.

Yes, but that's not what I'm suggesting. I'm just suggesting that we do the same workflow we did for PR #645 : 

- New branch
- Run bumpversion on new branch
- Push branch but not tag and open a PR
- PR gets approved
- From the command line merge the branch into `master` and then push
- Push the tag after everything else worked

src/pyhf/__init__.py
```diff
@@ -27,24 +28,63 @@ def get_backend():
...
 
     Returns:
         None
     """
     global tensorlib
     global optimizer
 
+    if isinstance(backend, (str, bytes)):
+        if isinstance(backend, bytes):
+            backend = backend.decode("utf-8")
```
matthewfeickert: The backend strings should get put in lowercase.

```suggestion
            backend = backend.decode("utf-8").lower()
```

kratsg:
> Yes, but that's not what I'm suggesting. I'm just suggesting that we do the same workflow we did for PR #645 :
> 
> * New branch
> * Run bumpversion on new branch
> * Push branch but not tag and open a PR
> * PR gets approved
> * From the command line merge the branch into `master` and then push
> * Push the tag after everything else worked

Right, but no v0.3.0 makes it to testpypi because we don't push the tag in the PR.

matthewfeickert:
> Right, but no v0.3.0 makes it to testpypi because we don't push the tag in the PR.

Yup. I'm just meaning for _this_ PR as this is going to need a release bump.

src/pyhf/__init__.py
```diff
@@ -27,24 +28,64 @@ def get_backend():
...
+    .. |tf.compat.v1.Session| replace:: ``tf.compat.v1.Session``
+    .. _tf.compat.v1.Session: https://www.tensorflow.org/api_docs/python/tf/compat/v1/Session
 
     Returns:
         None
     """
     global tensorlib
     global optimizer
 
+    if isinstance(backend, (str, bytes)):
```
lukasheinrich: can't we use `string_types` here? is this just tu support unicode (or rather to support 2.7?)

matthewfeickert: > is this just to support unicode (or rather to support 2.7?)

Correct. I originally had just `str` but then Girodon pointed out that we would need to check on this for Python 2.7. I can use`string_types` but I figured why use `six` if we don't need to. We can if you prefer.

lukasheinrich: ok fine with me

PR #676, opened 2019-12-09
docs: make all warnings errors, and fix all warnings

kratsg:
- [xref](https://github.com/michaeljones/sphinx-xref/blob/master/xref.py) handles cross-references globally in the docs

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use setup.py build_sphinx to build docs
* Make any warnings when building docs be errors raised in CI
* Migrate pvals_from_teststat and pvals_from_teststat_expected
* Add missing utilities to utilities
* Use pkg_resources to get dirty/clean version of docs (so we version correctly)
* Use globs for notebooks, explicitly list notebooks we're not building in `conf.py`
* Use setup.cfg to configure some sphinx build variables
* Add roadmap governance and update it
* Use cross-references to standardize the use of URLs across documentation
```


docs/api.rst
```diff
@@ -109,18 +109,20 @@ Interpolators
    code4
    code4p
 
-Statistics
-----------
+Inference
```
matthewfeickert: Thanks for doing this already, as I was going to ask for it. :)

matthewfeickert:
I'll review later as I'm working on something else now, but I think this is too many changes for a single PR &mdash; even though it is just docs. I'd keep this closer in scope to its original incarnation and then make the other changes in another PR. This is mixing style and function now.

kratsg:
> I'll review later as I'm working on something else now, but I think this is too many changes for a single PR — even though it is just docs. I'd keep this closer in scope to its original incarnation and then make the other changes in another PR. This is mixing style and function now.

I needed to fix the warnings because I wasn't able to get the docs to build on my computer anymore... so when I fixed the warnings, things worked again.

kratsg:
> > I'll review later as I'm working on something else now, but I think this is too many changes for a single PR — even though it is just docs. I'd keep this closer in scope to its original incarnation and then make the other changes in another PR. This is mixing style and function now.
> 
> I needed to fix the warnings because I wasn't able to get the docs to build on my computer anymore... so when I fixed the warnings, things worked again.

Basically, I found a bug where because of how `pkg_resources` works -- it was picking up multiple python installs of pyhf from different virtual environments -- and that has to be fixed for me to at least get docs starting to build.

Then, when I switched from `Statistics` to `Inference`, we noted that the original problem with this was that we didn't catch the warning for the missing import, so that needed to be set -- and the current recommendation for doing so is to configure this through setup.cfg -- but a problem with that is that the docs won't pick up the non-tagged versions if it's in setup.cfg (since bumpversion doesn't support it but setuptools_scm does) and the setuptools_scm repository actually [recommends the changes to sphinx's docs/conf.py](https://github.com/pypa/setuptools_scm#usage-from-sphinx).

The last warnings came from multiple references (and I have an issue on my computer with the bibliography of intro.rst linking to likelihood.rst -- and that needed to be fixed). Similarly, the `arXiv` links were not actually rendering at all on our docs...

matthewfeickert:
> Basically, I found a bug where because of how `pkg_resources` works -- it was picking up multiple python installs of pyhf from different virtual environments -- and that has to be fixed for me to at least get docs starting to build.

That's great that this is fixed it, but that seems unrelated from renaming `Statistics` to `Inference` and making sure the links got corrected as well. 

> Then, when I switched from `Statistics` to `Inference`, 

This makes me very happy to have (more) correct naming of things. But that alone should be a PR &mdash; a complete thought/idea executed: to correct the naming.

> we noted that the original problem with this was that we didn't catch the warning for the missing import, so that needed to be set -- and the current recommendation for doing so is to configure this through setup.cfg -- but a problem with that is that the docs won't pick up the non-tagged versions if it's in setup.cfg (since bumpversion doesn't support it but setuptools_scm does) and the setuptools_scm repository actually [recommends the changes to sphinx's docs/conf.py](https://github.com/pypa/setuptools_scm#usage-from-sphinx).
> 
> The last warnings came from multiple references (and I have an issue on my computer with the bibliography of intro.rst linking to likelihood.rst -- and that needed to be fixed). Similarly, the `arXiv` links were not actually rendering at all on our docs...

Sorry, not really sure what you mean about the arXiv links not rendering. They are working fine on our currently deployed website. Can you elaborate?

> tl;dr https://i.imgur.com/rQIb4Vw.gifv

My recommendation for [Hal](https://media.giphy.com/media/NUBp5KcV0PJBe/giphy.gif) would be the same as for here: each ~~household~~ codebase item in need of repair gets its own ~~afternoon project~~ PR for clarity.

This PR is obviously great in that it fixes many things and makes the docs more robust and better, but why not have each issue get a much clearer representation?

What about starting by splitting this out into multiple PRs starting with just creating a new PR with the commits:
- update inference in docs
- fix api linking

? I've already done this on the branch `docs/change-stats-to-inference` and it is [passing the CI](https://github.com/scikit-hep/pyhf/runs/342441800) and when I build it locally on my machine things work fine. The rest of this PR which does nice functional things could then be worked on in another.

docs/conf.py
```diff
@@ -109,7 +114,27 @@ def setup(app):
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 # This patterns also effect to html_static_path and html_extra_path
-exclude_patterns = ['_build', '**.ipynb_checkpoints', 'examples/experiments']
+exclude_patterns = [
+    '_build',
+    '**.ipynb_checkpoints',
+    'examples/experiments/edwardpyhf.ipynb',
+    'examples/notebooks/ImpactPlot.ipynb',
+    'examples/notebooks/ImpactPlot.ipynb',
```
matthewfeickert: This line looks like it snuck in here twice given the line above.
```suggestion
```

Also, can these be added as just a directory instead of individual files? So just `examples/notebooks/`?

docs/examples.rst
```diff
@@ -10,23 +10,7 @@ Notebooks:
...
-   examples/notebooks/histogrammar.ipynb
-   examples/notebooks/histosys-pytorch.ipynb
-   examples/notebooks/histosys.ipynb
-   examples/notebooks/importxml.ipynb
-   examples/notebooks/multichannel-coupled-normsys.ipynb
-   examples/notebooks/multichannel-normsys.ipynb
-   examples/notebooks/normsys.ipynb
-   examples/notebooks/pytorch_tests_onoff.ipynb
-   examples/notebooks/tensorflow-limit.ipynb
+   examples/notebooks/*
```
matthewfeickert: I'm not sure we want to add all the notebooks that are under `docs/examples/notebooks/`. We currently only have the notebooks that we test in the notebooks CI being shown as examples and I think that's a good pattern to keep. Also, I think there are some notebooks in there that either need to be overhauled or have huge slabs of text added to them to give better context.

kratsg:
> This PR is obviously great in that it fixes many things and makes the docs more robust and better, but why not have each issue get a much clearer representation?
> 
> What about starting by splitting this out into multiple PRs starting with just creating a new PR with the commits:

The problem here is that I can't get a PR passing CI when I enable warnings->errors -- because it means I need to fix _all_ warnings, which is what happened. I'm following TDD where I write a failing test (assert no warnings) and then make it pass.

docs/exts/xref.py
```diff
@@ -0,0 +1,44 @@
...
+
+def xref(typ, rawtext, text, lineno, inliner, options={}, content=[]):
+
+    title = target = text
+    titleistarget = True
+    # look if explicit title and target are given with `foo <bar>` syntax
+    brace = text.find('<')
+    if brace != -1:
+        titleistarget = False
+        m = caption_ref_re.match(text)
```
matthewfeickert: Nitpick: Unless variables are inside a list comprehension or some generator expression I'd ask that we abandon single letter variables. (This affects the following 3 lines too).

```suggestion
        match = caption_ref_re.match(text)
```

docs/index.rst
```diff
@@ -17,6 +17,7 @@
    cli
    api
    citations
+   governance/ROADMAP
```
matthewfeickert: It seems that adding the ROADMAP to the docs (a good idea! :+1:) should be its own PR that is separate from the maintenance going on here.

matthewfeickert:
> arXiv links are broken: https://scikit-hep.org/pyhf/api.html#statistics
> 
> Screenshot 2019-12-10 10 26 17

Ah okay, yeah they are there. I thought you meant universally &mdash; they render fine on the API pages. Thanks for clarifying.

matthewfeickert:
> The problem here is that I can't get a PR passing CI when I enable warnings->errors -- because it means I need to fix _all_ warnings, which is what happened. I'm following TDD where I write a failing test (assert no warnings) and then make it pass.

Yes, and I'm very glad that you are. I'm just suggesting that the changes that are not related to ensuring that all warnings are accounted for become their own separate PRs so that it is much more clear what was needed for what.

docs/examples.rst
matthewfeickert: Nevermind, I see now this is all controlled through `conf.py`

docs/exts/xref.py
kratsg: We can't change this code. https://github.com/michaeljones/sphinx-xref/blob/master/xref.py

docs/index.rst
kratsg: Can't. It causes a warning->error.

docs/exts/xref.py
matthewfeickert: Okay. Since it isn't clear from the PR code can you add a short paragraph to the PR body about what `xref` is doing?

docs/index.rst
matthewfeickert: Fair.

setup.py
```diff
@@ -118,5 +98,5 @@ def _is_test_pypi():
     extras_require=extras_require,
     entry_points={'console_scripts': ['pyhf=pyhf.cli:cli']},
     dependency_links=[],
-    use_scm_version=_is_test_pypi(),
+    use_scm_version=lambda: {'local_scheme': lambda version: ''},
```
matthewfeickert: Can you add a short paragraph or some links or code snippet to the PR body that outline why the `IS_COMMIT_TAGGED` environment var isn't needed anymore?

matthewfeickert:
Could also stand for a more descriptive PR title. It is absolutely 100% true as is, but not going to tell us in the future a drive by overview of the PR goal.

setup.py
kratsg: I don't think it was ever needed... This is the part I'm really confused about to be honest. I tried with just this, without needing the IS_COMMIT_TAGGED and we properly tag on releases, and properly add devN on non-releases.

kratsg:
> Could also stand for a more descriptive PR title. It is absolutely 100% true as is, but not going to tell us in the future a drive by overview of the PR goal.

done

setup.py
matthewfeickert: Cool! That's great news.

PR #680, opened 2019-12-11
fix: Add git to Docker image during build to avoid setuptools-scm error

matthewfeickert:
# Description

In PR #676 the `_is_test_pypi` was removed as it was discovered that it wasn't needed for CI to pass. However, removing it caused issues in the Docker image build on Docker Hub as a result of `setuptools-scm` looking for `.git`

>     LookupError: setuptools-scm was unable to detect version for '/code'.
>    
>    Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.

 Adding git to the image fixes these problems.

Also use `COPY` instead of `ADD` as `COPY` is considered best practice over using `ADD` unless `ADD`'s tar and remote URL capabilities are needed.
c.f.
- https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy
- https://stackoverflow.com/a/24958548/8931942

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* apk add git to image to avoid setuptools-scm error
* Use COPY for best practices
* Add 'xmlio' extra in install to add ROOT file read support
* Use python -m pip for best pip practices
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,3 +1,6 @@
 FROM alectolytic/scipy
-ADD . /code
-RUN cd /code;pip install -e .
+COPY . /code
+RUN cd /code && \
+    rm -rf .git && \
+    python -m pip install --upgrade --no-cache-dir pip setuptools wheel && \
+    python -m pip install --no-cache-dir -e .
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,5 +4,5 @@ RUN cd /code && \
     apk add --no-cache git && \
     rm -rf /var/cache/apk/* && \
     python -m pip install --upgrade --no-cache-dir pip setuptools wheel && \
-    python -m pip install --no-cache-dir -e . && \
+    python -m pip install --no-cache-dir -e .[xmlio] && \
     python -m pip list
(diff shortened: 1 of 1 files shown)


kratsg:
Do we have a way to test docker builds as part of PRs?

matthewfeickert:
> Do we have a way to test docker builds as part of PRs?

We can add it as a separate part of the CI, like how we do for docs now.

docker/Dockerfile
```diff
@@ -1,3 +1,8 @@
 FROM alectolytic/scipy
-ADD . /code
-RUN cd /code;pip install -e .
+COPY . /code
+RUN cd /code && \
+    apk add --no-cache git && \
+    rm -rf /var/cache/apk/* && \
+    python -m pip install --upgrade --no-cache-dir pip setuptools wheel && \
+    python -m pip install --no-cache-dir -e . && \
```
matthewfeickert: Maybe we also give people the ability to use the `xmlio` extra here? That makes it a much more useful image.
```suggestion
    python -m pip install --no-cache-dir -e .[xmlio] && \
```

matthewfeickert:
@lukasheinrich this is ready for review now.

As @kratsg said we should also add CI for testing the build, but that can be in a follow up PR. The scope of this PR is just to fix the build.

PR #683, opened 2019-12-11
docs: add babel to describe HistFitter to pyhf workflow

kratsg:
# Description

This adds documentation in `docs/babel.rst` that describe how to extract the HistFactory workspaces for use by `pyhf xml2json`.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add documentation for HistFitter users on extracting HistFactory workspaces
```

### DIFF
--- a/docs/babel.rst
+++ b/docs/babel.rst
@@ -17,7 +17,7 @@ In order to go from `HistFitter` to `pyhf`, the first step is to extract out the
 
 The name of output workspace files depends on four parameters you define in your ``config.py``:
 
-- ``analysisName`` is from ``configMgr.analysisName``,
+- ``analysisName`` is from ``configMgr.analysisName``
 - ``prefix`` is defined in ``configMgr.addFitConfig({prefix})``, and
 - ``measurementName`` is the first measurement you define via ``fitConfig.addMeasurement(name={measurementName},...)``
 - ``channelName`` are the names of channels you define via ``fitConfig.addChannel("cuts", [{channelName}], ...)``
--- a/docs/babel.rst
+++ b/docs/babel.rst
@@ -1,6 +1,10 @@
 Translations
 ============
-One key goal of `pyhf` is to provide seamless translations from one statistical framework to us. This page details the various ways to translate from a tool you might already be using as part of an existing analysis to `pyhf`. Many of these solutions involve extracting out the `HistFactory` workspace and then running `pyhf xml2json <cli.html#pyhf-xml2json>`_ which provides a single JSON workspace that can be loaded directly into `pyhf`.
+One key goal of ``pyhf`` is to provide seamless translations between other statistical frameworks and ``pyhf``.
+This page details the various ways to translate from a tool you might already be using as part of an existing analysis to ``pyhf``.
+Many of these solutions involve extracting out the ``HistFactory`` workspace and then running |pyhf xml2json|_ which provides a single JSON workspace that can be loaded directly into ``pyhf``.
+
+.. |pyhf xml2json| replace:: ``pyhf xml2json``
 
 HistFitter
 ----------
--- a/docs/babel.rst
+++ b/docs/babel.rst
@@ -18,7 +18,7 @@ In order to go from `HistFitter` to `pyhf`, the first step is to extract out the
 The name of output workspace files depends on four parameters you define in your ``config.py``:
 
 - ``analysisName`` is from ``configMgr.analysisName``
-- ``prefix`` is defined in ``configMgr.addFitConfig({prefix})``, and
+- ``prefix`` is defined in ``configMgr.addFitConfig({prefix})``
 - ``measurementName`` is the first measurement you define via ``fitConfig.addMeasurement(name={measurementName},...)``
 - ``channelName`` are the names of channels you define via ``fitConfig.addChannel("cuts", [{channelName}], ...)``
 - ``cachePath`` is where `HistFitter` stores the cached histograms, defined by ``configMgr.histCacheFile`` which defaults to ``data/{analysisName}.root``
--- a/docs/babel.rst
+++ b/docs/babel.rst
@@ -29,7 +29,8 @@ To dump the HistFactory workspace, you will modify the above to skip the fit ``-
 
   HistFitter.py -wx -F excl config.py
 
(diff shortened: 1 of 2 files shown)


docs/babel.rst
```diff
@@ -0,0 +1,64 @@
...
+  # ...
+  channel = fitConfig.addChannel("cuts", ['SR_0L'], 1, 0.5, 1.5)
+  # ...
+  meas1=fitConfig.addMeasurement(name="DefaultMeasurement",lumi=1.0,lumiErr=0.029)
+  meas1.addPOI("mu_SIG1")
+  # ...
+  meas2=fitConfig.addMeasurement(name="DefaultMeasurement",lumi=1.0,lumiErr=0.029)
+  meas2.addPOI("mu_SIG2")
+
+Then, you expect the following files to be made:
```
matthewfeickert: Maybe explicitly state what these files are?

docs/babel.rst
```diff
@@ -0,0 +1,64 @@
...
+- ``cache/3b_tag21.2.27-1_RW_ExpSyst_36100_multibin_bkg.root``
+- ``results/3b_tag21.2.27-1_RW_ExpSyst_36100_multibin_bkg/Excl_combined_DefaultMeasurement.root``
+
+These are all the files you need in order to use `pyhf xml2json <cli.html#pyhf-xml2json>`_. At this point, you could run
+
+.. code:: bash
+
+    pyhf xml2json config/3b_tag21.2.27-1_RW_ExpSyst_36100_multibin_bkg/Excl.xml
+
+which will read all of the XML files and load the histogram data from the histogram cache.
```
matthewfeickert: While this will be clear soon after running, and is already explicitly stated from the Command Line API section in the docs, maybe it would be mentioning that as `xml2json` doesn't have a default value for `--output-file` they should provide this option here.

docs/babel.rst
```diff
@@ -0,0 +1,64 @@
+Translations
+============
+One key goal of `pyhf` is to provide seamless translations from one statistical framework to us. This page details the various ways to translate from a tool you might already be using as part of an existing analysis to `pyhf`. Many of these solutions involve extracting out the `HistFactory` workspace and then running `pyhf xml2json <cli.html#pyhf-xml2json>`_ which provides a single JSON workspace that can be loaded directly into `pyhf`.
+
+HistFitter
+----------
+
```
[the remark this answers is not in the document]
kratsg: Combine doesn't reaaaaaally need anything more though. It provides the workspace object which you run `pyhf xml2json` on without anything else. It's working on the workspaces directly.

docs/babel.rst
kratsg: Not in scope. This is for people using HistFitter. Not for random users.

docs/babel.rst
kratsg: Not in scope. This guide is only to get to the point where they can run `pyhf xml2json`.

kratsg:
Docs can be found here: https://pyhf-staging.readthedocs.io/en/docs-babelhistfitter/babel.html

docs/babel.rst
matthewfeickert: Why is that out of scope? Maybe I'm missing what you want the user of these docs to do, but if I was running through this for the first time I would expect to end with a file that I can do things with, not output to screen. You already are giving them an example run command.

setup.py
```diff
@@ -14,12 +14,18 @@
     'xmlio': ['uproot'],
     'minuit': ['iminuit'],
 }
-extras_require['test'] = sorted(
+extras_require['backends'] = sorted(
```
matthewfeickert: I guess I understand that you're just adding the `backends` extra, while the `test` extra still has the same effect, but it still feels like this is something that should be a separate PR. Eg, it would be good to be able to need to revert that PR without also reverting these nice docs that were added.  In this proposed additional PR can we also update the [Installation docs for "...with all backends"](https://scikit-hep.org/pyhf/installation.html#with-all-backends) to 

```
pip install pyhf[backends]
```

which is better then what is already there also as it automatically brings in additional backends as we add them to `setup.py`.

docs/babel.rst
```diff
@@ -0,0 +1,67 @@
+Translations
+============
+One key goal of ``pyhf`` is to provide seamless translations between other statistical frameworks and ``pyhf``.
+This page details the various ways to translate from a tool you might already be using as part of an existing analysis to ``pyhf``.
+Many of these solutions involve extracting out the ``HistFactory`` workspace and then running `pyhf xml2json <cli.html#pyhf-xml2json>`_ which provides a single JSON workspace that can be loaded directly into ``pyhf``.
+
+HistFitter
+----------
+
+In order to go from `HistFitter` to `pyhf`, the first step is to extract out the `HistFactory` workspaces. Assuming you have an existing configuration file, ``config.py``, you likely run an exclusion fit like so:
```
matthewfeickert: As the double quotes are being used above for pyhf and HistFactory they should be used throughout. Since I asked for this but didn't change it anywhere I will go ahead and make this change.

matthewfeickert: Done.

setup.py
matthewfeickert: This is now implemented in the [`refactor/create-backend-extra`](https://github.com/scikit-hep/pyhf/compare/refactor/create-backend-extra) branch. c.f. my edited review comment below.

kratsg: ok. just get that PR merged in first. because those changes need to happen to get the RTD to build.

kratsg:
This will be rebased off of #685.

setup.py
kratsg: #685

docs/babel.rst
kratsg: If the user does not know how to redirect output to a file, I don't think they should be following these instructions blindly...

kratsg: Additionally, these docs won't be tested, so having only the necessary information (and ignoring the CLI options that could change in the future) is going to make these robust/future-proof.

matthewfeickert:
My changes got overwritten in the force push. I'll fix this.

kratsg:
> My changes got overwritten in the force push. I'll fix this.

Screenshot 2019-12-13 14 01 39

I thought i grabbed them all.

matthewfeickert:
Fixed.

kratsg:
> LGTM. Thanks for making this PR!

np.

I've also checked with HistFitter folks, as well as the other SUSY Combination team members -- and they're happy with the doc as it stands.

lukasheinrich:
LGTM!

PR #685, opened 2019-12-13
refactor: Create 'backends' extra

kratsg:
# Description

This allows RTD to only build and mock the necessary things (also makes it so docs can build without needing to install all the packages that pyhf depends on!)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Create a 'backends' extra for installing all the supported backends
* Make the docs and test options for installing separate
* Mock the packages that are not needed to build the docs, but to make sure all pyhf modules are imported
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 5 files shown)


matthewfeickert:
As @kratsg has discovered, it seems like until https://github.com/dotnet/core/issues/3988 is resolved all of our CI is broken (and so are millions of other people's). So looks like we're going to loose at least a day of being able to merge PRs.

Issue #687, opened 2019-12-14
add nicer API for returning post-fit uncertainties for optimizers that support it


lukasheinrich:
# Description

currently we need to point people to example notebooks to hack it.

PR #531, opened 2019-08-26
refactor: Create infer subpackage to hold our statistics code

kratsg:
# Description

This will also be a `bumpversion/minor` update.

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR
```
* Create a pyhf.infer subpackage
* Move cli.stats to cli.infer
* Bump version: 0.2.2 → 0.3.0
```

PR #690, opened 2019-12-16
refactor: MLE API

lukasheinrich:
# Description

changes the optimizer API to only be about  minimization. Actual fitting (in the sense of maximum likelihood estimation) is moved to `pyhf.infer.mle`

#531 
#687 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Simplify optimizer API to only have 'minimize()'
* Add option to return minimized function in 'minimize()'
* Add optin to return uncertainties in the parameters for 'minimize()'
* Move concrete fits that are needed for maximum likelihood estimation / inference to pyhf.infer.mle
```


lukasheinrich:
This works then

screenshot

lukasheinrich:
the only reason CI is failing is due to an addition of a new file as part of the refactoring

matthewfeickert:
> the only reason CI is failing is due to an addition of a new file as part of the refactoring

Later today I can work on having Codecov only fail if there is a greater than 0.2% change or something like that.

src/pyhf/optimize/opt_minuit.py
```diff
@@ -45,23 +45,25 @@ def f(pars):
...
-        return self.minimize(
-            objective,
-            data,
-            pdf,
-            init_pars,
-            par_bounds,
-            [(pdf.config.poi_index, constrained_mu)],
-        )
+        if return_uncertainties:
+            bestfit_pars = np.asarray([(v, mm.errors[k]) for k, v in mm.values.items()])
```
matthewfeickert: This line is apparently uncovered.

https://codecov.io/gh/scikit-hep/pyhf/compare/1d5098395d33c37b4dd1844c8b201f284adae7fc...4567aa82acd46dfdf474b73ea720a083c18a4ec1/diff#D6-63

Can we fix that?

lukasheinrich: done

src/pyhf/infer/mle.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def fit(data, pdf, init_pars=None, par_bounds=None, **kwargs):
+    _, opt = get_backend()
+    init_pars = init_pars or pdf.config.suggested_init()
+    par_bounds = par_bounds or pdf.config.suggested_bounds()
+    return opt.minimize(loglambdav, data, pdf, init_pars, par_bounds, **kwargs)
+
+
+def fixed_poi_fit(constrained_mu, data, pdf, init_pars=None, par_bounds=None, **kwargs):
```
kratsg:
```suggestion
def fixed_poi_fit(poi_val, data, pdf, init_pars=None, par_bounds=None, poi_index=None **kwargs):
```

and add sth like

```
poi_index = poi_index or pdf.config.poi_index
```

matthewfeickert:
As it seems that there is a lot of conversation around the `fval` naming convention (which I am still thinking about too) can we get some docstrings on all the class methods that have `return_fval` as an arg?

src/pyhf/infer/mle.py
lukasheinrich: hm I don't agree. the current API only knows a POI as a proprty of the model config. Allowing this to be overridden doesn't feel right (i.e. while it is `m.config.suggested_init()`, it's `m.config.poi_index` not `m.config.suggested_poi_index`). I would leave this overriding ability to whenever we do the larger refactor of passing in methods

lukasheinrich:
added  docstrings @matthewfeickert @kratsg

tests/test_optim.py
```diff
@@ -67,12 +67,67 @@ def test_optim(backend, source, spec, mu):
 
     optim = pyhf.optimizer
 
-    result = optim.unconstrained_bestfit(
-        pyhf.infer.utils.loglambdav, data, pdf, init_pars, par_bounds
+    result = optim.minimize(pyhf.infer.mle.loglambdav, data, pdf, init_pars, par_bounds)
```
matthewfeickert: I'm confused as to how this is passing given that

```
$ git checkout errorapi
$ git pull
$ git grep "loglambdav" src/
```

returns no results. Shouldn't these be using the `fit` API now? Also why do these test not fail?

docs/api.rst
```diff
@@ -119,7 +119,7 @@ Inference
 
    hypotest
    test_statistics.qmu
-   utils.loglambdav
+   mle.loglambdav
```
matthewfeickert: For reasons brought up below this should be
```suggestion
   mle.fit
```

and then 
```
mle.fixed_poi_fit
```

should also be added.

tests/test_optim.py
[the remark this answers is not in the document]
lukasheinrich: I just used the renamed function name

src/pyhf/infer/mle.py
```diff
@@ -0,0 +1,65 @@
+"""Module for Maximum Likelihood Estimation."""
+from .. import get_backend
+
+
+def twice_nll(pars, data, pdf):
```
kratsg: `nll` is a sliiiiiiightly bothersome name for HEP physicists (NLO, NLL, etc...). Is there no official name for this calculable?

src/pyhf/infer/mle.py
```diff
@@ -0,0 +1,65 @@
...
+
+    """
+    _, opt = get_backend()
+    init_pars = init_pars or pdf.config.suggested_init()
+    par_bounds = par_bounds or pdf.config.suggested_bounds()
+    return opt.minimize(twice_nll, data, pdf, init_pars, par_bounds, **kwargs)
+
+
+def fixed_poi_fit(
+    poi_val, data, pdf, init_pars=None, par_bounds=None, poi_index=None, **kwargs
```
kratsg: poi_index=None? I thought you said we weren't including it. If you are, don't we need to handle the default case when it's not set by the user (and grab it from the pdf)?

src/pyhf/infer/test_statistics.py
```diff
@@ -30,13 +30,11 @@ def qmu(mu, data, pdf, init_pars, par_bounds):
         Float: The calculated test statistic, :math:`q_{\mu}`
     """
     tensorlib, optimizer = get_backend()
-    mubhathat = optimizer.constrained_bestfit(
-        loglambdav, mu, data, pdf, init_pars, par_bounds
+    mubhathat, fixed_val = fixed_poi_fit(
```
kratsg:
```suggestion
    mubhathat, fixed_result = fixed_poi_fit(
```

src/pyhf/infer/test_statistics.py
```diff
@@ -30,13 +30,11 @@ def qmu(mu, data, pdf, init_pars, par_bounds):
...
+        mu, data, pdf, init_pars, par_bounds, return_fval=True
     )
-    muhatbhat = optimizer.unconstrained_bestfit(
-        loglambdav, data, pdf, init_pars, par_bounds
-    )
-    qmu = loglambdav(mubhathat, data, pdf) - loglambdav(muhatbhat, data, pdf)
+    muhatbhat, float_val = fit(data, pdf, init_pars, par_bounds, return_fval=True)
+    qmu = fixed_val - float_val
```
kratsg:
```suggestion
    qmu = fixed_result - fluid_result
```

src/pyhf/infer/test_statistics.py
```diff
@@ -30,13 +30,11 @@ def qmu(mu, data, pdf, init_pars, par_bounds):
...
-    mubhathat = optimizer.constrained_bestfit(
-        loglambdav, mu, data, pdf, init_pars, par_bounds
+    mubhathat, fixed_val = fixed_poi_fit(
+        mu, data, pdf, init_pars, par_bounds, return_fval=True
     )
-    muhatbhat = optimizer.unconstrained_bestfit(
-        loglambdav, data, pdf, init_pars, par_bounds
-    )
-    qmu = loglambdav(mubhathat, data, pdf) - loglambdav(muhatbhat, data, pdf)
+    muhatbhat, float_val = fit(data, pdf, init_pars, par_bounds, return_fval=True)
```
kratsg:
```suggestion
    muhatbhat, fluid_result = fit(data, pdf, init_pars, par_bounds, return_fit_result=True)
```

src/pyhf/infer/test_statistics.py
```diff
@@ -30,13 +30,11 @@ def qmu(mu, data, pdf, init_pars, par_bounds):
         Float: The calculated test statistic, :math:`q_{\mu}`
     """
     tensorlib, optimizer = get_backend()
-    mubhathat = optimizer.constrained_bestfit(
-        loglambdav, mu, data, pdf, init_pars, par_bounds
+    mubhathat, fixed_val = fixed_poi_fit(
+        mu, data, pdf, init_pars, par_bounds, return_fval=True
```
kratsg:
```suggestion
        mu, data, pdf, init_pars, par_bounds, return_fit_result=True
```

src/pyhf/optimize/autodiff.py
```diff
@@ -1,9 +1,32 @@
+"""Helper Classes for use of automatic differentiation."""
 import scipy
 from .. import get_backend
 
 
 class AutoDiffOptimizerMixin(object):
-    def minimize(self, objective, data, pdf, init_pars, par_bounds, fixed_vals=None):
+    """Mixin Class to build optimizers that use automatic differentiation."""
+
+    def __init__(*args, **kwargs):
```
kratsg: we don't need to define the `__init__` if you're not changing it.

src/pyhf/optimize/autodiff.py
```diff
@@ -1,9 +1,32 @@
...
+
+    def minimize(
+        self,
+        objective,
+        data,
+        pdf,
+        init_pars,
+        par_bounds,
+        fixed_vals=None,
+        return_fval=False,
```
kratsg:
```suggestion
        return_fit_result=False,
```

src/pyhf/infer/mle.py
lukasheinrich: NLL (see e.g. createNLL in RooFit) is fairly standard. If it were a likelihood ratio it'd be https://en.wikipedia.org/wiki/Deviance_(statistics) but it's not

src/pyhf/optimize/autodiff.py
lukasheinrich: hm, I feel like the "fit result" is more the bestfit parameters. The value of the objective at the best fit point is more additional info

src/pyhf/infer/test_statistics.py
lukasheinrich: something that floats is not really flows, no? :)

src/pyhf/infer/mle.py
lukasheinrich: yeah, you're right, removed the arg (wasn't using it)

src/pyhf/infer/test_statistics.py
kratsg: i misinterpreted this as a floating type... `float_val` or `int_val`. `fixed` and `fluid` borrows from web-design (e.g. [fixed vs fluid](https://www.techprevue.com/fixed-layout-vs-fluid-layout-web-design/)).

src/pyhf/optimize/autodiff.py
lukasheinrich: thought so at first as well but it's needed bkg as a mixin, arg,kwarg can be filled and if super().__init__ is called on `object` it balks

src/pyhf/optimize/autodiff.py
kratsg: to be fair, you called it "result" in the tests so... not sure... I would prefer `fitted_pars, fitted_vals` instead of `result, fval` or `pars, fval`. `fval` is just too ambiguous for me.

src/pyhf/infer/test_statistics.py
kratsg: or `varied_val`?

src/pyhf/optimize/autodiff.py
kratsg: oh, right, good point... we should just delete the `pass` line or have codefactor ignore it...

PR #694, opened 2019-12-17
feat: Add contrib module

lukasheinrich:
# Description

adds contrib directory and seeds it with a poi scan plot. the idea behind `contrib` is to have often-repeated code e.g. in the example noteboook centrally managed, however nothing in the non-`contrib` part of pyhf  (src/pyhf) must depend on on it.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add pyhf.contrib
* Add pyhf.contrib.viz.brazil for POI scan "Brazil band" plots
```


matthewfeickert:
While doing testing of visualization libraries is hard, it is possible. Can we get tests for `contrib/viz`?

lukasheinrich:
@matthewfeickert not sure.. any ideas?

lukasheinrich:
https://towardsdatascience.com/unit-testing-python-data-visualizations-18e0250430 ?

kratsg:
> @matthewfeickert not sure.. any ideas?

https://matplotlib.org/3.1.1/devel/testing.html#writing-an-image-comparison-test

Kinda works like betamax/vcr works with python requests... record and compare existing recording.

kratsg:
So one thing I was playing around with locally for `contrib` was using it to attach new functions to the existing classes we have. E.G. when importing `contrib`, it would automatically add a `parameter_scan` function like so

in `contrib/__init__.py`
```python
from ..pdf import Model
from .utils import parameter_scan
Model.parameter_scan = parameter_scan
```

since I kinda treated the `contrib` as a "decoration" of the existing code, rather than adding new things...

lukasheinrich:
monkey patching the rest of the codebase seems a bit too magical to me :).

matthewfeickert:
> @matthewfeickert not sure.. any ideas?

I think we can pass on having them in this PR and then make an Issue to get them in soon.

src/pyhf/contrib/viz/brazil.py
```diff
@@ -0,0 +1,13 @@
+import numpy as np
+
+
+def plot_results(ax, mutests, tests, test_size=0.05):
+    cls_obs = np.array([test[0] for test in tests]).flatten()
+    cls_exp = [np.array([test[1][i] for test in tests]).flatten() for i in range(5)]
+    ax.plot(mutests, cls_obs, c='k')
```
matthewfeickert: I'd suggest using verbose color names
```suggestion
    ax.plot(mutests, cls_obs, c='black')
```

src/pyhf/contrib/viz/brazil.py
```diff
@@ -0,0 +1,13 @@
...
+    for i, c in zip(range(5), ['k', 'k', 'k', 'k', 'k']):
+        ax.plot(mutests, cls_exp[i], c=c, linestyle='dotted' if i != 2 else 'dashed')
+    ax.fill_between(mutests, cls_exp[0], cls_exp[-1], facecolor='y')
```
matthewfeickert:
```suggestion
    ax.fill_between(mutests, cls_exp[0], cls_exp[-1], facecolor='yellow')
```

src/pyhf/contrib/viz/brazil.py
```diff
@@ -0,0 +1,13 @@
...
+    ax.fill_between(mutests, cls_exp[1], cls_exp[-2], facecolor='g')
```
matthewfeickert:
```suggestion
    ax.fill_between(mutests, cls_exp[1], cls_exp[-2], facecolor='green')
```

src/pyhf/contrib/viz/brazil.py
```diff
@@ -0,0 +1,13 @@
...
+    ax.plot(mutests, [test_size] * len(mutests), c='r')
```
matthewfeickert:
```suggestion
    ax.plot(mutests, [test_size] * len(mutests), c='red')
```

lukasheinrich:
I settled for https://github.com/matplotlib/pytest-mpl which seems to work

lukasheinrich:
I tested this with an intentionally failing build. The reference images are in `tests/contrib/baseline`

screenshot

maybe in a new PR we could add some github CI magic that would *post* the reference, new and diff image to the PR when a visual test fails.

kratsg:
> maybe in a new PR we could add some github CI magic that would _post_ the reference, new and diff image to the PR when a visual test fails.

I like this idea. It should be doable with the GitHub API + octokit.

matthewfeickert:
Why is the coverage dropping so sharply?

> Merging #694 into master will decrease coverage by 70.96%.

This seems wrong, but I'm confused as to what is happening.

lukasheinrich:
@matthewfeickert I think the comment  is a old run of codecoov, the latest check on codecov should be green

matthewfeickert:
> @matthewfeickert I think the comment is a old run of codecoov, the latest check on codecov should be green

I rebased the PR to trigger a new run of codecov. Hopefully once the CI runs all the way through it will change. If so, I'll merge in when back from dinner tonight (or one of you can also). :)

matthewfeickert:
Nope. Sadly Codecov is just being weird and is reporting a coverage of 21% now. WTF. :(

matthewfeickert:
I've reported this now: https://community.codecov.io/t/large-coverage-drop-after-addition-of-100-covered-file/842

PR #697, opened 2019-12-17
ci: Use Python 3 scipy Alpine image for base Docker image

### LINKED ISSUE #692: Python 3 base image for Docker image build
# Description

We will be dropping Python 2.7 at the end of the year, but that also means that our current Docker image is going to fail in the build as the base image of `alectolytic/scipy` is Python 2.7. So we will need to find an Alpine image with Python 3 and SciPy before the end of December.

@lukasheinrich Do you know of one?

matthewfeickert:
# Description

Resolves #692 

To be able to drop Python 2 we will need to move away from the Python 2.7 Alpine Docker image that is used as the base image for the pyhf Docker image. This replaces it with a [Python 3 Docker image](https://github.com/matthewfeickert/python3-scipy-alpine).

This image uses the `py3-scipy` `apk` package, and so while not quite as small as the installed form PyPI scipy image (Python 2.7) it is just a few MB larger and could be brought down

```
$ docker images | grep scipy
matthewfeickert/python3-scipy-alpine                     latest              fba7df6a582c        2 hours ago         292MB
alectolytic/scipy                                        latest              00f11a79976b        15 months ago       276MB
```

which I think is pretty good.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use a Python 3 Apline image with scipy as the base image for the pyhf Docker image
   - c.f. https://github.com/matthewfeickert/python3-scipy-alpine
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM matthewfeickert/python3-scipy-alpine:latest
+FROM matthewfeickert/python3-scipy-alpine:scipy-1.2.1
 COPY . /code
 RUN cd /code && \
     apk add --no-cache git && \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM matthewfeickert/python3-scipy-alpine:scipy-1.2.1 # PR 697
+FROM matthewfeickert/python3-scipy-alpine:scipy-1.2.1
 COPY . /code
 RUN cd /code && \
     apk add --no-cache git && \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM matthewfeickert/python3-scipy-alpine:scipy-1.2.1
+FROM matthewfeickert/python3-scipy-alpine:scipy-1.2.1 # PR 697
 COPY . /code
 RUN cd /code && \
     apk add --no-cache git && \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM alectolytic/scipy
+FROM matthewfeickert/python3-scipy-alpine:latest
 COPY . /code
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alectolytic/scipy
+FROM matthewfeickert/python3-scipy-alpine:scipy-1.2.1
```
lukasheinrich: Is there a github repo somewhere to see how this is built?

matthewfeickert: yup. Linked in the description, but also here: https://github.com/matthewfeickert/python3-scipy-alpine

lukasheinrich: ok will we find this again if we forget? maybe link PR number as comment in the dockerfile?

matthewfeickert: > ok will we find this again if we forget?

Sorry, I think I'm missing something simple. What is the "this" what we will forget? The Dockerfile for the base image?

lukasheinrich: was just thinking `FROM matthewfeickert/python3-scipy-alpine:scipy-1.2.1  # issue #697` but yeah no strong opinion

matthewfeickert: Done in https://github.com/scikit-hep/pyhf/pull/697/commits/745fed951d178e1b07b9bbf21349619a348d419a :+1:

matthewfeickert: Oh that actually breaks things. I'm going to revert it.

Uh, unless you have other ideas maybe we just go with `git log -- docker/Dockerfile` for future reminders?

lukasheinrich:
ok cool

docker/Dockerfile
matthewfeickert: @kratsg If you have thoughts here they'd be welcome.

kratsg: no ideas unfortunately

Issue #271, opened 2018-09-18
NumPy backend configurable to use single-float precision


kratsg:
# Description

When comparing many calculations between single-float and double-float precision backends, there will be a (sometimes non-negligible) deviation between the two because of this precision. This doesn't necessarily imply that the calculation is wrong, since it depends on the underlying precision of the floats (64-bit versus 32-bit). NumPy and Python by default use whatever is available on the system (e.g. 64-bit most of the time these days) while tensorflow and pytorch are using single-float precision (e.g. 32-bit).

We should make sure the numpy backend can be configured to use 32-bit precision through something like `pyhf.tensor.numpy_backend(use_single_floats=True)` as a flag that will pass in `dtype=np.float32` for all functions in the numpy backend.

# Relevant Issues and Pull Requests

- #251

lukasheinrich:
this won't work since fitting at single precisiono often fails without exact gradients, instead we have added double precision for autodiff backends

PR #701, opened 2019-12-18
feat: double precision for ML backends

lukasheinrich:
# Description

Moves to double precision in ML backends (though one of the advantages of AutoDiff was that optimization would success even when using single precison, what should be chose as default @matthewfeickert @kratsg ). In any case having the option to run double precision in all backends is good. The results match more closely (despite different minimization strategies 

Adresses #271 but in the opposite way, since single precision numpy isn't really feasible without autodiff.

```
{
    "CLs_exp": [
        1.94672389834305e-09,
        1.3010988044241188e-07,
        7.282370205974331e-06,
        0.000292182057317543,
        0.006623453577998972
    ],
    "CLs_obs": 3.6246795231118695e-06
}

real	0m4.507s
user	0m4.088s
sys	0m0.492s
 pyhfdevenv  lukasheinrich  mpb2019lheinric  ~  Code  pyhfdev  dev  pyhfsrc  master  6✎  13?  $  time pyhf cls lhood.json -p patch.json  --backend tf
/Users/lukasheinrich/Code/pyhfdev/dev/pyhfdevenv/lib/python3.6/site-packages/pandas/compat/__init__.py:85: UserWarning: Could not import the lzma module. Your installed Python is incomplete. Attempting to use lzma compression will result in a RuntimeError.
  warnings.warn(msg)
/Users/lukasheinrich/Code/pyhfdev/dev/pyhfdevenv/lib/python3.6/site-packages/pandas/compat/__init__.py:85: UserWarning: Could not import the lzma module. Your installed Python is incomplete. Attempting to use lzma compression will result in a RuntimeError.
  warnings.warn(msg)
2019-12-18 11:36:42.970841: I tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 FMA
2019-12-18 11:36:42.988807: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x7fadc66a8290 initialized for platform Host (this does not guarantee that XLA will be used). Devices:
2019-12-18 11:36:42.988823: I tensorflow/compiler/xla/service/service.cc:176]   StreamExecutor device (0): Host, Default Version
WARNING:tensorflow:From /Users/lukasheinrich/Code/pyhfdev/dev/pyhfsrc/src/pyhf/optimize/opt_tflow.py:40: The name tf.placeholder is deprecated. Please use tf.compat.v1.placeholder instead.

WARNING:tensorflow:From /Users/lukasheinrich/Code/pyhfdev/dev/pyhfdevenv/lib/python3.6/site-packages/tensorflow_core/python/ops/math_grad.py:281: setdiff1d (from tensorflow.python.ops.array_ops) is deprecated and will be removed after 2018-11-30.
Instructions for updating:
This op will be removed after the deprecation date. Please switch to tf.sets.difference().
WARNING:tensorflow:From /Users/lukasheinrich/Code/pyhfdev/dev/pyhfdevenv/lib/python3.6/site-packages/tensorflow_core/python/ops/math_grad.py:1375: where (from tensorflow.python.ops.array_ops) is deprecated and will be removed in a future version.
Instructions for updating:
Use tf.where in 2.0, which has the same broadcast rule as np.where
{
    "CLs_exp": [
        1.946725694260987e-09,
        1.301098823154145e-07,
        7.282370205968245e-06,
        0.00029218204623841445,
        0.006623453528608673
    ],
    "CLs_obs": 3.6246795231138554e-06
}

real	0m15.268s
user	0m14.458s
sys	0m1.245s
 pyhfdevenv  lukasheinrich  mpb2019lheinric  ~  Code  pyhfdev  dev  pyhfsrc  master  6✎  13?  $  time pyhf cls lhood.json -p patch.json  --backend numpy
/Users/lukasheinrich/Code/pyhfdev/dev/pyhfsrc/src/pyhf/tensor/numpy_backend.py:253: RuntimeWarning: invalid value encountered in log
  return n * np.log(lam) - lam - gammaln(n + 1.0)
{
    "CLs_exp": [
        1.94688222641348e-09,
        1.3011880504076847e-07,
        7.282784255220192e-06,
        0.0002921952842746209,
        0.00662367909427313
    ],
    "CLs_obs": 3.6247127954767586e-06
}

real	0m9.204s
user	0m8.781s
sys	0m0.397s
```

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add option to use double precision in ML backends
```

### DIFF
--- a/src/pyhf/tensor/tensorflow_backend.py
+++ b/src/pyhf/tensor/tensorflow_backend.py
@@ -13,9 +13,9 @@ class tensorflow_backend(object):
         self.session = kwargs.get('session')
         self.name = 'tensorflow'
         self.dtypemap = {
-            'float': getattr(tf,kwargs.get('float','float32')),
-            'int': getattr(tf,kwargs.get('int','int32')),
-            'bool': tf.bool
+            'float': getattr(tf, kwargs.get('float', 'float32')),
+            'int': getattr(tf, kwargs.get('int', 'int32')),
+            'bool': tf.bool,
         }
 
     def clip(self, tensor_in, min_value, max_value):
@@ -203,10 +203,10 @@ class tensorflow_backend(object):
         return tf.abs(tensor)
 
     def ones(self, shape):
-        return tf.ones(shape,dtype = self.dtypemap['float'])
+        return tf.ones(shape, dtype=self.dtypemap['float'])
 
     def zeros(self, shape):
-        return tf.zeros(shape,dtype = self.dtypemap['float'])
+        return tf.zeros(shape, dtype=self.dtypemap['float'])
 
     def power(self, tensor_in_1, tensor_in_2):
         return tf.pow(tensor_in_1, tensor_in_2)
@@ -478,8 +478,7 @@ class tensorflow_backend(object):
             TensorFlow Tensor: The CDF
         """
         normal = tfp.distributions.Normal(
-            self.astensor(mu,dtype='float'), 
-            self.astensor(sigma,dtype='float'), 
+            self.astensor(mu, dtype='float'), self.astensor(sigma, dtype='float'),
         )
         return normal.cdf(x)
 
--- a/src/pyhf/tensor/tensorflow_backend.py
+++ b/src/pyhf/tensor/tensorflow_backend.py
@@ -1,3 +1,4 @@
+"""Tensorflow Tensor Library Module."""
 import logging
 import tensorflow as tf
 import tensorflow_probability as tfp
@@ -11,6 +12,11 @@ class tensorflow_backend(object):
     def __init__(self, **kwargs):
         self.session = kwargs.get('session')
         self.name = 'tensorflow'
+        self.dtypemap = {
+            'float': getattr(tf,kwargs.get('float','float32')),
+            'int': getattr(tf,kwargs.get('int','int32')),
+            'bool': tf.bool
+        }
 
     def clip(self, tensor_in, min_value, max_value):
         """
@@ -158,9 +164,8 @@ class tensorflow_backend(object):
         Returns:
             `tf.Tensor`: A symbolic handle to one of the outputs of a `tf.Operation`.
         """
-        dtypemap = {'float': tf.float32, 'int': tf.int32, 'bool': tf.bool}
         try:
-            dtype = dtypemap[dtype]
+            dtype = self.dtypemap[dtype]
         except KeyError:
             log.error('Invalid dtype: dtype must be float, int, or bool.')
             raise
@@ -198,10 +203,10 @@ class tensorflow_backend(object):
         return tf.abs(tensor)
 
     def ones(self, shape):
-        return tf.ones(shape)
+        return tf.ones(shape,dtype = self.dtypemap['float'])
 
     def zeros(self, shape):
-        return tf.zeros(shape)
+        return tf.zeros(shape,dtype = self.dtypemap['float'])
 
     def power(self, tensor_in_1, tensor_in_2):
         return tf.pow(tensor_in_1, tensor_in_2)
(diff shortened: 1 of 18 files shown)


tests/conftest.py
```diff
@@ -44,13 +44,14 @@ def reset_backend():
...
         (pyhf.tensor.pytorch_backend(), None),
+        (pyhf.tensor.pytorch_backend(float='float64', int='int64'), None),
         (pyhf.tensor.tensorflow_backend(session=tf.compat.v1.Session()), None),
         (
             pyhf.tensor.numpy_backend(poisson_from_normal=True),
             pyhf.optimize.minuit_optimizer(),
         ),
     ],
-    ids=['numpy', 'pytorch', 'tensorflow', 'numpy_minuit'],
+    ids=['numpy', 'pytorch', 'pytorch64', 'tensorflow', 'numpy_minuit'],
```
matthewfeickert: Maybe I'm missing something obvious, but why don't we also have a `'tensorflow64'` opetion here?

lukasheinrich: we could add it, just worried it would increase testing time at some pooint?

matthewfeickert: Yeah, it definitely will. It just seems a bit strange to only have one of the backends be tested at 64. I'll approve for now and let @kratsg give further thoughts.

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -198,10 +204,10 @@ def abs(self, tensor):
         return tf.abs(tensor)
 
     def ones(self, shape):
-        return tf.ones(shape)
+        return tf.ones(shape, dtype=self.dtypemap['float'])
```
kratsg: aren't these ints by default?

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -198,10 +204,10 @@ def abs(self, tensor):
...
 
     def zeros(self, shape):
-        return tf.zeros(shape)
+        return tf.zeros(shape, dtype=self.dtypemap['float'])
```
kratsg: aren't these ints by default?

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -443,12 +445,11 @@ def normal(self, x, mu, sigma):
         normal = tfp.distributions.Normal(mu, sigma)
         return normal.prob(x)
 
-    def normal_cdf(self, x, mu=0, sigma=1):
+    def normal_cdf(self, x, mu=0.0, sigma=1):
```
kratsg: mu was changed to 0.0, but not sigma to 1.0?

tests/conftest.py
kratsg: i would rather have this split up into 64b backends and 32b backend parameterizations (using `backend32` and `backend64`). Since it's clear we really need to be concerned about consistency, we should be keeping things on equal footing.

src/pyhf/tensor/tensorflow_backend.py
lukasheinrich: nope, they need to be floats (for use in the modifiers)

tests/conftest.py
lukasheinrich: hm, but then do we need separate tests that request the backends64 fixture?

kratsg: I've been thinking about this for a while. I suspect that we will be able to test to a tighter precision with 64b than we can with 32b. In which case, I would want to see stronger tests for the 64b backend than the 32b. This could just be parameterized by adding information about the precision along with the backend?

lukasheinrich: yes we can absolutely compare numpy-torch-tf at higher precision when they're all at double precision (and perhaps even use the same optimizer)

PR #704, opened 2019-12-18
feat: Drop Python 2.7 support

### LINKED ISSUE #469: Drop Python 2.7 support

matthewfeickert:
# Description

Resolves #469

On ~~December 28th, 2019~~ January 13th, 2020 pyhf will drop support for Python 2.7. This PR implements that change.

This PR makes the codebase Python 3 _only_. I think that this is fine to do, given that pyhf is currently something a standalone tool that is run as in independent step in an analysis. So if a user only has access to Python 2 we can suggest that:

* If the user is on LXPLUS then they can setup Python 3 environment (there is a Red Hat Python 3.6 runtime with `source scl_source enable rh-python36`) and use that, as thanks to uproot any step that involves pyhf can be run in isolation from any ENV that requires ROOT
* If they are running on a local machine then we can try and point them to the best resources on how to get a Python 3 runtime with minimal pain
* Worst case scenario we can tell people to use our Docker image until there is a better solution

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Drop Python 2.7 support for pyhf
* Add no Python 2.7 support to FAQ and link to last supported release
```

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -14,6 +14,13 @@ Use the :code:`--backend` option for :code:`pyhf cls` to specify a tensor backen
 The default backend is NumPy.
 For more information see :code:`pyhf cls --help`.
 
+Does ``pyhf`` support Python 2?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+No.
+Like the rest of the Python community, as of January 2020 ``pyhf`` no longer supports Python 2.
+The last release of ``pyhf`` that was compatible with Python 2.7 is `v0.3.2 <https://pypi.org/project/pyhf/0.3.2/>`__.
+
+
 Troubleshooting
 ---------------
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -18,7 +18,7 @@ Does ``pyhf`` support Python 2?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 No.
 Like the rest of the Python community, as of January 2020 ``pyhf`` no longer supports Python 2.
-The last release of ``pyhf`` that was compatible with Python 2.7 is `v0.3.2 <https://pypi.org/project/pyhf/0.3.2/>`__.
+The last release of ``pyhf`` that was compatible with Python 2.7 is `v0.3.3 <https://pypi.org/project/pyhf/0.3.3/>`__.
 
 I only have access to Python 2. How can I use ``pyhf``?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -20,6 +20,15 @@ No.
 Like the rest of the Python community, as of January 2020 ``pyhf`` no longer supports Python 2.
 The last release of ``pyhf`` that was compatible with Python 2.7 is `v0.3.2 <https://pypi.org/project/pyhf/0.3.2/>`__.
 
+I only have access to Python 2. How can I use ``pyhf``?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``pyhf`` should be able to be used as a standalone step in any analysis, and its environment need not be the same as the rest of the analysis.
+As Python 2 is not supported it is suggested that you setup a Python 3 runtime on whatever machine you're using.
+If you're using a cluster, talk with your system administrators to get their help in doing so.
+If you are unable to get a Python 3 runtime, versioned Docker images of ``pyhf`` are distributed through `Docker Hub <https://hub.docker.com/r/pyhf/pyhf>`__.
+
+Once you have Python 3 installed, see the `installation <installation.rst>`__ page to get started.
 
 Troubleshooting
 ---------------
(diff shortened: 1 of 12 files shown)


matthewfeickert:
@lukasheinrich @kratsg I'm opening this PR now in draft mode to give us a full ~~10 days~~  days (more like with Christmas and travel in there) to address anything before I want this to go in. :snake:

matthewfeickert:
@lukasheinrich @kratsg I've requested your review of this, but I'm going to leave this PR as a draft until we cut a final release (maybe December 27th?) that still supports Python 2.7.

matthewfeickert:
@jpivarski and @henryiii have both suggested that it would be better to make dropping Python 2 support be a minor patch release (v0.4.0) even though we aren't changing the API. They mentioned that this both makes a clear signal that there has been a drop (`v0.4` onward is Python 3 only, `v0.3` and below also has Python 2), and that if there ever was a reason that we needed to go back and patch something for Python 2 users (which I personally am against doing) then we could have room in the `v0.3` releases to do this.

matthewfeickert:
@kratsg @lukasheinrich Can I get you both to re-review `docs/faq.rst` as well as prompting this PR from a patch release to a minor release?

lukasheinrich:
that seems reasonable to me

docs/faq.rst
```diff
@@ -14,6 +14,22 @@ Use the :code:`--backend` option for :code:`pyhf cls` to specify a tensor backen
 The default backend is NumPy.
 For more information see :code:`pyhf cls --help`.
 
+Does ``pyhf`` support Python 2?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+No.
+Like the rest of the Python community, as of January 2020 ``pyhf`` no longer supports Python 2.
```
kratsg: change to something like `pip install pyhf~=0.3`

docs/faq.rst
```diff
@@ -14,6 +14,22 @@ Use the :code:`--backend` option for :code:`pyhf cls` to specify a tensor backen
...
+The last release of ``pyhf`` that was compatible with Python 2.7 is `v0.3.4 <https://pypi.org/project/pyhf/0.3.4/>`__.
+
+I only have access to Python 2. How can I use ``pyhf``?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``pyhf`` should be able to be used as a standalone step in any analysis, and its environment need not be the same as the rest of the analysis.
```
kratsg:
```suggestion
It is recommended that ``pyhf`` is used as a standalone step in any analysis, and its environment need not be the same as the rest of the analysis.
```

docs/faq.rst
```diff
@@ -14,6 +14,22 @@ Use the :code:`--backend` option for :code:`pyhf cls` to specify a tensor backen
...
+As Python 2 is not supported it is suggested that you setup a Python 3 runtime on whatever machine you're using.
+If you're using a cluster, talk with your system administrators to get their help in doing so.
+If you are unable to get a Python 3 runtime, versioned Docker images of ``pyhf`` are distributed through `Docker Hub <https://hub.docker.com/r/pyhf/pyhf>`__.
+
+Once you have Python 3 installed, see the `installation <installation.rst>`__ page to get started.
```
kratsg: use a meaningful ref here instead of relying on the `.rst`. I'm not sure we catch invalid links as warnings when they're explicitly written like this.

pyproject.toml
```diff
@@ -5,9 +5,8 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 88
-py36 = false  # Python 2 is still supported
+target-version = ['py36', 'py37', 'py38']
```
kratsg: target-version here includes py38, even though we don't support that in testing...

matthewfeickert: This if for Black though:

```
  -t, --target-version [py27|py33|py34|py35|py36|py37|py38]
                                  Python versions that should be supported by
                                  Black's output. [default: per-file auto-
                                  detection]
```

so this isn't affecting the testing.

docs/faq.rst
```diff
@@ -14,6 +14,26 @@ Use the :code:`--backend` option for :code:`pyhf cls` to specify a tensor backen
...
+
+I only have access to Python 2. How can I use ``pyhf``?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is recommended that ``pyhf`` is used as a standalone step in any analysis, and its environment need not be the same as the rest of the analysis.
+As Python 2 is not supported it is suggested that you setup a Python 3 runtime on whatever machine you're using.
+If you're using a cluster, talk with your system administrators to get their help in doing so.
+If you are unable to get a Python 3 runtime, versioned Docker images of ``pyhf`` are distributed through `Docker Hub <https://hub.docker.com/r/pyhf/pyhf>`__.
+
+Once you have Python 3 installed, see the `installation <installation.html>`__ page to get started.
```
kratsg: this won't raise an error. An explicit anchor is better (via labels): http://udig.refractions.net/files/docs/latest/user/docguide/sphinxSyntax.html#page-labels

PR #707, opened 2019-12-21
docs: Apply pydocstyle to Workspace

lukasheinrich:
# Description

Makes `workspace.py` pass `pydocstyle`.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add docstrings to workspace.py that pass pydocstyle
* Add workspace.py to pydocstyle tests in CI
```


src/pyhf/workspace.py
```diff
@@ -31,20 +37,25 @@ def __init__(self, spec, **config_kwargs):
...
 
     def __repr__(self):
+        """Representation of the Workspace."""
         return object.__repr__(self)
 
     # NB: this is a wrapper function to validate the returned measurement object against the spec
     def get_measurement(self, **config_kwargs):
         """
-        Get (or create) a measurement object using the following logic:
+        Get (or create) a measurement objec.
```
matthewfeickert: Small typo
```suggestion
        Get (or create) a measurement object.
```

src/pyhf/workspace.py
```diff
@@ -61,15 +72,14 @@ def get_measurement(self, **config_kwargs):
...
         """
         m = self._get_measurement(**config_kwargs)
         utils.validate(m, 'measurement.json', self.version)
         return m
 
     def _get_measurement(self, **config_kwargs):
-        """
-        See `Workspace::get_measurement`.
-        """
+        """See `Workspace::get_measurement`."""
```
matthewfeickert: Can we use the `~` syntax to link this?

src/pyhf/workspace.py
```diff
@@ -10,11 +17,10 @@
 
 
 class Workspace(_ChannelSummaryMixin, dict):
-    """
-    A JSON-serializable object that is built from an object that follows the `workspace.json` schema.
-    """
+    """A JSON-serializable object that is built from an object that follows the `workspace.json` schema."""
```
matthewfeickert: Can we use the ~ syntax to link this?

src/pyhf/workspace.py
```diff
@@ -1,3 +1,10 @@
+"""
+pyhf workspaces hold the three data items.
```
matthewfeickert:
```suggestion
pyhf workspaces hold the three data items:
```

src/pyhf/workspace.py
matthewfeickert: Nevermind. Since the method starts with an underscore this will never show up in the docs.

src/pyhf/workspace.py
matthewfeickert: No, we can't. @kratsg Can you think of a better way to link this then the following?

```diff
diff --git a/src/pyhf/workspace.py b/src/pyhf/workspace.py
index ece88a9..34193f9 100644
--- a/src/pyhf/workspace.py
+++ b/src/pyhf/workspace.py
@@ -17,7 +17,11 @@ log = logging.getLogger(__name__)
 
 
 class Workspace(_ChannelSummaryMixin, dict):
-    """A JSON-serializable object that is built from an object that follows the `workspace.json` schema."""
+    """
+    A JSON-serializable object that is built from an object that follows the :obj:`workspace.json` schema_.
+
+    .. _schema: https://scikit-hep.org/pyhf/likelihood.html#workspace
+    """
 
     def __init__(self, spec, **config_kwargs):
         """Workspaces hold the model, data and measurements."""
```

matthewfeickert:
@lukasheinrich @kratsg How do you feel about making this PR the patch release `v0.3.2` that would be the last PyPI release to support Python 2.7?

lukasheinrich:
I'm fine with that @matthewfeickert

src/pyhf/workspace.py
matthewfeickert: Though this does introduce the following warning when running `make html` for the docs:

```
/home/feickert/Code/GitHub/pyhf/docs/api.rst:47:<autosummary>:1: WARNING: Unknown target name: "schema".
```

So that should get corrected.

matthewfeickert: Fixed by 572102db1353a7e09f66485907e50a41d7d6bf17

PR #709, opened 2019-12-27
feat: calculator API

lukasheinrich:
# Description

Introduces the "calculator API" in order to allow to run inference in various modes (toybased, asymptotics, etc).

This might be quite a big PR, we can consider splitting it up into various small PRs and rebasing along the way.

Demo notebook:

https://github.com/scikit-hep/pyhf/blob/calculator_api/docs/examples/notebooks/UsingCalculators.ipynb

Sub-changes:
* change `qmu` to return scalar instead of size-1-tensor #712 
* rearrange asymptotic CLs calculation #745 
* introduce AsymptoticCalculator #750 
* introduce ToyCalculator #790 

# Checklist Before Requesting Reviewer

- [ ] Tests are passing
- [ ] "WIP" removed from the title of the pull request
- [ ] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [ ] Summarize commit messages into a comprehensive review of the PR

PR #712, opened 2019-12-29
feat: change qmu API and remove test_stat return

lukasheinrich:
# Description

subchanges of #709 

* make `qmu` return a scalar
* remove API to return test statistics from hypothesis tests (which only makes sense for asymptotics)
* **changes public API** 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Make 'pyhf.infer.test_statistics.qmu' return a scalar
* Remove API to return test statistics from hypothesis tests (which only makes sense for asymptotics)
```


kratsg:
ok with these changes too

PR #721, opened 2020-01-03
docs: Add copy buttons to code snippets in docs with sphinx-copybutton

matthewfeickert:
# Description

Add [`sphinx-copybutton`](https://sphinx-copybutton.readthedocs.io/en/latest/) to the `docs` extra to add copy buttons to all code snippet blocks.

Thanks [for tweeting](https://twitter.com/choldgraf/status/1213167082176577542?s=11) and for the great work, @choldgraf!

This makes all of our docstrings even nicer now as the copy button removes the Python REPL prompts:

[cropped_view](https://scikit-hep.org/pyhf/_generated/pyhf.tensor.pytorch_backend.pytorch_backend.html#pyhf.tensor.pytorch_backend.pytorch_backend.clip)

so that when pasted you get
```python
import pyhf
pyhf.set_backend("pytorch")
a = pyhf.tensorlib.astensor([-2, -1, 0, 1, 2])
pyhf.tensorlib.clip(a, -1, 1)
```

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add sphinx-copybutton to docs extra
* Enable sphinx-copybutton in docs to add copy buttons to code snippets
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -53,6 +53,7 @@ extensions = [
     'nbsphinx',
     'm2r',
     'sphinx_issues',
+    'sphinx_copybutton',
     'xref',
 ]
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -290,6 +290,9 @@ html_extra_path = ['_extras']
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'pyhfdoc'
 
+# sphinx-copybutton configuration
+copybutton_prompt_text = ">>> "
+
 # -- Options for LaTeX output ---------------------------------------------
 
 latex_elements = {
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ extras_require['docs'] = sorted(
             'nbsphinx',
             'ipywidgets',
             'sphinx-issues',
-            'sphinx-copybutton',
+            'sphinx-copybutton>0.2.8',
             'm2r',
         ]
     )
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ extras_require['docs'] = sorted(
             'nbsphinx',
             'ipywidgets',
             'sphinx-issues',
+            'sphinx-copybutton',
             'm2r',
         ]
     )
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ extras_require['docs'] = sorted(
             'nbsphinx',
             'ipywidgets',
             'sphinx-issues',
-            'sphinx-copybutton>0.2.8',
+            'sphinx-copybutton>0.2.9',
             'm2r',
         ]
     )


lukasheinrich:
does this also copy the output of the command?  e.g. `tensor([-1., -1.,  0.,  1.,  1.])` if it's possible to just copy input data that'd be preferable

choldgraf:
Current sphinx-copybutton just copies whatever is inside the code cell (minus stripping out some characters like prompts). If there's a way to know what part of the code cell should be skipped entirely (e.g. because it's an output), I'd be open to a PR in sphinx-copybutton that implements it! Here's the JS function that does the text copying:

https://github.com/choldgraf/sphinx-copybutton/blob/master/sphinx_copybutton/_static/copybutton.js#L65

lukasheinrich:
at least for python shouldn't it be safe to say everything that does not start with `>>>` (the prompt)  or `...` (the continuation of a prompt) is an output?

choldgraf:
@lukasheinrich it would be if all projects used `>>>` for their inputs, but a lot of them don't :-/

kratsg:
> @lukasheinrich it would be if all projects used `>>>` for their inputs, but a lot of them don't :-/
> 
> A lot of projects just dump raw code in there (so no prompts at all)

@choldgraf , a simple way is to add a configuration to allow for the last line to be copied or skipped.  This should handle both cases. Additionally, could add a configuration indicating if the project wants to use `>>>` and `...` so then you can just not copy lines that don't start with that.

matthewfeickert:
> a simple way is to add a configuration to allow for the last line to be copied or skipped. This should handle both cases. Additionally, could add a configuration indicating if the project wants to use `>>>` and `...` so then you can just not copy lines that don't start with that.

Maybe we should make an issue over at [`sphinx-copybutton`](https://github.com/choldgraf/sphinx-copybutton/issues) and ask Chris to assign one of us to it.

choldgraf:
I think I'd be +1 on that idea @kratsg - we'd need to think through the potential use-cases. We'd need to consider whether there is a case that someone *wants* to copy the stuff after prompts, and *also* the stuff that comes after that. +1 on opening an issue to discuss there

matthewfeickert:
> Additionally, could add a configuration indicating if the project wants to use `>>>` and `...` so then you can just not copy lines that don't start with that.

Just checking up on this tonight, and as @choldgraf is :100:% awesome, he already did all this &mdash; thank you Chris! :tada: As of [`sphinx-copybutton` PR 58](https://github.com/choldgraf/sphinx-copybutton/pull/58) (which translated to [`sphinx-copybutton` `v0.2.9` on PyPI](https://pypi.org/project/sphinx-copybutton/0.2.9/))

> [By default, if sphinx-copybutton detects lines that begin with code prompts, it will only copy the text in those lines (after stripping the prompts). This assumes that the rest of the code block contains outputs that shouldn’t be copied.](https://sphinx-copybutton.readthedocs.io/en/latest/#configure-whether-only-lines-with-prompts-are-copied)

so the example for `clip`

[cropped_view](https://scikit-hep.org/pyhf/_generated/pyhf.tensor.pytorch_backend.pytorch_backend.html#pyhf.tensor.pytorch_backend.pytorch_backend.clip)

now gives when pasted
```python
import pyhf
pyhf.set_backend("pytorch")
a = pyhf.tensorlib.astensor([-2, -1, 0, 1, 2])
pyhf.tensorlib.clip(a, -1, 1)
```

Issue #726, opened 2020-01-12
Provide simple API to get pre/post fit distributions with errors


ntadej:
The usual HEP workflow is to create a workspace with several channels and several systematics. In the final paper we need to show the post-fit distributions with appropriate errors.

The request is to have a simple interface to get the channel distribution with errors, ideally just two arrays (one with yields and one with errors), supporting:
 - the whole background at once
 - individual background samples separately (errors might not be needed here)

Plotting the actual distributions is then certainly out of scope of pyhf.

alexander-held:
I believe this is a very important ingredient needed to allow people to fully move to `pyhf`. As @lukasheinrich mentions in #731, it's not so easy to calculate the post-fit uncertainty band and I've seen calculations starting to break especially with high parameter correlations. I think it would be useful to not only get the uncertainty of the total model prediction, but also per sample (e.g. needed for yield tables).

alexander-held:
Since GitHub does not seem to link discussions back to issues: #1414 is related to this request.

alexander-held:
`cabinetry` includes some functionality that might help with this, including the uncertainty calculation.

pre-fit version:
```python
model_prefit = cabinetry.model_utils.prediction(model)
_ = cabinetry.tabulate.yields(model_prefit, data)
```

post-fit:
```python
fit_results = cabinetry.fit.fit(model, data)
model_postfit = cabinetry.model_utils.prediction(model, fit_results=fit_results)
_ = cabinetry.tabulate.yields(model_postfit, data)
```

A more complete example is in [this notebook](https://github.com/cabinetry/cabinetry-tutorials/blob/7f1bb2a779c147e8c95e9a4eb1f61570f48efeea/HEPData_workspace.ipynb). See also the [`cabinetry.model_utils.prediction`](https://cabinetry.readthedocs.io/en/latest/api.html#cabinetry.model_utils.prediction) and [`cabinetry.tabulate`](https://cabinetry.readthedocs.io/en/latest/api.html#module-cabinetry.tabulate) documentation.

wiso:
interesting, but I would prefer not to use another tool to do such a simple thing. My main point was about documentation.

PR #731, opened 2020-01-13
feat: Add return_by_sample to expected_data API

lukasheinrich:
# Description

Addresses #726 partially as requested by  @ntadej . Returning uncertainties is not so trivial since we need to agree on a prescription of what that even means. ROOT has various modes, ones where "parameter vectors" are samples according to the uncertainties of the fit and an envelope is computed. Will leave this for later.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Promote 'expected_data' to public pdf._MainModel API
* Add 'return_by_sample' arg to 'expected_data' API
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -530,8 +530,7 @@ class _MainModel(object):
             batchfirst = tensorlib.swapaxes(newbysample, 0, 1)
             if self.batch_size is None:
                 return batchfirst[0]
-            else:
-                return batchfirst
+            return batchfirst
 
         newresults = tensorlib.sum(newbysample, axis=0)
         if self.batch_size is None:
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -527,7 +527,7 @@ class _MainModel(object):
 
         newbysample = tensorlib.product(allfac, axis=0)
         if return_by_sample:
-            batchfirst = tensorlib.swapaxes(newbysample, 0, 1)
+            batchfirst = tensorlib.einsum('ij...->ji...', newbysample)
             if self.batch_size is None:
                 return batchfirst[0]
             return batchfirst
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -453,7 +453,7 @@ class _MainModel(object):
         return True
 
     def make_pdf(self, pars):
-        lambdas_data = self._expected_data(pars)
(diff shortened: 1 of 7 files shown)


src/pyhf/pdf.py
```diff
@@ -383,6 +383,12 @@ def _expected_data(self, pars):
         allfac = tensorlib.concatenate(factors + [nom_plus_delta])
 
         newbysample = tensorlib.product(allfac, axis=0)
+        if return_by_sample:
+            batchfirst = tensorlib.swapaxes(newbysample, 0, 1)
```
kratsg: what does this do

lukasheinrich: the array dimensions are (sample,batch,bins), while what we want to return is (batch,sample,bins) and in case of batch=None (sample,bins)

lukasheinrich:
I would assume `np.swapaxes` is a zero-copy operation also swapping random axes seems clunky in `einsum` (i..e `axyzbmnop -> bxyzamnop`)

src/pyhf/pdf.py
kratsg: that's why i'm confused. i thought `batch` was always first dimension...

kratsg:
> I would assume `np.swapaxes` is a zero-copy operation also swapping random axes seems clunky in `einsum` (i..e `axyzbmnop -> bxyzamnop`)

https://github.com/numpy/numpy/blob/master/numpy/core/tests/test_einsum.py#L228-L230

ehh... einsum means a more robust (slimmer) API...

lukasheinrich:
you're right :)

src/pyhf/pdf.py
```diff
@@ -526,6 +526,12 @@ def _expected_data(self, pars):
         allfac = tensorlib.concatenate(factors + [nom_plus_delta])
 
         newbysample = tensorlib.product(allfac, axis=0)
+        if return_by_sample:
+            batchfirst = tensorlib.einsum('ij...->ji...', newbysample)
+            if self.batch_size is None:
+                return batchfirst[0]
+            return batchfirst
```
matthewfeickert, line 533: Codecov is reporting this is uncovered. https://codecov.io/gh/scikit-hep/pyhf/pull/731/diff Can we fix this?

kratsg: should be covered

paulgessinger:
🎉

PR #734, opened 2020-01-13
feat: Improve JAX caching

lukasheinrich:
# Description

np.asarray(jax_array) and jax.np.asarray(np_array) seem to be particularly expensive operations (compared to the same in e.g. torch)

This introduces a patch to cache the creation of _Tensorviewer (the same index combinations are used multiple times) in the JAX optimizer

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/optimize/opt_jax.py
+++ b/src/pyhf/optimize/opt_jax.py
@@ -75,6 +75,6 @@ class jax_optimizer(AutoDiffOptimizerMixin):
                 tuple(fixed_idx),
                 tuple(variable_idx),
             )
-            return onp.asarray(a), onp.asarray(b)
+            return onp.asarray(objective), onp.asarray(grad)
 
         return tv, fixed_values_tensor, func, variable_init, variable_bounds
--- a/src/pyhf/optimize/opt_jax.py
+++ b/src/pyhf/optimize/opt_jax.py
@@ -4,6 +4,7 @@ from .. import get_backend, default_backend
 from ..tensor.common import _TensorViewer
 from .autodiff import AutoDiffOptimizerMixin
 import jax
+import numpy as onp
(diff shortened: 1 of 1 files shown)


src/pyhf/optimize/opt_jax.py
```diff
@@ -50,14 +54,19 @@ def setup_minimize(
...
+            ] = _TensorViewer([fixed_idx, variable_idx])
+            tv = self.tv_cache[tuple(fixed_idx)][tuple(variable_idx)]
 
         data = tensorlib.astensor(data)
         fixed_values_tensor = tensorlib.astensor(fixed_values, dtype='float')
 
         def func(pars):
             # need to conver to tuple to make args hashable
-            return _jitted_objective_and_grad(
+            a, b = _jitted_objective_and_grad(
```
matthewfeickert:
```suggestion
            objective, grad = _jitted_objective_and_grad(
```

src/pyhf/optimize/opt_jax.py
```diff
@@ -66,5 +75,6 @@ def func(pars):
                 tuple(fixed_idx),
                 tuple(variable_idx),
             )
+            return onp.asarray(a), onp.asarray(b)
```
matthewfeickert:
```suggestion
            return onp.asarray(objective), onp.asarray(grad)
```

src/pyhf/optimize/opt_jax.py
matthewfeickert: More explicit variable names would be nice.

matthewfeickert:
@lukasheinrich What did this get closed in favor of (if anything)? Also, do we still need this branch or can we delete it?

PR #737, opened 2020-01-14
chore: Run pyflakes over the full codebase

lukasheinrich:
# Description

this runs pyflakes on the full source tree including docs.

@kratsg that variable seems to never have been used. What was this for originally?

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Run pyflakes over the full codebase, including docs
```


kratsg:
> @kratsg that variable seems to never have been used. What was this for originally?

`xref.py` comes from https://github.com/michaeljones/sphinx-xref/blob/master/xref.py -- I didn't write that code. But for reasons, we have to copy it into our codebase for now.

Issue #735, opened 2020-01-13
Update the tensor backend retriever code


kratsg:
# Description

Now that we've dropped python 2, we can go ahead and use some things added to python 3 to simplify some of the existing code for the backend retriever by using an actual module instead of a wrapper class and messing w/ imports.

matthewfeickert:
@kratsg Can you elaborate more on what we originally had planned here?

kratsg:
module level imports can exist without having a custom backend retriever. That's all.

kratsg:
https://scientific-python.org/specs/spec-0001/

PR #741, opened 2020-01-17
fix: better optimizers

lukasheinrich:
# Description

superseded #734 

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Improve caching for JAX optimizer
* Add learning rates for autodiff optimizers (defaults to 1.0)
```

### DIFF
--- a/src/pyhf/optimize/__init__.py
+++ b/src/pyhf/optimize/__init__.py
@@ -12,6 +12,13 @@ class _OptimizerRetriever(object):
             # for autocomplete and dir() calls
             self.scipy_optimizer = scipy_optimizer
             return scipy_optimizer
+        elif name == 'autodiff':
+            from . import autodiff
+
+            assert autodiff
+            # for autocomplete and dir() calls
+            self.autodiff = autodiff
+            return autodiff
         elif name == 'jax_optimizer':
             try:
                 from .opt_jax import jax_optimizer
--- a/src/pyhf/optimize/autodiff.py
+++ b/src/pyhf/optimize/autodiff.py
@@ -34,13 +34,7 @@ class AutoDiffOptimizerMixin(object):
         tv, fixed_values_tensor, func, init, bounds = self.setup_minimize(
             objective, data, pdf, init_pars, par_bounds, fixed_vals
         )
-        result = minimize(
-            func,
-            init,
-            method='SLSQP',
-            bounds=bounds,
-            jac=True,
-        )
+        result = minimize(func, init, method='SLSQP', bounds=bounds, jac=True,)
         try:
             assert result.success
         except AssertionError:
--- a/src/pyhf/optimize/autodiff.py
+++ b/src/pyhf/optimize/autodiff.py
@@ -1,5 +1,5 @@
 """Helper Classes for use of automatic differentiation."""
-from scipy.optimize import minimize
+import scipy.optimize
 from .. import get_backend
 import logging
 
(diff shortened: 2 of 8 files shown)


lukasheinrich:
thanks @kratsg  <3

src/pyhf/optimize/__init__.py
```diff
@@ -12,6 +12,13 @@ def __getattr__(self, name):
             # for autocomplete and dir() calls
             self.scipy_optimizer = scipy_optimizer
             return scipy_optimizer
+        elif name == 'autodiff':
```
lukasheinrich: the `autodiff` is just a mixin not a standalone optimimzer so prob shouldn't appear here @kratsg

kratsg: it does need to be there for now.. until we update to python3 and fix the imports.

src/pyhf/optimize/autodiff.py
```diff
@@ -1,13 +1,17 @@
 """Helper Classes for use of automatic differentiation."""
-import scipy
+from scipy.optimize import minimize
```
kratsg: I don't like this. I would prefer to see `scipy.optimize.minimize` because I get confused with the existing `minimize()` defined in this file as well.

src/pyhf/optimize/__init__.py
kratsg: #735 will address this.

src/pyhf/optimize/opt_tflow.py
```diff
@@ -8,6 +8,11 @@
 class tflow_optimizer(AutoDiffOptimizerMixin):
     """Tensorflow Optimizer Backend."""
 
+    def __init__(self, *args, **kwargs):
+        """Tensorflow based optimizer based on autograd."""
+        self.lr = kwargs.get('lr', 1.0)
```
kratsg: what does lr stand for?

lukasheinrich: learning late

matthewfeickert: `lr` seems to be the standard for learning rate across TensorFlow and PyTorch too.

tests/test_optim.py
```diff
@@ -151,3 +152,20 @@ def test_optim_uncerts(backend, source, spec, mu):
...
+@pytest.mark.skip_numpy_minuit
+def test_autodiff_minimize_failure(backend, source, spec):
+    pdf = pyhf.Model(spec)
+    data = source['bindata']['data'] + pdf.config.auxdata
+
+    init_pars = pdf.config.suggested_init()
+    par_bounds = pdf.config.suggested_bounds()
+
+    optim = pyhf.optimizer
+    with mock.patch('pyhf.optimize.autodiff.minimize') as mocked_minimize:
```
matthewfeickert: This is currently failing (https://github.com/scikit-hep/pyhf/runs/408386678?check_suite_focus=true#step:8:70) as

```pytb
        if not self.create and original is DEFAULT:
            raise AttributeError(
>               "{} does not have the attribute {!r}".format(target, name)
            )
E           AttributeError: <module 'pyhf.optimize.autodiff' from '/home/runner/work/pyhf/pyhf/src/pyhf/optimize/autodiff.py'> does not have the attribute 'minimize'
```

kratsg:
Closing in favor of #986 and #987.

PR #742, opened 2020-01-18
[WIP] differentiable model construction

lukasheinrich:
# Description

The JSON structure from which the model is constructed is essentially a collection of array data. If literally read from JSON, it is static data, however  that array data might also be the output of other programs and one would like to have a fully differentiable chain of operations *throughout* the JSON-like structure  (@phinate is looking into this more closely)

This PR makes especially `_create_nominal_and_modifiers` differentiable. Perhaps for now this might not be merged in directly, but will just provide @phinate with a functioning fork until we figure out how to do this properly / more cleanly.

PR #744, opened 2020-01-21
refactor: standalone spec parsing for parameters

lukasheinrich:
# Description

this simplifies the `_ModelConfig` `__init__` method and encapsulates the code that needs the JSON structure (somewhat related to #742 ) so that it could be replaced maybe by something non-JSON-y

This is a simple refactor.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* refactor spec usage into standalone parsing function
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -76,9 +76,16 @@ def _paramset_requirements_from_spec(spec, channel_nbins):
             )
         _paramsets_user_configs[parameter.pop('name')] = parameter
 
-    return reduce_paramsets_requirements(
+    _reqs = reduce_paramsets_requirements(
         _paramsets_requirements, _paramsets_user_configs
     )
+
+    _sets = {}
+    for param_name, paramset_requirements in _reqs.items():
+        paramset_type = paramset_requirements.get('paramset_type')
+        paramset = paramset_type(**paramset_requirements)
+        _sets[param_name] = paramset
+    return _sets
 
 
 class _ModelConfig(_ChannelSummaryMixin):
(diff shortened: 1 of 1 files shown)


matthewfeickert:
@lukasheinrich You had mentioned you were confused about the Codecov report. What parts in particular? The diff shows that all the unhit lines are the `raise RuntimeError` lines.

lukasheinrich:
I'm confused because in the github comment it says

```
Merging #744 into master will increase coverage by <.01%.
```

so I don't understand why it's complaining

matthewfeickert:
> I'm confused because in the github comment it says
> 
> ```
> Merging #744 into master will increase coverage by <.01%.
> ```
> 
> so I don't understand why it's complaining

Yes, but that's the total coverage. The part that is complaining is the **patch** coverage, which is low:

patch_coverage

lukasheinrich:
done

src/pyhf/pdf.py
```diff
@@ -17,90 +17,217 @@
...
+    _reqs = reduce_paramsets_requirements(
+        _paramsets_requirements, _paramsets_user_configs
+    )
+
+    _sets = {}
+    for param_name, paramset_requirements in _reqs.items():
+        paramset_type = paramset_requirements.get('paramset_type')
+        paramset = paramset_type(**paramset_requirements)
+        _sets[param_name] = paramset
+    return _sets
```
kratsg: These lines are specific to a given Model. If we care about efficiency and running things once and passing around that object, then we want to factor this out into a separate function called by the Model, with this existing function called by the Workspace, perhaps renaming this as `_paramset_requirements_from_channels`.

PR #748, opened 2020-01-24
feat: Add GPU enabled Dockerfile

### LINKED ISSUE #746: Add additional Dockerfiles
# Question

I've gone ahead and put the GPU enabled Docker images for the machine learning backends up on the Docker hub page (c.f. [tags with `cuda10.1-cudnn7-runtime-ubuntu18.04` in them](https://hub.docker.com/r/pyhf/pyhf/tags?page=1&name=cuda10.1-cudnn7-runtime-ubuntu18.04)). However, as can be seen, since they are based off of the (already quite large) NVIDIA CUDA images they are not even close to optimized for space (though I've tried to do my best at the moment with using multi-stage Docker builds combined with some smarter copying).

@lukasheinrich @kratsg I have all the Dockerfiles used to make these in a branch at the moment. Should I open a PR to add them in even though they images are unsatisfactorily large? This is my inclination but I thought I'd get your thoughts before asking you to review a PR full of Dockerfiles.

matthewfeickert:
# Description

Resolves #746 

Add [NVIDIA GPU enabled](https://github.com/NVIDIA/nvidia-docker) Dockerfile for the machine learning backends up on the Docker hub page (c.f. [tags with `cuda10.1-cudnn7-runtime-ubuntu18.04` in them](https://hub.docker.com/r/pyhf/pyhf/tags?page=1&name=cuda10.1-cudnn7-runtime-ubuntu18.04)).

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add NVIDIA GPU enabled Dockerfile for the machine learning framework backends
* Add .dockerignore to reduce build context size
```


matthewfeickert:
There is still a CodeFactor warning about hadolint warning

> [Avoid additional packages by specifying `--no-install-recommends` (DL3015)](https://github.com/hadolint/hadolint/wiki/DL3015)

but I will have to check that that does not affect the builds.

lukasheinrich:
how hard would it be to use build args (```ARG MLBACKEND```) to use just a single but parametrized dockerfile ```docker build --buildarg MLBACKEND=jax```

matthewfeickert:
> how hard would it be to use build args (`ARG MLBACKEND`) to use just a single but parametrized dockerfile `docker build --buildarg MLBACKEND=jax`

For TensorFlow and PyTorch this is super easy to do. The only problem is that JAX requires additional setup to get the GPU enabled wheel, and my memory is that trying to do `if` statements in Dockerfiles is frowned upon. So I think to get that sort of functionality we'd have to have all the pip installing be done by a Shell script that is called. Want me to try that out?

kratsg:
> For TensorFlow and PyTorch this is super easy to do. The only problem is that JAX requires additional setup to get the GPU enabled wheel, and my memory is that trying to do `if` statements in Dockerfiles is frowned upon. So I think to get that sort of functionality we'd have to have all the pip installing be done by a Shell script that is called. Want me to try that out?

Just use if/else in bash.

matthewfeickert:
> Just use if/else in bash.

Well, yeah. What I meant by "Do you want me to try that out?" is "Are we willing to accept a Bash script as another dependency of the Docker images?". My goal is generally to try and have Dockerfiles be as explicit as possible about what is happening, and disappearing into scripts can be frustrating for readers. That being said, as you point out there is no magic happening here.

matthewfeickert:
The following now works:

```
docker build . \
	--file docker/jax/Dockerfile \
	--build-arg BACKEND=jax \
	--tag pyhf/pyhf:jax \
	--tag pyhf/pyhf:jax-cuda10.1-cudnn7-runtime-ubuntu18.04 \
	--compress
```

docker/gpu/install_backend.sh
```diff
@@ -0,0 +1,40 @@
...
+  fi
+}
+
+function main() {
+  # 1: the backend option name in setup.py
+  local BACKEND="${1}"
+  install_backend "${BACKEND}"
+}
+
+main "$@" || exit 1
```
lukasheinrich: what a nice shell script  oO passes https://www.shellcheck.net/

matthewfeickert: I love ShellCheck! I'd be lying if I said that this passed on the first time writing it, but I run ShellCheck over all my shell scripts now for the past year or so. :) I think it is really nice that in addition to [Hadolint](https://github.com/hadolint/hadolint) CodeFactor also runs ShellCheck as well.

matthewfeickert:
@kratsg As this just adds the Dockerfile to make them easy to build if we want them to do you have any problems with having this PR go in? These won't become the default images.

PR #752, opened 2020-01-26
feat: Combine is taught about formal data (outer) join methods

### LINKED ISSUE #695: Combined workspaces have duplicated lumi/POI param configurations
# Description

When combining two workspaces with the same measurement name that has the same POI, the lumi and the POI param config are repeated twice. They should usually be compatible so we should just drop the extra copies.

# Expected Behavior

A single instance of the lumi and POI parameter configuration.

# Actual Behavior

Duplicated/extra instances of the lumi and POI parameter configuratoin.

# Steps to Reproduce

`pyhf combine a.json b.json`

# Checklist

### LINKED ISSUE #749: parameters overrides in  are not combined in "pyhf combine"
# Description

In a closure test for `pyhf combine` I noticed that parameter section overrides in the workspace json can lead to a crash

here's a 2 bin - 1 channel json

https://gist.github.com/lukasheinrich/e6a9c49780e06d617f763c4d5496a205

```
curl https://gist.githubusercontent.com/lukasheinrich/e6a9c49780e06d617f763c4d5496a205/raw/a737eb4a04f627daeaad7d6acd5e54183ed5994f/combine_test_target.json|pyhf cls
{
    "CLs_exp": [
        0.07377754112073907,
        0.16776812083252346,
        0.3508132196480346,
        0.6260263084704382,
        0.8769621351034118
    ],
    "CLs_obs": 0.60635945662
```

Here are two workspaces, each carries one of the bins of the above one 

https://gist.github.com/lukasheinrich/925437eb40ffbf19e8b8f180b0d67924
https://gist.github.com/lukasheinrich/ecad7225503616360a96a35ea77e87a2

```
curl https://gist.githubusercontent.com/lukasheinrich/925437eb40ffbf19e8b8f180b0d67924/raw/417737735d6358139a50437fb3fccaf1423fbd00/combine_test_one.json|pyhf cls
{
    "CLs_exp": [
        0.14521985430105866,
        0.2715371244860493,
        0.4739897985534657,
        0.7271619041517813,
        0.9213903347174244
    ],
    "CLs_obs": 0.6553036139942451
}
```

```
curl https://gist.githubusercontent.com/lukasheinrich/ecad7225503616360a96a35ea77e87a2/raw/e70c93ae2464ad0bbbe5d931284db4fa8ef66164/combine_test_two.json|pyhf cls
{
    "CLs_exp": [
        0.10991489771691786,
        0.2230264429815649,
        0.4196899637032293,
        0.685255927920688,
        0.9041426586275007
    ],
    "CLs_obs": 0.5878244654561694
}
```

In principle combination is easy, however the "lumi" sepc in the "parameter sections" of both worksapces trips up `pyhf combine`

```
pyhf combine \
<(curl https://gist.githubusercontent.com/lukasheinrich/925437eb40ffbf19e8b8f180b0d67924/raw/417737735d6358139a50437fb3fccaf1423fbd00/combine_test_one.json) \
<(curl https://gist.githubusercontent.com/lukasheinrich/ecad7225503616360a96a35ea77e87a2/raw/e70c93ae2464ad0bbbe5d931284db4fa8ef66164/combine_test_two.json) | \
pyhf cls
pyhf.exceptions.InvalidModel: Multiple parameter configurations for lumi were found.
```

kratsg:
# Description

This PR will resolve a nagging issue with @kratsg about dealing with combining workspaces. The hardest thing is that there is not necessarily one right way to combine (or join) two workspaces. Considering this, `pyhf combine` will now support (as of this PR) four methods.

1. none
2. outer
3. left outer
4. right outer

Before, `pyhf combine` was doing a mixture of (1) and (2) above: `none` and `outer`. Now, the code is cleaned up to separate the two logic better.

A `none` will only check for duplicated data structures (by name) between the two workspaces and refuse to combine if any names are in common. This uses the summary data provided by `_ChannelSummaryMixin`.

An `outer` join will join both workspaces and any duplicated data structures will be merged together if they are fully identical and not mutually incompatible. This is simply done by asking if the two underlying dictionary objects are equivalent.

A `left outer` (or `right outer`) will join both workspaces using the left (or right) workspace as preference in the case any conflicts arise during the procedure of an `outer` join.

Additionally, a lot of exceptions/error messages are clarified in the process.

This will fix #749 and fix #695.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Workspace combination now allows for more robust join operations with a variation of outer join supported
* Some refactoring of Workspace.combine was done to support a future clean-up
```

### DIFF
--- a/src/pyhf/workspace.py
+++ b/src/pyhf/workspace.py
@@ -395,6 +395,13 @@ class Workspace(_ChannelSummaryMixin, dict):
                 joined_items.append(right_item)
             return joined_items
 
+        def _join_versions(join, left_version, right_version):
+            if left_version != right_version:
+                raise exceptions.InvalidWorkspaceOperation(
+                    f"Workspaces of different versions cannot be combined: {left_version} != {right_version}"
+                )
+            return left_version
+
         def _join_channels(join, left_channels, right_channels):
             joined_channels = _join_items(join, left_channels, right_channels)
             if join == 'none':
@@ -419,6 +426,8 @@ class Workspace(_ChannelSummaryMixin, dict):
                     )
             return joined_channels
 
+        new_version = _join_versions(join, left['version'], right['version'])
+
         common_measurements = set(left.measurement_names).intersection(
             right.measurement_names
         )
@@ -435,12 +444,6 @@ class Workspace(_ChannelSummaryMixin, dict):
                         for m, i in zip(common_measurements, incompatible_poi)
                         if incompatible_poi
                     ]
-                )
-            )
-        if left.version != right.version:
-            raise exceptions.InvalidWorkspaceOperation(
-                "Workspaces of different versions cannot be combined: {} != {}".format(
-                    left.version, right.version
                 )
             )
 
@@ -509,6 +512,6 @@ class Workspace(_ChannelSummaryMixin, dict):
                 left_measurements + right_measurements + merged_measurements
             ),
             'observations': left['observations'] + right['observations'],
-            'version': left['version'],
+            'version': new_version,
         }
         return Workspace(newspec)
--- a/src/pyhf/workspace.py
+++ b/src/pyhf/workspace.py
@@ -377,6 +377,24 @@ class Workspace(_ChannelSummaryMixin, dict):
                 f"Workspaces must be joined using one of the valid join operations ({Workspace.valid_joins}): {join}"
             )
 
+        def _join_items(join, left_items, right_items):
+            if join == 'right outer':
+                left_items, right_items = right_items, left_items
+            joined_items = copy.deepcopy(left_items)
+            for right_item in right_items:
+                # outer join: merge left and right, matching where possible
+                if join == 'outer' and right_item in left_items:
+                    continue
+                # left(right) outer join: only add right(left) if existing item (by name) is not in left(right)
+                # NB: we switch left/right if using a right outer join so we only ever check the "right"
+                # NB: this will be slow for large numbers of items
+                elif join in ['left outer', 'right outer'] and right_item['name'] in [
+                    item['name'] for item in joined_items
+                ]:
+                    continue
+                joined_items.append(right_item)
+            return joined_items
+
         common_channels = set(left.channels).intersection(right.channels)
         if common_channels:
             raise exceptions.InvalidWorkspaceOperation(
@@ -419,28 +437,15 @@ class Workspace(_ChannelSummaryMixin, dict):
             for m in set(right.measurement_names) - set(common_measurements)
         ]
 
-        def _merge_parameter_configs(
+        def _join_parameter_configs(
             measurement_name, join, left_parameters, right_parameters
         ):
-            if join == 'right outer':
-                left_parameters, right_parameters = right_parameters, left_parameters
-
-            merged_parameter_configs = copy.deepcopy(left_parameters)
-            for right_parameter in right_parameters:
-                # outer join: merge left and right, matching where possible
-                if join == 'outer' and right_parameter in left_parameters:
-                    continue
-                # left(right) outer join: only add right(left) if existing parameter config's name is not in left(right)
-                # NB: we switch left/right if using a right outer join so we only ever check the "right"
-                # NB: this will be slow for large numbers of parameter configs
-                elif join in ['left outer', 'right outer'] and right_parameter[
-                    'name'
-                ] in [parameter['name'] for parameter in merged_parameter_configs]:
-                    continue
-                merged_parameter_configs.append(right_parameter)
+            joined_parameter_configs = _join_items(
+                join, left_parameters, right_parameters
+            )
             if join == 'outer':
                 counted_parameter_configs = collections.Counter(
-                    parameter['name'] for parameter in merged_parameter_configs
+                    parameter['name'] for parameter in joined_parameter_configs
                 )
                 incompatible_parameter_configs = [
                     parameter
@@ -451,14 +456,14 @@ class Workspace(_ChannelSummaryMixin, dict):
                     raise exceptions.InvalidWorkspaceOperation(
                         f"Workspaces cannot have a measurement ({measurement_name}) with incompatible parameter configs: {incompatible_parameter_configs}. You can also try a different join operation: {Workspace.valid_joins}."
                     )
-            return merged_parameter_configs
+            return joined_parameter_configs
 
         merged_measurements = [
             dict(
                 name=m,
                 config=dict(
                     poi=left.get_measurement(measurement_name=m)['config']['poi'],
-                    parameters=_merge_parameter_configs(
+                    parameters=_join_parameter_configs(
                         m,
                         join,
                         left.get_measurement(measurement_name=m)['config'][
--- a/src/pyhf/workspace.py
+++ b/src/pyhf/workspace.py
@@ -34,20 +34,21 @@ def _join_items(join, left_items, right_items, key='name'):
 
     """
     if join == 'right outer':
-        left_items, right_items = right_items, left_items
-    joined_items = copy.deepcopy(left_items)
-    for right_item in right_items:
-        # outer join: merge left and right, matching where possible
-        if join == 'outer' and right_item in left_items:
+        primary_items, secondary_items = right_items, left_items
+    else:
+        primary_items, secondary_items = left_items, right_items
+    joined_items = copy.deepcopy(primary_items)
+    for secondary_item in secondary_items:
+        # outer join: merge primary and secondary, matching where possible
+        if join == 'outer' and secondary_item in primary_items:
             continue
-        # left(right) outer join: only add right(left) if existing item (by name) is not in left(right)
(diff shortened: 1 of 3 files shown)


lukasheinrich:
thanks @kratsg. Can you give some details when you would need `none` and `left/right outer` shouldn't one always merge with `outer`?

kratsg:
> thanks @kratsg. Can you give some details when you would need `none` and `left/right outer` shouldn't one always merge with `outer`?

`none` is the default. If there's anything in common _at all_, it will complain. This is when you want to combine two workspaces that you think are completely different (and this will do checks like whether the channels have the same name or the measurements have the same name). It's a _very_ strict operation that does not attempt to merge workspaces and simply tries to concatenate or add.

`outer` is the join we've generally wanted, which will only merge channels of the same name if they have the same structure, or measurements, or observations, etc... but if the structure is different between left and right, then it will complain and error out... this leads us to

`left outer` and `right outer` which makes the statement of: "if there are any incompatibilities between things of the same name, please use the left workspace or the right workspace". 

So if you think about it in terms of joining ability/strictness checking:

- `left outer` and `right outer` will never fail to combine two workspaces.
- `outer` will sometimes fail to combine two workspaces if it runs into incompatible structures for items of the same name in the workspace.
- `none` will commonly fail to combine two workspaces if it runs into any items of the same name in the workspace.

kratsg:
Additionally, `none` is backwards compatible. Which is partly why I introduced it, since I feel like the existing functionality should be retained and enhanced with `outer`. We could certainly choose to have `outer` be the default.. [would need to update some tests to not rely on the default if we start changing it around]

lukasheinrich:
Yeah I can follow the logic, but was wondering whether we should just refuse to join any workspaces that are incompatible. After all a given workspace was designed with a certain parameter in mind, so overriding that would be strange

kratsg:
And as for why left/right joins? I ran into issues like this:

```
$ pyhf combine -j outer 3Loffshell/125_100/3Loffshell_winobino_125_100_divided4.json compressed/compressed_winobino_125_100_divided4.json > combined_divided4.json 
  File "/Users/kratsg/.virtualenvs/pyhf-dev/bin/pyhf", line 11, in <module>
    load_entry_point('pyhf', 'console_scripts', 'pyhf')()
  File "/Users/kratsg/.virtualenvs/pyhf-dev/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/kratsg/.virtualenvs/pyhf-dev/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/kratsg/.virtualenvs/pyhf-dev/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/kratsg/.virtualenvs/pyhf-dev/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/kratsg/.virtualenvs/pyhf-dev/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/kratsg/pyhf/src/pyhf/cli/spec.py", line 267, in combine
    combined_ws = Workspace.combine(ws_one, ws_two, join=join)
  File "/Users/kratsg/pyhf/src/pyhf/workspace.py", line 503, in combine
    for m in common_measurements
  File "/Users/kratsg/pyhf/src/pyhf/workspace.py", line 503, in <listcomp>
    for m in common_measurements
  File "/Users/kratsg/pyhf/src/pyhf/workspace.py", line 482, in _join_parameter_configs
    f"Workspaces cannot have a measurement ({measurement_name}) with incompatible parameter configs: {incompatible_parameter_configs}. You can also try a different join operation: {Workspace.valid_joins}."
pyhf.exceptions.InvalidWorkspaceOperation: Workspaces cannot have a measurement (NormalMeasurement) with incompatible parameter configs: ['lumi']. You can also try a different join operation: ['none', 'outer', 'left outer', 'right outer'].
```

and realized that even slight differences in the lumi between the two will break things:

```
{"bounds": [[0.915, 1.085]], "inits": [1.0], "auxdata": [1.0], "sigmas": [0.017], "name": "lumi"}
{"auxdata": [1.0], "bounds": [[0.915, 1.085]], "fixed": true, "inits": [1.0], "name": "lumi", "sigmas": [0.017]}
```

where in this case, one analysis had `fixed: true` and the other didn't.

kratsg:
> Yeah I can follow the logic, but was wondering whether we should just refuse to join any workspaces that are incompatible. After all a given workspace was designed with a certain parameter in mind, so overriding that would be strange

Yeah. I think on some level, we should allow the override to exist, but we shouldn't disallow it. People _are_ going to want to join workspaces that are incompatible to check something quickly -- and then go back and start figuring out why they're incompatible and fix it. E.G. in the case I put above, I just want to combine and get an idea for the fit and then go back and fix things.

lukasheinrich:
hm, yeah but maybe I would add like a `unsafe-outer-left` or `unsafe-outer-right`  or something that makes clear that this is only for temporary studies?

kratsg:
> hm, yeah but maybe I would add like a `unsafe-outer-left` or `unsafe-outer-right` or something that makes clear that this is only for temporary studies?

I can have it print a warning in those situations. Would that work? I prefer not to deviate from standard SQL terminology [which is where I borrowed it from].

lukasheinrich:
yeah I think a warning in case there are inconsistensies is fine, maybe with `--yes` to confirm? I"m fine either way.

src/pyhf/workspace.py
```diff
@@ -16,11 +18,262 @@
...
+        join (str): The join operation to apply. See ~pyhf.workspace.Workspace for valid join operations.
+        left_items (list): A list of dictionaries to join on the left
+        right_items (list): A list of dictionaries to join on the right
+
+    Returns:
+        :obj:`list`: A joined list of dictionaries.
+
+    """
+    if join == 'right outer':
+        left_items, right_items = right_items, left_items
```
lukasheinrich: maybe better to call one `primary` and the other `secondary` and assign `left` `right` according to the passed preference order?

src/pyhf/workspace.py
```diff
@@ -16,11 +18,263 @@
...
+        right_observations (list): The right observation specification.
+
+    Returns:
+        :obj:`list`: A joined list of observations. Each observation follows the :obj:`defs.json#observation` `schema <https://scikit-hep.org/pyhf/likelihood.html#observations>`__
+
+    """
+    joined_observations = _join_items(join, left_observations, right_observations)
+    if join == 'none':
+        common_observations = set(c['name'] for c in left_observations).intersection(
+            c['name'] for c in right_observations
```
matthewfeickert: I might be tempted to use `obs` instead of `c` here for clarity, but this is more personal preference.

src/pyhf/workspace.py
```diff
@@ -16,11 +18,263 @@
...
+
+    Returns:
+        :obj:`list`: A joined list of measurements. Each measurement follows the :obj:`defs.json#measurement` `schema <https://scikit-hep.org/pyhf/likelihood.html#measurements>`__
+
+    """
+    joined_measurements = _join_items(join, left_measurements, right_measurements)
+    if join == 'none':
+        common_measurements = set(c['name'] for c in left_measurements).intersection(
+            c['name'] for c in right_measurements
+        )
```
matthewfeickert: Similarly I might be tempted to use `meas` or `m` instead of `c`.

src/pyhf/workspace.py
```diff
@@ -362,74 +604,34 @@ def combine(cls, left, right):
         Args:
             left (~pyhf.workspace.Workspace): A workspace
             right (~pyhf.workspace.Workspace): Another workspace
+            join (:obj:`str`): How to join the two workspaces. Pick from "none", "outer", "left outer", or "right outer".
```
matthewfeickert: > Pick from "none", "outer", "left outer", or "right outer".

@kratsg Do you think that we should explain somewhere that this is using SQL syntax? Or link to some documentation on what this means? I can see that we might be able to say "yo, if you're joining then you should know what you're doing" but I'm thinking about the uninitiated grad student that has been told to go do something.

matthewfeickert:
@kratsg Also why is this a "fix"? :) Seems like this is a pretty nice "feat" to me!

src/pyhf/workspace.py
kratsg: dunno. i'm kinda pessimistic and go with "nah" but...

lukasheinrich: can't we just let this be called `prefer_left` and `prefer_right`?

kratsg: I mean, this is a very _common_ concept: https://en.wikipedia.org/wiki/Join_(SQL) -- so I'd rather stick with the definition of it. We can add an additional option to feed in "prefer left outer" and "prefer right outer" (because there are also inner joins) and those translate to left outer / right outer.

PR #766, opened 2020-02-08
fix: Uncorrelated shape systematic with proper masking

### LINKED ISSUE #404: Fix ShapeSys masking
# Description

https://github.com/diana-hep/pyhf/blob/49f30cd52bd543e67b1464d0ce916e15014a6f5b/pyhf/modifiers/shapesys.py#L106-L119

is a bit hacky right now by letting `==0` mean that the bin does not participate. Should be masking instead.

# Expected Behavior

Expect to do something more like

```
factors = nom_sq / unc_sq
```

and then mask `factors` after the fact.

# Actual Behavior

No masking is happening.

# Steps to Reproduce

See the code.

# Checklist

kratsg:
# Description

Resolves #404 (issue not found 😉). It's 11:30pm and I've stared about this piece of code in the shapesys for probably way too long and I've either gone crazy or I've discovered a new dimension. In either case, I hit a eureka moment (maybe?) or I'm just fooling myself...

I've somehow simplified the shapesys `finalize()` call which is what was causing the crash to happen. To make this work, a few things had to be realized:

1. We need to distinguish between shapesys modifiers which are subscribed to a sample (in a channel) but set their uncertainty to zero; and to shapesys modifiers which are not subscribed to a sample. This is done by using the provided shapesys mask.
1. We need to ensure the denominator (sigma**2) should be set to `1` by default in cases where the shapesys modifier does not affect that sample(in that channel). This is done by setting the default from 0 to -1.
1. We need to make sure we don't crash in cases where the shapesys uncertainty is zero. This is done by setting the denominator equal to `1` in these cases.
1. We need to make sure we don't crash in cases where the shapesys uncertainty is zero AND the nominal expected rate is also zero. This is done by setting the numerator to `1` in these cases.

As it turns out, we can not only speed up `shapesys` by making specific decisions, but we can be clever about our calculations to also reduce the amount of work done in `finalize()`!

Related: #774, #775.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Apply a mask to skip cases where given shapesys modifier doesn't affect any samples
* Remove RuntimeError check that is never reached
```

### DIFF
--- a/src/pyhf/modifiers/shapesys.py
+++ b/src/pyhf/modifiers/shapesys.py
@@ -128,11 +128,12 @@ class shapesys_combined(object):
             ]
 
             # if empty: a case where given shapesys modifier affects zero samples
-            if default_backend.shape(sample_uncert_info)[0] == 0: continue
+            if default_backend.shape(sample_uncert_info)[0] == 0:
+                continue
 
             # sample_uncert_info = (bin_mask, nominal rate, uncertainty)
             bin_mask = default_backend.astensor(sample_uncert_info[0][0], dtype='bool')
-            nom_unc = sample_uncert_info[0,1:]
+            nom_unc = sample_uncert_info[0, 1:]
 
             # TODO: understand why this works??? what the fuck though -- must
             # be something with setting the default (not affects sample == -1,
--- a/src/pyhf/modifiers/shapesys.py
+++ b/src/pyhf/modifiers/shapesys.py
@@ -125,11 +125,14 @@ class shapesys_combined(object):
             # shapesys is not shared, so there should only ever be at most one sample
             sample_uncert_info = mod_uncert_info[
                 default_backend.sum(mod_uncert_info[:, 0] > 0, axis=1, dtype='bool')
-            ][0]
+            ]
+
+            # if empty: a case where given shapesys modifier affects zero samples
+            if default_backend.shape(sample_uncert_info)[0] == 0: continue
 
             # sample_uncert_info = (bin_mask, nominal rate, uncertainty)
-            bin_mask = default_backend.astensor(sample_uncert_info[0], dtype='bool')
-            nom_unc = sample_uncert_info[1:]
+            bin_mask = default_backend.astensor(sample_uncert_info[0][0], dtype='bool')
+            nom_unc = sample_uncert_info[0,1:]
 
             # TODO: understand why this works??? what the fuck though -- must
             # be something with setting the default (not affects sample == -1,
--- a/src/pyhf/modifiers/shapesys.py
+++ b/src/pyhf/modifiers/shapesys.py
@@ -49,12 +49,13 @@ class shapesys_combined(object):
         self._shapesys_mask = [
             [[mega_mods[m][s]['data']['mask']] for s in pdfconfig.samples] for m in keys
         ]
-        self.__shapesys_uncrt = default_backend.astensor(
+        self.__shapesys_info = default_backend.astensor(
             [
                 [
                     [
-                        mega_mods[m][s]['data']['uncrt'],
+                        mega_mods[m][s]['data']['mask'],
                         mega_mods[m][s]['data']['nom_data'],
+                        mega_mods[m][s]['data']['uncrt'],
                     ]
                     for s in pdfconfig.samples
                 ]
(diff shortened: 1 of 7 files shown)


src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -156,7 +156,8 @@ def astensor(self, tensor_in, dtype='float'):
             tensor = tf.cast(tensor, dtype)
         return tensor
 
-    def sum(self, tensor_in, axis=None):
+    def sum(self, tensor_in, axis=None, dtype=None):
+        # TODO: implement dtype cast conversion??
```
kratsg: TODO

src/pyhf/modifiers/shapesys.py
```diff
@@ -90,28 +91,30 @@ def _precompute(self):
...
+            # shapesys is not shared, so there should only ever be at most one sample
+            sample_uncert_info = mod_uncert_info[
+                default_backend.sum(mod_uncert_info[:, 0] > 0, axis=1, dtype='bool')
+            ][0]
+
+            # sample_uncert_info = (bin_mask, nominal rate, uncertainty)
+            bin_mask = default_backend.astensor(sample_uncert_info[0], dtype='bool')
+            nom_unc = sample_uncert_info[1:]
+
+            # TODO: understand why this works??? what the fuck though -- must
```
kratsg: TODO

src/pyhf/modifiers/shapesys.py
```diff
@@ -115,30 +116,36 @@ def _precompute(self):
...
+            if default_backend.shape(sample_uncert_info)[0] == 0:
+                continue
+
+            # sample_uncert_info = (bin_mask, nominal rate, uncertainty)
+            bin_mask = default_backend.astensor(sample_uncert_info[0][0], dtype='bool')
+            nom_unc = sample_uncert_info[0, 1:]
+
+            # Why this works: setting the default/not-affecting-sample == -1
+            # means that squaring it gives +1 by default)
+
```
matthewfeickert:
```suggestion
            # means that squaring it gives +1 by default
            # (done in _nominal_and_modifiers_from_spec)
```

src/pyhf/modifiers/shapesys.py
```diff
@@ -115,30 +116,36 @@ def _precompute(self):
...
+            # identify the sample that the given parameter affects
+            # shapesys is not shared, so there should only ever be at most one sample
+            sample_uncert_info = mod_uncert_info[
+                default_backend.sum(mod_uncert_info[:, 0] > 0, axis=1, dtype='bool')
+            ]
+
+            # if empty: a case where given shapesys modifier affects zero samples
+            if default_backend.shape(sample_uncert_info)[0] == 0:
+                continue
+
```
lukasheinrich: this case shouold not happen anymore / raise an exception once we have 

```
if not pdfconfig.param_set(pname).n_parameters
``` 
?

src/pyhf/pdf.py
```diff
@@ -187,13 +187,17 @@ def _nominal_and_modifiers_from_spec(config, spec):
...
                     if mtype == 'shapesys':
-                        maskval = [(x > 0 and y > 0) for x, y in zip(uncrt, nom)]
+                        maskval = [(x > 0 and y > 0) for x, y in zip(uncrts, nom)]
                     else:
                         maskval = [True if thismod else False] * len(nom)
                     mega_mods[key][s]['data']['mask'] += maskval
-                    mega_mods[key][s]['data']['uncrt'] += uncrt
+                    mega_mods[key][s]['data']['uncrt'] += uncrts
                     mega_mods[key][s]['data']['nom_data'] += nom
                 else:
```
lukasheinrich: maybe we should split this up cleanly infoo `elif mtype == shapesys` and `elif mtype == staterror`

src/pyhf/modifiers/shapesys.py
```diff
@@ -115,30 +116,36 @@ def _precompute(self):
...
-                unc_sq > 0, unc_sq, default_backend.ones(unc_sq.shape)
-            )
-
-            factors = numerator / denominator
-            factors = factors[factors > 0]
+
+            # identify the sample that the given parameter affects
+            # shapesys is not shared, so there should only ever be at most one sample
+            sample_uncert_info = mod_uncert_info[
+                default_backend.astensor(
```
lukasheinrich: we already solve a similar problem here `singular_sample_index` maybe we justs save it as an attribute?

lukasheinrich: would allow us to avoid adding the `sum` extension to new dtypes

src/pyhf/modifiers/shapesys.py
```diff
@@ -115,30 +116,36 @@ def _precompute(self):
...
+                    default_backend.sum(mod_uncert_info[:, 0] > 0, axis=1), dtype='bool'
+                )
+            ]
+
+            # sample_uncert_info: (bin_mask, nominal rate, uncertainty)
+            bin_mask = default_backend.astensor(sample_uncert_info[0][0], dtype='bool')
```
lukasheinrich: shouldn't this be `sample_uncert_info[0]` instead of `sample_uncert_info[0][0]` ?

src/pyhf/modifiers/shapesys.py
```diff
@@ -115,30 +116,36 @@ def _precompute(self):
...
+            nom_unc = sample_uncert_info[0, 1:]
```
lukasheinrich: and this `sample_uncert_info[1]` ?

src/pyhf/modifiers/shapesys.py
kratsg: #785 is filed to handle later

matthewfeickert:
**Posterity:**

To help me understand the 

```python
            sample_uncert_info = mod_uncert_info[
                default_backend.astensor(
                    default_backend.sum(mod_uncert_info[:, 0] > 0, axis=1), dtype='bool'
                )
            ][0]
```

snippet in `finalize` @kratsg gave me the following example of how the `sum` over the axis with the mask applies it:

>```
>(Pdb) pdfconfig.channels
>['channel1']
>(Pdb) pdfconfig.samples
>['background1', 'signal']
>(Pdb) pdfconfig.channel_nbins
>{'channel1': 6}
>(Pdb) self._shapesys_combined__shapesys_info
>array([[[[ 1. , 1. , 0. , 0. , 1. , 1. ],
>          [100. , 90. , 0. , 70. , 0.1, 50. ],
>          [ 10. , 9. , 1. , 0. , 0.1, 5. ]],
>          
>          [[ 0. , 0. , 0. , 0. , 0. , 0. ],
>          [ 20. , 10. , 5. , 3. , 2. , 1. ],
>          [ 0. , 0. , 0. , 0. , 0. , 0. ]]],
>          
>          
>          [[[ 0. , 0. , 0. , 0. , 0. , 0. ],
>          [100. , 90. , 0. , 70. , 0.1, 50. ],
>          [ 0. , 0. , 0. , 0. , 0. , 0. ]],
>          
>          [[ 0. , 0. , 0. , 0. , 0. , 0. ],
>          [ 20. , 10. , 5. , 3. , 2. , 1. ],
>          [ 0. , 0. , 0. , 0. , 0. , 0. ]]]])
>```
>1 channel, 2 samples, 6 bins, and two shape systematics. (one called "syst" and the other called "lowsyst") where syst looks like
>```
>(Pdb) pname
> 'syst'
> (Pdb) mod_uncert_info
> array([[[ 1. , 1. , 0. , 0. , 1. , 1. ],
>         [100. , 90. , 0. , 70. , 0.1, 50. ],
>         [ 10. , 9. , 1. , 0. , 0.1, 5. ]],
>         
>         [[ 0. , 0. , 0. , 0. , 0. , 0. ],
>         [ 20. , 10. , 5. , 3. , 2. , 1. ],
>         [ 0. , 0. , 0. , 0. , 0. , 0. ]]])
>```
> you can see that for the first sample, it affects some but not all bins.. but it's not in the second sample at all (look at the mask). The sum does
>```
>(Pdb) default_backend.sum(mod_uncert_info[:, 0] > 0, axis=1)
>array([4, 0])
>(Pdb) default_backend.astensor(default_backend.sum(mod_uncert_info[:, 0] > 0, axis=1), dtype='bool')
>array([ True, False])
>```
> converted to bools, it says: this shape systematic called syst only subscribes to the first sample.
> For the next one
>```
>(Pdb) pname
>'syst_lowstats'
>(Pdb) pdfconfig.param_set(pname).n_parameters
>0
>```
> you can see, that we'll skip the next one -- because it has no parameters.

PR #768, opened 2020-02-11
test: Add test of independence of pyhf.infer from pyhf details

lukasheinrich:
# Description

this is a test that pyhf.infer.hypotest is really independent of pyhf. This is basically preparing a out-sourcing of the inference package

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add test that pyhf.infer.hypotest is independent of pyhf to ensure that inference is decoupled
```

### DIFF
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -85,3 +85,53 @@ def test_hypotest_return_expected_set(tmpdir, hypotest_args):
     assert isinstance(result[2], type(tb.astensor(result[2])))
     assert len(result[3]) == 5
     assert check_uniform_type(result[3])
+
+def test_inferapi_pyhf_independence():
+    '''
+    pyhf.infer should eventually be factored out so it should be
+    infependent from pyhf internals. This is testing that
+    a much simpler model still can run through pyhf.infer.hypotest
+    '''
+    from pyhf import get_backend
+    class _NonPyhfConfig(object):
+        def __init__(self):
+            self.poi_index = 0
+            self.npars = 2
+        def suggested_init(self):
+            return [1.0,1.0]
+        def suggested_bounds(self):
+            return [[0.,10.],[0.,10.]]
+
+    class NonPyhfModel(object):
+        def __init__(self,spec):
+            self.sig,self.nominal,self.uncert = spec 
+            self.factor = (self.nominal/self.uncert)**2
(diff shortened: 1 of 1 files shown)


lukasheinrich:
I guess we could have abstract base classes

```
class _configApi(object):
    def suggested_init(self) -> 'tensor':
        raise NotImplementedError('not implemented')

    def suggested_bounds(self) -> 'tensor':
        raise NotImplementedError('not implemented')

class _modelApi(object):
    def expected_data(self,pars: 'tensor', include_auxdata=True) -> 'tensor':
        raise NotImplementedError('not implemented')

    def logpdf(self, pars: 'tensor', data: 'tensor') -> 'tensor':
        raise NotImplementedError('not implemented')

```

tests/test_infer.py
```diff
@@ -85,3 +85,61 @@ def test_hypotest_return_expected_set(tmpdir, hypotest_args):
...
+            return tensorlib.concatenate([expected_main, aux_data])
+
+        def logpdf(self, pars, data):
+            tensorlib, _ = get_backend()
+            maindata, auxdata = data
+            main = self._make_main_pdf(pars).log_prob(maindata)
+            constraint = self._make_constraint_pdf(pars).log_prob(auxdata)
+            return tensorlib.astensor([main + constraint])
+
+    m = NonPyhfModel([5, 50, 7])
```
matthewfeickert:
```suggestion
    model = NonPyhfModel([5, 50, 7])
```

Prefer to not have single letter variables. :)

tests/test_infer.py
```diff
@@ -85,3 +85,61 @@ def test_hypotest_return_expected_set(tmpdir, hypotest_args):
...
+    print(pyhf.infer.hypotest(1.0, m.expected_data(m.config.suggested_init()), m))
```
matthewfeickert:
```suggestion
    print(pyhf.infer.hypotest(1.0, model.expected_data(model.config.suggested_init()), model))
```

matthewfeickert: Why use `print` here instead of an `assert`?

kratsg:
Overall, this is mostly just using python's duck-typing here. Not that it's necessarily a bad thing and I only have a minor concern that this test is not as useful as it appears to be... that said, more tests are always good :)

PR #771, opened 2020-02-14
fix: Add key to workflow dictionary to make it valid JavaScript

kratsg:
# Pull Request Description

Following up PR #770, valid JavaScript doesn't allow a single string in a dictionary, assign a key to it fix it

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Valid JavaScript doesn't allow a single string in a dictionary, assign a key to it fix it
* Amends PR #770
```


matthewfeickert:
@kratsg Can you check this in `stare` first?

kratsg:
> @kratsg Can you check this in `stare` first?

I checked it against octokit/rest.js -- we're good.

kratsg:
> I checked it against octokit/rest.js -- we're good.

Ironically, it's ok if this part breaks, because the CI is meant to crash either way. The only difference is whether or not a comment is left in the PR indicating there was an issue, and the bumpversion label being removed. This should at least make this valid javascript so things work.

matthewfeickert:
> > I checked it against octokit/rest.js -- we're good.
> 
> Ironically, it's ok if this part breaks, because the CI is meant to crash either way. The only difference is whether or not a comment is left in the PR indicating there was an issue, and the bumpversion label being removed. This should at least make this valid javascript so things work.

Thanks for this info and check!

PR #774, opened 2020-02-20
feat: Update required_parset to pass in sample data and modifier data

### LINKED ISSUE #773: change required_parset API to receive  full nominal and modifier data
# Description

some modifiers (shapesys) require access to the moodifier details in order to work out how many nuisance parameters to allocate. Thus we need to pass more information than just the number of bins in the channel.

kratsg:
# Pull Request Description

This is related to a few issues on shapesys, but overall, we need to be able to hav ethe fine-control/fine-tuning to mask individual bins of a given parameter.

Resolves #773.

- Poisson(0) is undefined
- Poisson(1/0) is poorly defined
- we don't like 1/0 or similarly asymptotic values
- sometimes an analysis with n-bins in a channel defines a shapesys with data value = 0 somewhere -- maybe all bins -- maybe one bin
- and sometimes an analysis has a sample with 0 expected events (poor stats)
- shapesys sucks in this cases, and ROOT prunes out these cases, but we don't
- so we should
- the PR is the first step by expanding the api to pass around more information so we can handle it.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Modify required_parset API to allow sample data and modifier data to be passed through
```

### DIFF
--- a/src/pyhf/cli/spec.py
+++ b/src/pyhf/cli/spec.py
@@ -43,7 +43,7 @@ def inspect(workspace, output_file, measurement):
         (
             parname,
             modifiers.registry[result['modifiers'][parname]]
-            .required_parset(0)['paramset_type']
+            .required_parset([], [])['paramset_type']
             .__name__,
         )
         for parname in ws.parameters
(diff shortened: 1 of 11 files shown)


lukasheinrich:
#773

lukasheinrich:
@kratsg some tests seem too be failing

PR #775, opened 2020-02-21
feat: Only allocate shapesys parameter if yield and uncrt are positive, nonzero

lukasheinrich:
# Description

from discussions with @kratsg this fixes shapesys by only allocating nuisance parameters for bins in shapesys where the yield and the uncrt is > 0

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- improve handling of shapesys by only allocating nuisance parameters for valid, physically meaningful bins (yield and uncertainty are nonzero and positive)
- add documentation about this particular "feature" of shapesys
```

### DIFF
--- a/src/pyhf/modifiers/shapesys.py
+++ b/src/pyhf/modifiers/shapesys.py
@@ -93,14 +93,14 @@ class shapesys_combined(object):
                     len(batch_access)
                 )
                 singular_sample_index = [
-                    i
-                    for i, x in enumerate(
+                    idx
+                    for idx, syst in enumerate(
                         default_backend.astensor(self._shapesys_mask)[s, :, 0]
                     )
-                    if any(x)
+                    if any(syst)
                 ][-1]
-                this_sample_mask = self._shapesys_mask[s][singular_sample_index][0]
-                access_field_for_syst_and_batch[this_sample_mask] = selection
+                sample_mask = self._shapesys_mask[s][singular_sample_index][0]
+                access_field_for_syst_and_batch[sample_mask] = selection
                 self._access_field[s, t] = access_field_for_syst_and_batch
 
     def _precompute(self):
--- a/src/pyhf/modifiers/shapesys.py
+++ b/src/pyhf/modifiers/shapesys.py
@@ -78,7 +78,7 @@ class shapesys_combined(object):
             for t, batch_access in enumerate(syst_access):
                 selection = self.param_viewer.index_selection[s][t]
                 access_field_for_syst_and_batch = default_backend.astensor(
-                    [-1] * len(batch_access)
+                    [0] * len(batch_access)
                 )
                 singular_sample_index = [
                     i
--- a/src/pyhf/modifiers/shapesys.py
+++ b/src/pyhf/modifiers/shapesys.py
@@ -73,6 +73,8 @@ class shapesys_combined(object):
         # now overwrite with right index in parameter
 
         for s, syst_access in enumerate(self._access_field):
+            if not pdfconfig.param_set(self._shapesys_mods[s]).n_parameters:
+                continue
             for t, batch_access in enumerate(syst_access):
                 selection = self.param_viewer.index_selection[s][t]
                 access_field_for_syst_and_batch = default_backend.astensor(
@@ -105,6 +107,8 @@ class shapesys_combined(object):
 
     def finalize(self, pdfconfig):
         for uncert_this_mod, pname in zip(self.__shapesys_uncrt, self._shapesys_mods):
+            if not pdfconfig.param_set(pname).n_parameters:
+                continue
             unc_nom = default_backend.astensor(
                 [x for x in uncert_this_mod[:, :, :] if any(x[0][x[0] > 0])]
             )
--- a/src/pyhf/modifiers/shapesys.py
+++ b/src/pyhf/modifiers/shapesys.py
@@ -74,12 +74,12 @@ class shapesys_combined(object):
         # access field is shape (sys, batch, globalbin)
 
         # reindex it based on current masking
-        self._reindex_access_field()
+        self._reindex_access_field(pdfconfig)
 
         self._precompute()
         events.subscribe('tensorlib_changed')(self._precompute)
 
-    def _reindex_access_field(self):
+    def _reindex_access_field(self, pdfconfig):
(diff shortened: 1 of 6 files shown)


lukasheinrich:
this requires changing in the access_fields of shapesys

matthewfeickert:
@lukasheinrich is this still WIP?

src/pyhf/modifiers/shapesys.py
```diff
@@ -66,14 +68,22 @@ def __init__(self, shapesys_mods, pdfconfig, mega_mods, batch_size=None):
...
-
+                access_field_for_syst_and_batch = default_backend.astensor(
+                    [-1] * len(batch_access)
+                )
+                singular_sample_index = [
+                    i for i, x in enumerate(self._shapesys_mask[s][0]) if any(x)
+                ][-1]
+                this_sample_mask = self._shapesys_mask[s][0][singular_sample_index]
+                access_field_for_syst_and_batch[this_sample_mask] = selection
+                self._access_field[s, t] = access_field_for_syst_and_batch
```
kratsg: what's going on here to make this... more complicated?

src/pyhf/pdf.py
```diff
@@ -188,7 +188,7 @@ def _nominal_and_modifiers_from_spec(config, spec):
                     )  # broadcasting
                 elif mtype in ['shapesys', 'staterror']:
                     uncrt = thismod['data'] if thismod else [0.0] * len(nom)
-                    maskval = [True if thismod else False] * len(nom)
+                    maskval = [(x > 0 and y > 0) for x, y in zip(uncrt, nom)]
```
kratsg: this is affecting `staterror` too. Is this intended?

lukasheinrich: the logic should be the same but I haven't looked at staterror, prob adding the mast doesn't hurt

lukasheinrich: staterror will only hit this if the entire bin is zero (i.e. none of the samples contribute) which is probably rare

src/pyhf/modifiers/shapesys.py
```diff
@@ -66,16 +72,36 @@ def __init__(self, shapesys_mods, pdfconfig, mega_mods, batch_size=None):
...
+
+        # reindex it based on current masking
+        self._reindex_access_field(pdfconfig)
+
+        self._precompute()
+        events.subscribe('tensorlib_changed')(self._precompute)
+
+    def _reindex_access_field(self, pdfconfig):
+        # access field maps to bin index, but needs to map to right index in
+        # the parameter set
```
matthewfeickert: I know that this was approved when added in PR 519, but maybe the enumerate returns can be named more clearly to make it easier to follow. Though this would include also renaming the `s` and `t` perhaps to `syst_idx` and `batch_idx`. Though maybe I need a refresher on what the `self._access_field` is actually mapping.

src/pyhf/modifiers/shapesys.py
```diff
@@ -66,16 +72,36 @@ def __init__(self, shapesys_mods, pdfconfig, mega_mods, batch_size=None):
...
             for t, batch_access in enumerate(syst_access):
                 selection = self.param_viewer.index_selection[s][t]
-                for b, bin_access in enumerate(batch_access):
-                    self._access_field[s, t, b] = (
-                        selection[bin_access] if bin_access < len(selection) else 0
+                access_field_for_syst_and_batch = default_backend.zeros(
+                    len(batch_access)
+                )
+                singular_sample_index = [
+                    i
```
matthewfeickert:
```suggestion
                    idx
```

src/pyhf/modifiers/shapesys.py
```diff
@@ -66,16 +72,36 @@ def __init__(self, shapesys_mods, pdfconfig, mega_mods, batch_size=None):
...
+                    for i, x in enumerate(
```
matthewfeickert:
```suggestion
                    for idx, syst in enumerate(
```

Assuming I'm reading what is being unpacked correctly.

src/pyhf/modifiers/shapesys.py
```diff
@@ -66,16 +72,36 @@ def __init__(self, shapesys_mods, pdfconfig, mega_mods, batch_size=None):
...
+                        default_backend.astensor(self._shapesys_mask)[s, :, 0]
                     )
-
-        self._precompute()
-        events.subscribe('tensorlib_changed')(self._precompute)
+                    if any(x)
```
matthewfeickert:
```suggestion
                    if any(syst)
```

src/pyhf/modifiers/shapesys.py
```diff
@@ -66,16 +72,36 @@ def __init__(self, shapesys_mods, pdfconfig, mega_mods, batch_size=None):
...
+                ][-1]
+                this_sample_mask = self._shapesys_mask[s][singular_sample_index][0]
```
matthewfeickert: There phrase `sample_mask` doesn't appear anywhere else in the file, so unless I'm missing the importance of the "this" then perhaps just keep it as `sample_mask` to avoid confusion in the future?

```suggestion
                sample_mask = self._shapesys_mask[s][singular_sample_index][0]
```

src/pyhf/modifiers/shapesys.py
```diff
@@ -66,16 +72,36 @@ def __init__(self, shapesys_mods, pdfconfig, mega_mods, batch_size=None):
...
+                access_field_for_syst_and_batch[this_sample_mask] = selection
```
matthewfeickert:
```suggestion
                access_field_for_syst_and_batch[sample_mask] = selection
```

src/pyhf/modifiers/shapesys.py
kratsg: technically, the `this` comes from the fact that it's a loop over samples. So you're saying _this_ sample's mask.

src/pyhf/modifiers/shapesys.py
kratsg: done

kratsg:
> then this is ready to :shipit:

needs an integration test. i'm probably getting around to it Sunday night / Monday...

lukasheinrich:
@kratsg I  added integration test. Should we merge?

PR #790, opened 2020-03-04
feat: Add toy calculator, empirical distribution, and toy example notebook

lukasheinrich:
# Description

this adds the toy calculator.

Needs: #1147.

ReadTheDocs build: 
- https://pyhf.readthedocs.io/en/toycalc/
- https://pyhf.readthedocs.io/en/toycalc/api.html#inference
- https://pyhf.readthedocs.io/en/toycalc/_generated/pyhf.infer.hypotest.html#pyhf.infer.hypotest
- https://pyhf.readthedocs.io/en/toycalc/_generated/pyhf.infer.calculators.EmpiricalDistribution.html#pyhf.infer.calculators.EmpiricalDistribution
- https://pyhf.readthedocs.io/en/toycalc/_generated/pyhf.infer.calculators.ToyCalculator.html#pyhf.infer.calculators.ToyCalculator
- https://pyhf.readthedocs.io/en/toycalc/_generated/pyhf.infer.utils.create_calculator.html#pyhf.infer.utils.create_calculator

- [x] Tests are passing
- [x] Verified [notebook workflow tests](https://github.com/scikit-hep/pyhf/actions?query=workflow%3ANotebooks) are passing

# Before Merging

For the PR Assignees:

```
* Add EmpiricalDistribution and ToyCalculator classes
* Move top level infer functions under 'infer.calculators'
* Add tests for EmpiricalDistribution and ToyCalculator
* Add EmpiricalDistribution and ToyCalculator to docs
* Add example notebook on how to use the ToyCalculator

```


src/pyhf/infer/calculators.py
```diff
@@ -168,3 +174,75 @@ def _false_case():
...
+    def pvalue(self, value):
+        tensorlib, _ = get_backend()
+        return (
+            tensorlib.where(self.samples >= value, 1, 0).sum()
+            / tensorlib.shape(self.samples)[0]
+        )
+
+    def expected_value(self, nsigma):
+        tensorlib, _ = get_backend()
+        import numpy as np
```
lukasheinrich: could be hardcoded for the few nsigma values we need, avoiding extension to tensorlib and importing np

src/pyhf/infer/calculators.py
```diff
@@ -168,3 +174,75 @@ def _false_case():
...
+        print('signal!')
+        signal_qtilde = []
+        for i, sample in enumerate(signal_sample):
+            if i % 100 == 0:
+                print(i)
+            signal_qtilde.append(
+                qmu(poi_test, sample, self.pdf, signal_pars, self.par_bounds)
+            )
+        signal_qtilde = tensorlib.astensor(signal_qtilde)
+
```
lukasheinrich: the progress reporting here should be improved.. alsoo progress bars? something else? I want to avoid cluttering up the stdout but even w/ pyhf toys can be slow

also do we want parallelism a la joblib?

src/pyhf/infer/calculators.py
kratsg: hard-coding still won't get rid of dependence on `np.percentile` or making `tensorlib.percentile`. The hard-coding on `tensorlib.normal_cdf` should be possible -- although I should point out that nothing needs `*Distribution.expected_value` right now...

src/pyhf/infer/mixins.py
```diff
@@ -0,0 +1,116 @@
...
+            raise RuntimeError('need to call .teststatistic(poi_test) first')
+
+        if self.distribution is None:
+            raise RuntimeError('need to call this from a mixin\'d class')
+
+        s_plus_b = self.distribution(signal_qtilde)
+        b_only = self.distribution(bkg_qtilde)
+        return s_plus_b, b_only
+
+    def qmu(self, mu, data=None, pdf=None, init_pars=None, par_bounds=None):
```
lukasheinrich: this was named teststatistic, no?

kratsg: qtilde wasn't used in an equivalent way. This has all the same functionality as the rest of the code. This is calling `from .test_statistics import qmu`

src/pyhf/infer/calculators.py
matthewfeickert: > progress bars?

Would be nice. :) I don't think there are too many times that I've seen a progress bar and thought to myself "oh I hate that".

> do we want parallelism a la joblib?

Parallelism sounds generally preferable.

src/pyhf/infer/mixins.py
```diff
@@ -0,0 +1,116 @@
...
+        self.ntoys = ntoys
+
+    def distributions(self, poi_test):
+        if self.something_signal is None or self.something_bkg is None:
+            raise RuntimeError('need to call .teststatistic(poi_test) first')
+
+        if self.distribution is None:
+            raise RuntimeError('need to call this from a mixin\'d class')
+
+        s_plus_b = self.distribution(signal_qtilde)
```
lukasheinrich: I don't think we have a guarantee that the init signature of distribution have the same semantics so I wouldn't add this here

src/pyhf/infer/calculators.py
kratsg: let's pull this feature into a new issue. out of scope for this PR i think.

kratsg: #807 is filed.

kratsg: Progress bars via tqdm added in this PR.

src/pyhf/infer/calculators.py
```diff
@@ -168,3 +169,160 @@ def _false_case():
                 (sqrtqmu_v < self.sqrtqmuA_v), _true_case, _false_case
             )
         return teststat
+
+
+class EmpiricalDistribution(object):
+    """
+    The empirical distribution of the test statistic.
+
+    Unlike ~pyhf.infer.calculators.AsymptoticTestStatDistribution where the
```
matthewfeickert: I'm not sure why, but if I add `calculators.EmpiricalDistribution` to the `api.rst` in docs this rendering doesn't come out correctly.
EmpiricalDistribution_rendering_error

src/pyhf/infer/calculators.py
```diff
@@ -168,3 +169,160 @@ def _false_case():
...
+        Args:
+            nsigma (`int` or `tensor`): The number of standard deviations.
+
+        Returns:
+            Float: The expected value of the test statistic.
+        """
+        tensorlib, _ = get_backend()
+        import numpy as np
+
+        # TODO: tensorlib.percentile function
```
matthewfeickert: Can you make an Issue for this?

src/pyhf/infer/calculators.py
```diff
@@ -98,7 +99,7 @@ def __init__(self, data, pdf, init_pars=None, par_bounds=None, qtilde=False):
             pdf (~pyhf.pdf.Model): The statistical model adhering to the schema ``model.json``.
             init_pars (`tensor`): The initial parameter values to be used for fitting.
             par_bounds (`tensor`): The parameter value bounds to be used for fitting.
-            qtilde (`bool`): Whether to use qtilde as the test statistic.
+            qtilde (Bool): When ``True`` perform the calculation using the alternative test statistic, :math:`\\tilde{q}`, as defined in Equation (62) of :xref:`arXiv:1007.1727`.
```
matthewfeickert:
```suggestion
            qtilde (`bool`): When ``True`` perform the calculation using the alternative test statistic, :math:`\\tilde{q}`, as defined in Equation (62) of :xref:`arXiv:1007.1727`.
```

I'd keep it as the same style as

```python
>>> print(type(True))
<class 'bool'>
```

If we change to using Bool then we should make sure this gets changed everywhere and not just here.

src/pyhf/infer/utils.py
```diff
@@ -0,0 +1,140 @@
...
+
+from .. import get_backend
+from .calculators import AsymptoticCalculator, ToyCalculator
+
+
+def create_calculator(calctype, *args, **kwargs):
+    """
+    Creates a calculator object of the specified `calctype`.
+
+    See ~pyhf.infer.calculators.AsymptoticCalculator and ~pyhf.infer.calculators.ToyCalculator on additional arguments to be specified.
```
matthewfeickert: These also aren't getting rendered out correctly when `utils.create_calculator` is added to `api.rst`.

src/pyhf/infer/utils.py
```diff
@@ -0,0 +1,140 @@
...
+    pdf,
+    init_pars=None,
+    par_bounds=None,
+    calctype='asymptotics',
+    **kwargs,
+):
+    r"""
+    Compute :math:`p`-values and test statistics for a single value of the parameter of interest.
+
+    See ~pyhf.infer.calculators.AsymptoticCalculator and ~pyhf.infer.calculators.ToyCalculator on additional keyword arguments to be specified.
```
matthewfeickert: These also aren't getting rendered out correctly in docs

matthewfeickert:
@kratsg I made the following edits to the notebook and have some questions. I'm happy to discuss or roll back any of them.

- Instead of using pylab just import numpy and matplotlib. I'm not really a big fan of pylab and I'd prefer to be very explicit to users what is happening.
- Mention that mu' is the signal strength of the model being hypothesis being considered
- Compute the Poisson uncertainty so that if the model were to change it is updated correctly
```
background_uncertainty = int(np.sqrt(background))
```
- Once the pyhf model is introduced switch to code formatting to channels and signal strengths
- I'd suggest first using the word "pseudo-experiments" which is more generally known, and then say '(or "toys" as particle physicists would say)'). Also introduce the term "throwing toys" ala Glen and Kyle.
- Set a parameter to control the number of samples (`n_samples`)
- > This means to sample N=1000  from the pdfs

 Shouldn't this be"N=2000" given the numbers that show up in the notebook?
- Given the API choice in the PR (introducing `utils.create_calculator`) need to change to using `pyhf.infer.utils.create_calculator`

- > For now, we will create a toy-based calculator and evaluate it for data expected if the background-only hypothesis were true ( 𝜇=1.0 )

 Shouldn't this be "the nominal signal+background model were true"? This is `q_1`/`\tilde{q}_1`. Or have I confused myself with notation again? 

- Increased the font size of the plots
```python
plt.rcParams.update({"font.size": 14})
```
- Set the x-axes tick marks to match Figure 5 of the Asymptoics paper more.
- Make the plots larger (was `12,5`) but keep spacing
```python
fig.set_size_inches(13.5,6)
fig.tight_layout(pad=2.0)
```
- Add titles to the plots
- Add vertical-axis labels
- Ran Black over it

Let me know your thoughts and if you disagree with anything that I can fix up. I think this is a _really_ nice notebook and I'm very excited that we finally have this made now.

kratsg:
> * Compute the Poisson uncertainty so that if the model were to change it is updated correctly
> 
> ```
> background_uncertainty = int(np.sqrt(background))
> ```

we want to set the uncertainty explicitly. Revert this.



> Shouldn't this be"N=2000" given the numbers that show up in the notebook?
> 
> *

should be 2k.



> * Given the API choice in the PR (introducing `utils.create_calculator`) need to change to using `pyhf.infer.utils.create_calculator`
> * >

no. `pyhf.infer.__init__` imports this correctly. `utils` are meant to be used internally (imo).



> * > For now, we will create a toy-based calculator and evaluate it for data expected if the background-only hypothesis were true ( 𝜇=1.0 )
> 
> Shouldn't this be "the nominal signal+background model were true"? This is `q_1`/`\tilde{q}_1`. Or have I confused myself with notation again?

See chat with Lukas offline.



> * Make the plots larger (was `12,5`) but keep spacing

you lose the square ratio. Should be at least 14,6 if you're doing that.


> * Increased the font size of the plots
> 
> ```python
> plt.rcParams.update({"font.size": 14})
> ```

this is nitpicking for the notebook and should be removed.

> * Set the x-axes tick marks to match Figure 5 of the Asymptoics paper more.

nitpicking as well. should be removed.



> ```python
> ```python
> fig.tight_layout(pad=2.0)
> ```
> ```

we don't need tight layout do we? the artists fit fine in the boundary box by default.

matthewfeickert:
> we want to set the uncertainty explicitly. Revert this.

Done
 
> no. `pyhf.infer.__init__` imports this correctly. `utils` are meant to be used internally (imo).

Seem we have a problem then
not_working

 
> > * > For now, we will create a toy-based calculator and evaluate it for data expected if the background-only hypothesis were true ( 𝜇=1.0 )
> > 
> > Shouldn't this be "the nominal signal+background model were true"? This is `q_1`/`\tilde{q}_1`. Or have I confused myself with notation again?
> 
> See chat with Lukas offline.

The Skype chat still doesn't make sense. I'll follow up again.
  
> you lose the square ratio. Should be at least 14,6 if you're doing that.

Done. Changed to `14,6 `.

> > * Increased the font size of the plots
> > 
> > ```python
> > plt.rcParams.update({"font.size": 14})
> > ```
> 
> this is nitpicking for the notebook and should be removed.

The font is pretty small without it. :/

> > * Set the x-axes tick marks to match Figure 5 of the Asymptoics paper more.
> 
> nitpicking as well. should be removed.

Is the spacing too tight?

> we don't need tight layout do we? the artists fit fine in the boundary box by default.

I think that these plots can still be confusing (_even though_ there are the nice plot legends) without axis labels. If you don't set padding and apply a vertical-axis label then the right hand plot's label gets scrunched up against the left hand side. The padding solves this.

kratsg:
> Seem we have a problem then

I'm realizing that create_calculator shouldn't be exposed by default unless people want to do more advanced things. The interface via `pyhf.infer.hypotest` exists, so this is good.

kratsg:
> The font is pretty small without it. :/
> Is the spacing too tight?

ok. I guess as long as it's defined at the top of the notebook, then it's fine.

matthewfeickert:
This is more for me, but this is a [nbviewer render of the toys notebook in this PR](https://nbviewer.jupyter.org/github/scikit-hep/pyhf/blob/toycalc/docs/examples/notebooks/toys.ipynb?flush_cache=false): (also [weirdly you have to set `flush_cache` to `false` to get it to update](https://github.com/jupyter/nbviewer/issues/295#issuecomment-591660582)).

src/pyhf/infer/calculators.py
```diff
@@ -168,3 +169,160 @@ def _false_case():
...
+
+        bkg_qtilde = []
+        for sample in tqdm.tqdm(bkg_sample, **tqdm_options, desc='Background-like'):
+            bkg_qtilde.append(
+                qmu(poi_test, sample, self.pdf, bkg_pars, self.par_bounds)
+            )
+        bkg_qtilde = tensorlib.astensor(bkg_qtilde)
+
+        s_plus_b = EmpiricalDistribution(signal_qtilde)
+        b_only = EmpiricalDistribution(bkg_qtilde)
```
matthewfeickert, line 313:
```suggestion
        signal_test_stat = []
        for sample in tqdm.tqdm(signal_sample, **tqdm_options, desc='Signal-like'):
            signal_test_stat.append(
                qmu(poi_test, sample, self.pdf, signal_pars, self.par_bounds)
            )

        bkg_test_stat = []
        for sample in tqdm.tqdm(bkg_sample, **tqdm_options, desc='Background-like'):
            bkg_test_stat.append(
                qmu(poi_test, sample, self.pdf, bkg_pars, self.par_bounds)
            )

        s_plus_b = EmpiricalDistribution(tensorlib.astensor(signal_test_stat))
        b_only = EmpiricalDistribution(tensorlib.astensor(bkg_test_stat))
```

I think that we should use use the more generic "test statistic" (`test_stat`) here rather than `qtilde` as that is making an assumption about the test being done. Seems like it could lead to confusion later and that a more explicit example in the docstring that uses `qtilde` would be more enlightening.

src/pyhf/infer/calculators.py
matthewfeickert: Resolved with 
```
:py:class:`target-goes-here`
```

matthewfeickert:
At the moment the toys notebook is getting ignored in tests (given the `slow` marker I added). There should be a CI job that is just a nightly CRON schedule that will run all the `slow` tests.

src/pyhf/infer/calculators.py
matthewfeickert: This is Issue #815

src/pyhf/infer/calculators.py
matthewfeickert: Question RE: this here: https://github.com/scikit-hep/pyhf/issues/815#issuecomment-609107946

kratsg:
> A note that as this is going to be part of minor release `v0.5.0` we should add tests for all the calculators into the public API tests (`tests/test_public_api.py`).

something similar to `toys.ipynb` would be good.

src/pyhf/infer/calculators.py
lukasheinrich: what do we do with this? wait for torch 1.7.0 or wrap/accept this use of `default_backend` / `np`. Alternatively we implement percentile ourselves.

matthewfeickert: @lukasheinrich I would vote that we just wrap this for the time being and then come back and remove it when we can, but I don't think we should hold this up any longer than I already have by asking us to wait for other things.

src/pyhf/infer/utils.py
```diff
@@ -0,0 +1,220 @@
...
+
+    calc = create_calculator(
+        calctype,
+        data,
+        pdf,
+        init_pars,
+        par_bounds,
+        fixed_params,
+        qtilde=qtilde,
+        **kwargs,
```
matthewfeickert, line 177: At the moment all the `kwargs` are getting passed through and not checked which is then resulting in errors as the `hypotest` specific `kwargs` are not known to `create_calculator`. Example:

```python
>>> import pyhf
>>> test_mu = 1.0
>>> pyhf.infer.hypotest(test_mu, data, model, qtilde=True, return_expected=True)
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/infer/utils.py", line 177, in hypotest
    **kwargs,
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/infer/utils.py", line 39, in create_calculator
    ](*args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'return_expected'
```

matthewfeickert: This _should_ now be resolved by adding the following above the `create_calculator` call

```python
    # Get the hypotest return specific kwargs
    return_tail_probs = kwargs.pop("return_tail_probs", None)
    return_expected = kwargs.pop("return_expected", None)
    return_expected_set = kwargs.pop("return_expected_set", None)
```

matthewfeickert:
PR #1114 will change the default `qtilde` to `True` so the screaming of doctest can be ignored.

src/pyhf/infer/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        ... )
+        >>> qmu_sig, qmu_bkg = toy_calculator.distributions(mu_test)
+        >>> qmu_sig.pvalue(mu_test), qmu_bkg.pvalue(mu_test)
+        (0.14, 0.76)
+
+    Args:
+        calctype (`str`): The calculator to create. Choose either 'asymptotics' or 'toybased'.
+
+    Returns:
+        calculator (`object`): A calculator.
```
matthewfeickert, line 34: Note to self for later make these

```suggestion
        calctype (:obj:`str`): The calculator to create. Choose either 'asymptotics' or 'toybased'.

    Returns:
        calculator (:obj:`object`): A calculator.
```

and check elsewhere for similar.

src/pyhf/infer/utils.py
```diff
@@ -0,0 +1,41 @@
...
+        >>> import numpy.random as random
+        >>> random.seed(0)
+        >>> model = pyhf.simplemodels.hepdata_like(
+        ...     signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0],
+        ... )
+        >>> observations = [51, 48]
+        >>> data = observations + model.config.auxdata
+        >>> mu_test = 1.0
+        >>> toy_calculator = pyhf.infer.utils.create_calculator(
+        ...     "toybased", data, model, ntoys=100, track_progress=False
```
matthewfeickert, line 26: Should make this explicit in the example

```suggestion
        ...     "toybased", data, model, ntoys=100, qtilde=True, track_progress=False
```

tests/test_notebooks.py
```diff
@@ -62,6 +62,11 @@ def test_impactplot(common_kwargs):
     pm.execute_notebook('docs/examples/notebooks/ImpactPlot.ipynb', **common_kwargs)
 
 
+@pytest.mark.slow
```
matthewfeickert, line 65: Do we still need this anymore? Perhaps so if we want to be able to run all the notebooks locally? Not sure.

matthewfeickert:
@lukasheinrich @kratsg I think we can review this now. Also, as ReviewNB wasn't added to the repo before this PR was started here's a [`nbviewer` render of the `toys` notebook in this PR](https://nbviewer.jupyter.org/github/scikit-hep/pyhf/blob/toycalc/docs/examples/notebooks/toys.ipynb?flush_cache=1).

src/pyhf/infer/calculators.py
kratsg: See #817.

kratsg (approved):

matthewfeickert (approved):
thank you @lukasheinrich for your work on this and a huge amount of patience.

PR #794, opened 2020-03-07
chore: Avoid mutable defaults

matthewfeickert:
# Description

Given a Tweet by @jakevdp [about mutable defaults](https://twitter.com/jakevdp/status/1235271748867612673) that got me thinking along with a message with @mickypaganini, this PR makes all function defaults that are empty lists be `None` instead and follow the
```python
x = [] if x is None else x
```
pattern. However, this doesn't actually really do anything, as the design of the code so far is well thought through and any mutable defaults can't propagate through to [produce any of the usual errors](http://effbot.org/zone/default-values.htm). This PR exists to essentially future proof any changes and to try to mindless adopt 'best patterns' (if you will).

As can be seen by `git grep "=\[\]"`, `git grep "={}"`, (thank you [Black](https://github.com/psf/black)!) and 

```
$ git grep ".update" src/
src/pyhf/optimize/opt_minuit.py:            kwargs.update(**d)
src/pyhf/readxml.py:                        result['config']['parameters'][0].update(overall_param_obj)
src/pyhf/readxml.py:                        param_obj.update(overall_param_obj)
src/pyhf/tensor/jax_backend.py:        config.update('jax_enable_x64', True)
src/pyhf/writexml.py:            sample.attrib.update({'NormalizeByTheory': 'True'})
```

none of these changes will affect the current code. 

 this only affects `pyhf.workspace` and the `spec` CLI (hence why I mention above this change can't actually affect anything now). `setup.py` requires the `[]` and it is justified in the tests.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Avoid mutable defaults in all function definitions to follow 'best patterns'
   - c.f. https://twitter.com/jakevdp/status/1235271748867612673
```

### DIFF
--- a/docs/exts/xref.py
+++ b/docs/exts/xref.py
@@ -5,8 +5,9 @@ from docutils import nodes
 from sphinx.util import caption_ref_re
 
 
-def xref(typ, rawtext, text, lineno, inliner, options={}, content=None):
+def xref(typ, rawtext, text, lineno, inliner, options=None, content=None):
     # avoid mutable defaults
+    options = {} if options is None else options
     content = [] if content is None else content
 
     title = target = text
--- a/docs/exts/xref.py
+++ b/docs/exts/xref.py
@@ -5,7 +5,9 @@ from docutils import nodes
 from sphinx.util import caption_ref_re
 
 
-def xref(typ, rawtext, text, lineno, inliner, options={}, content=[]):
+def xref(typ, rawtext, text, lineno, inliner, options={}, content=None):
+    # avoid mutable defaults
+    content = [] if content is None else content
 
     title = target = text
     # titleistarget = True
--- a/src/pyhf/cli/spec.py
+++ b/src/pyhf/cli/spec.py
@@ -133,17 +133,17 @@ def inspect(workspace, output_file, measurement):
     help='The location of the output json file. If not specified, prints to screen.',
     default=None,
 )
-@click.option('-c', '--channel', default=None, multiple=True, metavar='<CHANNEL>...')
-@click.option('-s', '--sample', default=None, multiple=True, metavar='<SAMPLE>...')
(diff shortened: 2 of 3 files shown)


matthewfeickert:
CodeFactor is complaining that `_prune_and_rename` is too complex but that isn't something we care about and also not the scope of the PR so that can be ignored.

kratsg:
I have a hard time imagining a situation where the `cli` has a mutable default that gets modified. It just doesn't make sense to me... Did you grep for `={}` as well?

lukasheinrich:
I usually use the 

```
def foo(x = None):
    x = x or []
```

pattern which I find a bit conciser. What do you think?

matthewfeickert:
> I have a hard time imagining a situation where the `cli` has a mutable default that gets modified.

I agree that I couldn't think of a situation in which this would be possible, but I thought I would include it to see if you could think of some bizarre abuse case. As you can't I'm happy to revert this section.

>  Did you grep for `={}` as well?

No, but should for same reasons

```
>>> def f(str="a", x={}):
...     x.update({str: 1})
...     return x
... 
>>> print(f(), f("b"))
{'a': 1, 'b': 1} {'a': 1, 'b': 1}
```

matthewfeickert:
> I usually use the
> 
> ```
> def foo(x = None):
>     x = x or []
> ```
> 
> pattern which I find a bit conciser. What do you think?

The _very_ small problem with this is that it treats `False` the same as `None`

```python
>>> def foo(x=None):
...     x = x or []
...     return x
... 
>>> print(foo(),foo(1),foo(False),foo(None))
[] 1 [] []
>>> def f(x=None):
...     x = [] if x is None else x
...     return x
... 
>>> print(f(),f(1),f(False),f(None))
[] 1 False []
```

matthewfeickert:
I can't seem to figure out how to get CodeFactor to ignore that segment...but everything else is passing CI.

PR #795, opened 2020-03-09
chore: Use Debian based python:3.7-slim as base Docker image

matthewfeickert:
# Description

For less than 100MB more in size using [`python:3.7-slim`](https://hub.docker.com/layers/python/library/python/3.7-slim/images/sha256-e65d472af32c1f51cca9350de6bd63cf4ce440f2e68fb9301bcf12cdba1b0acd?context=explore) as 
a base image for the pyhf Docker images gets a much more common environment for reproduction and testing. For example, it can reproduce the failing error of the [Minimal pyhf example failing with 'Inequality constraints incompatible'](https://stackoverflow.com/questions/60514470/minimal-pyhf-example-failing-with-inequality-constraints-incompatible) Stack Overflow issue instantly rather then running for minutes and then failing on an iteration limit exceeded error as the Alpine based image does. It also allows for direct installation of SciPy and NumPy.

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Use python:3.7-slim as base image for Docker image builds
   - Allows for greater flexibility and easier reproducibility of user results and errors
```

PR #796, opened 2020-03-10
build: Require minimum SciPy version of v1.4.0

matthewfeickert:
# Description

- Requires PR #795 to be resolved first

In [SciPy `v1.4.0`](https://github.com/scipy/scipy/blob/master/doc/release/1.4.0-notes.rst#scipy-optimize-improvements) the following fix is added in https://github.com/scipy/scipy/pull/10739

> SLSQP can run to singular BFGS updates, which it doesn't handle, and may exit with nan outputs and successful exit code. To avoid this, do the most trivial thing and just reset the matrix. This might not lead to convergence, but gets rid of false successful exits.
> ...
> The linalg changes are no-ops if there are no nans, and with nans generally give error exits. The BFGS update tweak is only triggered in cases where the code previously divided by zero. The SLSQP exit-on-too-many-resets check is also a corner case. So there should be no impact in non-problematic cases.

Adopting SciPy `v1.4.0` as the new minimum version will lead to faster fails and fewer 'Iteration limit exceeded' failures after multiple minutes. This change was discovered empirically in PR #795 and then hunted down by @kratsg for this PR.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Require SciPy v1.4.0 or newer
   - Avoids some 'Iteration limit exceeded' errors and false exits with nans
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ packages = find:
 include_package_data = True
 python_requires = >=3.6
 install_requires =
-    scipy  # requires numpy, which is required by pyhf and tensorflow
+    scipy>=1.4.0  # requires numpy, which is required by pyhf and tensorflow
     click>=6.0  # for console scripts,
     tqdm  # for readxml
     jsonschema>=3.2.0  # for utils


matthewfeickert:
This should probably also be a patch release.

matthewfeickert:
Maybe this should wait for PR #799 to go in, so that the patch release nicely grabs it too.

matthewfeickert:
@lukasheinrich To try to move forward on this, can you remind me about your ideas here on adding extras for a "application" mode vs. a (default) "library" mode?

matthewfeickert:
@lukasheinrich @kratsg I'd like to restart the discussion on if we need an extra that doesn't require this recent version of SciPy. My argument is that we don't need one, as people should be running pyhf in virtual environments anyway so their dependencies shouldn't be limited by other projects. Though it has been long enough that I don't remember the specifics of the discussion.

kratsg:
The main argument is that there’s not enough new scipy versions so there’s no wiggle room or flexibility. I personally am fine restricting now but...

matthewfeickert:
> The main argument is that there’s not enough new scipy versions so there’s no wiggle room or flexibility.

This should be fixed simply by the progression of time. I also don't really see this being an issue at all, as the only common dependency across everything is NumPy, and the latest versions of SciPy [just push this up to `numpy>=1.14.5`](https://github.com/scipy/scipy/commit/c33e9d3cbf69bb692c4487308ba6dbc805e89b3f) &mdash; not a very aggressive release number. So I don't see any real situation in which a user of pyhf is able to install everything they need _except_ for a recent version of SciPy.

Though maybe I'm missing the main point of concern.

PR #798, opened 2020-03-10
refactor: Use pathlib over os.path for path

### LINKED ISSUE #614: Use pathlib over os.path
# Description

After Issue #469 is resolved, pyhf will be Python 3.6 or higher. At this point, it is possible to then use `pathlib` instead of `os.path` which is generally nice.

Examples and discussions of why this is preferred can be seen in Trey Hunner's post on the subject: [Why you should be using pathlib](https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/)

kanishk16:
# Description

This updates all files except `test_notebooks.py` to use `pathlib` over `os.path` for path related functionality.

Resolves #614 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Refactor to use pathlib over os.path module for path related functions
* Add tests for readxml.import_root_histogram
```


kanishk16:
Changes to test_notebooks.py introduced errors, so currently this PR caters to updates to all files except the test_notebook file. I would update changes if anyone reviews and verifies if it introduces errors only on my system or others could reproduce the errors as well.

docs/conf.py
```diff
@@ -14,14 +14,14 @@
...
 #
-import os
+from pathlib import Path
 import sys
 from pkg_resources import get_distribution
 
-sys.path.insert(0, os.path.abspath('../src'))
-sys.path.insert(1, os.path.abspath('./exts'))
+sys.path.insert(0, Path('../src').resolve().as_posix())
+sys.path.insert(1, Path('./exts').resolve().as_posix())
```
kratsg: Please don't use the `.resolve().as_posix()` since we don't want to assume a filesystem. This should work fine with just the `str(Path(...).resolve())` or something similar.

setup.py
```diff
@@ -1,8 +1,8 @@
 from setuptools import setup, find_packages
-from os import path
+from pathlib import Path
 
-this_directory = path.abspath(path.dirname(__file__))
-with open(path.join(this_directory, 'README.md'), encoding='utf-8') as readme_md:
+this_directory = Path(__file__).parent.resolve()
+with open(Path().joinpath(this_directory, 'README.md'), encoding='utf-8') as readme_md:
```
kratsg: Is there a difference between this and `Path(this_directory).joinpath('README.md')`?

src/pyhf/readxml.py
```diff
@@ -48,12 +48,11 @@ def import_root_histogram(rootdir, filename, path, name, filecache=None):
         h = f[name]
     except KeyError:
         try:
-            h = f[os.path.join(path, name)]
+            h = f[Path().joinpath(path, name).as_posix()]
```
kratsg: Does this not work?

```suggestion
            h = f[Path(path).joinpath(name)]
```

Otherwise, wrap it in `str`? I'm not sure about the `as_posix` because we don't test on Windows.

src/pyhf/readxml.py
```diff
@@ -311,7 +310,7 @@ def parse(configfile, rootdir, track_progress=False):
     for inp in inputs:
         inputs.set_description('Processing {}'.format(inp))
         channel, data, samples, channel_parameter_configs = process_channel(
-            ET.parse(os.path.join(rootdir, inp)), rootdir, track_progress
+            ET.parse(Path().joinpath(rootdir, inp).as_posix()), rootdir, track_progress
```
kratsg: same comment about posix here

src/pyhf/utils.py
```diff
@@ -17,14 +17,12 @@ def load_schema(schema_id, version=None):
...
-        return SCHEMA_CACHE[
-            "{0:s}{1:s}".format(SCHEMA_BASE, os.path.join(version, schema_id))
-        ]
+        return SCHEMA_CACHE[f'{SCHEMA_BASE}{Path().joinpath(version, schema_id)}']
     except KeyError:
         pass
 
     path = pkg_resources.resource_filename(
-        __name__, os.path.join('schemas', version, schema_id)
+        __name__, Path().joinpath('schemas', version, schema_id).as_posix()
```
kratsg: same comment about posix here

src/pyhf/writexml.py
```diff
@@ -242,16 +242,21 @@ def writexml(spec, specdir, data_rootdir, resultprefix):
...
     )
     combination = ET.Element(
-        "Combination", OutputFilePrefix=os.path.join('.', specdir, resultprefix)
+        "Combination",
+        OutputFilePrefix=Path().joinpath('.', specdir, resultprefix).as_posix(),
     )
 
-    with uproot.recreate(os.path.join(data_rootdir, 'data.root')) as _ROOT_DATA_FILE:
+    with uproot.recreate(
+        Path(data_rootdir).joinpath('data.root').as_posix()
```
kratsg: same comment about posix here

src/pyhf/writexml.py
```diff
@@ -242,16 +242,21 @@ def writexml(spec, specdir, data_rootdir, resultprefix):
...
+    ) as _ROOT_DATA_FILE:
         for channelspec in spec['channels']:
-            channelfilename = os.path.join(
-                specdir, '{0:s}_{1:s}.xml'.format(resultprefix, channelspec['name'])
+            channelfilename = (
+                Path(specdir)
+                .joinpath(f'{resultprefix}_{channelspec["name"]}.xml')
+                .as_posix()
```
kratsg: same comment about posix here

tests/test_import.py
```diff
@@ -232,7 +232,7 @@ def test_import_filecache(mocker):
 
     # check if uproot.open was only called once with the expected root file
     pyhf.readxml.uproot.open.assert_called_once_with(
-        os.path.join("validation/xmlimport_input", "./data/example.root")
+        Path().joinpath("validation/xmlimport_input", "./data/example.root").as_posix()
```
kratsg: same comment about posix here

setup.py
kanishk16: I believe no.

kanishk16:
I would like to know why empty `Path()` calls are not recommended or a bad coding practice. What about pyhf being the default starting path as `Path('../pyhf').resolve()` ? Also, explicitly specifying any path as the argument won't make it system dependent?

kratsg:
> I would like to know why empty `Path()` calls are not recommended or a bad coding practice. 

I don't think it's bad coding practice, but it looks a little weird to me style-wise. If `from pathlib import joinpaths` existed, then maybe that makes sense. The default I think is the current directory so that `Path() == Path('./')` but not sure here.

> What about pyhf being the default starting path as `Path('../pyhf').resolve()` ?

Usually, if we want to resolve files located within the package, you use the `pkg_resources`.

> Also, explicitly specifying any path as the argument won't make it system dependent?

The `as_posix` makes it system-dependent. From what I understand, `pathlib` consumes both posix/windows-style paths.

kanishk16:
After making changes in a way similar to `h = f[str(Path(path).joinpath(name))]`, there are no empty path calls left. 
 
>  The default I think is the current directory so that `Path() == Path('./')` but not sure here.

I reconfirmed from the official docs and you are on point.

src/pyhf/writexml.py
```diff
@@ -242,16 +242,18 @@ def writexml(spec, specdir, data_rootdir, resultprefix):
...
-    with uproot.recreate(os.path.join(data_rootdir, 'data.root')) as _ROOT_DATA_FILE:
+    with uproot.recreate(
+        str(Path(data_rootdir).joinpath('data.root'))
+    ) as _ROOT_DATA_FILE:
         for channelspec in spec['channels']:
-            channelfilename = os.path.join(
-                specdir, '{0:s}_{1:s}.xml'.format(resultprefix, channelspec['name'])
+            channelfilename = str(
+                Path(specdir).joinpath(f'{resultprefix}_{channelspec["name"]}.xml')
             )
```
kratsg:
```suggestion
            channelfilename = Path(specdir).joinpath(f'{resultprefix}_{channelspec["name"]}.xml')
```

kratsg: Don't forget to put `str(channelfilename)` down below.

kratsg: Never mind. I think this is fine as it is.

matthewfeickert:
Oh, weird. Codecov has never been run on this PR. I'm going to do an empty commit to bump and see if it triggers.

matthewfeickert:
The CI is failing and I'm not sure why, so I'm going to drop the commit I added and do a force push.

kanishk16:
From what I observed, one of the errors that occurred were `git checkout failed with exit code 128`, not sure about if this is the exact reason for the CI/CD failing but this error has been in discussion in this [thread](https://github.com/actions/checkout/issues/23) by Github Actions which I believe you would like to take a look at.

Just another question, why is that in some cases the Codcov does not trigger (observed this issue in another PR)?

src/pyhf/readxml.py
```diff
@@ -48,12 +48,11 @@ def import_root_histogram(rootdir, filename, path, name, filecache=None):
...
         try:
-            h = f[os.path.join(path, name)]
+            h = f[str(Path(path).joinpath(name))]
         except KeyError:
             raise KeyError(
-                'Both {0:s} and {1:s} were tried and not found in {2:s}'.format(
-                    name, os.path.join(path, name), os.path.join(rootdir, filename)
-                )
+                f'Both {name} and {Path(path).joinpath(name)} were tried and not found'
+                f' in {Path(rootdir).joinpath(filename)}'
```
matthewfeickert: This try except block is now not getting covered in the tests for reasons that aren't clear to me: https://codecov.io/gh/scikit-hep/pyhf/pull/798/diff?src=pr&el=tree#D3-49

It would be good to fix these if possible.

matthewfeickert:
> From what I observed, one of the errors that occurred were `git checkout failed with exit code 128`, not sure about if this is the exact reason for the CI/CD failing but this error has been in discussion in this [thread](https://github.com/actions/checkout/issues/23) by Github Actions which I believe you would like to take a look at.

Yeah, this seemed to go away after I dropped the empty commit I added to try to tirgger Codecov. Not sure exactly.

> Just another question, why is that in some cases the Codecov does not trigger (observed this issue in another PR)?

That's a good question. Codecov seems odd at times, but it should always trigger after a successful run of the CI where it gets to the point where it pytest reports the coverage and it can upload the coverage file. If the CI fails then there is no coverage report. We'll need to check if this is only generated on commits on the main project though. :?

src/pyhf/readxml.py
kanishk16: I have restructured 54-55 into a single line for now and would probably revert to using os.path if it still does not get covered in the tests.

kratsg:
@matthewfeickert are we ok here?

PR #802, opened 2020-03-14
fix: Set correct workspace.version upon loading a workspace

### LINKED ISSUE #801: Workspace.version is not set when loading in a workspace with a version
# Description

See the title.

# Expected Behavior

`Workspace.version is not None`

# Actual Behavior

`Workspace.version is None`

# Steps to Reproduce

```python
import pyhf, json

ws = pyhf.Workspace(json.load(open('workspace.json')))
assert ws.version is not None
assert ws.version == ws['version']
```

# Checklist

kanishk16:
# Description

This PR fixes the bug `Workspace.version is not None` upon loading a workspace with a version by assigning the correct version to the instance variable `version` accordingly as per the arguments.

Resolves #801 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Set the workspace version while loading a workspace
* Add test to check loaded workspace for version
```


src/pyhf/workspace.py
```diff
@@ -267,7 +267,12 @@ def __init__(self, spec, **config_kwargs):
         """Workspaces hold the model, data and measurements."""
         super(Workspace, self).__init__(spec, channels=spec['channels'])
         self.schema = config_kwargs.pop('schema', 'workspace.json')
-        self.version = config_kwargs.pop('version', None)
+
+        if 'version' in spec:
+            self.version = spec['version']
+        else:
+            self.version = config_kwargs.pop('version', None)
+
```
kratsg: this should be done after schema validation which guarantees a `version` in the spec.

src/pyhf/workspace.py
```diff
@@ -267,10 +267,15 @@ def __init__(self, spec, **config_kwargs):
...
         self.schema = config_kwargs.pop('schema', 'workspace.json')
-        self.version = config_kwargs.pop('version', None)
+
         # run jsonschema validation of input specification against the (provided) schema
-        log.info("Validating spec against schema: {0:s}".format(self.schema))
-        utils.validate(self, self.schema, version=self.version)
+        log.info(f"Validating spec against schema: {self.schema}")
+        utils.validate(self, self.schema)
+
+        if 'version' in spec:
```
kratsg: this check is not needed since `version` must be in `spec` to pass schema validation.

kanishk16: Yes, my bad. I'll make the changes.

kratsg:
I'm realizing we're going to make this harder in the future unfortunately, this way. Let's move it back to before the schema validation and have the one-liner be

`self.version = config_kwargs.pop('version', spec.get('version', None))`

kanishk16:
> I'm realizing we're going to make this harder in the future unfortunately, this way.
Would it be reasonable to ask for the intuition or the logic behind it?

kratsg:
> Would it be reasonable to ask for the intuition or the logic behind it?

we need to know the version of the workspace to validate against the right version of the schema.

kanishk16:
> Need a test to check for regression in `tests/test_workspace.py`

Now on to it.

PR #804, opened 2020-03-16
refactor: Use _TensorViewer in ParamViewer

lukasheinrich:
# Description

moves `ParamViewer` to use `_TensorViewer` so that we re-use that logic instead of repeating ourselves

```
* Remove unused constraint_model.expected_data method (used some details in ParamViewer that were removed
* Refactor ParamViewer to use TensorViewer remove .slices attribute
* Add ability in TensorViewer to attach names to subspaces and return only subset of tensors when splitting
* Add tensorviewer tests
* Modify paramviewer tests to test more interesting cases
* simpify _TensorViewer construction for full pdf using new helper methods
```

### DIFF
--- a/src/pyhf/parameters/paramview.py
+++ b/src/pyhf/parameters/paramview.py
@@ -47,7 +47,6 @@ class ParamViewer(object):
     """
 
     def __init__(self, shape, par_map, selection):
-        db = default_backend
         self.shape = shape
         self.selection = selection
 
--- a/src/pyhf/parameters/paramview.py
+++ b/src/pyhf/parameters/paramview.py
@@ -70,7 +70,7 @@ class ParamViewer(object):
 
     def __init__(self, shape, par_map, selection):
 
-        sbatch = shape[0] if len(shape) > 1 else None
+        batch = shape[0] if len(shape) > 1 else None
 
         fullsize = default_backend.product(default_backend.astensor(shape))
         flat_indices = default_backend.astensor(range(int(fullsize)), dtype='int')
--- a/src/pyhf/parameters/paramview.py
+++ b/src/pyhf/parameters/paramview.py
@@ -2,16 +2,43 @@ from .. import get_backend, default_backend, events
 from ..tensor.common import _TensorViewer
 
 
-def index_helper(name, tensor_shape, batch_shape, par_map):
-    if isinstance(name, list):
-        return [index_helper(x, tensor_shape, batch_shape, par_map) for x in name]
-    x = list(range(int(default_backend.product(tensor_shape))))
-    indices = default_backend.reshape(x, tensor_shape)
-    parfield_slice = tuple(slice(None, None) for x in batch_shape) + (
-        par_map[name]['slice'],
+def _tensorviewer_from_parmap(par_map, batch_size):
+    db = default_backend
+    # prepares names and per-parset ranges
+    # in the order or the parameters
+    names, indices, starts = list(
+        zip(
+            *sorted(
+                [
+                    (
+                        k,
+                        db.astensor(range(v['slice'].start, v['slice'].stop)),
+                        v['slice'].start,
+                    )
+                    for k, v in par_map.items()
+                ],
+                key=lambda x: x[2],
+            )
+        )
     )
-    indices = indices[parfield_slice]
-    return default_backend.tolist(indices)
+    return _TensorViewer(indices, names=names, batch_size=batch_size)
+
+
+def _tensorviewer_from_slices(start_slices, batch_size):
+    target_slices = []
+    start = 0
+    for sl in start_slices:
+        stop = start + (sl.stop - sl.start)
+        target_slices.append(slice(start, stop))
+        start = stop
+
+    db = default_backend
+    ranges = []
+    for sl in target_slices:
+        ranges.append(db.astensor(range(sl.start, sl.stop)))
+    if not ranges:
+        return (target_slices, None)
+    return target_slices, _TensorViewer(ranges, batch_size=batch_size)
 
 
 class ParamViewer(object):
@@ -24,48 +51,21 @@ class ParamViewer(object):
         self.shape = shape
         self.selection = selection
 
-        # prepares names and per-parset ranges
-        # in the order or the parameters
-        names, indices, starts = list(
-            zip(
-                *sorted(
-                    [
-                        (
-                            k,
-                            db.astensor(range(v['slice'].start, v['slice'].stop)),
-                            v['slice'].start,
-                        )
-                        for k, v in par_map.items()
-                    ],
-                    key=lambda x: x[2],
-                )
-            )
-        )
-
         self.batch = shape[0] if len(shape) > 1 else None
 
         # a tensor viewer that can split and stitch parameters
-        self.allpar_viewer = _TensorViewer(indices, names=names, batch_size=self.batch)
+        self.allpar_viewer = _tensorviewer_from_parmap(par_map, self.batch)
 
         # to combine the selected
         # parameters into a overall tensor
         # we need to prep some ranges
-        slices = []
-        ranges = []
-        start = 0
-        for s in selection:
-            sl = par_map[s]['slice']
-            stop = start + (sl.stop - sl.start)
-            ranges.append(db.astensor(range(start, stop)))
-            slices.append(slice(start, stop))
-            start = stop
 
-        # used in tests
-        self.slices = slices
+        start_slices = [par_map[s]['slice'] for s in selection]
 
-        if self.selection:
-            # a tensor viewer that can split and stitch the selected parameters
-            self.selected_viewer = _TensorViewer(ranges, batch_size=self.batch)
+        # a tensor viewer that can split and stitch the selected parameters
+        self.slices, self.selected_viewer = _tensorviewer_from_slices(
+            start_slices, self.batch
+        )
 
         self._precompute()
         events.subscribe('tensorlib_changed')(self._precompute)
--- a/src/pyhf/parameters/paramview.py
+++ b/src/pyhf/parameters/paramview.py
@@ -1,5 +1,5 @@
 from .. import get_backend, default_backend, events
-from ..tensor.common import _TensorViewer
+from ..tensor.common import _TensorViewer, _tensorviewer_from_slices
 
 
 def _tensorviewer_from_parmap(par_map, batch_size):
@@ -22,23 +22,6 @@ def _tensorviewer_from_parmap(par_map, batch_size):
         )
     )
     return _TensorViewer(indices, names=names, batch_size=batch_size)
-
-
-def _tensorviewer_from_slices(slices, names, batch_size):
-    target_slices = []
-    start = 0
-    for sl in slices:
-        stop = start + (sl.stop - sl.start)
-        target_slices.append(slice(start, stop))
-        start = stop
-
-    db = default_backend
-    ranges = []
-    for sl in target_slices:
-        ranges.append(db.astensor(range(sl.start, sl.stop)))
-    if not ranges:
-        return None
-    return _TensorViewer(ranges, names=names, batch_size=batch_size)
 
 
 def extract_index_access(
--- a/src/pyhf/parameters/paramview.py
+++ b/src/pyhf/parameters/paramview.py
@@ -1,4 +1,5 @@
 from .. import get_backend, default_backend, events
+from ..tensor.common import _TensorViewer
 
 
 def index_helper(name, tensor_shape, batch_shape, par_map):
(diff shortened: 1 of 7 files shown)


tests/test_paramviewer.py
```diff
@@ -10,14 +10,14 @@ def test_paramviewer_simple_nonbatched(backend):
...
     )
     par_slice = view.get(pars)
-    assert pyhf.tensorlib.tolist(par_slice[view.slices[0]]) == [1, 2]
+    assert pyhf.tensorlib.tolist(par_slice[slice(2, 4)]) == [1, 2]
 
-    assert pyhf.tensorlib.tolist(par_slice[view.slices[1]]) == [6, 7]
+    assert pyhf.tensorlib.tolist(par_slice[slice(0, 2)]) == [6, 7]
 
-    assert pyhf.tensorlib.tolist(par_slice) == [1, 2, 6, 7]
+    assert pyhf.tensorlib.tolist(par_slice) == [6, 7, 1, 2]
```
matthewfeickert: As the choice in ordering here results in changes beyond just `view.slices` -> `slice` is there a pedagogical, style, or preference for doing so? If so that's fine, just curious. :)

tests/test_paramviewer.py
```diff
@@ -28,7 +28,7 @@ def test_paramviewer_simple_batched(backend):
     view = ParamViewer(
         parshape,
         {'hello': {'slice': slice(0, 2)}, 'world': {'slice': slice(3, 4)}},
-        ['hello', 'world'],
+        ['world', 'hello'],
```
matthewfeickert: Same comment as above

tests/test_paramviewer.py
```diff
@@ -38,7 +38,7 @@ def test_paramviewer_simple_batched(backend):
     )  # first dimension is batch dim
 
     assert pyhf.tensorlib.shape(par_slice) == (3, 3)
-    assert pyhf.tensorlib.tolist(par_slice[view.slices[0]]) == [[1, 5, 9], [2, 6, 10]]
-    assert pyhf.tensorlib.tolist(par_slice[view.slices[1]]) == [[4, 8, 12]]
+    assert pyhf.tensorlib.tolist(par_slice[slice(1, 3)]) == [[1, 5, 9], [2, 6, 10]]
+    assert pyhf.tensorlib.tolist(par_slice[slice(0, 1)]) == [[4, 8, 12]]
 
-    assert pyhf.tensorlib.tolist(par_slice) == [[1, 5, 9], [2, 6, 10], [4, 8, 12]]
+    assert pyhf.tensorlib.tolist(par_slice) == [[4, 8, 12], [1, 5, 9], [2, 6, 10]]
```
matthewfeickert: Same comment as above

tests/test_paramviewer.py
lukasheinrich: there is a feature in paramviewers that does some ordering, and the change in the test is explicitly chosen now to make clear that the ordering happens

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
-    if isinstance(name, list):
-        return [index_helper(x, tensor_shape, batch_shape, par_map) for x in name]
-    x = list(range(int(default_backend.product(tensor_shape))))
-    indices = default_backend.reshape(x, tensor_shape)
-    parfield_slice = tuple(slice(None, None) for x in batch_shape) + (
-        par_map[name]['slice'],
+def _tensorviewer_from_parmap(par_map, batch_size):
+    db = default_backend
+    # prepares names and per-parset ranges
+    # in the order or the parameters
```
kratsg:
```suggestion
    # in the order of the parameters
```

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
+    names, indices, starts = list(
+        zip(
+            *sorted(
+                [
+                    (
+                        k,
+                        db.astensor(range(v['slice'].start, v['slice'].stop)),
+                        v['slice'].start,
+                    )
+                    for k, v in par_map.items()
```
kratsg: why are you not using `par_order` here?

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
+                ],
+                key=lambda x: x[2],
```
kratsg: order based on the slice.start?

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
-        return [index_helper(x, tensor_shape, batch_shape, par_map) for x in name]
-    x = list(range(int(default_backend.product(tensor_shape))))
-    indices = default_backend.reshape(x, tensor_shape)
-    parfield_slice = tuple(slice(None, None) for x in batch_shape) + (
-        par_map[name]['slice'],
+def _tensorviewer_from_parmap(par_map, batch_size):
+    db = default_backend
+    # prepares names and per-parset ranges
+    # in the order or the parameters
+    names, indices, starts = list(
```
kratsg:
```suggestion
    names, indices, _ = list(
```

perhaps? do we need the starts or is it just for sorting correctly?

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
+                    )
+                    for k, v in par_map.items()
+                ],
+                key=lambda x: x[2],
+            )
+        )
     )
-    indices = indices[parfield_slice]
-    return default_backend.tolist(indices)
+    return _TensorViewer(indices, names=names, batch_size=batch_size)
```
kratsg: I would strongly prefer that you just get this to extract out the slices/names/batch_size and call the below `return _tensorviewer_from_slices(slices, names, batch_size)` to reduce code

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
+
+
+def _tensorviewer_from_slices(slices, names, batch_size):
```
kratsg: this makes sense to me to make this a `_TensorViewer` class method e.g. `_TensorViewer.from_slices(slices, names, batch_size)` which returns a `_TensorViewer` object.

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
+    db = default_backend
+    ranges = []
+    for sl in target_slices:
+        ranges.append(db.astensor(range(sl.start, sl.stop)))
+    if not ranges:
+        return None
+    return _TensorViewer(ranges, names=names, batch_size=batch_size)
+
+
+def extract_index_access(
```
kratsg: what does this do?

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
-        last = 0
-        sl = []
-        for s in [
-            self._par_map[x]['slice'].stop - self._par_map[x]['slice'].start
-            for x in (name if isinstance(name, list) else [name])
-        ]:
-            sl.append(slice(last, last + s))
-            last += s
-        self._slices = sl
+        batch = shape[0] if len(shape) > 1 else None
```
kratsg: call this `batch_size`. Is the `shape` passed in here just the parameters and batched parameters are more than 1D? That's not obvious. What if we end up having N-dimensional parameters, and batching is N+1-dimensional?

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
+
+        fullsize = default_backend.product(default_backend.astensor(shape))
```
kratsg: i guess because we don't have the underlying `tensor` to call `tensor.size`, we need to resort to doing it this way?

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
+        flat_indices = default_backend.astensor(range(int(fullsize)), dtype='int')
```
kratsg: why a call to `int()` here? you are typecasting twice which isn't a big deal performance-wise but...

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
-    def __init__(self, tensor_shape, par_map, name):
-        self.tensor_shape = tensor_shape
-        self.batch_shape = tensor_shape[:-1]
-        # for more general batch shapes we need to revisit
-        assert len(self.batch_shape) <= 1
-        self._par_map = par_map
-        self._index_selection = index_helper(
-            name, tensor_shape, self.batch_shape, self._par_map
-        )
+    def __init__(self, shape, par_map, selection):
```
kratsg:
```suggestion
    def __init__(self, shape, par_map, par_names):
```

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
         tensorlib, _ = get_backend()
-        if not self.batch_shape:
-            return tensorlib.gather(
-                tensor, indices if indices is not None else self.indices_concatenated
-            )
-        return tensorlib.gather(
-            tensorlib.reshape(tensor, (-1,)),
-            indices if indices is not None else self.indices_concatenated,
-        )
+        indices = indices if indices is not None else self.indices_concatenated
```
kratsg:
```suggestion
        indices = indices or self.indices_concatenated
```

what happens if `indices = []`?

src/pyhf/parameters/paramview.py
```diff
@@ -1,114 +1,107 @@
...
-        """
-        Args:
-            tensor (`tensor`): the data tensor to extract a view from
-            indices (`tensor`): an optional index selection (default behavior is to use self.indices_concatenated)
-        Returns:
-            filtered set of parameter field/array :
-                type when batched: (sum of slice sizes, batchsize) tensor
-                type when not batched: (sum of slice sizes, ) tensors
-        """
+    def get(self, data, indices=None):
```
kratsg: which indices are being passed in here? the all_indices which can be > npars (if batched) or corresponding to the pars, and tiled across all batched?

src/pyhf/tensor/common.py
```diff
@@ -45,12 +47,17 @@ def stitch(self, data):
             stitched = tensorlib.einsum('j...->...j', stitched)
         return stitched
 
-    def split(self, data):
+    def split(self, data, selection=None):
```
kratsg:
```suggestion
    def split(self, data, names=None):
```

tests/test_paramviewer.py
```diff
@@ -10,14 +10,14 @@ def test_paramviewer_simple_nonbatched(backend):
...
         {'hello': {'slice': slice(0, 2)}, 'world': {'slice': slice(5, 7)}},
-        ['hello', 'world'],
+        ['world', 'hello'],
     )
     par_slice = view.get(pars)
-    assert pyhf.tensorlib.tolist(par_slice[view.slices[0]]) == [1, 2]
+    assert pyhf.tensorlib.tolist(par_slice[slice(2, 4)]) == [1, 2]
 
-    assert pyhf.tensorlib.tolist(par_slice[view.slices[1]]) == [6, 7]
+    assert pyhf.tensorlib.tolist(par_slice[slice(0, 2)]) == [6, 7]
```
kratsg: these two tests don't really do anything that's not covered by the third. I suggest maybe testing something like `par_slice['world'] == [1,2]` but it doesn't seem like that's possible. I suppose that's not the point since `ParamViewer` already filters out the `hello` and `world` for you...

tests/test_paramviewer.py
```diff
@@ -38,7 +38,7 @@ def test_paramviewer_simple_batched(backend):
     )  # first dimension is batch dim
 
     assert pyhf.tensorlib.shape(par_slice) == (3, 3)
-    assert pyhf.tensorlib.tolist(par_slice[view.slices[0]]) == [[1, 5, 9], [2, 6, 10]]
-    assert pyhf.tensorlib.tolist(par_slice[view.slices[1]]) == [[4, 8, 12]]
+    assert pyhf.tensorlib.tolist(par_slice[slice(1, 3)]) == [[1, 5, 9], [2, 6, 10]]
```
kratsg: is it more useful to have the transposed version? I naively expected

```
[[1, 2], [5, 6], [9, 10]]
```

to maintain batching as the first dimension since you're slicing. In this case, we're slicing _AND_ transposing.

src/pyhf/parameters/paramview.py
lukasheinrich: because we receive only the par_map object.

src/pyhf/parameters/paramview.py
lukasheinrich: range requires an int and fullsize is a np.int64

src/pyhf/parameters/paramview.py
lukasheinrich: it's just for sorting

src/pyhf/parameters/paramview.py
lukasheinrich: yes

src/pyhf/parameters/paramview.py
lukasheinrich: yes & done

src/pyhf/parameters/paramview.py
lukasheinrich: yeah we don't have a size method

src/pyhf/parameters/paramview.py
lukasheinrich: yes in this case these are different cases so the `is None` check is more precise

src/pyhf/parameters/paramview.py
lukasheinrich: it prepares index tensors for `tb.gather(parameters,indextensor)`-like operations

src/pyhf/parameters/paramview.py
lukasheinrich: good point! done

tests/test_paramviewer.py
lukasheinrich: yeah the modifiers need it transposed

src/pyhf/parameters/paramview.py
kratsg: We need a way to ensure that if we end up changing the order in `par_order` that it is reflected here somehow. I'm worried that we've disentangled things which is tricky.

tests/test_paramviewer.py
lukasheinrich: I didn't want to remove the tests just b/c I removed the attr, but I agree they're not adding anything

tests/test_paramviewer.py
kratsg: Yeah, this is a bit of a shock to me. I would expect you adhere to the same dimension ordering as the input parameters if that's possible? It's not obvious that this would transpose...

tests/test_paramviewer.py
kratsg: ok, i agree. it doesn't necessarily hurt, just extra unneeded code.

PR #813, opened 2020-03-21
feat: Catch unknown kwargs while creating a model

### LINKED ISSUE #668: Catch unspecified/unknown kwargs
# Description

Came up in #620. The "bug" there is that the line had a typo:

```
p = w.model(measurement_name=None, patches=patches, modifiers_settings=msettings)
                                                            ^
```

instead of

```
p = w.model(measurement_name=None, patches=patches, modifier_settings=msettings)
```

We should probably catch unknown kwargs (e.g. pop them all off and make sure there's no unconsumed kwargs).

Typos happen and we should probably help identify the typos.

Check if all kwargs are used up and raise an error if not.

None.

- #620

kanishk16:
# Description

This PR catches unspecified or unknown kwargs due to typos while creating a model object.
 
Resolves #668 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Catch unknown kwargs while creating a model object  
* Add test to check any unknown keyword argument
```


matthewfeickert:
@kanishk16 Thanks for working on this. Is this ready for review now or are you planning on doing more work on this?

kanishk16:
Yes, it's ready for review and would like to know if I possibly missed anything.

src/pyhf/pdf.py
```diff
@@ -222,12 +222,22 @@ def __init__(self, spec, **config_kwargs):
             spec, self.channel_nbins
         )
 
-        poiname = config_kwargs.get('poiname', 'mu')
+        poiname = config_kwargs.pop('poiname', 'mu')
+
+        # measurement_name is explicitly removed in cases where the model is created using poi_name, multiple
+        # measurements are there but the first one is taken or if measurement is defined as None in the argument
+        if 'measurement_name' in config_kwargs:
+            del config_kwargs['measurement_name']
```
matthewfeickert: Is the only reason this is done to ensure that `config_kwargs` is empty after the `"modifier_settings"` are popped off? If so, then this could probably be moved to just above the check and its positioning alone would make the comment unnecessary.

src/pyhf/pdf.py
```diff
@@ -222,12 +222,22 @@ def __init__(self, spec, **config_kwargs):
...
         default_modifier_settings = {'normsys': {'interpcode': 'code1'}}
-        self.modifier_settings = (
-            config_kwargs.get('modifier_settings') or default_modifier_settings
+        self.modifier_settings = config_kwargs.pop(
+            'modifier_settings', default_modifier_settings
         )
 
+        assert (
+            not config_kwargs
+        ), f"""Unexpected keyword argument(s) '{"', '".join(config_kwargs.keys())}'"""
```
matthewfeickert:
```suggestion
        ), f"""Unexpected keyword argument(s): '{"', '".join(config_kwargs.keys())}'"""
```

src/pyhf/pdf.py
```diff
@@ -222,12 +222,22 @@ def __init__(self, spec, **config_kwargs):
...
+            del config_kwargs['measurement_name']
+
         default_modifier_settings = {'normsys': {'interpcode': 'code1'}}
-        self.modifier_settings = (
-            config_kwargs.get('modifier_settings') or default_modifier_settings
+        self.modifier_settings = config_kwargs.pop(
+            'modifier_settings', default_modifier_settings
         )
 
+        assert (
```
matthewfeickert: In general I think it is a good idea to avoid the use of `assert` outside of tests, as [they can be removed unintentionally](https://stackoverflow.com/a/1838411/8931942). It would be better to do a check here and then raise the proper error.

src/pyhf/workspace.py
```diff
@@ -325,15 +325,15 @@ def get_measurement(self, **config_kwargs):
 
     def _get_measurement(self, **config_kwargs):
         """See `Workspace::get_measurement`."""
-        poi_name = config_kwargs.get('poi_name')
+        poi_name = config_kwargs.pop('poi_name', None)
```
matthewfeickert: Maybe I'm missing something (@kratsg?) but as no check is performed anywhere here (in `workspace.py`) what is the advantage of switching from `get` to `pop` here?

src/pyhf/pdf.py
matthewfeickert: It would be nice to get a test added to `test_pdf.py` that checks this behavior and the error type.

src/pyhf/workspace.py
kratsg: well, the advantage is we know when the configuration parameter is used. `config_kwargs` will be modified by deleting the `poi_name` key, so we can check if there's extra keys used... such as perhaps a typo.

src/pyhf/pdf.py
kanishk16: Yes, I believe we can move it up. On second thoughts, wouldn't it be nice to keep the comment for future as we are already using pop while `_get_measurement()` is called?

kanishk16:
I would be moving it to WIP since I'll be working on this in a day or two as I'm on the verge of completing my draft for the proposal.

src/pyhf/pdf.py
kanishk16: Pinging @matthewfeickert, just to highlight the above query.

kratsg: keep the comment and change to `pop`

src/pyhf/pdf.py
```diff
@@ -222,12 +222,22 @@ def __init__(self, spec, **config_kwargs):
             spec, self.channel_nbins
         )
 
-        poiname = config_kwargs.get('poiname', 'mu')
+        # measurement_name is explicitly removed in cases where the model is created using poi_name, multiple
+        # measurements are there but the first one is taken or if measurement is defined as None in the argument
+        config_kwargs.pop('measurement_name', None)
```
kratsg, line 227: these lines could be removed entirely. `_ModelConfig()` should never be called with `measurement_name`. if this is the case, then logic needs to be fixed, but we can file an issue to fix this beyond the PR.

matthewfeickert: Yeah, I would remove the comment. I don't think it is helpful and currently as is seems confusing.

kratsg: I've filed #836 to keep track.

src/pyhf/pdf.py
```diff
@@ -222,12 +222,22 @@ def __init__(self, spec, **config_kwargs):
             spec, self.channel_nbins
         )
 
-        poiname = config_kwargs.get('poiname', 'mu')
+        # measurement_name is explicitly removed in cases where the model is created using poi_name, multiple
+        # measurements are there but the first one is taken or if measurement is defined as None in the argument
```
kratsg, line 226:
```suggestion
        # measurement_name is passed in via Workspace::model and this is a bug. We'll remove it here for now
        # but needs to be fixed upstream. #836 is filed to keep track.
```

src/pyhf/pdf.py
kanishk16: Indeed as in the first place, this should already be handled.

src/pyhf/pdf.py
kanishk16: I believe this seems apt.

PR #819, opened 2020-04-06
docs: Use rST instead of Markdown for README and ROADMAP

kratsg:
# Description

Resolves #818. We drop the `m2r` dependency. This fixes the sphinx builds.

Docs build at: https://pyhf.readthedocs.io/en/docs-newsphinx/

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Write README and ROADMAP in rST instead of Markdown
* Drop m2r dependency
* Avoid unused arguments error from pyflakes (new in pyflakes v2.2.0)
* Run apt-get clean on Linux runners due to GitHub Actions disk space regression
   - c.f. https://github.community/t5/GitHub-Actions/GitHub-Actions-Failing-with-Errno-28-No-space-left-on-device/m-p/54516
```

PR #825, opened 2020-04-19
fix: Replace use of raw directive in README.rst

### LINKED ISSUE #824: Use of raw directive in README.rst not allowed on PyPI
# Description

In PR #819 the use of the `raw` directive in rST was used to render images correctly. However, this is not allowed on PyPI and breaks the publication checks. The `raw` directive must be fixed with `image` directives like @kratsg originally had. However, this is not rendered correctly on GitHub.

# Expected Behavior

Both GitHub and PyPI render the `READEM.rst` correctly.

matthewfeickert:
# Description

Resolves #824

In PR #819 the use of the `raw` directive in rST was used to render images correctly. However, this is not allowed on PyPI and breaks the publication checks.

ReadTheDocs build: https://pyhf.readthedocs.io/en/fix-fix-raw-html-in-readme.rst/

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Remove use of raw directive in README.rst as not allowed on PyPI
   - Amends PR #819
* Resize image files for Brazil Band plots in README to be 500 px wide
   - c.f. https://github.com/github/markup/issues/295
* Add twine check of dist to publishing CI tests
```


matthewfeickert:
@kratsg you had a suggestion here about 64 bit string encoding of images for the resize? Can you elaborate?

kratsg:
> There is an example on the Stack Overflow question "[Can you resize or change resolution of base64 image through manipulating the base64 code?](https://stackoverflow.com/q/33291436/8931942)" but I'm not sure how to be able to do this without editing the source images themselves as we can't use raw HTML or javascript on PyPI.

this is only for the SO tag. Can you just not resize the other images and save the resized images in the repo? Then we don't need to explicitly resize them. Or we get rid of the README on github and move the README into the docs, because GitHub is clearly the problem here...

matthewfeickert:
> Can you just not resize the other images and save the resized images in the repo? Then we don't need to explicitly resize them.

This certainly is the easiest, but I wanted to avoid having to do this as then we have to downsize the logo, which is used in other places online from this repo and it would be nice to keep it large.

> Or we get rid of the README on github and move the README into the docs, because GitHub is clearly the problem here...

I think it is better to have a README with weird sized images than no README in the top level of the repo at all. We could move the rST README into the docs and then have a Markdown README at the top level, but then we're maintaining two versions of the same file.

matthewfeickert:
> This certainly is the easiest, but I wanted to avoid having to do this as then we have to downsize the logo, which is used in other places online from this repo and it would be nice to keep it large.

I guess the only thing I really care about is the pyhf logo. To me, the only images in the GitHub render of the README that are offensive are the CLs plots. I've resized the ones that are the worst offenders to have widths of 500 pixels (I didn't bother with the one that is 514 pixels wide) using GIMP. If we can live with a slightly gaudily large pyhf logo and big Stack Overflow logo in the README then I think this fixes things.

kratsg:
> This certainly is the easiest, but I wanted to avoid having to do this as then we have to downsize the logo, which is used in other places online from this repo and it would be nice to keep it large.

provide an svg version of the logo and a png version that's "logo-small.png" or similar -- this is probably better.

kratsg:
Screenshot 2020-04-21 08 17 28

checkboxes look consistent btw. The image should show as long as it's included in master.

PR #833, opened 2020-04-21
fix: Get full history and all tags for TestPyPI and PyPI publishing workflow

### LINKED ISSUE #832: Bug: scm versioning no longer being used during TestPyPI builds
# Description

In PR #826, the [`checkout` GitHub Action](https://github.com/actions/checkout) was updated to stable release `v2`. However, [this broke the semantic versioning](https://github.com/scikit-hep/pyhf/runs/605413197#step:8:12) that is relied upon for the TestPyPI uploads to work. The repo seems to have lost understanding of what the version is as the wheel that is built is `pyhf-0.1.dev1-py2.py3-none-any.whl`, while the TestPyPI release just before that from PR #823 (still on [`actions/checkout@v1.2.0`](https://github.com/actions/checkout/releases/tag/v1.2.0)) correctly gave the version as `0.4.2.dev20`.

# Expected Behavior

Have the repo know its current release version so that it can use the scm dev release versioning.

# Actual Behavior

The repo doesn't know its current version state and defaults to version `0.1.dev1`.

# Steps to Reproduce

Run the Build and publish Python distro to (Test)PyPI CI workflow

# Checklist

matthewfeickert:
# Description

Resolves #832 

Have [`actions/checkout@v2`](https://github.com/actions/checkout/releases/tag/v2) unshallow the history and get all of the tags. This allows `pep517` to be able to use the repo information to assign the version number of the build properly.

Here are examples of the verification check:
- [Failing properly (as expected)](https://github.com/scikit-hep/pyhf/runs/606545321?check_suite_focus=true#step:7:13)
- [Succeeding](https://github.com/scikit-hep/pyhf/runs/606768192?check_suite_focus=true#step:8:13)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Get all history and all tags for pep517 to determine version of the built distribution
   - Amends PR #826
* Add versioning verification check
```

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -30,16 +30,28 @@ jobs:
     - name: Build a binary wheel and a source tarball
       run: |
         python -m pep517.build --source --binary --out-dir dist/ .
-    - name: Verify versioning is working
-      if: github.event_name == 'push' && "!(startsWith(github.ref, 'refs/tags/'))" && github.repository == 'scikit-hep/pyhf'
+    - name: Verify untagged commits have dev versions
+      if: "!(startsWith(github.ref, 'refs/tags/'))" && github.repository == 'scikit-hep/pyhf'
       run: |
         wheel_name=$(find dist/ -iname "*.whl" -printf "%f\n")
-        echo "pep517.build named built distribution: ${wheel_name}"
-        if [[ "${wheel_name}" == *"pyhf-0.1.dev"* ]]; then
+        if [[ "${wheel_name}" == *"pyhf-0.1.dev"* || "${wheel_name}" != *"dev"* ]]; then
+          echo "pep517.build incorrectly named built distribution: ${wheel_name}"
           echo "pep517 is lacking the history and tags required to determine version number"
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-package.yml
```diff
@@ -26,6 +30,15 @@ jobs:
     - name: Build a binary wheel and a source tarball
       run: |
         python -m pep517.build --source --binary --out-dir dist/ .
+    - name: Verify versioning is working
```
kratsg, line 33: shouldn't this only be run for non-tags? something like

`if: "!startsWith(github.ref, 'refs/tags')"`

matthewfeickert: Actually, is this necessary to not have it run on some jobs? All it is doing is checking to make sure that the wheel name **does not** contain the string "pyhf-0.1.dev". If it doesn't then it doesn't matter.

kratsg: my point is that versioning should be confirmed by checking that the version is not "pyhf-0.1.dev" and that it matches "pyhf-*.dev" which is always true for non-tagged.

PR #838, opened 2020-04-22
refactor: Simplify Workspace::get_measurement and harmonize poi_name

### LINKED ISSUE #836: Simplify Workspace::_get_measurement()
# Description

At the moment, some of the functionality of `get_measurement()` on the `Workspace` is likely a bit too smart/magic. This should be trimmed/cleaned up to be simpler and less complex. This came out of some changes in #813 which reveal that we're passing in unused keyword arguments into `_ModelConfig` (namely: `measurement_name`).

kratsg:
# Pull Request Description

Resolves #836.

This PR will simplify `Workspace::get_measurement` by removing `_get_measurement` and rewriting into conditionals instead for clarity.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Workspace::_get_measurement is removed and merged with Workspace::get_measurement
* measurement_name is not passed as a keyword argument into pyhf::Model
* poiname is renamed to poi_name for consistency everywhere
```

### DIFF
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -701,3 +701,11 @@ def test_combine_workspace(workspace_factory, join):
     assert set(combined.channels) == set(ws.channels + new_ws.channels)
     assert set(combined.samples) == set(ws.samples + new_ws.samples)
     assert set(combined.parameters) == set(ws.parameters + new_ws.parameters)
+
+
+def test_workspace_equality(workspace_factory):
+    ws = workspace_factory()
+    ws_other = workspace_factory()
+    assert ws == ws
+    assert ws == ws_other
+    assert ws != 'not a workspace'
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -78,17 +78,17 @@ def test_get_measurement_nonexist(workspace_factory):
 
 
 def test_get_measurement_index_outofbounds(workspace_factory):
-    w = workspace_factory()
+    ws = workspace_factory()
     with pytest.raises(pyhf.exceptions.InvalidMeasurement) as excinfo:
-        w.get_measurement(measurement_index=9999)
+        ws.get_measurement(measurement_index=9999)
     assert 'out of bounds' in str(excinfo.value)
 
 
 def test_get_measurement_no_measurements_defined(workspace_factory):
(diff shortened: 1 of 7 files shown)


kratsg:
NB: I'm not sure if the public API really changes. We don't explicitly list `poiname`/`poi_name` anywhere so...

tests/test_workspace.py
```diff
@@ -77,6 +77,21 @@ def test_get_measurement_nonexist(workspace_factory):
...
+    with pytest.raises(pyhf.exceptions.InvalidMeasurement) as excinfo:
+        w.get_measurement(measurement_index=9999)
+    assert 'out of bounds' in str(excinfo.value)
+
+
+def test_get_measurement_no_measurements_defined(workspace_factory):
+    w = workspace_factory()
+    w.measurement_names = []
+    with pytest.raises(pyhf.exceptions.InvalidMeasurement) as excinfo:
+        w.get_measurement()
```
matthewfeickert, line 91:
```suggestion
    ws = workspace_factory()
    with pytest.raises(pyhf.exceptions.InvalidMeasurement) as excinfo:
        ws.get_measurement(measurement_index=9999)
    assert 'out of bounds' in str(excinfo.value)


def test_get_measurement_no_measurements_defined(workspace_factory):
    ws = workspace_factory()
    ws.measurement_names = []
    with pytest.raises(pyhf.exceptions.InvalidMeasurement) as excinfo:
        ws.get_measurement()
```

nitpicky annoying request for no single letter variables.

Though as I realize there are other instances where this occurs in this test file and other instances that use `ws` so if you'd prefer I just make a harmonizing PR I can do that too.

matthewfeickert:
> NB: I'm not sure if the public API really changes. We don't explicitly list `poiname`/`poi_name` anywhere so...

I think the tag is still valid as if a user ran `help()` they would see changes in `get_measurement` at least.

I'm also happy with this style change. :+1:

kratsg:
> I think the tag is still valid as if a user ran `help()` they would see changes in `get_measurement` at least.

There's no changes to `get_measurement` except really documentation-wise. It's more explicit which kwargs are being allowed, but that's it.

PR #839, opened 2020-04-22
fix: Have release versioning work in docs build CI

matthewfeickert:
# Description

Similar to PR #833, the docs build requires information on the version of pyhf is is reporting on to be able to properly display the version on the page. This PR unshallows the clone of pyhf used for the docs build and gets the tags as well.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Get all history and all tags for docs build to determine version of pyhf
   - Amends PR #826
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      # Get history and tags for versioning to work
+    - run: |
+        git fetch --prune --unshallow
+        git fetch --depth=1 origin +refs/tags/*:refs/tags/*
     - name: Set up Python 3.7
       uses: actions/setup-python@v1
       with:


matthewfeickert:
This test run in another branch [shows that this PR will work](https://github.com/scikit-hep/pyhf/runs/610052052?check_suite_focus=true#step:7:260) as adding

```yaml
grep "pyhf .*.dev" docs/_build/html/index.html
```

after the build in the CI results in

```
  <title>pure-python fitting/limit-setting/interval estimation HistFactory-style &mdash; pyhf 0.4.2.dev29 documentation</title>.
```

matthewfeickert:
@kratsg @lukasheinrich I'm pretty sure this is the only other place we need to get the repo history and tags in the checkout. The docker builds are working as

```
$ docker pull pyhf/pyhf:latest
$ docker run --rm pyhf/pyhf:latest --version
pyhf, version 0.4.1
```

so I think only here and the publishing to PyPI were affected.

I think we can leave everywhere else with just

```yaml
- uses: actions/checkout@v2
```

for faster checkouts and builds. Let me know if you have any concerns on this.

PR #826, opened 2020-04-19
chore: Use GitHub Actions checkout action v2

matthewfeickert:
# Description

Update the [`checkout` GitHub Action](https://github.com/actions/checkout) to use stable release `v2`.

Requires PR #825 to go in first.

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Update checkout action to v2 (stable major release v2.0.0)
```

PR #842, opened 2020-04-24
fix: Get full history for Tag Creator workflow

matthewfeickert:
# Description

In PR #826 all uses of `actions/checkout` were updated to `v2`. This was also done to the Tag Creator GitHub Action, which [resulted in a break](https://github.com/scikit-hep/pyhf/pull/796/checks?check_run_id=614484399#step:7:1). This change already caused trouble in PR #833 and PR #839, so the same patch that was applied to PR #833 is also applied here.

An example of the [patched workflow is done on this test branch here](https://github.com/scikit-hep/pyhf/runs/615811679?check_suite_focus=true#step:7:1).

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Get all history for checkout master to work in Tag Creator
   - Amends PR #826
```

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -43,7 +43,7 @@ jobs:
           echo "::set-env name=BV_PART::patch"
         fi
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v1.2.0
     - name: Set up git user
       run: |
         git config --local user.email "action@github.com"
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -44,10 +44,9 @@ jobs:
         fi
     - name: Checkout repository
       uses: actions/checkout@v2
-      # Get history and tags for checkout master to work
+      # Get history for checkout master to work
     - run: |
         git fetch --prune --unshallow
-        git fetch --depth=1 origin +refs/tags/*:refs/tags/*
     - name: Set up git user
       run: |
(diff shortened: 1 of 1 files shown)


matthewfeickert:
@kratsg if you have a good way of debugging this now though we can just add the necessary changes and switch this from a `revert` to a `fix`. I think all it needs is just the changes from PR #839 also applied to Tag Creator, but I'm not 100% this is all that is needed.

kratsg:
> I think all it needs is just the changes from PR #839 also applied to Tag Creator

we need the unshallow changes, yes... this difference in the checkout action is rather inane as it's clearly very backwards-incompatible.

kratsg:
> his was also done to the Tag Creator GitHub Action, which is complicated and brittle to change [resulting in a break](https://github.com/scikit-hep/pyhf/pull/796/checks?check_run_id=614484399#step:7:1).

It's just unexpected that a stage for checking out the master branch would actually break...

matthewfeickert:
> It's just unexpected that a stage for checking out the master branch would actually break...

Sorry, I realize that it seems like I'm blaming Tag Creator when this is fully the fault of `checkout`. I simply meant that Tag Creator is complicated. If you agree I'll make this a fix PR instead.

matthewfeickert:
@kratsg this should now work given this example of the [patched workflow is done on this test branch here](https://github.com/scikit-hep/pyhf/runs/615758409?check_suite_focus=true#step:7:1).

PR #846, opened 2020-04-27
feat: fixed_vals

cgottard:
## fixed_vals argument in infer functions

I propagated to the `infer` functions the capability of fixing some parameters to a constant the value. The optimizer already implemented this functionality so it was only a matter of interfacing.
I tested the changes running both CLs calculations and MLE fits. The CI succeeds.

_Note_: before these changes it was possible to perform a MLE fit passing the list of fixed parameters via **kwargs. The same was not true for hypotest. I found that adding an explicit function argument for both cases was appropriate for such a common task.

@kratsg, we discussed the use case of these changed via e-mail. Could you review this MR?

## Checklist Before Requesting Reviewer

- [x] Tests are passing

## Before Merging

For the PR Assignees:


tests/test_backend_consistency.py
```diff
@@ -115,7 +115,12 @@ def test_hypotest_q_mu(
         pyhf.set_backend(backend)
 
         q_mu = pyhf.infer.test_statistics.qmu(
-            1.0, data, pdf, pdf.config.suggested_init(), pdf.config.suggested_bounds(),
+            1.0,
+            data,
+            pdf,
+            pdf.config.suggested_init(),
+            pdf.config.suggested_bounds(),
+            None,
```
matthewfeickert, line 123: There should be a test that actually tests the use of `fixed_vals` and not just the `None` case.

lukasheinrich:
Hi @cgottard thanks a lot for this PR. It's been something we wanted to add for some time. I wonder whether we could somehow streamline the API further.  As it is, we do a lot of passing around for these items

* init values
* bounds values
* fixed_values

and perhaps we should be rather moving around `pdfconfig` objects 

```
sb_config = model.make_config()
sb_config.set_poival(1.0)

b_config = model.make_config()
b_config.set_poival(0.0)

pyhf.infer.hypotest(sb_config, data, pdf)
```

what do you think?

src/pyhf/infer/__init__.py
```diff
@@ -17,6 +24,7 @@ def hypotest(
         pdf (~pyhf.pdf.Model): The HistFactory statistical model
         init_pars (Array or Tensor): The initial parameter values to be used for minimization
         par_bounds (Array or Tensor): The parameter value bounds to be used for minimization
+        fixed_vals (list of tuples): Parameters to be held constant and their value
```
kratsg, line 27: need to be careful, since this will be confusing given the existing `poi_test` argument.

kratsg:
I think we should have this in for 0.6.0 because I don't think trying to keep the same API is a good idea.

lukasheinrich:
@cgottard would you be up for shepherding a larger change to the API  that enables this feature?

cgottard:
Dear all,

thanks for the feedback. I was addressing @matthewfeickert's comment about the test but locally pytest is not running as it should and I see the errors only from the CI.

Anyway, I am ok with changing the API as @lukasheinrich suggested. I agree that it makes everything clearer and more transparent. In the meantime I think we can keep this MR open to discuss the progress, then we'll see if we want to close it and create a new one using a new feature branch name.

kratsg:
> Anyway, I am ok with changing the API as @lukasheinrich suggested. I agree that it makes everything clearer and more transparent. In the meantime I think we can keep this MR open to discuss the progress, then we'll see if we want to close it and create a new one using a new feature branch name.

I think this PR should go in as close to how it is, but with tests and coverage as part of 0.5.0 -> 0.6.0 (we don't want very large PRs) and you've pointed out this is a relatively quick change. The API is backwards compatible here with it. Then we change the API in another PR. Having very large PRs in general scares me.

lukasheinrich:
that sounds good to me.

src/pyhf/infer/__init__.py
cgottard: It can be confusing because `fixed_poi_fit` is not strictly necessary anymore. Anyway in `fixed_poi_fit` the mu is fixed via the `fixed_vals` and if additional parameters are set to constant they're added to the list, see https://github.com/cgottard/pyhf/blob/fixed_vals/src/pyhf/infer/mle.py#L56
So there should be no problem nor ambiguity.

cgottard:
Ok, I'll ping you when this is ready. I implemented a new background uncertainty in test_backend_consistency which is then fixed and shifted.
The syntax is correct but I see AssertionErrors that I need to investigate. I'll fix my local pytest first, so to avoid all these useless commits.

cgottard:
I found why the test is failing. The mapping between the parameters defined in the model and those entering the minimization is broken by the gammas. The shapesys appers as a single NP in pdf.config.par_map and is expanded into a NP per bin in a later stage shifting the indices of the following NPs by the number of bins.
I'm looking in how to retrieve the final mapping.

kratsg:
> I found why the test is failing. The mapping between the parameters defined in the model and those entering the minimization is broken by the gammas. The shapesys appers as a single NP in pdf.config.par_map and is expanded into a NP per bin in a later stage shifting the indices of the following NPs by the number of bins.

You can ask for the number of parameters as part of that par_map. E.G.

```
pdf = Model(spec)
pdf.config.param_set('shapesys_name').n_parameters
```

but also additionally just the slicing information from

```
pdf = Model(spec)
pdf.config.par_slice('shapesys_name')
```

The reason the count is different is because there are likely zero-valued uncertainties in some of the shapesys bins.

https://github.com/scikit-hep/pyhf/blob/8fb4f6e54dbd5f2c485014fd5ceb0b16866dbd4f/src/pyhf/pdf.py#L189-L197

cgottard:
Is there a  check that even the commented pseudo-code runs?!

___________________ [doctest] pyhf.infer.test_statistics.qmu ___________________
024         >>> model = pyhf.simplemodels.hepdata_like(
025         ...     signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0]
026         ... )
027         >>> observations = [51, 48]
028         >>> data = pyhf.tensorlib.astensor(observations + model.config.auxdata)
029         >>> test_mu = 1.0
030         >>> init_pars = model.config.suggested_init()
031         >>> par_bounds = model.config.suggested_bounds()
032         >>> fixed_vals = [(2, 0)]
033         >>> pyhf.infer.test_statistics.qmu(test_mu, data, model, init_pars, par_bounds, fixed_vals)
Expected:
    3.938244920380498
Got:
    0.0

kratsg:
> Is there a check that even the commented pseudo-code runs?!

we have doctesting, yes.

cgottard:
The feature is now implemented and is used in `test_backend_consistency.py`.

I added a new NP on the background of `test_backend_consistency.py` and a test parametrisation on it. The fits are repeated with this NP "floating" and fixed to its central value, +1 sigma and -1 sigma. 

I checked the values of all the parameters before and after each fit (fit as in `fixed_poi_fit` and` fit`) and they make sense to me. However in some configurations the difference between backends exceeds the tolerance previously set (1 percent for Numpy/SciPy and 5 permille among Tensors). In particular in one of the cases with 500 bins and no fixed NP Tensorflow returns a value 5% different w.r.t. to the reference backend. In my opinion this discrepancy is unrelated to the new feature (the NP is not even fixed!) and is due to the fact that the model fitted is more complex than before.

I think that we are ready to merge. Any difference between backends can be investigated separately.

tests/test_backend_consistency.py
cgottard: Included in the test, see https://github.com/scikit-hep/pyhf/pull/846#issuecomment-624550955

kratsg:
@cgottard the reason why the tests are failing is because the consistency should be across backends, not across hypotest configurations. By adding the parameterization on fixed values, it breaks the test on purpose since now you're trying to compare the result with fixed values vs the result without -- and they shouldn't really be expected to match within tolerance.

cgottard:
Hi @kratsg
in my understanding the comparison among backends is repeated for each given configuration.
Isn't this the case?

kratsg:
> in my understanding the comparison among backends is repeated for each given configuration.

I see now. This is correct. In this case then, we should not expect the backends to be inconsistent. This likely indicates a bug somewhere. We definitely do not want to lower tolerances.

cgottard:
I think the way to proceed is to first decouple the backend consistency from this MR. I'll checkout the last tag and implement in `test_backend_consistency.py` the statistical model I'm using now. 
If the inconsistency is gone I'll look into my changes, otherwise I'll test different models to see if there's a relation with the kind of NPs defined.

cgottard:
Dear all,

I **cloned** the repo from **master** and run locally the `test_backend_consistency.py`. As it is everything succeds. 
Then I changed the statistical model to match what I have in this MR.
<details>
<summary>Model</summary>

    source = generate_source_static(n_bins)
    bkg_unc_5pc_up = [x + 0.05 * x for x in source['bindata']['bkg']]
    bkg_unc_5pc_dn = [x - 0.05 * x for x in source['bindata']['bkg']]
    signal_sample = {
        'name': 'signal',
        'data': source['bindata']['sig'],
        'modifiers': [{'name': 'mu', 'type': 'normfactor', 'data': None}],
    }

    background_sample = {
        'name': 'background',
        'data': source['bindata']['bkg'],
        'modifiers': [
            {
                'name': 'uncorr_bkguncrt',
                'type': 'shapesys',
                'data': source['bindata']['bkgerr'],
            },
            {
                'name': 'norm_bkgunc',
                'type': 'histosys',
                'data': {'hi_data': bkg_unc_5pc_up, 'lo_data': bkg_unc_5pc_dn,},
            },
        ],
    }


</details>
and pytorch returns a result slightly above tolerance.

`
FAILED test_backend_consistency.py::test_hypotest_q_mu[normal-500_bins] - assert False
FAILED test_backend_consistency.py::test_hypotest_q_mu[inverted-500_bins] - assert False
`
because
`
array([0.00000000e+00, 1.59306155e-03, 1.06477109e-02, 2.60919799e-06]) < 0.01.all
 array([0.00000000e+00, 1.59260676e-03, 1.06472520e-02, 2.14421824e-06]) < 0.01.all
`

Then, I **updated my feature branch to master** and run the test with the fixed values:

`
FAILED tests/test_backend_consistency.py::test_hypotest_q_mu[none-normal-500_bins] - assert False
FAILED tests/test_backend_consistency.py::test_hypotest_q_mu[none-inverted-500_bins] - assert False
`
because

`
array([0.00000000e+00, 1.59306155e-03, 1.06477109e-02, 2.60919799e-06]) < 0.01.all
array([0.00000000e+00, 1.59260676e-03, 1.06472520e-02, 2.14421824e-06]) < 0.01.all
`

Discrepancies among tensors are gone. Possibly because I am using a conda env correctly configured for cuda and TF. And yes, I am running from my branch as I can verify from the stdout and already the "[none-" in the name of the test.

Given the small excess in the tolerance w.r.t 0.01 shall we increase the tolerance to 1.5%?

matthewfeickert:
@cgottard Sorry to have left you hanging here on this. If you have time can you rebase this so that we can try to get this in for `v0.5.2`? If you don't one of the core devs can do it.

kratsg:
Closing in favor of #1051. Thanks a lot @cgottard for the initial push to get this working. I cleaned it up a lot into the new PR and rebased it all.

Will be migrating the tests shortly.

PR #847, opened 2020-04-30
feat: Add patchset schema and datadir pytest fixture

### LINKED ISSUE #631: Develop PatchSet JSON Schema
# Description

distributing a tarball with O(100) files is a bit overwhelming. I think it would be better to define a patch set, which could just be 

```
[
  {
    "metadata": {
        "sbottom_mass": 100,
        "neutralino_mass": 300
    },
    "patch": [
        {"op": "replace"},
        {"op": "replace"}
    ]
  },
  {
    "metadata": {
        "sbottom_mass": 200,
        "neutralino_mass": 400
    },
    "patch": [
        {"op": "replace"},
        {"op": "replace"}
    ]
  }
]
```

to get a specific patch one could do:
```
cat patchset.json|ca|jq '.[]|select(
.metadata.sbottom_mass == 100 and \
.metadata.neutralino_mass == 300
).patch'
```
This way for example the sbottom hepdata tarball would only be 6 files

* RegionA/BkgOnly.json
* RegionA/SignalPatchSet.json
* RegionB/BkgOnly.json
* RegionB/SignalPatchSet.json
* RegionC/BkgOnly.json
* RegionC/SignalPatchSet.json

kratsg:
# Description

Adds the `jsonpatch` schema from http://json.schemastore.org/json-patch and build up the `patchset` schema for encoding all signal patches for a given background-only file. No functionality in `pyhf` is developed yet to handle the `patchset` files.

- Resolves #631 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add jsonpatch schema
* Add patchset schema
* Add tests to ensure schemas behave as expected
* Define fixture for setting up sandboxed, temporary directories of data for tests
```


lukasheinrich:
this looks great. should we at some point add metadata also to the workspace.json schema?

tests/conftest.py
```diff
@@ -138,3 +140,22 @@ def backend(request):
...
+    tests can use them freely.
+    '''
+    # this gets the module name (e.g. /path/to/pyhf/tests/test_schema.py)
+    # and then gets the directory by removing the suffix (e.g. /path/to/pyhf/tests/test_schema)
+    test_dir = pathlib.Path(request.module.__file__).with_suffix('')
+
+    if test_dir.is_dir():
+        distutils.dir_util.copy_tree(test_dir, tmpdir.strpath)
+        # shutil is nicer, but doesn't work: https://bugs.python.org/issue20849
+        # shutil.copytree(test_dir, tmpdir)
```
matthewfeickert, line 159: Maybe I'm missing something, but why doesn't the linked PR fix this?

kratsg:
> this looks great. should we at some point add metadata also to the workspace.json schema?

Which metadata would you be talking about? You mean the concept of "metadata"?

tests/conftest.py
kratsg: I believe it's only in 3.8. My version of 3.7.3 has this error:

```
        if test_dir.is_dir():
>           shutil.copytree(test_dir, tmpdir, dirs_exist_ok=True)
E           TypeError: copytree() got an unexpected keyword argument 'dirs_exist_ok'
```

matthewfeickert: Ah, sorry. You're correct. It only shows up in the `3.8` version of the docs: https://docs.python.org/3.8/library/shutil.html Thanks!

matthewfeickert:
@kratsg As you've marked this for minor release `v0.5.0` do you want this to get moved to draft status until `v0.4.2` is out so that it doesn't get merged in early? Or does it not matter?

kratsg:
> @kratsg As you've marked this for minor release `v0.5.0` do you want this to get moved to draft status until `v0.4.2` is out so that it doesn't get merged in early? Or does it not matter?

I guess earlier would be better (re: hosted docs) for hepdata to start work.

PR #852, opened 2020-05-04
feat: Add utility for computing object digests ("hashing")

kratsg:
# Pull Request Description

NB: an unreleased schema update was renamed `hash` to `digests` to further clarify the usage and unify it across the codebase as part of this PR.

Docs: https://pyhf.readthedocs.io/en/feat-workspacehash

This PR adds the utility functionality `pyhf.utils.digest` which allows for calculating the digest of an arbitrary JSON-serializable object (e.g. a workspace!). This additionally adds a `pyhf digest` command line interface to compute digests on demand for provided input workspaces -- output as both plaintext "codec" as well as JSON (via a flag `-j`).

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add digest utility for computing digest of JSON-serializable objects
* Add workspace digest shortcut from CLI to compute workspace digest
* Update schema to rename hash -> digest for patchsets metadata
```


src/pyhf/workspace.py
```diff
@@ -295,6 +297,16 @@ def __repr__(self):
...
+    @property
+    def hash(self):
+        """
+        A :obj:`dict` mapping the hashing algorithms defined in :attr:`~pyhf.workspace.Workspace.hash_algs` to the computed hash value for the workspace.
+        """
+        return {
+            hash_alg: utils.hash(self, algorithm=hash_alg)
+            for hash_alg in Workspace.hash_algs
+        }
+
```
lukasheinrich, line 309: do we need this to be a method of the workspace. I think this might be used that keeping it a standlone method suffices.

```
pyhf.utils.hash(wspace)
```

kratsg: This provides the quick utility to get something that can plug right in the patchset -- but it's a fair point. I could promote this directly to having a CLI option instead, so that it's easier for @danikam's work.

src/pyhf/utils.py
```diff
@@ -67,3 +68,35 @@ def convert(self, value, param, ctx):
...
+
+    Raises:
+        ValueError: If the object is not JSON-serializable or if the algorithm is not supported.
+
+    Args:
+        obj (`obj`): A JSON-serializable object to compute the digest of. Usually a :class:`~pyhf.workspace.Workspace` object.
+        algorithm (`str`): The hashing algorithm to use.
+
+    Returns:
+        digest (`str`): The digest for the JSON-serialized object provided and hash algorithm specified.
```
matthewfeickert, line 87: Can we add an example here?

src/pyhf/utils.py
```diff
@@ -67,3 +68,35 @@ def convert(self, value, param, ctx):
...
+    """
+
+    try:
+        stringified = json.dumps(obj, sort_keys=True, ensure_ascii=False).encode('utf8')
```
matthewfeickert, line 91: Not super crazy about "stringified" as a name. Would prefer something along the lines of "json_as_str" but either will be fine.

tests/test_scripts.py
```diff
@@ -446,3 +446,43 @@ def test_combine_outfile(tmpdir, script_runner):
...
+def test_workspace_digest(tmpdir, script_runner, algorithms, do_json):
+    results = {
+        'md5': '202eb7615102c35ba86be47eb6fa5e78',
+        'sha256': '7c32ca3b8db75cbafcf5cd7ed4672fa2b1fa69e391c9b89068dd947a521866ec',
+    }
+
+    temp = tmpdir.join("parsed_output.json")
+    command = 'pyhf xml2json validation/xmlimport_input/config/example.xml --basedir validation/xmlimport_input/ --output-file {0:s} --hide-progress'.format(
+        temp.strpath
+    )
```
matthewfeickert, line 464: You had gotten in the habit of using f-strings. Can we keep this going here? I like the adoption! :)

src/pyhf/utils.py
kratsg: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

I like stringified :)

src/pyhf/utils.py
```diff
@@ -67,3 +68,46 @@ def convert(self, value, param, ctx):
             self.fail(
                 '{0:s} is not a valid equal-delimited string'.format(value), param, ctx
             )
+
+
+def digest(obj, algorithm='sha256'):
+    """
+    Get the digest for the provided object. Note: object must be JSON-serializable.
+
+    The hashing algorithms supported are in :module:`hashlib`, part of Python's Standard Libraries.
```
matthewfeickert, line 77: `module:` isn't getting rendered correctly on Read The Docs. :? https://pyhf.readthedocs.io/en/feat-workspacehash/_generated/pyhf.utils.digest.html We can debug this tomorrow.

src/pyhf/utils.py
```diff
@@ -67,3 +68,46 @@ def convert(self, value, param, ctx):
...
+        '2c0633f242928eb55c3672fed5ba8612'
+        >>> pyhf.utils.digest(obj, algorithm='sha1')
+        '49a27f499e763766c9545b294880df277be6f545'
+
+    Raises:
+        ValueError: If the object is not JSON-serializable or if the algorithm is not supported.
+
+    Args:
+        obj (`obj`): A JSON-serializable object to compute the digest of. Usually a :class:`~pyhf.workspace.Workspace` object.
+        algorithm (`str`): The hashing algorithm to use.
```
matthewfeickert, line 95: Maybe list the supported algorithms?

kratsg: There's a whole bunch. That's why I mentioned the hashlib documentation.

src/pyhf/utils.py
kratsg: `:module:` -> `:mod:`

src/pyhf/utils.py
matthewfeickert: Woops. Guess I was tired when I wrote that. I missed that this was using the Python library, and not just what we allow.

PR #873, opened 2020-05-20
chore: Migrate metadata and setup options to setup.cfg

### LINKED ISSUE #872: Move metadata from setup.py to setup.cfg
# Description

@henryiii has documented on the [Scikit-HEP packaging information page](https://scikit-hep.org/developer/packaging#setup-configuration-medium-priority) how one can move from having PyPI metadata be stored in `setup.py` to `setup.cfg`. We've known about this for sometime but haven't taken the time to do it yet, but it is probably worth doing.

@henryiii Can you comment on how to deal with `project_urls`?

https://github.com/scikit-hep/pyhf/blob/3e1f157119dbcb4d8db8ffd8c98e16a2d12d0239/setup.py#L82-L86

matthewfeickert:
# Description

Resolves #872 

Use `setup.cfg` as the main source of configuration and install information by migrating metadata and setup options from `setup.py` to `setup.cfg`. This is done mainly by following @henryiii's examples on the [Scikit-HEP packaging information page](https://scikit-hep.org/developer/packaging#setup-configuration-medium-priority) and by looking at the [gwcelery `setup.cfg` file](https://git.ligo.org/emfollow/gwcelery/-/blob/46374fad07099fcc38cc6e620c6c7fc7afc736b2/setup.cfg) and `setuptools` docs on [Configuring `setup()` using `setup.cfg` files](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files).

The basic argument for doing this is `setup.py` is for logic as it is Python, and `setup.cfg` is for static configuration options.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Move PyPI metadata from setup.py to setup.cfg
* Move setup options from setup.py to setup.cfg
* Remove now unused pathlib.Path and setuptools.find_packages imports
* Remove dependency_links as there are none
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,7 +2,7 @@
 name = pyhf
 version = 0.4.1
 description = (partial) pure Python HistFactory implementation
-long_description = file:README.rst
+long_description = file: README.rst
 long_description_content_type = text/x-rst
 url = https://github.com/scikit-hep/pyhf
 author = Lukas Heinrich, Matthew Feickert, Giordon Stark
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,30 @@
+[metadata]
+name = pyhf
+version = 0.4.1
+description = (partial) pure Python HistFactory implementation
+long_description = file:README.rst
+long_description_content_type = text/x-rst
+url = https://github.com/scikit-hep/pyhf
+author = Lukas Heinrich, Matthew Feickert, Giordon Stark
+author_email = lukas.heinrich@cern.ch, matthew.feickert@cern.ch, gstark@cern.ch
+license = Apache
+license_file = LICENSE
+keywords = physics fitting numpy scipy tensorflow pytorch jax
+project_urls =
+    Documentation = https://scikit-hep.org/pyhf/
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -1,13 +1,56 @@
-[bdist_wheel]
-universal = 1
-
 [metadata]
+name = pyhf
+version = 0.4.1
+description = (partial) pure Python HistFactory implementation
+long_description = file:README.rst
```
henryiii, line 5:
```suggestion
long_description = file: README.rst
```

matthewfeickert: @henryiii I'm happy to change this, but just to make sure I'm not missing anything, this is just style/readability right?

henryiii: Just following style here: https://setuptools.readthedocs.io/en/latest/setuptools.html#id47 probably works without a space, never tried. .cfg is not really a well defined language.

matthewfeickert: > probably works without a space

Yeah, it was passing CI but I'm always happy to follow recommended style. Thanks for the review! :)

PR #875, opened 2020-05-21
fix: Sanitize double quotes for Tag Creator

### LINKED ISSUE #874: Tag creator is failing
# Description

On PR #796 the tag creator is failing to make a new patch release tag. From the logs

https://github.com/scikit-hep/pyhf/pull/796/checks?check_run_id=696984693

this is occurring at the "Comment on issue" stage
https://github.com/scikit-hep/pyhf/blob/2b0f3bd638613f49473c7b8c99873d9cf459b0db/.github/workflows/tag.yml#L88-L95

matthewfeickert:
# Description

Resolves #874 

Tag creator breaks due to

> feat: Add utility for computing object digests ("hashing")

as a result of the `"`s, so simply use `sed` to replace `"` with `'` in the sanitize stage.

```
sed -e 's/"/'"'"'/g'
```

This is not super elegant, but it works. Additionally, a `name` field is added to `github.issues.removeLabel` so that it actually removed the bumpversion label when the CI fails.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Update to use github-script v1.1.0
* Replace double quotes with single quotes to avoid errors
* Add 'name' field to removeLabel to actually remove bumpversion label if tag creator fails
```

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -120,6 +120,7 @@ jobs:
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
+              name: "bumpversion/${{ env['BV_PART'] }}"
             })
 
   always_job:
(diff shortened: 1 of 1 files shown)


matthewfeickert:
cc @andrzejnovak as this is probably of interest.

PR #877, opened 2020-05-23
docs: Add simple examples for inspect, cls, and digest CLI

matthewfeickert:
# Description

Add some simple examples on how to use the `pyhf` CLI for `inspect`, `cls` and `digest`. This is not comprehensive, but just an in passing through PR.

ReadTheDocs build: https://pyhf.readthedocs.io/en/docs-add-inspect-example/cli.html

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add simple example for pyhf inspect, pyhf cls, and pyhf digest CLI
```

### DIFF
--- a/src/pyhf/cli/infer.py
+++ b/src/pyhf/cli/infer.py
@@ -47,7 +47,25 @@ def cls(
     optimizer,
     optconf,
 ):
-    """Compute CLs value(s) for a given pyhf workspace."""
+    """
+    Compute CLs value(s) for a given pyhf workspace.
+
+    Example:
+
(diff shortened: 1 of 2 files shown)


matthewfeickert:
> Only concern as with anything depending on master... if the example is removed in a PR -- we won't catch those changes until after it's merged into master which then breaks on master.

Along with that, as these examples aren't Python code blocks they aren't actually being tested. I'm not sure how to avoid that though. I also don't know how to avoid something like the `master` issue for files that aren't hosted externally.

PR #880, opened 2020-05-27
ci: Use checkout v2.2.0 fetch-depth option to unshallow all

matthewfeickert:
# Description

Use the `fetch-depth: 0` option introduced in [`v2.2.0` of the `checkout` action](https://github.com/actions/checkout/releases/tag/v2.2.0) to fetch all history for all tags and branches in a much cleaner fashion.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use checkout fetch-depth API to unshallow all branches and tags
   - Introduced in v2.2.0
   - c.f. https://github.com/actions/checkout/releases/tag/v2.2.0
* Explicitly set checkout version to v2.2.0 to avoid floating v2 release
   - c.f. https://github.com/actions/toolkit/blob/master/docs/action-versioning.md#recommendations
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,10 +88,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-      # Get history and tags for versioning to work
-    - run: |
-        git fetch --prune --unshallow
-        git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+      with:
+        fetch-depth: 0
     - name: Set up Python 3.7
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -15,10 +15,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-      # Get history and tags for dev versioning to work
-    - run: |
-        git fetch --prune --unshallow
-        git fetch --depth=1 origin +refs/tags/*:refs/tags/*
+      with:
+        fetch-depth: 0
     - name: Set up Python 3.7
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -44,9 +44,8 @@ jobs:
         fi
     - name: Checkout repository
       uses: actions/checkout@v2.2.0
-      # Get history for checkout master to work
-    - run: |
-        git fetch --prune --unshallow
+      with:
+        fetch-depth: 0
     - name: Set up git user
       run: |
         git config --local user.email "action@github.com"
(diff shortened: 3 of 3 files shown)


matthewfeickert:
Something that isn't clear to me about how GitHub Actions does versioning, is that this change was introduced in `v2.2.0` but the docs show that to [fetch all history for all tags and branches](https://github.com/actions/checkout/tree/e52d022eb52c224e5f2201beb687a66849e3b200#fetch-all-history-for-all-tags-and-branches) one would use

```yaml
- uses: actions/checkout@v2
  with:
    fetch-depth: 0
```

So does this mean that using `@v2` will resolve to the last `2.X.Y` tag that was released? This doesn't seem to be well documented.

@ericsciple, if you have time to comment that would be highly appreciated.

matthewfeickert:
@lukasheinrich @kratsg I want to understand this `v2` vs `v2.2.0` issue before merging, but I think that this can still be reviewed now.

matthewfeickert:
Okay, I was able to find [on the GitHub Community Forums](https://github.community/t/referencing-major-version-of-github-action-only-fails-to-download-action/17129/2?u=matthewfeickert) that [the recommendations](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md#recommendations) are that for new (minor or patch releases) to:

> **Make the new release available to those binding to the major version tag**: Move the major version tag (v1, v2, etc.) to point to the ref of the current release. This will act as the stable release for that major version. You should keep this tag updated to the most recent stable minor/patch release.

So for the latest release of the `checkout` action we just need to stay on `v2`.

This is really poorly documented in the GitHub Actions public documentation, but that's sadly the norm it seems.

ericsciple:
@matthewfeickert yes i force pushed the v2 tag, after publishing v2.2.0. It's a convention so not every repo follows the convention. The goal is to be semver-like (get bug fixes)

I can add something to the README to make it more clear. I have seen at least one other user confused by the same thing.

ericsciple:
fwiw, we try to encourage action-authors to adopt the convention ([action-author guidance here](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md))

matthewfeickert:
> @matthewfeickert yes i force pushed the v2 tag, after publishing v2.2.0. It's a convention so not every repo follows the convention. The goal is to be semver-like (get bug fixes)

Thanks very much for this info, @ericsciple, and for taking time to respond. Also, thank you very much for your work on `checkout` &mdash; it is truly fantastic. Thank you also for being so responsive to the requests for the support of unshallowing by default with fetch-detph of `0`. :rocket: 

> I can add something to the README to make it more clear.

To be clear, my frustration with the documentation is not on _your_ actions end, but that it is not better documented in the public documentation. So I don't think it is actually necessary for you to have to add anything to your `README`.

> fwiw, we try to encourage action-authors to adopt the convention ([action-author guidance here](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md))

Thank you also very much for this link, as this is good to know. If this is a convention that the GitHub Actions team is trying to get universally adopted then I can understand that it might be hard to publish public docs that detail specific conventions like this if it is more just a recommendation at this point.

Issue #896, opened 2020-06-16
Add API for switching between GPU and CPU


matthewfeickert:
# Description

Given the desire to be able to test the performance of the CPU and GPU backends, and that by default the TensorFlow, PyTorch, and JAX backends all attempt to run in GPU mode, it would be helpful to add the ability to set the "mode" (if you will) to both the Python and CLI API.

@coolalexzb can probably also give relevant advice on what is needed in this area.

matthewfeickert:
With regards to JAX, it seems that the [recommended way to do this is by setting the `JAX_PLATFORM_NAME` environment variable](https://github.com/google/jax/issues/696#issuecomment-642827038). So an example of this might be

```python
import os
mode = user_input_mode.lower() # "gpu" or "cpu"
os.environ["JAX_PLATFORM_NAME"] = mode
```

but for TensorFlow it might be a bit more complicated as to run in CPU mode I've seen people suggest simply masking the existence of GPUs from the visible devices

```python
import tensorflow as tf
tf.config.set_visible_devices([], "GPU")
```

but this isn't great as if a user has switch from GPU to CPU, we want the user to be able to go back from CPU to GPU and that requires being able to find the GPUs in

```python
tf.config.get_visible_devices()
# example output
# [PhysicalDevice(name='/physical_device:CPU:0', device_type='CPU'),
# PhysicalDevice(name='/physical_device:GPU:0', device_type='GPU')]
```

coolalexzb:
To [test the performance of the CPU and GPU backends](https://github.com/pyhf/pyhf-benchmark), we will firstly use `pyhf.set_backend(backend)` in [run.py](https://github.com/pyhf/pyhf-benchmark/blob/master/scripts/run.py) to set computation backend. From my suggestion, we can add CPU or GPU option as a parameter in `pyhf.set_backend` function and add "set mode" code in `_OptimizerRetriever` class.

coolalexzb:
As you mentioned 

> we want the user to be able to go back from CPU to GPU

Can you show me how users will interact to switch between CPU and GPU @matthewfeickert

coolalexzb:
For PyTorch, from [Easy way to switch between CPU and cuda](https://github.com/pytorch/pytorch/issues/1668)

```
mode = user_input_mode.lower()
if mode == "gpu":
    import torch.cuda as t
else:
    import torch as t

x = t.zeros()
y = Variable(x)
```

[...]

PR #899, opened 2020-06-18
feat: Add CPU&GPU option

coolalexzb:
This PR is used to resolve issue #896. The approach to set CPU and GPU mode is mentioned in #896. For Tensorflow and Pytorch, we can set the tensor over GPU or CPU. As for JAX, the best way to switch between CPU and GPU is to change environment path.  I leave the same interface for numpy backend but it is always in CPU mode.

I also add test script in `tests` directory.


src/pyhf/tensor/pytorch_backend.py
```diff
@@ -118,6 +127,10 @@ def astensor(self, tensor_in, dtype='float'):
             tensor.shape[0]
         except IndexError:
             tensor = tensor.expand(1)
+
+        if self._mode == "CPU":
```
lukasheinrich, line 131: should this be `==GPU`?

coolalexzb: Oh, I made a mistake. I will modify it in the next commit.

coolalexzb:
How can I track this bug in `CI/CD / docs (3.7) (pull_request)`
屏幕快照 2020-06-18 上午11 49 57

coolalexzb:
What is the possible reason I cannot pass CI/CD / test (ubuntu-latest, 3.7) (pull_request)?  It shows that there is an error in `test_paramviewer_order`,  but does not denote any information about which line has problem. It seems an HTTP related error.

src/pyhf/tensor/pytorch_backend.py
```diff
@@ -118,6 +127,10 @@ def astensor(self, tensor_in, dtype='float'):
             tensor.shape[0]
         except IndexError:
             tensor = tensor.expand(1)
+
+        if self._mode == "GPU":
+            if torch.cuda.is_available():
+                tensor.cuda()
```
matthewfeickert, line 133: It looks like this line doesn't actually do anything given that [`torch.Tensor.cuda()`](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.cuda)

> Returns a copy of this object in CUDA memory. If this object is already in CUDA memory and on the correct device, then no copy is performed and the original object is returned.

As an example
```python
import torch
example = torch.rand(3,3)

print(example)
print("First just call it")
if torch.cuda.is_available():
  example.cuda()
  print(example)

print("Now return it")
if torch.cuda.is_available():
  example = example.cuda()
  print(example)
```
returns (something like)
```
tensor([[0.7907, 0.6151, 0.5249],
        [0.9019, 0.4719, 0.4160],
        [0.8475, 0.7123, 0.6323]])
First just call it
tensor([[0.7907, 0.6151, 0.5249],
        [0.9019, 0.4719, 0.4160],
        [0.8475, 0.7123, 0.6323]])
Now return it
tensor([[0.7907, 0.6151, 0.5249],
        [0.9019, 0.4719, 0.4160],
        [0.8475, 0.7123, 0.6323]], device='cuda:0')
```

matthewfeickert:
> What is the possible reason I cannot pass CI/CD / test (ubuntu-latest, 3.7) (pull_request)?

I restarted the CI and it passed.

src/pyhf/tensor/pytorch_backend.py
coolalexzb: Updated!

src/pyhf/__init__.py
```diff
@@ -28,7 +28,7 @@ def get_backend():
 
 
 @events.register('change_backend')
-def set_backend(backend, custom_optimizer=None):
+def set_backend(backend, custom_optimizer=None, mode="GPU"):
```
kratsg, line 31:
```suggestion
def set_backend(backend, custom_optimizer=None, mode="CPU"):
```

I think CPU should be the default. Additionally, we should allow for lowercase `cpu` and `gpu` too. The easiest way to accommodate this is to do something like

```python
if mode not in ['cpu', 'gpu', 'CPU', 'GPU']:
  raise ValueError(f"The mode must be either 'cpu' or 'gpu', not {mode}.")
tensorlib.mode = mode.lower()
```

kratsg: Note that we should only allow `mode` to be used if the backend specified is a string -- because we'll be instantiating the backend for the user. Otherwise, we shouldn't do anything if the user passes in a custom backend.

src/pyhf/tensor/jax_backend.py
```diff
@@ -42,8 +43,19 @@ class jax_backend(object):
...
         self.name = 'jax'
+        self._mode = 'GPU'
+        os.environ["JAX_PLATFORM_NAME"] = self._mode.lower()
         config.update('jax_enable_x64', True)
 
+    @property
+    def mode(self):
+        return self._mode
+
+    @mode.setter
```
kratsg, line 54: To make this easier to deal with -- we shouldn't allow the mode to be changed after initialization. If the user wants to change the mode, they should supply a new backend...

src/pyhf/tensor/numpy_backend.py
```diff
@@ -38,6 +38,15 @@ class numpy_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'numpy'
+        self._mode = 'CPU'
+
+    @property
+    def mode(self):
+        return self._mode
+
+    @mode.setter
```
kratsg, line 47: To make this easier to deal with -- we shouldn't allow the mode to be changed after initialization. If the user wants to change the mode, they should supply a new backend...

src/pyhf/tensor/pytorch_backend.py
```diff
@@ -11,12 +11,21 @@ class pytorch_backend(object):
...
             'float': getattr(torch, kwargs.get('float', 'float32')),
             'int': getattr(torch, kwargs.get('float', 'int32')),
             'bool': torch.bool,
         }
 
+    @property
+    def mode(self):
+        return self._mode
+
+    @mode.setter
```
kratsg, line 25: To make this easier to deal with -- we shouldn't allow the mode to be changed after initialization. If the user wants to change the mode, they should supply a new backend...

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -11,12 +11,21 @@ class tensorflow_backend(object):
...
             'float': getattr(tf, kwargs.get('float', 'float32')),
             'int': getattr(tf, kwargs.get('int', 'int32')),
             'bool': tf.bool,
         }
 
+    @property
+    def mode(self):
+        return self._mode
+
+    @mode.setter
```
kratsg, line 25: To make this easier to deal with -- we shouldn't allow the mode to be changed after initialization. If the user wants to change the mode, they should supply a new backend...

coolalexzb:
It seems that I fail to resolve conflicts in tensorflow_backend.py, what is the possible solution now? Do I need to replace my local `tensorflow_backend.py` with the latest `tensorflow_backend.py`?

src/pyhf/__init__.py
```diff
@@ -101,7 +104,7 @@ def set_backend(backend, custom_optimizer=None):
 
     optimizer_changed = bool(optimizer != new_optimizer)
     # set new backend
-    tensorlib = backend
+    tensorlib = backend(mode=mode.lower())
```
matthewfeickert, line 107: `backend` isn't a callable. This isn't valid code.

matthewfeickert, line 107: `backend` isn't a callable. This isn't valid code.

matthewfeickert:
> What is the possible solution now?

Just rebase your branch against the project `master`. If you called our project `upstream` then on your branch

```
git fetch --all
git rebase upstream/master # Resolve any merge conflicts during the rebase
git push --force-with-lease
```

src/pyhf/tensor/jax_backend.py
```diff
@@ -42,6 +43,8 @@ class jax_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'jax'
+        self._mode = kwargs['mode'] if 'mode' in kwargs else 'cpu'
```
kratsg, line 46:
```suggestion
        self._mode = kwargs.get('mode', 'cpu')
```

src/pyhf/tensor/numpy_backend.py
```diff
@@ -38,6 +38,7 @@ class numpy_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'numpy'
+        self._mode = kwargs['mode'] if 'mode' in kwargs else 'cpu'
```
kratsg, line 41:
```suggestion
        self._mode = kwargs.get('mode', 'cpu')
```

src/pyhf/tensor/pytorch_backend.py
```diff
@@ -11,6 +11,7 @@ class pytorch_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'pytorch'
+        self._mode = kwargs['mode'] if 'mode' in kwargs else 'cpu'
```
kratsg, line 14:
```suggestion
        self._mode = kwargs.get('mode', 'cpu')
```

src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -11,6 +11,7 @@ class tensorflow_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'tensorflow'
+        self._mode = kwargs['mode'] if 'mode' in kwargs else 'cpu'
```
kratsg, line 14:
```suggestion
        self._mode = kwargs.get('mode', 'cpu')
```

src/pyhf/__init__.py
coolalexzb: How can I set the mode variable in the backend class? Use `tensorlib.mode = mode.lower()`?

coolalexzb:
The `set_backend` function in pyhf `__init__.py `, will set a new backend using the following code, which I think is hard to initialize backend with parameters like mode. So, I set `mode` value outside. 
```
    # set new backend
    tensorlib = backend
    tensorlib.mode = mode.lower()
```
Can I add set function in jax_backend since in jax_backend switch by modify env path. My current solution is checking when calling `astensor` function
```
   if os.environ["JAX_PLATFORM_NAME"] != self.mode:
       os.environ["JAX_PLATFORM_NAME"] = self.mode
```

nhartman94:
Hi, is there any progress or status on when to expect this functionality?

PR #902, opened 2020-06-21
fix: Change p-value calculation from 1-cdf(x) to cdf(-x)

### LINKED ISSUE #901: Numerical unstable conversion of significances to p-values
# Description

The conversion of significances to p-values for the asymptotic methods is done by 1 - cdf

https://github.com/scikit-hep/pyhf/blob/73312338ef563305813198e5abc019de46fbbbc2/src/pyhf/infer/calculators.py#L74

This is numerically unstable, since we potentially want to get a very small number by subtracting almost 1 from 1. At a certain point (above ~8 sigma) it will cancel out to 0.

# Expected Behavior

I should get a non-zero p-value.

# Actual Behavior

I got zero.

# Steps to Reproduce

```python
>>> import pyhf
>>> pyhf.__version__ # using current master
'0.4.3'
>>> pyhf.set_backend("numpy")
>>> model = pyhf.simplemodels.hepdata_like(
...     signal_data=[120.0, 110.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0]
... )
>>> observations = [51, 48]
>>> data = pyhf.tensorlib.astensor(observations + model.config.auxdata)
>>> pyhf.infer.hypotest(1.0, data, model)
array([0.])
```

# Checklist

- [X] Run `git fetch` to get the most up to date version of `master`

# Possible fix

For the numpy backend one could use `scipy.stats.norm.sf` which seems to be a more numerically stable implementation. Comparing `1-norm.cdf` with `norm.sf` one can also see how the precision is already degraded before the point where the p-value is cancelled out to completely 0:

```python
>>> import matplotlib.pyplot as plt
>>> import numpy as np
>>> from scipy.stats import norm
>>> sigma = np.arange(0, 10, 0.1)
>>> one_minus_cdf = 1 - norm.cdf(sigma)
>>> sf = norm.sf(sigma)
>>> plt.plot(sigma, np.abs(one_minus_cdf - sf) / sf)
>>> plt.yscale("log")
>>> plt.show()
```
image

nikoladze:
# Description

Resolves #901 

By looking at the [implementation of `scipy.stats.norm.sf`](https://github.com/scipy/scipy/blob/0a7bc723d105288f4b728305733ed8cb3c8feeb5/scipy/stats/_continuous_distns.py#L262)  I noticed we can fix #901 by changing from `1-cdf(x)` to `cdf(-x)` which gives a more numerically stable result for most backends except pytorch. The pytorch function uses `erf`, but i changed that to using `erfc` (see https://en.wikipedia.org/wiki/Q-function#Definition_and_basic_properties) and after that it also seems to produce better results. I wrote it analogous to the [implementation of `torch.Normal.cdf`](
https://github.com/pytorch/pytorch/blob/3bbedb34b9b316729a27e793d94488b574e1577a/torch/distributions/normal.py#L78-L81)

Now, as a side effect it seems one of the validation tests does not pass anymore since the relative difference now is `1.3651304883824116e-06` instead of the required `1e-6`. What to do about that? In theory the values should be calculated more accurately now, so it could be the better match was accidentally?

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use Normal cdf(-x) to compute p-value for improved numerical stability
* Add tests for extremely small p-values
```

### DIFF
--- a/src/pyhf/tensor/pytorch_backend.py
+++ b/src/pyhf/tensor/pytorch_backend.py
@@ -335,7 +335,10 @@ class pytorch_backend(object):
         # we get a more numerically stable variant for low p-values/high significances using erfc(x) := 1 - erf(x)
         # since erf(-x) = -erf(x) we can replace
         # 1 + erf(x) = 1 - erf(-x) = 1 - (1 - erfc(-x)) = erfc(-x)
-        mu, sigma = broadcast_all(mu, sigma)
+        mu, sigma = [
+            torch.as_tensor(_par, dtype=self.dtypemap["float"])
+            for _par in broadcast_all(mu, sigma)
+        ]
         return 0.5 * torch.erfc(-((x - mu) * sigma.reciprocal() / math.sqrt(2)))
 
     def poisson_dist(self, rate):
--- a/src/pyhf/tensor/pytorch_backend.py
+++ b/src/pyhf/tensor/pytorch_backend.py
@@ -324,8 +324,11 @@ class pytorch_backend(object):
         Returns:
             PyTorch FloatTensor: The CDF
         """
-        # the following variant is numerically more stable for low p-values
-        # than the default version that uses torch.erf
+        # the implementation of torch.Normal.cdf uses torch.erf:
+        # 0.5 * (1 + torch.erf((value - self.loc) * self.scale.reciprocal() / math.sqrt(2)))
+        # (see https://github.com/pytorch/pytorch/blob/3bbedb34b9b316729a27e793d94488b574e1577a/torch/distributions/normal.py#L78-L81)
+        # we get a more numerically stable variant for low p-values/high significances
+        # by replacing erf with (1 - erfc) since by definition
         # erfc(x) = 1 - erf(x)
         mu, sigma = broadcast_all(mu, sigma)
         return 0.5 * torch.erfc(-((x - mu) * sigma.reciprocal() / math.sqrt(2)))
--- a/src/pyhf/tensor/pytorch_backend.py
+++ b/src/pyhf/tensor/pytorch_backend.py
@@ -1,7 +1,9 @@
 """PyTorch Tensor Library Module."""
 import torch
 import torch.autograd
+from torch.distributions.utils import broadcast_all
 import logging
+import math
 
 log = logging.getLogger(__name__)
 
@@ -322,8 +324,11 @@ class pytorch_backend(object):
         Returns:
             PyTorch FloatTensor: The CDF
         """
-        normal = torch.distributions.Normal(mu, sigma)
-        return normal.cdf(x)
+        # the following variant is numerically more stable for low p-values
+        # than the default version that uses torch.erf
+        # erfc(x) = 1 - erf(x)
+        mu, sigma = broadcast_all(mu, sigma)
+        return 0.5 * torch.erfc(-((x - mu) * sigma.reciprocal() / math.sqrt(2)))
 
     def poisson_dist(self, rate):
         r"""
--- a/src/pyhf/tensor/pytorch_backend.py
+++ b/src/pyhf/tensor/pytorch_backend.py
@@ -337,10 +337,7 @@ class pytorch_backend(object):
         # we get a more numerically stable variant for low p-values/high significances using erfc(x) := 1 - erf(x)
         # since erf(-x) = -erf(x) we can replace
         # 1 + erf(x) = 1 - erf(-x) = 1 - (1 - erfc(-x)) = erfc(-x)
-        mu, sigma = [
-            torch.as_tensor(_par, dtype=self.dtypemap["float"])
-            for _par in broadcast_all(mu, sigma)
-        ]
+        mu, sigma = broadcast_all(mu, sigma)
         return 0.5 * torch.erfc(-((x - mu) * sigma.reciprocal() / math.sqrt(2)))
 
     def poisson_dist(self, rate):
--- a/src/pyhf/tensor/pytorch_backend.py
+++ b/src/pyhf/tensor/pytorch_backend.py
@@ -327,9 +327,9 @@ class pytorch_backend(object):
         # the implementation of torch.Normal.cdf uses torch.erf:
         # 0.5 * (1 + torch.erf((value - self.loc) * self.scale.reciprocal() / math.sqrt(2)))
         # (see https://github.com/pytorch/pytorch/blob/3bbedb34b9b316729a27e793d94488b574e1577a/torch/distributions/normal.py#L78-L81)
-        # we get a more numerically stable variant for low p-values/high significances
-        # by replacing erf with (1 - erfc) since by definition
-        # erfc(x) = 1 - erf(x)
+        # we get a more numerically stable variant for low p-values/high significances using erfc(x) := 1 - erf(x)
+        # since erf(-x) = -erf(x) we can replace
+        # 1 + erf(x) = 1 - erf(-x) = 1 - (1 - erfc(-x)) = erfc(-x)
         mu, sigma = broadcast_all(mu, sigma)
         return 0.5 * torch.erfc(-((x - mu) * sigma.reciprocal() / math.sqrt(2)))
 
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -176,7 +176,7 @@ def test_calculator_distributions_without_teststatistic(qtilde):
 )
 def test_asymptotic_dist_low_pvalues(backend, nsigma, expected_pval):
     rtol = 1e-8
-    if backend[0].float_precision != '64b':
+    if backend[0].precision != '64b':
         rtol = 1e-5
     dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
     assert np.isclose(np.array(dist.pvalue(nsigma)), expected_pval, rtol=rtol, atol=0)
@@ -184,7 +184,7 @@ def test_asymptotic_dist_low_pvalues(backend, nsigma, expected_pval):
 
 def test_significance_to_pvalue_roundtrip(backend):
     rtol = 1e-15
-    if backend[0].float_precision != '64b':
+    if backend[0].precision != '64b':
         rtol = 1e-6
     sigma = np.arange(0, 10, 0.1)
     dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -2,6 +2,8 @@ import pytest
 import pyhf
 import numpy as np
 import scipy.stats
+import torch
+import tensorflow as tf
 
 
 @pytest.fixture(scope='module')
@@ -172,35 +174,24 @@ def test_calculator_distributions_without_teststatistic(qtilde):
         (9, 1.1285884059538408e-19),
     ],
 )
-@pytest.mark.parametrize(
-    "backend,rtol",
-    [
-        (pyhf.tensor.numpy_backend(), 1e-8),
-        (pyhf.tensor.tensorflow_backend(), 1e-5),
-        (pyhf.tensor.pytorch_backend(), 1e-5),
-        (pyhf.tensor.jax_backend(), 1e-8),
-    ],
-)
-def test_asymptotic_dist_low_pvalues(backend, rtol, nsigma, expected_pval):
-    atol = 0
-    pyhf.set_backend(backend)
+def test_asymptotic_dist_low_pvalues(backend, nsigma, expected_pval):
+    rtol = 1e-8
+    if hasattr(backend[0], "dtypemap") and (
+        backend[0].dtypemap["float"] == torch.float32
+        or backend[0].dtypemap["float"] == tf.float32
+    ):
+        rtol = 1e-5
     dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
-    assert np.isclose(
-        np.array(dist.pvalue(nsigma)), expected_pval, rtol=rtol, atol=atol
-    )
+    assert np.isclose(np.array(dist.pvalue(nsigma)), expected_pval, rtol=rtol, atol=0)
 
 
-@pytest.mark.parametrize(
-    "backend,rtol",
-    [
-        (pyhf.tensor.numpy_backend(), 1e-15),
-        (pyhf.tensor.tensorflow_backend(), 1e-6),
-        (pyhf.tensor.pytorch_backend(), 1e-6),
-        (pyhf.tensor.jax_backend(), 1e-15),
-    ],
-)
-def test_significance_to_pvalue_roundtrip(backend, rtol):
-    pyhf.set_backend(backend)
+def test_significance_to_pvalue_roundtrip(backend):
+    rtol = 1e-15
+    if hasattr(backend[0], "dtypemap") and (
+        backend[0].dtypemap["float"] == torch.float32
+        or backend[0].dtypemap["float"] == tf.float32
+    ):
+        rtol = 1e-6
     sigma = np.arange(0, 10, 0.1)
     dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
     pvalue = dist.pvalue(pyhf.tensorlib.astensor(sigma))
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -176,10 +176,7 @@ def test_calculator_distributions_without_teststatistic(qtilde):
 )
 def test_asymptotic_dist_low_pvalues(backend, nsigma, expected_pval):
     rtol = 1e-8
-    if hasattr(backend[0], "dtypemap") and (
-        backend[0].dtypemap["float"] == torch.float32
-        or backend[0].dtypemap["float"] == tf.float32
-    ):
+    if backend[0].float_precision != '64b':
         rtol = 1e-5
     dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
     assert np.isclose(np.array(dist.pvalue(nsigma)), expected_pval, rtol=rtol, atol=0)
@@ -187,10 +184,7 @@ def test_asymptotic_dist_low_pvalues(backend, nsigma, expected_pval):
 
 def test_significance_to_pvalue_roundtrip(backend):
     rtol = 1e-15
-    if hasattr(backend[0], "dtypemap") and (
-        backend[0].dtypemap["float"] == torch.float32
-        or backend[0].dtypemap["float"] == tf.float32
-    ):
+    if backend[0].float_precision != '64b':
         rtol = 1e-6
     sigma = np.arange(0, 10, 0.1)
     dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -159,13 +159,31 @@ def test_calculator_distributions_without_teststatistic(qtilde):
         calc.distributions(1.0)
 
 
-def test_asymptotic_dist_low_pvalues():
-    rtol = 1e-6
+@pytest.mark.parametrize(
+    "backend",
+    [
+        pyhf.tensor.numpy_backend(),
+        pyhf.tensor.tensorflow_backend(),
+        pyhf.tensor.pytorch_backend(),
+        pyhf.tensor.jax_backend(),
+    ],
+)
+@pytest.mark.parametrize(
+    "nsigma,expected_pval",
+    [
+        # values tabulated using ROOT.RooStats.SignificanceToPValue
+        (5, 2.866515718791945e-07),
+        (6, 9.865876450377018e-10),
+        (7, 1.279812543885835e-12),
+        (8, 6.220960574271829e-16),
+        (9, 1.1285884059538408e-19),
+    ],
+)
+def test_asymptotic_dist_low_pvalues(backend, nsigma, expected_pval):
+    rtol = 1e-5
     atol = 0
+    pyhf.set_backend(backend)
     dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
-    # values tabulated using ROOT.RooStats.SignificanceToPValue
-    assert np.isclose(dist.pvalue(5), 2.866515718791945e-07, rtol=rtol, atol=atol)
-    assert np.isclose(dist.pvalue(6), 9.865876450377018e-10, rtol=rtol, atol=atol)
-    assert np.isclose(dist.pvalue(7), 1.279812543885835e-12, rtol=rtol, atol=atol)
-    assert np.isclose(dist.pvalue(8), 6.220960574271829e-16, rtol=rtol, atol=atol)
-    assert np.isclose(dist.pvalue(9), 1.1285884059538408e-19, rtol=rtol, atol=atol)
+    assert np.isclose(
+        np.array(dist.pvalue(nsigma)), expected_pval, rtol=rtol, atol=atol
+    )
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -172,6 +172,7 @@ def test_calculator_distributions_without_teststatistic(qtilde):
     "nsigma,expected_pval",
     [
         # values tabulated using ROOT.RooStats.SignificanceToPValue
+        # they are consistent with relative difference < 1e-14 with scipy.stats.norm.sf
         (5, 2.866515718791945e-07),
         (6, 9.865876450377018e-10),
         (7, 1.279812543885835e-12),
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -157,3 +157,15 @@ def test_calculator_distributions_without_teststatistic(qtilde):
     )
     with pytest.raises(RuntimeError):
         calc.distributions(1.0)
+
+
+def test_asymptotic_dist_low_pvalues():
+    rtol = 1e-6
+    atol = 0
+    dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
+    # values tabulated using ROOT.RooStats.SignificanceToPValue
+    assert np.isclose(dist.pvalue(5), 2.866515718791945e-07, rtol=rtol, atol=atol)
+    assert np.isclose(dist.pvalue(6), 9.865876450377018e-10, rtol=rtol, atol=atol)
+    assert np.isclose(dist.pvalue(7), 1.279812543885835e-12, rtol=rtol, atol=atol)
+    assert np.isclose(dist.pvalue(8), 6.220960574271829e-16, rtol=rtol, atol=atol)
+    assert np.isclose(dist.pvalue(9), 1.1285884059538408e-19, rtol=rtol, atol=atol)
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -2,8 +2,6 @@ import pytest
 import pyhf
 import numpy as np
 import scipy.stats
-import torch
-import tensorflow as tf
 
 
 @pytest.fixture(scope='module')
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -1,6 +1,7 @@
 import pytest
 import pyhf
 import numpy as np
+import scipy.stats
 
 
 @pytest.fixture(scope='module')
@@ -160,12 +161,12 @@ def test_calculator_distributions_without_teststatistic(qtilde):
 
 
 @pytest.mark.parametrize(
-    "backend",
+    "backend_rtol",
     [
-        pyhf.tensor.numpy_backend(),
-        pyhf.tensor.tensorflow_backend(),
-        pyhf.tensor.pytorch_backend(),
-        pyhf.tensor.jax_backend(),
+        (pyhf.tensor.numpy_backend(), 1e-8),
+        (pyhf.tensor.tensorflow_backend(), 1e-5),
+        (pyhf.tensor.pytorch_backend(), 1e-5),
+        (pyhf.tensor.jax_backend(), 1e-8),
     ],
 )
 @pytest.mark.parametrize(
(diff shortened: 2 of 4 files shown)


lukasheinrich:
Thanks @nikoladze. This is great! Can you add a similar explanatory command to calculators.py (for future reference)

tests/test_infer.py
```diff
@@ -157,3 +157,33 @@ def test_calculator_distributions_without_teststatistic(qtilde):
...
+        pyhf.tensor.numpy_backend(),
+        pyhf.tensor.tensorflow_backend(),
+        pyhf.tensor.pytorch_backend(),
+        pyhf.tensor.jax_backend(),
+    ],
+)
+@pytest.mark.parametrize(
+    "nsigma,expected_pval",
+    [
+        # values tabulated using ROOT.RooStats.SignificanceToPValue
```
matthewfeickert, line 174: ROOT should not be used for calculating anything here. These should be calculated with NumPy/SciPy.

src/pyhf/tensor/pytorch_backend.py
```diff
@@ -322,8 +324,11 @@ def normal_cdf(self, x, mu=0.0, sigma=1.0):
         Returns:
             PyTorch FloatTensor: The CDF
         """
-        normal = torch.distributions.Normal(mu, sigma)
-        return normal.cdf(x)
+        # the following variant is numerically more stable for low p-values
+        # than the default version that uses torch.erf
+        # erfc(x) = 1 - erf(x)
```
matthewfeickert, line 329: Can there be a plot that is added to the PR similar to how was done in Issue #901 that demonstrates the improved stability?

src/pyhf/tensor/pytorch_backend.py
```diff
@@ -322,8 +324,11 @@ def normal_cdf(self, x, mu=0.0, sigma=1.0):
...
+        mu, sigma = broadcast_all(mu, sigma)
+        return 0.5 * torch.erfc(-((x - mu) * sigma.reciprocal() / math.sqrt(2)))
```
matthewfeickert, line 331:
```suggestion
        return 0.5 * torch.erfc(-((x - mu) * sigma.reciprocal() / torch.sqrt(torch.tensor(2.))))
```
Any reason to not just use `torch`'s versions?

tests/test_infer.py
lukasheinrich: my read is that this is for xcheck purposes (i.e. use ROOT to xcheck numpy)

matthewfeickert:
> Now, as a side effect it seems one of the validation tests does not pass anymore since the relative difference now is `1.3651304883824116e-06` instead of the required `1e-6`. What to do about that? 

Should just need to update the values here

https://github.com/scikit-hep/pyhf/blob/73312338ef563305813198e5abc019de46fbbbc2/tests/test_validation.py#L188-L199

tests/test_infer.py
nikoladze: yes, that was the intention

src/pyhf/tensor/pytorch_backend.py
nikoladze: I was roughly following how [torch.Normal.cdf](https://github.com/pytorch/pytorch/blob/3bbedb34b9b316729a27e793d94488b574e1577a/torch/distributions/normal.py#L78-L81) is implemented. They also use `math.sqrt` - i guess it is not nescessary here to create a Torch tensor since this is a constant.

tests/test_infer.py
matthewfeickert: Why are we using ROOT to validate anything related to just numbers and not implementation though? For something like a p-value I don't think that ROOT should be viewed as Truth.

nikoladze:
One way to check this is now more stable is by doing a "roundtrip" (converting the p-value back into significance and see if it is reasonably close to what we started with.

```pycon
>>> import pyhf
>>> import numpy as np
>>> from scipy.stats import norm
>>> import matplotlib.pyplot as plt
>>> sigma = np.arange(0, 10, 0.1)
>>> dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
>>> roundtrip = -norm.ppf(dist.pvalue(sigma))
>>> plt.plot(sigma, np.abs(roundtrip - sigma) / sigma)
>>> plt.yscale("log")
```

Before this PR this looked like this

image

Now it looks like this (linscale)

image

src/pyhf/tensor/pytorch_backend.py
matthewfeickert: Okay that's good to know.

nikoladze:
> Thanks @nikoladze. This is great! Can you add a similar explanatory command to calculators.py (for future reference)

@lukasheinrich  what do you mean by that?

src/pyhf/tensor/pytorch_backend.py
nikoladze: i added a comment below

matthewfeickert: Thanks! It is exactly what I was hoping for. :)

tests/test_infer.py
nikoladze: It was just the first one i looked at that didn't seem to have this issue. Anyway, from a quick look, the low-level code use in that function in [scipy](https://github.com/scipy/scipy/blob/0a7bc723d105288f4b728305733ed8cb3c8feeb5/scipy/special/cephes/ndtri.c#L134) and [ROOT](https://root.cern/doc/master/SpecFuncCephesInv_8cxx_source.html#l00137) seems to be basically the same.

lukasheinrich:
> > Thanks @nikoladze. This is great! Can you add a similar explanatory command to calculators.py (for future reference)
> 
> @lukasheinrich what do you mean by that?

I just meant sth short a la `# computing cdf(-x) instead of 1-cdf(x) for right-tail p-value for improved numerical stability`

tests/test_infer.py
lukasheinrich: I don't think it's inherently bad to take the numbers from ROOT. it's not like we're adding a dependency here.

lukasheinrich:
> > Now, as a side effect it seems one of the validation tests does not pass anymore since the relative difference now is `1.3651304883824116e-06` instead of the required `1e-6`. What to do about that?
> 
> Should just need to update the values here
> 
> https://github.com/scikit-hep/pyhf/blob/73312338ef563305813198e5abc019de46fbbbc2/tests/test_validation.py#L188-L199

yes I think we should change the value. The change here is mathematically sound and should serve as the new reference

tests/test_infer.py
matthewfeickert: It just seems like a strange choice to me given that all our other numbers that we used throughout for testing are taken from NumPy/SciPy and we only use ROOT for validation testing. I doubt there is a difference, but I think we should be consistent and more importantly clear.

nikoladze:
> yes I think we should change the value. The change here is mathematically sound and should serve as the new reference

ok, i can change it - however i'm a bit sceptical since these tests are supposed to validate against ROOT HistFactory?

nikoladze:
> I just meant sth short a la `# computing cdf(-x) instead of 1-cdf(x) for right-tail p-value for improved numerical stability`

is this enough or did you have another place in mind where i could mention this?

https://github.com/scikit-hep/pyhf/pull/902/files#diff-01dfd8b557c51a235c398dcab3b7dcdeR327-R329

lukasheinrich:
> > I just meant sth short a la `# computing cdf(-x) instead of 1-cdf(x) for right-tail p-value for improved numerical stability`
> 
> is this enough or did you have another place in mind where i could mention this?
> 
> https://github.com/scikit-hep/pyhf/pull/902/files#diff-01dfd8b557c51a235c398dcab3b7dcdeR327-R329

I would add it in https://github.com/scikit-hep/pyhf/pull/902/files#diff-89ee98d597780a13ab5b4e79e38d2c88R74 
since people might expect a standard righ-tail pvalue computation and wonder why another form is used.

(I would make a github suggestions but somehow I can't find this feature)

src/pyhf/tensor/pytorch_backend.py
```diff
@@ -322,8 +324,11 @@ def normal_cdf(self, x, mu=0.0, sigma=1.0):
         Returns:
             PyTorch FloatTensor: The CDF
         """
-        normal = torch.distributions.Normal(mu, sigma)
-        return normal.cdf(x)
+        # the following variant is numerically more stable for low p-values
+        # than the default version that uses torch.erf
+        # erfc(x) = 1 - erf(x)
+        mu, sigma = broadcast_all(mu, sigma)
```
kratsg, line 330:
```suggestion
        mu, sigma = self.broadcast(mu, sigma)
```

Not sure what is more appropriate here.

src/pyhf/tensor/pytorch_backend.py
kratsg, line 331: Can you provide a citation for this calculation for reference?

kratsg:
@nikoladze I really appreciate the PR. I think I want to try and summarize/clarify all the discussion about the accuracy by proposing a change to the test if you don't mind doing it.

1) consistency with ROOT(Stats)
2) self-consistency amongst the backends

I'm hoping with option (2), we can reduce the delta to 1e-8 amongst all the backends (ignoring ROOT) or at least keep it at 1e-6. We can then fix (1) in the existing test by bumping it to 1e-5 [that is, that all backends are within 1e-5 of ROOT; but within 1e-8 of each other). That is, `tests/test_infer.py` should be updated to 1e-5 to test against ROOT. Then we add something to `tests/test_backend_consistency.py` that will check consistency in this calculation.

Do you think this is something we could add as a test? You can look at this test to get an idea of what I mean: https://github.com/scikit-hep/pyhf/blob/03b227373b3f893eb501178c3613e0919d5198d8/tests/test_backend_consistency.py#L106-L148

Hopefully, this should be pretty straightforward to implement.

Additionally, I like the idea of round-tripping here and maybe we can add that as a test into `tests/test_infer.py` and assert that `np.abs(roundtrip - sigma) / sigma` is less than `1e-15`. It's a pretty novel approach and we don't have coverage for this sort of testing (roundtripping) in test statistics calculations and I like it...

nikoladze:
@kratsg i think in this case we don't need to separate out the consistency check with ROOT - we can tabulate the hardcoded values from either ROOT or scipy, it doesn't really matter since they under the hood both seem to use the code from the cephes library (see [ROOT implementation,](https://github.com/scipy/scipy/blob/0a7bc723d105288f4b728305733ed8cb3c8feeb5/scipy/special/cephes/ndtri.c#L134) [scipy implementation](https://root.cern/doc/master/SpecFuncCephesInv_8cxx_source.html#l00137)) and when i compare, the numbers seem to match up to ~double precision

```pycon
>>> from scipy.stats import norm
>>> import ROOT
>>> for i in range(10):
...     r = ROOT.RooStats.SignificanceToPValue(i)
...     s = norm.sf(i)
...     print((s - r) / r)
...
0.0
0.0
-9.150136693333652e-16
-1.9276161258069336e-15
-3.637258097853269e-15
-4.247699929554903e-15
-7.336226432840038e-15
0.0
-1.4265779501062304e-14
-7.465913580401231e-15
```

The lower tolerance for the backend consistency comes from the fact that the tensorflow and pytorch backends use single precision (`float32`, `int32`) by default. So either we could a) set that precision to 64 bits for this particular test, b) keep the tolerance at 1e-5 or c) have tolerance 1e-5 for tensorflow and torch and 1e-8 for numpy and jax. I have no clear preference, what do you think?

src/pyhf/tensor/pytorch_backend.py
nikoladze: mhh in principle one could do
```python
mu, sigma = self.simple_broadcast(self.astensor(mu), self.astensor(sigma))
```
I used `broadcast_all`, since it is also done in [torch.Normal](https://github.com/pytorch/pytorch/blob/3bbedb34b9b316729a27e793d94488b574e1577a/torch/distributions/normal.py#L44). There is a slight difference - `broadcast_all` produces scalar Tensors for scalar numbers whereas `self.astensor` produces arrays of length 1. Not sure if this has any impact. The [docstring example for torch.Normal](https://github.com/pytorch/pytorch/blob/3bbedb34b9b316729a27e793d94488b574e1577a/torch/distributions/normal.py#L17) also uses arrays of length 1.

src/pyhf/tensor/pytorch_backend.py
nikoladze: since erf(-x) = -erf(x) and erfc(x) := 1 - erf(x) we can replace 1 + erf(x) by erfc(-x). The rest is just copied from the [implementation of torch.Normal.cdf](https://github.com/pytorch/pytorch/blob/3bbedb34b9b316729a27e793d94488b574e1577a/torch/distributions/normal.py#L78-L81). I expanded a bit on that in the comment.

src/pyhf/tensor/pytorch_backend.py
kratsg: I see benefits in either direction.

- minimizing the necessary interface with a backend makes it easier in the long run to maintain...
- broadcast_all is likely faster/more efficient than our broadcast implementations...

kratsg:
> The lower tolerance for the backend consistency comes from the fact that the tensorflow and pytorch backends use single precision (`float32`, `int32`) by default. So either we could a) set that precision to 64 bits for this particular test, b) keep the tolerance at 1e-5 or c) have tolerance 1e-5 for tensorflow and torch and 1e-8 for numpy and jax.

I would be ok splitting this up based on the 64b/32b tolerance. Let's do that.

Can you add the test for the roundtripping as well? That is a useful one to have to ensure we're still able to roundtrip the pvalue.

tests/test_infer.py
```diff
@@ -160,12 +161,12 @@ def test_calculator_distributions_without_teststatistic(qtilde):
 
 
 @pytest.mark.parametrize(
-    "backend",
+    "backend_rtol",
```
kratsg, line 164:
```suggestion
    "backend,rtol",
```

tests/test_infer.py
```diff
@@ -180,11 +181,30 @@ def test_calculator_distributions_without_teststatistic(qtilde):
         (9, 1.1285884059538408e-19),
     ],
 )
-def test_asymptotic_dist_low_pvalues(backend, nsigma, expected_pval):
-    rtol = 1e-5
+def test_asymptotic_dist_low_pvalues(backend_rtol, nsigma, expected_pval):
```
kratsg, line 184:
```suggestion
def test_asymptotic_dist_low_pvalues(backend, rtol, nsigma, expected_pval):
```

tests/test_infer.py
```diff
@@ -180,11 +181,30 @@ def test_calculator_distributions_without_teststatistic(qtilde):
...
+    backend, rtol = backend_rtol
```
kratsg, line 185:
```suggestion
```

tests/test_infer.py
```diff
@@ -180,11 +181,30 @@ def test_calculator_distributions_without_teststatistic(qtilde):
...
     atol = 0
     pyhf.set_backend(backend)
     dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
     assert np.isclose(
         np.array(dist.pvalue(nsigma)), expected_pval, rtol=rtol, atol=atol
     )
+
+
+@pytest.mark.parametrize(
+    "backend_rtol",
```
kratsg, line 195:
```suggestion
    "backend,rtol",
```

tests/test_infer.py
```diff
@@ -180,11 +181,30 @@ def test_calculator_distributions_without_teststatistic(qtilde):
...
+    [
+        (pyhf.tensor.numpy_backend(), 1e-15),
+        (pyhf.tensor.tensorflow_backend(), 1e-6),
+        (pyhf.tensor.pytorch_backend(), 1e-6),
+        (pyhf.tensor.jax_backend(), 1e-15),
+    ],
+)
+def test_significance_to_pvalue_roundtrip(backend_rtol):
```
kratsg, line 203:
```suggestion
def test_significance_to_pvalue_roundtrip(backend, rtol):
```

tests/test_infer.py
```diff
@@ -180,11 +181,30 @@ def test_calculator_distributions_without_teststatistic(qtilde):
...
+    backend, rtol = backend_rtol
```
kratsg, line 204:
```suggestion
```

tests/test_infer.py
nikoladze: sorry, forgot to mention that - i tried to do it like that before, but it did not work. I get the following error:
```
=========================================================================================== ERRORS ============================================================================================
____________________________________________________________________________ ERROR collecting tests/test_infer.py _____________________________________________________________________________
../../venv/pyhf/lib/python3.8/site-packages/pluggy/hooks.py:286: in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
../../venv/pyhf/lib/python3.8/site-packages/pluggy/manager.py:93: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
../../venv/pyhf/lib/python3.8/site-packages/pluggy/manager.py:84: in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
../../venv/pyhf/lib/python3.8/site-packages/_pytest/python.py:224: in pytest_pycollect_makeitem
    res = list(collector._genfunctions(name, obj))
../../venv/pyhf/lib/python3.8/site-packages/_pytest/python.py:409: in _genfunctions
    self.ihook.pytest_generate_tests(metafunc=metafunc)
../../venv/pyhf/lib/python3.8/site-packages/pluggy/hooks.py:286: in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
../../venv/pyhf/lib/python3.8/site-packages/pluggy/manager.py:93: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
../../venv/pyhf/lib/python3.8/site-packages/pluggy/manager.py:84: in <lambda>
    self._inner_hookexec = lambda hook, methods, kwargs: hook.multicall(
../../venv/pyhf/lib/python3.8/site-packages/_pytest/python.py:133: in pytest_generate_tests
    metafunc.parametrize(*marker.args, **marker.kwargs)
../../venv/pyhf/lib/python3.8/site-packages/_pytest/python.py:966: in parametrize
    newcallspec.setmulti2(
../../venv/pyhf/lib/python3.8/site-packages/_pytest/python.py:841: in setmulti2
    self._checkargnotcontained(arg)
../../venv/pyhf/lib/python3.8/site-packages/_pytest/python.py:825: in _checkargnotcontained
    raise ValueError("duplicate %r" % (arg,))
E   ValueError: duplicate 'backend'
```

For the other test, where i don't have these "stacked" decorators it seems to work

nikoladze: ah - i think i have it now. The order matters - need to apply the two decorators in different order

nikoladze: unfortunately i don't really know why or where to look it up :wink:

kratsg: call it something other than `backend` and you're fine. the problem is we have a global fixture called `backend` that allows us to run a test automatically across all backends by specifying backend as a parameter.

nikoladze: oh - i think then it is maybe better to use that fixture here as well. I did that now - switching the tolerances based on the set dtypes of torch or tensorflow. However, the tests for pytorch64 only pass with the fix from #906.

tests/test_infer.py
nikoladze: i added a comment stating that these numbers are consistent with `scipy.stats.norm.sf` and the other test for the "roundtrip" directly uses scipy

src/pyhf/tensor/pytorch_backend.py
lukasheinrich: I would do broadcast_all

tests/test_infer.py
lukasheinrich: LGTM

tests/test_infer.py
```diff
@@ -172,35 +174,24 @@ def test_calculator_distributions_without_teststatistic(qtilde):
...
-)
-def test_asymptotic_dist_low_pvalues(backend, rtol, nsigma, expected_pval):
-    atol = 0
-    pyhf.set_backend(backend)
+def test_asymptotic_dist_low_pvalues(backend, nsigma, expected_pval):
+    rtol = 1e-8
+    if hasattr(backend[0], "dtypemap") and (
+        backend[0].dtypemap["float"] == torch.float32
+        or backend[0].dtypemap["float"] == tf.float32
+    ):
```
kratsg, line 182: `if backend.name in ["pytorch", "tensorflow"]` is easier to check for fyi (and less imports)

tests/test_infer.py
```diff
@@ -172,35 +174,24 @@ def test_calculator_distributions_without_teststatistic(qtilde):
...
-    ],
-)
-def test_significance_to_pvalue_roundtrip(backend, rtol):
-    pyhf.set_backend(backend)
+def test_significance_to_pvalue_roundtrip(backend):
+    rtol = 1e-15
+    if hasattr(backend[0], "dtypemap") and (
+        backend[0].dtypemap["float"] == torch.float32
+        or backend[0].dtypemap["float"] == tf.float32
+    ):
```
kratsg, line 193: `if backend.name in ["pytorch", "tensorflow"]` is easier to check for fyi (and less imports)

kratsg:
Once the above changes for the tests are in, I think we're ready to merge this in.

tests/test_infer.py
nikoladze: mhh my idea was to check for the dtypes since the achievable precision depends on this. There is for example also a case for pytorch with 64bits in the backend fixture
https://github.com/scikit-hep/pyhf/blob/cb4d37ba89e2aa9b43f6db3e168841b97d7079e6/tests/conftest.py#L82

lukasheinrich: alternatively we should add a dtypemap for all backends (i'm surprised we didn't add this)

nikoladze: there is a mapping for all backends, but only torch and tf have it as an argument in the constructor, numpy and jax have it as an argument for `astensor`. I agree it might make sense to unify it and have a single argument/property saying what precision it should use. Should we do it in this PR or a separate one?

lukasheinrich: i can do it in an upcoming PR. opening an issue

kratsg: opened up #910 to deal with this for @nikoladze . We'll merge that in asap.

tests/test_infer.py
kratsg: needs #910.

kratsg:
@nikoladze , #910 got merged in, which lets you grab the precision information from the backend (you can rely on just `tb.float_precision` for now as #912 will fix this up moving forward.

src/pyhf/tensor/pytorch_backend.py
```diff
@@ -327,8 +329,17 @@ def normal_cdf(self, x, mu=0.0, sigma=1.0):
...
-        normal = torch.distributions.Normal(mu, sigma)
-        return normal.cdf(x)
+        # the implementation of torch.Normal.cdf uses torch.erf:
+        # 0.5 * (1 + torch.erf((value - self.loc) * self.scale.reciprocal() / math.sqrt(2)))
+        # (see https://github.com/pytorch/pytorch/blob/3bbedb34b9b316729a27e793d94488b574e1577a/torch/distributions/normal.py#L78-L81)
+        # we get a more numerically stable variant for low p-values/high significances using erfc(x) := 1 - erf(x)
+        # since erf(-x) = -erf(x) we can replace
+        # 1 + erf(x) = 1 - erf(-x) = 1 - (1 - erfc(-x)) = erfc(-x)
+        mu, sigma = [
+            torch.as_tensor(_par, dtype=self.dtypemap["float"])
```
nikoladze, line 339: added this, since we can't rely on the `torch.set_default_dtype(self.dtypemap["float"])` above since this is changing global state, but global state should only change when `pyhf.set_backend(...)` is called. Maybe we can discuss in a separate issue.

tests/test_validation.py
```diff
@@ -711,7 +711,7 @@ def validate_hypotest(pdf, data, mu_test, expected_result, tolerance=1e-6):
     [
         (setup_1bin_shapesys(), 1e-6),
         (setup_1bin_lumi(), 4e-6),
-        (setup_1bin_normsys(), 1e-6),
+        (setup_1bin_normsys(), 2e-6),
```
nikoladze, line 714: increased the tolerance for now - no idea why the discrepancy increased here. Also in this case the discrepancy does not become significantly lower when using the minuit optimizer ...

lukasheinrich: i think i'd be fine with this increase (or alternatively we re-base it to the new numebers as a reference)

tests/test_infer.py
```diff
@@ -157,3 +160,34 @@ def test_calculator_distributions_without_teststatistic(qtilde):
...
+        (5, 2.866515718791945e-07),
+        (6, 9.865876450377018e-10),
+        (7, 1.279812543885835e-12),
+        (8, 6.220960574271829e-16),
+        (9, 1.1285884059538408e-19),
+    ],
+)
+def test_asymptotic_dist_low_pvalues(backend, nsigma, expected_pval):
+    rtol = 1e-8
+    if backend[0].float_precision != '64b':
```
kratsg, line 179: rebase to latest and this line simplifies further to

```suggestion
    if backend[0].precision != '64b':
```

tests/test_infer.py
```diff
@@ -157,3 +160,34 @@ def test_calculator_distributions_without_teststatistic(qtilde):
...
+        rtol = 1e-5
+    dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0)
+    assert np.isclose(np.array(dist.pvalue(nsigma)), expected_pval, rtol=rtol, atol=0)
+
+
+def test_significance_to_pvalue_roundtrip(backend):
+    rtol = 1e-15
+    if backend[0].float_precision != '64b':
```
kratsg, line 187: rebase to latest and this line simplifies further to

```suggestion
    if backend[0].precision != '64b':
```

tests/test_validation.py
kratsg: I would be wary about just rebasing to the new numbers. What if we added another test with the new numbers and tighter tolerances?

src/pyhf/tensor/pytorch_backend.py
kratsg: I don't understand this to be honest. Can you not use `self.astensor(_par, dtype='float')` ?

tests/test_infer.py
```diff
@@ -1,6 +1,9 @@
 import pytest
 import pyhf
 import numpy as np
+import scipy.stats
+import torch
+import tensorflow as tf
```
kratsg, line 6:
```suggestion
```

src/pyhf/tensor/pytorch_backend.py
nikoladze: unfortunately, it changes when the torch backend is **initialized**. In practice one probably doesn't create multiple backend objects, but it's done for the tests
https://github.com/scikit-hep/pyhf/blob/64dbce070156945f6bb524f79ee6e23d26b52f01/tests/conftest.py#L79-L89
So when a test runs, the global state is set to 64b precision. So also the 32b pytroch test will run with 64b. Even weirder, if it happens that another test that creates a pytorch backend, e.g. [`test_set_tensor_precision`](https://github.com/scikit-hep/pyhf/blob/master/tests/test_tensor.py#L387) runs before then the tests for both pytorch backends will be both 32b (since that was the last one tested there).
So i thought the easiest solution would be not to rely on such a global setting. The alternative would be to build something in that ensures this call is made once the backend is **changed**.

kratsg:
```python
>>> import torch
>>> torch.set_default_dtype(torch.float64)
>>> torch.as_tensor([1.0])
tensor([1.])
>>> torch.as_tensor([1.0]).dtype
torch.float64
>>> torch.set_default_dtype(torch.float32)
>>> torch.as_tensor([1.0]).dtype
torch.float32
```

it seems to be switching for me when the dtype changes for the backend.

```python
>>> import pyhf
>>> pyhf.set_backend(pyhf.tensor.pytorch_backend(precision='32b'))
>>> pyhf.tensorlib.astensor([1.]).dtype
torch.float32
>>> pyhf.set_backend(pyhf.tensor.pytorch_backend(precision='64b'))
>>> pyhf.tensorlib.astensor([1.]).dtype
torch.float64
```

Can you clarify?

nikoladze: The problem occurs when you create the backend objects both before you do the first call to `set_backend`

```pycon
>>> import pyhf
>>> backend64 = pyhf.tensor.pytorch_backend(precision="64b")
>>> backend32 = pyhf.tensor.pytorch_backend(precision="32b")
>>> pyhf.set_backend(backend64)
>>> pyhf.tensorlib.astensor([1.]).dtype
torch.float32
```

kratsg: Ok, I understand now. Can you file an issue with that minimal code example? I know how to fix it.

nikoladze: see #921

matthewfeickert: @nikoladze Can you rebase this off of (for you) `upstream/master`? I think that given the resolution in Issue #921 this should be resolved.

nikoladze: Since the resolution in #922 i merged `upstream/master` and changed back to the simpler version that relies on the pytorch default dtype and should work now.

tests/test_validation.py
kratsg: Everything else looks great. This is the last missing piece.

matthewfeickert: I'm running some tests right now, but I agree with @kratsg. This is a beautiful PR, so once we double check this is needed we're GTG!

matthewfeickert: As we're talking about at `1e-6` change, is it possible that the difference in representation is enough in some evaluation cases to bump this? I think this is fine though.

PR #903, opened 2020-06-23
ci: Separate linting from testing

matthewfeickert:
# Description

Create a separate `lint` extra that can be used to lint the codebase in a separate GitHub Actions workflow then the testing of the codebase. This should allow for faster testing overall, and it also allows for linting to not block development flow until a PR is finished.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add 'lint' extra to setup.py
* Move linting into a separate GitHub Actions workflow
   - Only run lint workflow on pull_request events
* Move MANIFEST check to publish-package workflow
   - Move check-manifest to 'develop' extra
```

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -21,11 +21,14 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.8
-    - name: Install pep517 and twine
+    - name: Install pep517, check-manifest, and twine
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install pep517 --user
-        python -m pip install twine
+        python -m pip install check-manifest twine
+    - name: Check MANIFEST
+      run: |
+        check-manifest
     - name: Build a binary wheel and a source tarball
       run: |
         python -m pep517.build --source --binary --out-dir dist/ .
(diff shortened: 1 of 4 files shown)


matthewfeickert:
The one thing that I've put in here that I'm not sure if it is a good idea or not is that linting only runs on PRs now. This shouldn't be an issue given our workflow, but it is a departure from what we've been doing.

matthewfeickert:
This requires PR #904 to go in first.

.github/workflows/publish-package.yml
```diff
@@ -21,11 +21,14 @@ jobs:
       uses: actions/setup-python@v2
       with:
         python-version: 3.8
-    - name: Install pep517 and twine
+    - name: Install pep517, check-manifest, and twine
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install pep517 --user
-        python -m pip install twine
+        python -m pip install check-manifest twine
```
kratsg, line 28: can we migrate this into a [publish] part of the setup.py?

matthewfeickert: We could, but then we'd also need to install `pyhf` as well to use the extras API with `pip` (right?). Seems a bit unnecessary to install multiple things when we just need 2. Though I guess the argument for it is that another 10 seconds doesn't matter one way or another?

PR #905, opened 2020-06-24
docs: Extend infer docstrings with mathematical examples

### LINKED ISSUE #601: Inconsistency in CLs definition in hypotest
If I look at the [hypotest docs](https://diana-hep.org/pyhf/_generated/pyhf.utils.hypotest.html), it seems to me that what is called `p_b` is incorrect. In the docs just linked to, you can see that `CLb = 1 - p_b = 1 - \int_{-infinity}^{q_obs} f(q|b) dq`, which appears to be defining `p_b` as if it were a _left _tailed__ probability, which it is not. All p-values, including `p_b`, should be defined, I believe, as `p_b = \int_{q_obs}^{+infinity} f(q|b) dq`. 

If I look at the implementation of the `pvals_from_teststat` methods [here](https://github.com/diana-hep/pyhf/blob/d9a9ad13fbb6336a73ce9a6d53a6c044f8807a49/src/pyhf/utils.py#L171),

https://github.com/diana-hep/pyhf/blob/d9a9ad13fbb6336a73ce9a6d53a6c044f8807a49/src/pyhf/utils.py#L171

 the quantity with variable name `CLb` is given as `1 - CDF(q)`, which is actually `p_b` (using the right-tailed p-value definition). And so, the [CLs value](https://github.com/diana-hep/pyhf/blob/d9a9ad13fbb6336a73ce9a6d53a6c044f8807a49/src/pyhf/utils.py#L172) 

https://github.com/diana-hep/pyhf/blob/d9a9ad13fbb6336a73ce9a6d53a6c044f8807a49/src/pyhf/utils.py#L172

should be corrected as `CLs = CLsb / (1-CLb)`, where I use the same variable name of `CLb` as [here](https://github.com/diana-hep/pyhf/blob/d9a9ad13fbb6336a73ce9a6d53a6c044f8807a49/src/pyhf/utils.py#L197), 

https://github.com/diana-hep/pyhf/blob/d9a9ad13fbb6336a73ce9a6d53a6c044f8807a49/src/pyhf/utils.py#L197

which is _not_ what CLb is in the literature, at least not how it is defined when constructing CLs.

I think the implementation in `pvals_from_teststat` is taken from the discussion in the [asymptotics paper](https://arxiv.org/abs/1007.1727) from the discussion based on the Tevatron test statistic, which does not follow the same behavior as the profile likelihood ratio test statistic we use. But who knows, because I argue that Equation 76 abuses notation like nobody's business.

Let me know if I'm misinterpreting anything! I was just going through the [implementation](https://github.com/diana-hep/pyhf/blob/d9a9ad13fbb6336a73ce9a6d53a6c044f8807a49/src/pyhf/utils.py#L171) and the fact that the mathematical form for CLsb and CLb were the same

https://github.com/diana-hep/pyhf/blob/d9a9ad13fbb6336a73ce9a6d53a6c044f8807a49/src/pyhf/utils.py#L170-L171

 caught my eye.

matthewfeickert:
# Description

Resolves #601 

Add mathematical examples/foundations of the use of the infer API in [arXiv:1007.1727](https://arxiv.org/abs/1007.1727) as well as provide additional API examples. This is mainly a port of work started on [`docs/add-docstrings-to-utils`](https://github.com/scikit-hep/pyhf/compare/docs/add-docstrings-to-utils), which was incomplete and had errors.

ReadTheDocs build: https://pyhf.readthedocs.io/en/docs-update-maths-in-infer-docstrings/api.html#inference

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add maths examples to the infer API docstrings
   - Based on relation to equations in https://arxiv.org/abs/1007.1727
* Correct hypotest docstring to use the profile likelihood ratio test statistic
* Add API example to twice_nll docstring
```

### DIFF
--- a/src/pyhf/infer/mle.py
+++ b/src/pyhf/infer/mle.py
@@ -18,8 +18,8 @@ def twice_nll(pars, data, pdf):
 
        \lambda\left(\mu\right) = \frac{L\left(\mu, \hat{\hat{\boldsymbol{\theta}}}\right)}{L\left(\hat{\mu}, \hat{\boldsymbol{\theta}}\right)}
 
-    It serves as the objective function to minimize in ~pyhf.infer.mle.fit and
-    ~pyhf.infer.mle.fixed_poi_fit.
+    It serves as the objective function to minimize in :func:`~pyhf.infer.mle.fit`
+    and :func:`~pyhf.infer.mle.fixed_poi_fit`.
 
     Example:
         >>> import pyhf
--- a/src/pyhf/infer/mle.py
+++ b/src/pyhf/infer/mle.py
@@ -50,7 +50,7 @@ def twice_nll(pars, data, pdf):
 def fit(data, pdf, init_pars=None, par_bounds=None, **kwargs):
     r"""
     Run a unconstrained maximum likelihood fit.
-    This is done by minimizing the objective function of twice the negative log-likelihood
+    This is done by minimizing the objective function :func:`~pyhf.infer.mle.twice_nll`
     of the model parameters given the observed data.
     This is used to produce the maximal likelihood :math:`L\left(\hat{\mu}, \hat{\boldsymbol{\theta}}\right)`
     in the profile likelihood ratio in Equation (7) in :xref:`arXiv:1007.1727`
@@ -102,7 +102,7 @@ def fit(data, pdf, init_pars=None, par_bounds=None, **kwargs):
 def fixed_poi_fit(poi_val, data, pdf, init_pars=None, par_bounds=None, **kwargs):
     r"""
     Run a maximum likelihood fit with the POI value fixed.
-    This is done by minimizing the objective function of twice the negative log-likelihood
+    This is done by minimizing the objective function of :func:`~pyhf.infer.mle.twice_nll`
     of the model parameters given the observed data, for a given fixed value of :math:`\mu`.
     This is used to produce the constrained maximal likelihood for the given :math:`\mu`
     :math:`L\left(\mu, \hat{\hat{\boldsymbol{\theta}}}\right)` in the profile
--- a/src/pyhf/infer/mle.py
+++ b/src/pyhf/infer/mle.py
@@ -4,16 +4,32 @@ from ..exceptions import UnspecifiedPOI
 
 
 def twice_nll(pars, data, pdf):
-    """
-    Twice the negative Log-Likelihood.
+    r"""
+    Two times the negative log-likelihood of the model parameters, :math:`\left(\mu, \boldsymbol{\theta}\right)`, given the observed data
+
+    .. math::
+
+        -2\ln L\left(\mu, \boldsymbol{\theta}\right)
+
+    It is used in the calculation of the test statistic, :math:`t_{\mu}`, as defiend in Equation (8) in :xref:`arXiv:1007.1727`
+
+    .. math::
+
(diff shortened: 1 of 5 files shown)


src/pyhf/infer/mle.py
```diff
@@ -3,23 +3,61 @@
 
 
 def twice_nll(pars, data, pdf):
-    """
-    Twice the negative Log-Likelihood.
+    r"""
+    Two times the negative log-likelihood of the model parameters, :math:`\left(\mu, \boldsymbol{\theta}\right)`, given the observed data.
```
kratsg, line 7: should we mention this is to maximize the likelihood by minimizing the objective function?

matthewfeickert: Hm. We mention below that

```
It serves as the objective function to minimize in :func:`~pyhf.infer.mle.fit` and :func:`~pyhf.infer.mle.fixed_poi_fit`.
```

and those methods' docstrings make it clear that the likelihood is being maximized, so it feels unnecessary to me. What are your thoughts here?

matthewfeickert:
@lukasheinrich @kratsg As this is a lot of formalism changes it is maybe easier to focus on reading the pages changed and look for errors:

- [`pyhf.infer.hypotest`](https://pyhf.readthedocs.io/en/docs-update-maths-in-infer-docstrings/_generated/pyhf.infer.hypotest.html)
- [`pyhf.infer.test_statistics.qmu`](https://pyhf.readthedocs.io/en/docs-update-maths-in-infer-docstrings/_generated/pyhf.infer.test_statistics.qmu.html#pyhf.infer.test_statistics.qmu)
- [`pyhf.infer.test_statistics.qmu_tilde`](https://pyhf.readthedocs.io/en/docs-update-maths-in-infer-docstrings/_generated/pyhf.infer.test_statistics.qmu_tilde.html#pyhf.infer.test_statistics.qmu_tilde)
- [`pyhf.infer.test_statistics.tmu`](https://pyhf.readthedocs.io/en/docs-update-maths-in-infer-docstrings/_generated/pyhf.infer.test_statistics.tmu.html#pyhf.infer.test_statistics.tmu)
- [`pyhf.infer.test_statistics.tmu_tilde`](https://pyhf.readthedocs.io/en/docs-update-maths-in-infer-docstrings/_generated/pyhf.infer.test_statistics.tmu_tilde.html#pyhf.infer.test_statistics.tmu_tilde)

I'll note there is also lots of redundancy in the (re)definition of lambda, but we don't really define the likelihood ratio anywhere by itself.

kratsg:
hypotest seems to be the only page broken for me. We can increase the width of the page via a solution like sphinx_rtd_theme#295.

Screen Shot 2020-08-14 at 7 26 57 AM

matthewfeickert:
> hypotest seems to be the only page broken for me. We can increase the width of the page via a solution like sphinx_rtd_theme#295.

Yeah, I only noticed this when looking at it on RTD. If I build locally things seem fine.

matthewfeickert:
@kratsg @lukasheinrich I've added a note to hypotest as well to mention that we're using the LHC era definition test statistic (which should be obvious, but apparently past Matthew missed this point when giving the CLs definitions).

matthewfeickert:
@dantrim, as you opened up Issue #601 your review is welcome here as well.

PR #906, opened 2020-06-25
fix: make pytorch_backend use configured dtypes

nikoladze:
# Description

This fixes a small typo in the constructor of `pytorch_backend`. The kwargs key for the int type was wrongly queried as "float".

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* typo in pytorch backend where one was unable to specify integer dtype
```

### DIFF
--- a/src/pyhf/tensor/pytorch_backend.py
+++ b/src/pyhf/tensor/pytorch_backend.py
@@ -13,7 +13,7 @@ class pytorch_backend(object):
         self.name = 'pytorch'
         self.dtypemap = {
             'float': getattr(torch, kwargs.get('float', 'float32')),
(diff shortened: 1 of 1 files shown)


nikoladze:
just noticed that in order to actually get pytorch to use the set dtypes  one also needs to run `torch.set_default_dtype(self.dtypemap["float"])`

lukasheinrich:
LGTM! Thanks a lot @nikoladze !

lukasheinrich:
hm codecov seems to be stuck. @matthewfeickert any way we can fix this?

kratsg:
> hm codecov seems to be stuck. @matthewfeickert any way we can fix this?

meh. it's slow, not stuck.

PR #910, opened 2020-06-30
feat: Add precision information to tensorlib backends

### LINKED ISSUE #907: report floating point precision of backends
# Description

tensor backends should have an easy way to report the floating point precisino
of their tensors (when using float tensors) 

This was raised by @nikoladze  in #902

kratsg:
# Description

This provides an API for `tb.xxxx_precision` to report the precision of the `float`/`int`. (I don't think we should mix and match precision for ints/floats so we should revisit how we set them).

This resolves #907.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Provide API for tensorlib precision of ints and floats
```

### DIFF
--- a/src/pyhf/tensor/jax_backend.py
+++ b/src/pyhf/tensor/jax_backend.py
@@ -42,7 +42,8 @@ class jax_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'jax'
-        self.precision = '64b'
+        self.float_precision = '64b'
+        self.int_precision = '64b'
         config.update('jax_enable_x64', True)
 
     def clip(self, tensor_in, min_value, max_value):
--- a/src/pyhf/tensor/jax_backend.py
+++ b/src/pyhf/tensor/jax_backend.py
@@ -42,6 +42,7 @@ class jax_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'jax'
+        self.precision = '64b'
         config.update('jax_enable_x64', True)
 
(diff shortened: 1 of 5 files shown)


tests/test_tensor.py
```diff
@@ -372,3 +372,37 @@ def test_pdf_eval_2(backend):
...
     )
+
+
+def test_tensor_precision(backend):
+    tb, _ = backend
+    assert tb.float_precision in ['32b', '64b']
+    assert tb.int_precision in ['32b', '64b']
+
+
+def test_set_tensor_precision():
```
matthewfeickert, line 383: Can this be parameterized? Or is that not any better here??

kratsg: not 100% sure it's worth it. It'll end up getting reduced massively when we merge int/float into a single precision field.

src/pyhf/tensor/jax_backend.py
```diff
@@ -42,6 +42,8 @@ class jax_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'jax'
+        self.float_precision = '64b'
+        self.int_precision = '64b'
```
lukasheinrich, line 46: do we want to take this opportunity to enable 32-bit jax? Or leave for later?

lukasheinrich:
other than my comment (that i'm fine with either way) this LGTM.

src/pyhf/tensor/jax_backend.py
kratsg: leave for later

Issue #913, opened 2020-07-02
cloudpickle v1.5.0 breaks testing


matthewfeickert:
# Description

With the release of [`cloudpickle` `v1.5.0`](https://pypi.org/project/cloudpickle/1.5.0/) on 2020-07-01 the CI is broken in testing as the following error is raised

```pytb
ImportError while loading conftest '/home/runner/work/pyhf/pyhf/tests/conftest.py'.
tests/conftest.py:83: in <module>
    (pyhf.tensor.tensorflow_backend(), None),
src/pyhf/tensor/__init__.py:44: in __getattr__
    e,
E   pyhf.exceptions.ImportBackendError: ('There was a problem importing TensorFlow. The tensorflow backend cannot be used.', ImportError("cannot import name 'CloudPickler' from 'cloudpickle.cloudpickle' (/opt/hostedtoolcache/Python/3.7.7/x64/lib/python3.7/site-packages/cloudpickle/cloudpickle.py)"))
##[error]Process completed with exit code 4.
```

`cloudpickle` is a required dependency of TensorFlow Probability and in TFP `v0.10.0` it is set to [`cloudpickle >= 1.2.2`](https://github.com/tensorflow/probability/blob/f051e03dd3cc847d31061803c2b31c564562a993/setup.py#L34).

This has been reported in:
- [TensorFlow Probability Issue 991](https://github.com/tensorflow/probability/issues/991)
- [`cloudpickle` Issue 390](https://github.com/cloudpipe/cloudpickle/issues/390)

# Expected Behavior

For no error to be raised

# Actual Behavior

c.f. above

# Steps to Reproduce

This was found in CI, but the minimal test case is just to install TensorFlow and TensorFlow Probability and then try to import TFP:

```
$ python -m pip install tensorflow tensorflow-probability
$ python -c "import tensorflow_probability"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/feickert/.venvs/debug-this/lib/python3.7/site-packages/tensorflow_probability/__init__.py", line 76, in <module>
    from tensorflow_probability.python import *  # pylint: disable=wildcard-import
  File "/home/feickert/.venvs/debug-this/lib/python3.7/site-packages/tensorflow_probability/python/__init__.py", line 23, in <module>
    from tensorflow_probability.python import distributions
[...]

PR #914, opened 2020-07-02
feat: Single precision setting for backends

### LINKED ISSUE #911: Provide 32-bit jax
Do we want to take this opportunity to enable 32-bit jax? Or leave for later?

_Originally posted by @lukasheinrich in https://github.com/scikit-hep/pyhf/pull/910_

### LINKED ISSUE #912: Merge precision configuration of backends into a single config
# Description

Instead of having separate int/float precision configurations -- just have them merged into a single configuration argument.

```
pyhf.set_backend('foo', precision='32b')
```

- #910 
- #911

kratsg:
# Pull Request Description

- Resolves #911
- Resolves #912 

This will unify our backends further by allowing for precision configuration of ints and floats simultaneously, rather than configuring each one separately. Allowing a mix of precisions doesn't quite make sense and we should not encourage this behavior.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Unify setting precision for all tensorlib backends
* Provide 32/64 settings for all backends while keeping same defaults
```

### DIFF
--- a/src/pyhf/tensor/numpy_backend.py
+++ b/src/pyhf/tensor/numpy_backend.py
@@ -38,8 +38,7 @@ class numpy_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'numpy'
-        self.float_precision = '64b'
-        self.int_precision = '64b'
+        self.precision = '64b'
 
     def clip(self, tensor_in, min_value, max_value):
         """
--- a/src/pyhf/tensor/numpy_backend.py
+++ b/src/pyhf/tensor/numpy_backend.py
@@ -38,7 +38,12 @@ class numpy_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'numpy'
-        self.precision = '64b'
+        self.precision = kwargs.get('precision', '64b')
+        self.dtypemap = {
+            'float': np.float64 if self.precision == '64b' else np.float32,
+            'int': np.int64 if self.precision == '64b' else np.int32,
+            'bool': np.bool_,
+        }
 
     def clip(self, tensor_in, min_value, max_value):
         """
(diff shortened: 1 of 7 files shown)


matthewfeickert:
This looks fine, but we should have a PR that temporarily patches Issue #913 before this goes in. I can open that up.

kratsg:
> This looks fine, but we should have a PR that temporarily patches Issue #913 before this goes in. I've opened up PR #915 for this.

rebased and pushed and fixed.

src/pyhf/tensor/numpy_backend.py
```diff
@@ -38,8 +38,12 @@ class numpy_backend(object):
 
     def __init__(self, **kwargs):
         self.name = 'numpy'
-        self.float_precision = '64b'
-        self.int_precision = '64b'
+        self.precision = kwargs.get('precision', '64b')
+        self.dtypemap = {
+            'float': np.float64 if self.precision == '64b' else np.float32,
+            'int': np.int64 if self.precision == '64b' else np.int32,
+            'bool': np.bool_,
```
matthewfeickert, line 45: @kratsg one question: Why is this `np.bool_` (the underscore)?

matthewfeickert: This is just a "Matthew doesn't know" question, not a "this is a problem" question.

kratsg: https://stackoverflow.com/a/6205237/1532974 -- this is a numpy thing.

matthewfeickert: Answer: https://stackoverflow.com/questions/6205020/numpy-types-with-underscore-int-float-etc/6205237#6205237

PR #919, opened 2020-07-03
feat: Enable shell completion

lukasheinrich:
# Description

do we want to add shell completion?

works like this

```
eval "$(pyhf shell-completion)"
pyhf c<tab><tab>
cls      combine
```

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add shell completion with click_completion
* Add tests for the CLI
```


matthewfeickert:
> do we want to add shell completion?

Yes please! :smile:

kratsg:
it would be great to... ... test it somehow(???). Can it be done without an eval (e.g. via pip install.)? https://pip.pypa.io/en/stable/user_guide/#command-completion

matthewfeickert:
Also, is the current PR supposed to work yet? On my machine I get

```
$ git checkout shellcompletion 
$ eval "$(pyhf shell-completion)"
Usage: pyhf [OPTIONS] COMMAND [ARGS]...
Try 'pyhf -h' for help.

Error: No such command 'shell-completion'.
```

lukasheinrich:
ok I added a test (that checks the stdout of the `pyhf shellcomplete bash` command) this reminds me that we should probably generally test the CLI in some ways (without testing the base APIs all over again)

lukasheinrich:
@matthewfeickert I added the setup.py dependencies

src/pyhf/cli/cli.py
```diff
@@ -31,3 +39,18 @@ def pyhf():
 pyhf.add_command(infer.cls)
 
 pyhf.add_command(patchset.cli)
+
+try:
```
lukasheinrich, line 43: should we escape it like this? or have `pyhf shellcomplete` always be there but then say sth like `Install pyhf[shellcomplete]` in case the import fails?

also should this be in a separate file?

setup.py
```diff
@@ -1,6 +1,7 @@
 from setuptools import setup
 
 extras_require = {
+    'shellcomplete': ['click_completion'],
```
kratsg, line 4:
```suggestion
    'autocomplete': ['click_completion'],
```

setup.py
```diff
@@ -26,6 +27,7 @@
         extras_require['backends']
         + extras_require['xmlio']
         + extras_require['contrib']
+        + extras_require['shellcomplete']
```
kratsg, line 30:
```suggestion
        + extras_require['autocomplete']
```

src/pyhf/cli/cli.py
kratsg: separate file I think. A cleaner variant might be

```python
try:
    import click_completion

    click_completion.init()

    from . import shellcomplete
    pyhf.add_command(shellcomplete.cli)

except ImportError:
    pass
```

or something...

setup.py
lukasheinrich: i like shellcomplete better :)

src/pyhf/cli/cli.py
lukasheinrich: done

PR #930, opened 2020-07-08
chore: Future proof release against uproot4 API changes

matthewfeickert:
# Description

As @jpivarski mentioned that `uproot4` is going to have some API changes from the `uproot` `v3` releases, require that the version of `uproot` that is included in the releases until `uproot4` is released for broad use is using the `uproot` `v3` API.

Once `uproot4` is released and stable this can get changed to using

```python
'xmlio': ['uproot~=4.0']
```

to ensure that all future releases of `pyhf` get the `uproot4` API.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Ensure that the next release series of pyhf isn't broken in the future by API changes in uproot4
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ extras_require = {
     ],
     'torch': ['torch~=1.2'],
     'jax': ['jax~=0.1,>0.1.51', 'jaxlib~=0.1,>0.1.33'],
-    'xmlio': ['uproot'],
+    'xmlio': ['uproot<4.0'],  # Future proof against uproot4 API changes
     'minuit': ['iminuit'],
 }
 extras_require['backends'] = sorted(
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ extras_require = {
     ],
     'torch': ['torch~=1.2'],
     'jax': ['jax~=0.1,>0.1.51', 'jaxlib~=0.1,>0.1.33'],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -7,7 +7,7 @@
     ],
     'torch': ['torch~=1.2'],
     'jax': ['jax~=0.1,>0.1.51', 'jaxlib~=0.1,>0.1.33'],
-    'xmlio': ['uproot'],
+    'xmlio': ['uproot~=3.6'],  # Future proof against uproot4 API changes
```
jpivarski, line 10: I think it should be `'uproot<4.0'`. I don't plan on anything other than bug-fixes for the Uproot 3 branch, but in principle there could be an Uproot 3.7.x. The space under 4.0 is squeezed, though: Uproot 3-style API will never be named "4" (or higher).

matthewfeickert: @jpivarski  Hm, I guess I'm missing something. This accomplishes the same thing while establishing a minimum `uproot` release (of `v3.6`), as the version of `uproot` that is installed in CI for this PR is 

```
uproot                 3.11.7
uproot-methods         0.7.4
```

jpivarski: I didn't realize that's what the syntax meant. Okay!

matthewfeickert: Yup! I think I only learned about the it last year (c.f. PR #424) &mdash; super handy!

PR #931, opened 2020-07-08
docs: Add TRExFitter babel translation

kratsg:
# Pull Request Description

This adds information about translating to pyhf from TRExFitter. See https://pyhf.readthedocs.io/en/docs-trexfitter-babel/babel.html#trexfitter for preview.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add documentation on translation from TRExFitter to pyhf
```

### DIFF
--- a/docs/babel.rst
+++ b/docs/babel.rst
@@ -65,3 +65,40 @@ These are all the files you need in order to use `pyhf xml2json <cli.html#pyhf-x
 which will read all of the XML files and load the histogram data from the histogram cache.
 
 The ``HistFactory`` workspace in ``results/`` contains all of the information necessary to rebuild the XML files again. For debugging purposes, the ``pyhf`` developers will often ask for your workspace file, which means ``results/3b_tag21.2.27-1_RW_ExpSyst_36100_multibin_bkg/Excl_combined_DefaultMeasurement.root``. If you want to generate the XML, you can open this file in ``ROOT`` and run ``DefaultMeasurement->PrintXML()`` which puts all of the XML files into the current directory you are in.
+
+
+TRExFitter
+----------
+
+.. note::
+
+    For more details on this section, please refer to the ATLAS-internal `TRExFitter documentation <https://trexfitter-docs.web.cern.ch/trexfitter-docs/advanced_topics/pyhf/>`_.
+
+In order to go from ``TRExFitter`` to ``pyhf``, the good news is that the RooWorkspace files (``XML`` and ``ROOT``) are already made for you. For a given configurastion which looks like
+
+.. code:: yaml
+
+    Job: "pyhf_example"
+    Label: "..."
+
+You can expect some files to be made like so:
+
+- ``pyhf_example/RooStats/pyhf_example.xml``
+- ``pyhf_example/RooStats/pyhf_example_Signal_region.xml``
+- ``pyhf_example/Histograms/pyhf_example_Signal_region_histos.root``
+
+These are all the files you need in order to use `pyhf xml2json <cli.html#pyhf-xml2json>`_. At this point, you could run
+
+.. code:: bash
+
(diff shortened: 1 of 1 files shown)


kratsg:
./cc @alexander-held

docs/babel.rst
```diff
@@ -65,3 +65,40 @@ These are all the files you need in order to use `pyhf xml2json <cli.html#pyhf-x
...
+
+
+TRExFitter
+----------
+
+.. note::
+
+    For more details on this section, please refer to the ATLAS-internal `TRExFitter documentation <https://trexfitter-docs.web.cern.ch/trexfitter-docs/advanced_topics/pyhf/>`_.
+
+In order to go from ``TRExFitter`` to ``pyhf``, the good news is that the RooWorkspace files (``XML`` and ``ROOT``) are already made for you. For a given configurastion which looks like
```
alexander-held: small typo `configurastion`

docs/babel.rst
```diff
@@ -65,3 +65,40 @@ These are all the files you need in order to use `pyhf xml2json <cli.html#pyhf-x
...
+
+.. code:: yaml
+
+    Job: "pyhf_example"
+    Label: "..."
+
+You can expect some files to be made like so:
```
alexander-held, line 84: I think for a `TRExFitter` user this is clear already, but to be very explicit maybe it is a good idea to put 

> (after the `n`/`h` and `w` steps)

after "so". The histograms and xml files are produced at different stages in `TRExFitter` and a user may wonder why they see the histograms already and not the `xml` files yet, this would clarify that they need to run an additional step in that case.
This is a minor point however, since they need to run those steps for a fit with `TRExFitter` anyway, so it is very unlikely they would run into this issue.

docs/babel.rst
```diff
@@ -65,3 +65,40 @@ These are all the files you need in order to use `pyhf xml2json <cli.html#pyhf-x
...
+
+- ``pyhf_example/RooStats/pyhf_example.xml``
+- ``pyhf_example/RooStats/pyhf_example_Signal_region.xml``
+- ``pyhf_example/Histograms/pyhf_example_Signal_region_histos.root``
+
+These are all the files you need in order to use `pyhf xml2json <cli.html#pyhf-xml2json>`_. At this point, you could run
```
matthewfeickert, line 90: It would be nice if the `pyhf xml2json` could be code stylized, but I don't think that and the hyperlink are possible at the same time(?) and the hyperlink is the important part.

docs/babel.rst
```diff
@@ -65,3 +65,40 @@ These are all the files you need in order to use `pyhf xml2json <cli.html#pyhf-x
...
+
+which will read all of the XML files and load the histogram data from the histogram cache.
+
+.. warning::
+
+    There are a few caveats one needs to be aware of with this conversion:
+
+    - Custom parameters cannot be held constant (e.g. lumi), see :pr:`846` and :issue:`739`.
+    - Uncorrelated shape systematics cannot be pruned, see :issue:`662`.
+    - Custom expressions for normalization factors cannot be used, see :issue:`850`.
```
matthewfeickert, line 104:
```suggestion
    - Custom parameters cannot be held constant (e.g. lumi), see PR :pr:`846` and Issue :issue:`739`.
    - Uncorrelated shape systematics cannot be pruned, see Issue :issue:`662`.
    - Custom expressions for normalization factors cannot be used, see Issue :issue:`850`.
```

I think it would be nice to say if things are actually PRs or Issues given that `:issue:` and `:pr:` seem to do the same thing of just adding the `#` and hyperlinking.

docs/babel.rst
matthewfeickert: Oh, seems we were reviewing at the same time. I already got that one. :+1:

docs/babel.rst
kratsg: it's code style afterwards.

PR #937, opened 2020-07-11
docs: Add hepdata_like model API to docs

### LINKED ISSUE #928: Document simplemodels API
# Description

In discussion today with @coolalexzb, I realized that the [`pyhf.simplemodels`](https://github.com/scikit-hep/pyhf/blob/79984be837ef6e53bdd12a82163c34d47d507dba/src/pyhf/simplemodels.py) API is not documented in our docs. Even thought this isn't something we want people to really use, we still show it in our examples and so it needs documentation.

matthewfeickert:
# Description

Resolves #928 

Add `pyhf.simplemodels.hepdata_like` to the docs as it is in the public API and used in many of the `pyhf` examples.

ReadTheDocs build: https://pyhf.readthedocs.io/en/docs-add-simplemodel-docs/_generated/pyhf.simplemodels.hepdata_like.html

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add docstring with examples to simplemodels.hepdata_like
* Add simplemodels.hepdata_like API to docs
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -46,6 +46,7 @@ Making Models from PDFs
    ~workspace.Workspace
    ~patchset.PatchSet
    ~patchset.Patch
+   simplemodels.hepdata_like
 
 Backends
 --------
--- a/src/pyhf/simplemodels.py
+++ b/src/pyhf/simplemodels.py
@@ -2,6 +2,38 @@ from . import Model
 
 
 def hepdata_like(signal_data, bkg_data, bkg_uncerts, batch_size=None):
+    """
+    Construct a simple single channel :class:`~pyhf.pdf.Model` with a
+    :class:`~pyhf.modifiers.shapesys` modifier representing an uncorrelated
+    background uncertainty.
+
+    Example:
+        >>> import pyhf
+        >>> pyhf.set_backend("numpy")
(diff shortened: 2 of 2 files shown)


docs/api.rst
```diff
@@ -46,6 +46,7 @@ Making Models from PDFs
    ~workspace.Workspace
    ~patchset.PatchSet
    ~patchset.Patch
+   simplemodels.hepdata_like
```
matthewfeickert, line 49: The built docs incorrectly classify `hepdata_like` as a class: https://pyhf.readthedocs.io/en/docs-add-simplemodel-docs/_generated/pyhf.simplemodels.hepdata_like.html @kratsg do you know how to fix this? Or should this get moved to the top level API docs?

kratsg: this is because of the template used.

matthewfeickert: @kratsg Okay. Are you okay with the docs being slightly wrong here? Or do you want me to move it someplace else where the `modifierclass.rst` template isn't used?

kratsg: we need to start probably removing the template in some cases.

kratsg: Opened #943 for this.

PR #944, opened 2020-07-14
refactor: Make hypotest return CLs as 0-d tensor

### LINKED ISSUE #714: API: change CLs to be scalar
# Description

right now it's returning (1,) vectors

matthewfeickert:
# Description

Resolves #714

Results in the following behavior for the CLs values

```python
>>> import pyhf
>>> test_mu = 1.0
>>> CLs_obs, CLs_exp = pyhf.infer.hypotest(test_mu, data, model, qtilde=True, return_expected=True)
>>> CLs_obs
array(0.05251554)
>>> CLs_exp
array(0.06445521)
```

previous behavior was

```python
>>> import pyhf
>>> test_mu = 1.0
>>> CLs_obs, CLs_exp = pyhf.infer.hypotest(test_mu, data, model, qtilde=True, return_expected=True)
>>> CLs_obs
array([0.05251554])
>>> CLs_exp
array([0.06445521])
```

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Have hypotest return the CLs as a 0-d tensor
   - Important for fully differential likelihoods
* Update the docs to reflect changes
* Update tests to use return type of 0-d tensor
```

### DIFF
--- a/docs/examples/notebooks/hello-world.ipynb
+++ b/docs/examples/notebooks/hello-world.ipynb
@@ -41,14 +41,16 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Observed: 0.05290116224852556, Expected: 0.06445521290832805\n"
+      "Observed: 0.052515541856109765, Expected: 0.06445521290832805\n"
      ]
     }
    ],
    "source": [
-    "pdf = pyhf.simplemodels.hepdata_like(signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0])\n",
-    "CLs_obs, CLs_exp = pyhf.infer.hypotest(1.0, [51, 48] + pdf.config.auxdata, pdf, return_expected=True)\n",
-    "print('Observed: {}, Expected: {}'.format(CLs_obs, CLs_exp))"
+    "model = pyhf.simplemodels.hepdata_like(signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0])\n",
+    "data = [51, 48] + model.config.auxdata\n",
+    "test_mu = 1.0\n",
+    "CLs_obs, CLs_exp = pyhf.infer.hypotest(test_mu, data, model, qtilde=True, return_expected=True)\n",
+    "print(f\"Observed: {CLs_obs}, Expected: {CLs_exp}\")"
    ]
   },
   {
@@ -67,13 +69,13 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Observed CL_s: 0.05290116224852556, CL_sb: 0.023599998519978738, CL_b: 0.4461149342826869\n"
+      "Observed CL_s: 0.052515541856109765, CL_sb: 0.023324961200974572, CL_b: 0.44415349012077077\n"
      ]
     }
    ],
    "source": [
-    "CLs_obs, p_values = pyhf.infer.hypotest(1.0, [51, 48] + pdf.config.auxdata, pdf, return_tail_probs=True)\n",
-    "print('Observed CL_s: {}, CL_sb: {}, CL_b: {}'.format(CLs_obs, p_values[0], p_values[1]))"
+    "CLs_obs, p_values = pyhf.infer.hypotest(test_mu, data, model, qtilde=True, return_tail_probs=True)\n",
+    "print(f\"Observed CL_s: {CLs_obs}, CL_sb: {p_values[0]}, CL_b: {p_values[1]}\")"
    ]
   },
   {
@@ -120,7 +122,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Observed CL_s: 0.05290116224852556\n",
+      "Observed CL_s: 0.052515541856109765\n",
       "\n",
       "Expected CL_s(-2 σ): 0.0026064088679947964\n",
       "Expected CL_s(-1 σ): 0.013820657528619273\n",
@@ -131,10 +133,10 @@
     }
    ],
    "source": [
-    "CLs_obs, CLs_exp_band = pyhf.infer.hypotest(1.0, [51, 48] + pdf.config.auxdata, pdf, return_expected_set=True)\n",
-    "print('Observed CL_s: {}\\n'.format(CLs_obs))\n",
+    "CLs_obs, CLs_exp_band = pyhf.infer.hypotest(test_mu, data, model, qtilde=True, return_expected_set=True)\n",
+    "print(f\"Observed CL_s: {CLs_obs}\\n\")\n",
     "for p_value, n_sigma in enumerate(np.arange(-2,3)):\n",
-    "    print('Expected CL_s{}: {}'.format('      ' if n_sigma==0 else '({} σ)'.format(n_sigma),CLs_exp_band[p_value]))"
+    "    print(\"Expected CL_s{}: {}\".format(\"      \" if n_sigma==0 else \"({} σ)\".format(n_sigma),CLs_exp_band[p_value]))"
    ]
   }
  ],
--- a/docs/examples/notebooks/hello-world.ipynb
+++ b/docs/examples/notebooks/hello-world.ipynb
@@ -46,10 +46,14 @@
     }
    ],
    "source": [
-    "model = pyhf.simplemodels.hepdata_like(signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0])\n",
+    "model = pyhf.simplemodels.hepdata_like(\n",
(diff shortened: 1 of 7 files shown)


matthewfeickert:
Still need to correct PyTorch's behavior of returning a Tensor.

lukasheinrich:
this will likely require a major version bump

kratsg:
> this will likely require a major version bump

you want this to be v1.0.0? or v0.5.0?

lukasheinrich:
meh as long as we're still 0.X we can do minor bumps I guess

lukasheinrich:
As part of this we probably need to fix this as well

```


import jax
import torch
import tensorflow as tf
print(jax.numpy.asarray(0.1))
pyhf.set_backend('jax')
print(pyhf.tensorlib.astensor(0.1))
print(np.asarray(0.1))
pyhf.set_backend('numpy')
print(pyhf.tensorlib.astensor(0.1))
print(torch.tensor(0.1))
pyhf.set_backend('pytorch')
print(pyhf.tensorlib.astensor(0.1))
print(tf.constant(0.1))
pyhf.set_backend('tensorflow')
print(pyhf.tensorlib.astensor(0.1))
```

```






import jax
import torch
import tensorflow as tf
print(jax.numpy.asarray(0.1))
pyhf.set_backend('jax')
print(pyhf.tensorlib.astensor(0.1))
print(np.asarray(0.1))
pyhf.set_backend('numpy')
print(pyhf.tensorlib.astensor(0.1))
print(torch.tensor(0.1))
pyhf.set_backend('pytorch')
print(pyhf.tensorlib.astensor(0.1))
print(tf.constant(0.1))
pyhf.set_backend('tensorflow')
print(pyhf.tensorlib.astensor(0.1))

0.1
[0.1]
0.1
[0.1]
tensor(0.1000)
tensor([0.1000])
tf.Tensor(0.1, shape=(), dtype=float32)
tf.Tensor([0.1], shape=(1,), dtype=float32)
```

matthewfeickert:
> As part of this we probably need to fix this as well

Ah, I guess so. I think we did this in the past to intentionally enforce uniformity across all backends, but I guess they're all the same now and so enforcing shape isn't helping.

```python
import pyhf
import numpy as np
import torch

print("numpy")
example = np.asarray(0.1)
print(f"example {example} is a {type(example)} with shape {example.shape}")
pyhf.set_backend("numpy")
example = pyhf.tensorlib.astensor(0.1)
print(f"example {example} is a {type(example)} with shape {example.shape}")

print("\ntorch")
example = torch.tensor(0.1)
print(f"example {example} is a {type(example)} with shape {example.shape}")
pyhf.set_backend("numpy")
example = pyhf.tensorlib.astensor(0.1)
print(f"example {example} is a {type(example)} with shape {example.shape}")
```

```
numpy
example 0.1 is a <class 'numpy.ndarray'> with shape ()
example [0.1] is a <class 'numpy.ndarray'> with shape (1,)

torch
example 0.10000000149011612 is a <class 'torch.Tensor'> with shape torch.Size([])
example [0.1] is a <class 'numpy.ndarray'> with shape (1,)
```

So we should probably just fix this first in a separate PR.

matthewfeickert:
@lukasheinrich @kratsg Related to Issue #974 and [my last question in Issue #714](https://github.com/scikit-hep/pyhf/issues/714#issuecomment-657978569), do we want this to return a `0`-d Tensor or do we want a float?

matthewfeickert:
The current state of the PR results in the following

```python
import pyhf
)
test_mu = 1.0
for backend in ["numpy", "jax", "tensorflow", "pytorch"]:
    pyhf.set_backend(backend)
    CLs_obs, CLs_exp = pyhf.infer.hypotest(
        test_mu, data, model, qtilde=True, return_expected=True
    )
    print(
        f"Observed: {CLs_obs} is of type {type(CLs_obs)}, Expected: {CLs_exp} is of type {type(CLs_exp)}"
    )
```
giving
```
Observed: 0.052515541856109765 is of type <class 'numpy.ndarray'>, Expected: 0.06445521290832805 is of type <class 'numpy.ndarray'>
Observed: 0.05251554103025729 is of type <class 'jax.interpreters.xla.DeviceArray'>, Expected: 0.06445520998614704 is of type <class 'jax.interpreters.xla.DeviceArray'>
Observed: 0.05251520499587059 is of type <class 'tensorflow.python.framework.ops.EagerTensor'>, Expected: 0.06445307284593582 is of type <class 'tensorflow.python.framework.ops.EagerTensor'>
Observed: 0.052520569413900375 is of type <class 'torch.Tensor'>, Expected: 0.06446529179811478 is of type <class 'torch.Tensor'>
```

matthewfeickert:
@phinate @WolfgangWaltenberger can you give feedback on if this will be problematic?

phinate:
@matthewfeickert it’s gonna break neos temporarily, but it’s the smallest fix, so I say just go for it. I’m in the middle of a refactor, so I can add this as part of it.
(Well, I guess we’re actually using pyhf@diffable_json, so it won’t break until that gets rebased)

WolfgangWaltenberger:
> @phinate @WolfgangWaltenberger can you give feedback on if this will be problematic?

If with "this" you refer to the change of return types, then no, thats not a problem.

matthewfeickert:
Thanks for the prompt feedback @phinate and @WolfgangWaltenberger. We just wanted to make sure that having the CLs values be 0-d tensors (shape `()`) instead of `(1,)` tensors wouldn't be an issue and make sure that a 0-d tensor is still okay as compared to just going to a float.

matthewfeickert:
> This change in return type should probably result in a release, even though the public API tests still pass.

@lukasheinrich @kratsg Given the feedback from @phinate and @WolfgangWaltenberger, maybe it is okay to not make this its own release and just slip it in with the rest of the PRs going towards `v0.5.2`?

WolfgangWaltenberger:
> Thanks for the prompt feedback @phinate and @WolfgangWaltenberger. We just wanted to make sure that having the CLs values be 0-d tensors (shape `()`) instead of `(1,)` tensors wouldn't be an issue and make sure that a 0-d tensor is still okay as compared to just going to a float.

In SModelS, we have
 
try:
   return float(pyhf_value)
except:
   return pyhf_value[0]


And while purists might frown, this solves any possible issue with the return type. We are pragmatists, after all :)

matthewfeickert:
> And while purists might frown, this solves any possible issue with the return type. We are pragmatists, after all :)

Sounds fine to me. I just like keeping things as tensors as it makes it easier to perform operations on it you want to after the fact, while a 0-d tensor still basically behaves like a float, but I'm more wondering out loud if this is the correct design decision or if we should just say "the CLs is a scalar and so it logically should only ever be a float". I'm happy to hear anyone give an argument that we should abandon 0-d tensors as soon as we return something to the user from the high level APIs like `hypotest` (though keeping 0-d tensors also keeps a parity between `hypotest` and the test statistics).

WolfgangWaltenberger:
> > And while purists might frown, this solves any possible issue with the return type. We are pragmatists, after all :)
> 
> Sounds fine to me. I just like keeping things as tensors as it makes it easier to perform operations on it you want to after the fact, while a 0-d tensor still basically behaves like a float, but I'm more wondering out loud if this is the correct design decision or if we should just say "the CLs is a scalar and so it logically should only ever be a float". I'm happy to hear anyone give an argument that we should abandon 0-d tensors as soon as we return something to the user from the high level APIs like `hypotest` (though keeping 0-d tensors also keeps a parity between `hypotest` and the test statistics).

I think either way, its not a biggie but maybe the pythonic notion of duck-typing is in fact a good guiding principle here? I.e. leave the tensors, but make sure they are always -- with all backends and versions -- castable to floats and thus behave just like floats? My 2 cents, you guys surely have more informed opinions.

matthewfeickert:
@alexander-held Will this affect `cabinetry`?

alexander-held:
Hypothesis tests are not yet interfaced, so no impact on `cabinetry`.

docs/examples/notebooks/hello-world.ipynb
```diff
@@ -120,7 +128,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Observed CL_s: 0.05290116224852556\n",
+      "Observed CL_s: 0.052515541856109765\n",
```
lukasheinrich, line 131: this seems like a comparatively big change of 7e-3.. do we have any indication where this comes from?

matthewfeickert: @lukasheinrich I agree, but that actually isn't the result of this PR. If you launch into the `hello-world` notebook on `master`: [Binder](https://mybinder.org/v2/gh/scikit-hep/pyhf/master?filepath=docs%2Fexamples%2Fnotebooks%2Fhello-world.ipynb)

we can see that this behavior is reproducible there given that `qtilde_True` isn't being used in the original example

qtilde_difference

I think this was introduced in PR #1005 as we can see comparable differences in the docstring examples: https://github.com/scikit-hep/pyhf/pull/1005/files#diff-9f6c3ebe7e5f12b3657aaa7b3949eaf4R80

PR #948, opened 2020-07-14
feat: Type hints

kratsg:
# Description

See:

- https://www.python.org/dev/peps/pep-0484/
- https://www.python.org/dev/peps/pep-0544/
- https://pypi.org/project/typing-extensions/
- https://github.com/scikit-hep/hist/issues/64

`typing-extensions` is used for python 3.6, 3.7 for certain things such as `Protocol` classes.

- [ ] Tests are passing

# Before Merging

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,6 +44,7 @@ install_requires =
     jsonschema>=3.2.0  # for utils
     jsonpatch
     pyyaml  # for parsing CLI equal-delimited options
+    typing_extensions; python_version<"3.8"
 
 [options.packages.find]
 where = src
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,12 @@ classifiers =
 [bdist_wheel]
 universal = 1
 
+[mypy]
+warn_unused_configs = True
+pretty = True
+files = src
+check_untyped_defs = True
+
 [options]
 setup_requires =
     setuptools_scm>=1.15.0
--- a/src/pyhf/tensor/jax_backend.py
+++ b/src/pyhf/tensor/jax_backend.py
@@ -57,6 +57,13 @@ class jax_backend(object):
         Run any global setups for the jax lib.
         """
 
+    @property
+    def tensor(self):
+        """
+        Provide the underlying tensor type.
+        """
+        return np.array
+
     def clip(self, tensor_in, min_value, max_value):
         """
         Clips (limits) the tensor values to be within a specified min and max.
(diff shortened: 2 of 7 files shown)


setup.cfg
```diff
@@ -28,6 +28,12 @@ classifiers =
 [bdist_wheel]
 universal = 1
 
+[mypy]
+warn_unused_configs = True
+pretty = True
+files = src
+check_untyped_defs = True
```
matthewfeickert, line 35: Can these live in `pyproject.toml`?

henryiii: PyTest 6.0 is adding pyproject.toml support. Black I believe always worked either place, IIRC.

henryiii: I suppose you like [PEP 621](https://www.python.org/dev/peps/pep-0621) then :)

src/pyhf/tensor/jax_backend.py
```diff
@@ -57,6 +57,13 @@ def _setup(self):
         Run any global setups for the jax lib.
         """
 
+    @property
+    def tensor(self):
+        """
+        Provide the underlying tensor type.
+        """
+        return np.array
```
matthewfeickert, line 65: I think this should be
```suggestion
        return jax.interpreters.xla.DeviceArray
```

given

```python
>>> import pyhf
>>> pyhf.set_backend("numpy")
>>> x = pyhf.tensorlib.astensor([10.])
>>> print(f"x is a {type(x)}")
x is a <class 'numpy.ndarray'>
>>> pyhf.set_backend("jax")
>>> y = pyhf.tensorlib.astensor([10.])
>>> print(f"y is a {type(y)}")
y is a <class 'jax.interpreters.xla.DeviceArray'>
```

alexander-held:
#1472 provides some motivation for typing of the tensor return values from the `pyhf` API.

mypy.ini
```diff
@@ -0,0 +1,232 @@
...
+[mypy-tests.test_workspace.*]
+check_untyped_defs = False
+disallow_incomplete_defs = False
+disallow_untyped_calls = False
+disallow_untyped_defs = False
+
+
+# Third-party module rule relaxations
+[mypy-iminuit.*]
+ignore_missing_imports = True
```
alexander-held, line 211: As I discovered via @henryiii, these ignores can be specified in a nice and compact manner in `pyproject.toml`, see https://github.com/scikit-hep/hist/pull/230.

PR #951, opened 2020-07-16
refactor: Optimizer API harmonization, minuit autodifferentiation support

### LINKED ISSUE #754: allow minuit to use gradients from autodiff
# Description

minuit has a not often used mode of being able to use externally provided gradients. So we could have optimizers that are

* scipy.optimize.minimize with (jax, tf, torch) and without gradients (numpy)
* minuit  with (jax, tf, torch) and without gradients (numpy)

### LINKED ISSUE #977: Instantiation of Minuit from arrays
# Question

Currently the opt_minuit backend initializes `Minuit` with lots of keyword arguments that need to be constructed first:
https://github.com/scikit-hep/pyhf/blob/c429268e8f1b19e2d6e938f1ff35cddcc5ce29cf/src/pyhf/optimize/opt_minuit.py#L34-L55

There is an alternative available, which allows to use arrays to set the relevant values: [from_array_func](https://iminuit.readthedocs.io/en/stable/reference.html#iminuit.Minuit.from_array_func).

@kratsg suggested there might be interest to use it in `pyhf`. An example that uses `from_array_func` can be found [here](https://github.com/alexander-held/cabinetry/blob/061a2de19adb762e1f9fd31840e3ce8bfae822f4/src/cabinetry/fit.py#L95-L119).

none

kratsg:
# Description

Closes #754. Resolves #977.

As an example of behavior change:

```python
>>> import pyhf
>>> pyhf.get_backend()
(<pyhf.tensor.numpy_backend.numpy_backend object at 0x7f33e6692dc0>, <pyhf.optimize.scipy_optimizer object at 0x7f33e668d460>)
>>> pyhf.set_backend("numpy", "minuit")
>>> pyhf.get_backend()
(<pyhf.tensor.numpy_backend.numpy_backend object at 0x7f339d705c80>, <pyhf.optimize.minuit_optimizer object at 0x7f339d706410>)
>>> # NumPy doesn't support autodiff
>>> pyhf.tensorlib.default_do_grad
False
>>> model = pyhf.simplemodels.hepdata_like([50.0], [100.0], [10])
>>> data = [125.0] + model.config.auxdata
>>> pyhf.infer.mle.fit(data, model, do_grad=False)
array([0.50004936, 1.00000438])
>>> # This will fail given that pyhf.tensorlib.default_do_grad == False for NumPy
>>> pyhf.infer.mle.fit(data, model, do_grad=True)
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/infer/mle.py", line 61, in fit
    return opt.minimize(twice_nll, data, pdf, init_pars, par_bounds, **kwargs)
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/optimize/mixins.py", line 133, in minimize
    do_stitch=do_stitch,
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/optimize/common.py", line 119, in shim
    'do_stitch': do_stitch,
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/optimize/opt_numpy.py", line 25, in wrap_objective
    raise exceptions.Unsupported("Numpy does not support autodifferentiation.")
pyhf.exceptions.Unsupported: Numpy does not support autodifferentiation.
>>> pyhf.set_backend("pytorch", "minuit")
>>> pyhf.tensorlib.default_do_grad
True
>>> pyhf.infer.mle.fit(data, model) # gradient is passed in by default
tensor([0.5018, 0.9998])
>>> pyhf.infer.mle.fit(data, model, do_grad=True)
tensor([0.5018, 0.9998])
>>> pyhf.infer.mle.fit(data, model, do_grad=False)
tensor([0.9685, 0.9171]) # bad result, but more due to 32b precision
>>> pyhf.set_backend(pyhf.tensor.pytorch_backend(precision="64b"))
>>> pyhf.get_backend()
(<pyhf.tensor.pytorch_backend.pytorch_backend object at 0x7fbe659b9f00>, <pyhf.optimize.scipy_optimizer object at 0x7fbe65951c30>)
>>> pyhf.tensorlib.precision
'64b'
>>> pyhf.infer.mle.fit(data, model)
tensor([0.5000, 1.0000])
>>> pyhf.infer.mle.fit(data, model, do_grad=False)
tensor([0.5000, 1.0000])
>>> pyhf.set_backend("numpy", "scipy")
>>> pyhf.infer.mle.fit(data, model)
array([0.49998815, 0.9999697 ])
```

Docs are built here: https://pyhf.readthedocs.io/en/feat-autogradminuit/

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Rewrite/reorganize the optimization functionality
* Individual tensor backend optimizers are dropped in favor of shims passed to the optimizer object
* Change the API for optimizer to align more closely with how tensor backends are set/used
* Allow for configurability of enabling stitching values, using gradients
* Use the from_array_func initialization of Minuit
* Add intersphinx_mapping between projects for documentation
```

### DIFF
--- a/src/pyhf/__init__.py
+++ b/src/pyhf/__init__.py
@@ -14,8 +14,11 @@ def get_backend():
 
     Example:
         >>> import pyhf
-        >>> pyhf.get_backend()
-        (<pyhf.tensor.numpy_backend.numpy_backend object at 0x...>, <pyhf.optimize.scipy_optimizer object at 0x...>)
+        >>> backend, optimizer = pyhf.get_backend()
+        >>> backend
+        <pyhf.tensor.numpy_backend.numpy_backend object at 0x...>
+        >>> optimizer
+        <pyhf.optimize.scipy_optimizer object at 0x...>
 
     Returns:
         backend, optimizer
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -49,7 +49,7 @@ def shim(
     Args:
         objective (`func`): objective function
         data (`list`): observed data
-        pdf (`pyhf.pdf.Model`): model
+        pdf (~pyhf.pdf.Model): model
         init_pars (`list`): initial parameters
         par_bounds (`list`): parameter boundaries
         fixed_vals (`list`): fixed parameter values
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -75,7 +75,7 @@ def shim(
         ``do_stitch`` will modify the ``init_pars``, ``par_bounds``, and ``fixed_vals`` by stripping away the entries associated with fixed parameters. The parameters can be stitched back in via ``stitch_pars``.
 
     Returns:
-        minimizer_kwargs (`dict`): arguments to pass to a minimizer following the ``scipy.optimize.minimize`` API (see notes)
+        minimizer_kwargs (`dict`): arguments to pass to a minimizer following the :func:`scipy.optimize.minimize` API (see notes)
         stitch_pars (`func`): callable that stitches fixed parameters into the unfixed parameters
     """
     tensorlib, _ = get_backend()
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -47,19 +47,29 @@ def shim(
     Prepare Minimization for Optimizer.
 
     Args:
-        objective: objective function
-        data: observed data
-        pdf: model
-        init_pars: initial parameters
-        par_bounds: parameter boundaries
-        fixed_vals: fixed parameter values
+        objective (`func`): objective function
+        data (`list`): observed data
+        pdf (`pyhf.pdf.Model`): model
+        init_pars (`list`): initial parameters
+        par_bounds (`list`): parameter boundaries
+        fixed_vals (`list`): fixed parameter values
+
+    .. note::
+
+        ``stitch_pars(pars, stitch_with=None)`` is a callable that will
+        stitch the fixed parameters of the minimization back into the unfixed
+        parameters.
+
+    .. note::
+
+        ``do_stitch`` will modify the ``init_pars`` and ``par_bounds`` by stripping away the entries associated with fixed parameters. The parameters can be stitched back in via ``stitch_pars``.
 
     Returns:
-        tv: tensor viewer
-        fixed_values_tensor: constant parameters in the fit
-        func: tensor backend wrapped function,gradient pair
-        variable_init: initializations for minimizer
-        variable_bounds: bounds for minimizer
+        stitch_pars (`func`): callable that stitches fixed parameters into the unfixed parameters
+        wrapped_objective (`func`): backend-wrapped ``objective`` function
+        jac (`func`) callable that accepts same parameters as the input ``objective`` but returns the gradient
+        variable_init (`list`): modified initializations for minimizer
+        variable_bounds (`list`): modified bounds for minimizer
     """
     tensorlib, _ = get_backend()
 
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -110,17 +110,10 @@ def shim(
         objective, tensorlib.astensor(data), pdf, stitch_pars, do_grad=do_grad,
     )
 
-    if do_grad:
-        wrapped_objective = lambda pars: objective_and_grad(pars)[0]
-        jac = lambda pars: objective_and_grad(pars)[1]
-    else:
-        wrapped_objective = objective_and_grad
-        jac = None
-
     minimizer_kwargs = dict(
-        func=wrapped_objective,
+        func=objective_and_grad,
         x0=variable_init,
-        jac=jac,
+        jac=do_grad,
         bounds=variable_bounds,
         fixed_vals=minimizer_fixed_vals,
     )
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -58,9 +58,9 @@ def shim(
 
         ``minimizer_kwargs`` is a dictionary containing
 
-          - ``func`` (`func`): backend-wrapped ``objective`` function
+          - ``func`` (`func`): backend-wrapped ``objective`` function (potentially with gradient)
           - ``x0`` (`list`):  modified initializations for minimizer
-          - ``jac`` (`func`): callable that accepts same parameters as the input ``objective`` but returns the gradient
+          - ``do_grad`` (`bool`): whether or not gradient is used
           - ``bounds`` (`list`): modified bounds for minimizer
           - ``fixed_vals`` (`list`): modified fixed values for minimizer
 
@@ -113,7 +113,7 @@ def shim(
     minimizer_kwargs = dict(
         func=objective_and_grad,
         x0=variable_init,
-        jac=do_grad,
+        do_grad=do_grad,
         bounds=variable_bounds,
         fixed_vals=minimizer_fixed_vals,
     )
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -49,7 +49,7 @@ def shim(
     Args:
         objective (`func`): objective function
         data (`list`): observed data
-        pdf (~pyhf.pdf.Model): model
+        pdf (~pyhf.pdf.Model): The statistical model adhering to the schema model.json
         init_pars (`list`): initial parameters
         par_bounds (`list`): parameter boundaries
         fixed_vals (`list`): fixed parameter values
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -63,9 +63,8 @@ def shim(
     """
     tensorlib, _ = get_backend()
 
-    all_init = default_backend.astensor(init_pars)
-
     if do_stitch:
+        all_init = default_backend.astensor(init_pars)
         all_idx = default_backend.astensor(range(pdf.config.npars), dtype='int')
 
         fixed_vals = fixed_vals or []
@@ -73,7 +72,7 @@ def shim(
         fixed_idx = [x[0] for x in fixed_vals]
 
         variable_idx = [x for x in all_idx if x not in fixed_idx]
-        variable_init = all_init[variable_idx]
+        variable_init = default_backend.tolist(all_init[variable_idx])
         variable_bounds = [par_bounds[i] for i in variable_idx]
 
         tv = _TensorViewer([fixed_idx, variable_idx])
@@ -82,12 +81,12 @@ def shim(
     else:
         tv = None
         fixed_values_tensor = tensorlib.astensor([], dtype='float')
-        variable_init = all_init
+        variable_init = init_pars
         variable_bounds = par_bounds
         build_pars = lambda pars: pars
 
-    func = _get_tensor_shim()(
+    wrapped_objective = _get_tensor_shim()(
         objective, tensorlib.astensor(data), pdf, build_pars, do_grad=do_grad,
     )
 
-    return tv, fixed_values_tensor, func, variable_init, variable_bounds
+    return tv, fixed_values_tensor, wrapped_objective, variable_init, variable_bounds
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -33,7 +33,16 @@ def _get_tensor_shim():
     raise ValueError(f'No optimizer shim for {tensorlib.name}.')
 
 
-def shim(objective, data, pdf, init_pars, par_bounds, fixed_vals=None, do_grad=False):
+def shim(
+    objective,
+    data,
+    pdf,
+    init_pars,
+    par_bounds,
+    fixed_vals=None,
+    do_grad=False,
+    do_stitch=False,
+):
     """
     Prepare Minimization for Optimizer.
 
@@ -54,32 +63,31 @@ def shim(objective, data, pdf, init_pars, par_bounds, fixed_vals=None, do_grad=F
     """
     tensorlib, _ = get_backend()
 
-    all_idx = default_backend.astensor(range(pdf.config.npars), dtype='int')
     all_init = default_backend.astensor(init_pars)
 
-    fixed_vals = fixed_vals or []
-    fixed_values = [x[1] for x in fixed_vals]
-    fixed_idx = [x[0] for x in fixed_vals]
+    if do_stitch:
+        all_idx = default_backend.astensor(range(pdf.config.npars), dtype='int')
 
-    variable_idx = [x for x in all_idx if x not in fixed_idx]
-    variable_init = all_init[variable_idx]
-    variable_bounds = [par_bounds[i] for i in variable_idx]
+        fixed_vals = fixed_vals or []
+        fixed_values = [x[1] for x in fixed_vals]
+        fixed_idx = [x[0] for x in fixed_vals]
 
-    tv = _TensorViewer([fixed_idx, variable_idx])
+        variable_idx = [x for x in all_idx if x not in fixed_idx]
+        variable_init = all_init[variable_idx]
+        variable_bounds = [par_bounds[i] for i in variable_idx]
 
-    data = tensorlib.astensor(data)
-    fixed_values_tensor = tensorlib.astensor(fixed_values, dtype='float')
+        tv = _TensorViewer([fixed_idx, variable_idx])
+        fixed_values_tensor = tensorlib.astensor(fixed_values, dtype='float')
+        build_pars = lambda pars: tv.stitch([fixed_values_tensor, pars])
+    else:
+        tv = None
+        fixed_values_tensor = tensorlib.astensor([], dtype='float')
+        variable_init = all_init
+        variable_bounds = par_bounds
+        build_pars = lambda pars: pars
 
-    # NB: need to pass in fixed_idx, variable_idx for jax to jit correctly
     func = _get_tensor_shim()(
-        objective,
-        data,
-        pdf,
-        tv,
-        fixed_values_tensor,
-        fixed_idx=fixed_idx,
-        variable_idx=variable_idx,
-        do_grad=do_grad,
+        objective, tensorlib.astensor(data), pdf, build_pars, do_grad=do_grad,
     )
 
     return tv, fixed_values_tensor, func, variable_init, variable_bounds
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -101,7 +101,6 @@ def shim(
             return tv.stitch([tb.astensor(fixed_values, dtype='float'), pars])
 
     else:
-        tv = None
         variable_init = init_pars
         variable_bounds = par_bounds
         minimizer_fixed_vals = fixed_vals
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -63,9 +63,8 @@ def shim(
     """
     tensorlib, _ = get_backend()
 
-    all_init = default_backend.astensor(init_pars)
-
     if do_stitch:
+        all_init = default_backend.astensor(init_pars)
         all_idx = default_backend.astensor(range(pdf.config.npars), dtype='int')
 
         fixed_vals = fixed_vals or []
@@ -73,21 +72,30 @@ def shim(
         fixed_idx = [x[0] for x in fixed_vals]
 
         variable_idx = [x for x in all_idx if x not in fixed_idx]
-        variable_init = all_init[variable_idx]
+        variable_init = default_backend.tolist(all_init[variable_idx])
         variable_bounds = [par_bounds[i] for i in variable_idx]
 
         tv = _TensorViewer([fixed_idx, variable_idx])
-        fixed_values_tensor = tensorlib.astensor(fixed_values, dtype='float')
-        build_pars = lambda pars: tv.stitch([fixed_values_tensor, pars])
+        # NB: this is a closure, tensorlib needs to be accessed at a different point in time
+        def stitch_pars(pars, stitch_with=fixed_values):
+            tb, _ = get_backend()
+            return tv.stitch([tb.astensor(fixed_values, dtype='float'), pars])
+
     else:
         tv = None
-        fixed_values_tensor = tensorlib.astensor([], dtype='float')
-        variable_init = all_init
+        variable_init = init_pars
         variable_bounds = par_bounds
-        build_pars = lambda pars: pars
+        stitch_pars = lambda pars, stitch_with=None: pars
 
-    func = _get_tensor_shim()(
-        objective, tensorlib.astensor(data), pdf, build_pars, do_grad=do_grad,
+    objective_and_grad = _get_tensor_shim()(
+        objective, tensorlib.astensor(data), pdf, stitch_pars, do_grad=do_grad,
     )
 
-    return tv, fixed_values_tensor, func, variable_init, variable_bounds
+    if do_grad:
+        wrapped_objective = lambda pars: objective_and_grad(pars)[0]
+        jac = lambda pars: objective_and_grad(pars)[1]
+    else:
+        wrapped_objective = objective_and_grad
+        jac = None
+
+    return stitch_pars, wrapped_objective, jac, variable_init, variable_bounds
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -63,8 +63,9 @@ def shim(
     """
     tensorlib, _ = get_backend()
 
+    all_init = default_backend.astensor(init_pars)
+
     if do_stitch:
-        all_init = default_backend.astensor(init_pars)
         all_idx = default_backend.astensor(range(pdf.config.npars), dtype='int')
 
         fixed_vals = fixed_vals or []
@@ -72,7 +73,7 @@ def shim(
         fixed_idx = [x[0] for x in fixed_vals]
 
         variable_idx = [x for x in all_idx if x not in fixed_idx]
-        variable_init = default_backend.tolist(all_init[variable_idx])
+        variable_init = all_init[variable_idx]
         variable_bounds = [par_bounds[i] for i in variable_idx]
 
         tv = _TensorViewer([fixed_idx, variable_idx])
@@ -81,12 +82,12 @@ def shim(
     else:
         tv = None
         fixed_values_tensor = tensorlib.astensor([], dtype='float')
-        variable_init = init_pars
+        variable_init = all_init
         variable_bounds = par_bounds
         build_pars = lambda pars: pars
 
-    wrapped_objective = _get_tensor_shim()(
+    func = _get_tensor_shim()(
         objective, tensorlib.astensor(data), pdf, build_pars, do_grad=do_grad,
     )
 
-    return tv, fixed_values_tensor, wrapped_objective, variable_init, variable_bounds
+    return tv, fixed_values_tensor, func, variable_init, variable_bounds
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -80,15 +80,16 @@ def shim(
     """
     tensorlib, _ = get_backend()
 
+    all_init = tensorlib.astensor(init_pars)
+    all_idx = tensorlib.astensor(range(pdf.config.npars), dtype='int')
+
+    fixed_vals = fixed_vals or []
+    fixed_values = [x[1] for x in fixed_vals]
+    fixed_idx = [x[0] for x in fixed_vals]
+
+    variable_idx = [x for x in all_idx if x not in fixed_idx]
+
     if do_stitch:
-        all_init = tensorlib.astensor(init_pars)
-        all_idx = tensorlib.astensor(range(pdf.config.npars), dtype='int')
-
-        fixed_vals = fixed_vals or []
-        fixed_values = [x[1] for x in fixed_vals]
-        fixed_idx = [x[0] for x in fixed_vals]
-
-        variable_idx = [x for x in all_idx if x not in fixed_idx]
         variable_init = tensorlib.tolist(all_init[variable_idx])
         variable_bounds = [par_bounds[i] for i in variable_idx]
         # stitched out the fixed values, so we don't pass any to the underlying minimizer
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -56,20 +56,27 @@ def shim(
 
     .. note::
 
+        ``minimizer_kwargs`` is a dictionary containing
+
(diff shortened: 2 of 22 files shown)


src/pyhf/optimize/autodiff_minuit.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        tensorlib, _ = get_backend()
+        tv, fixed_values_tensor, func_and_grad, init, bounds = self.setup_minimize(
+            objective, data, pdf, init_pars, par_bounds, fixed_vals
+        )
+        func = lambda pars: func_and_grad(pars)[0]
+        grad = lambda pars: func_and_grad(pars)[1]
+
+        fitresult = iminuit.minimize(
+            func, init, method='SLSQP', jac=grad, bounds=bounds
```
lukasheinrich, line 39: the method kwarg doesn't make any sense here (maybe it's just ignored by iminuit)

kratsg: no worries. the entire file is gone now :)

README.rst
```diff
@@ -41,7 +41,7 @@ Hello World
    >>> pdf = pyhf.simplemodels.hepdata_like(signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0])
    >>> CLs_obs, CLs_exp = pyhf.infer.hypotest(1.0, [51, 48] + pdf.config.auxdata, pdf, return_expected=True)
    >>> print('Observed: {}, Expected: {}'.format(CLs_obs, CLs_exp))
-   Observed: [0.05290116], Expected: [0.06445521]
+   Observed: [0.05290051], Expected: [0.0644532]
```
lukasheinrich, line 44: why does this refactor change results?

kratsg: If I'm honest, I don't know. The best fit pars are the same before and after refactor


```python
#before
>>> pyhf.infer.mle.fit([51, 48] + pdf.config.auxdata, pdf)
array([0.        , 1.0030512 , 0.96266961])

#after
>>> pyhf.infer.mle.fit([51, 48] + pdf.config.auxdata, pdf)
array([0.        , 1.0030512 , 0.96266961])
```

kratsg: casting

```python
#before
>>> pdf = pyhf.simplemodels.hepdata_like(signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0])
>>> pyhf.infer.mle.fit([51, 48] + pdf.config.auxdata, pdf).tolist()
[0.0, 1.00305120262146, 0.9626696070391361]

#after
>>> pdf = pyhf.simplemodels.hepdata_like(signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0])
>>> pyhf.infer.mle.fit([51, 48] + pdf.config.auxdata, pdf).tolist()
[0.0, 1.00305120262146, 0.9626696070391361]
```

kratsg: This is split out into #970.

src/pyhf/optimize/mixins.py
```diff
@@ -0,0 +1,140 @@
...
+            jac = lambda pars: func_and_grad(pars)[1]
+        else:
+            func = func_and_grad
+            jac = None
+
+        self._setup_minimizer(func, data, pdf, init_pars, par_bounds, fixed_vals)
+
+        minimizer_kwargs = dict(method=method, bounds=bounds, options=kwargs, jac=jac)
+        if not do_stitch:
+            minimizer_kwargs.update(dict(fixed_vals=fixed_vals))
```
lukasheinrich, line 113: this seems to late to do. E.g. for minuit all fixed vals are already set in`_setup_minimizer` and passing `fixed_vals` in .minimize() is not doing anything.

I would suggest sth like (as discussed)

```
objective, fixed_pars, postprocess = prepare_minimze(do_stitch, do_grad)
self.setup_minimize(objective, fixed_pars, )
result = self._internal_minimize()
fitted_pars = postprocess(result)
```

src/pyhf/optimize/mixins.py
```diff
@@ -0,0 +1,140 @@
...
+        result = self._internal_minimize(func, init, **minimizer_kwargs)
+
```
lukasheinrich, line 115: we need to somehow bridge the gap that minuit is stateful and scipy is stateless

```
self. _setup_minimizer(func, data, pdf, init_pars, par_bounds, fixed_vals) #prep self._minimizer obj
self._minimize() #use self.minimizer
```

could wrap scipy.optimize call into a stateless thin

```
self.minimizer = lambda : scipy.optimize.minimize(...)
```

which in `_internal_minimize()` might just be used via `self.minimizer()`

src/pyhf/optimize/mixins.py
```diff
@@ -0,0 +1,140 @@
...
+        nonfixed_vals = tensorlib.astensor(result.x)
+        fitted_val = result.fun
+        # stitch things back up if needed
+        if do_stitch:
+            fitted_pars = tv.stitch([fixed_values_tensor, nonfixed_vals])
+        else:
+            fitted_pars = nonfixed_vals
+
```
lukasheinrich, line 123: this should be wrapped into a `postprocess` function

src/pyhf/optimize/opt_scipy.py
```diff
@@ -1,53 +1,58 @@
...
+        self.name = 'scipy'
+        super(scipy_optimizer, self).__init__(*args, **kwargs)
 
-    def __init__(self, **kwargs):
-        """Create scipy.optimize-based Optimizer."""
-        self.maxiter = kwargs.get('maxiter', 100000)
+    def _setup_minimizer(
+        self, objective, data, pdf, init_pars, par_bounds, fixed_vals=None
+    ):
+        self._minimizer = scipy.optimize.minimize
```
lukasheinrich, line 23: maybe this should rather be 


```
self._minimizer = lambda : scipy.optimize.minimize(
            func,
            init,
            method=method,
            jac=jac,
            bounds=bounds,
            constraints=constraints,
            options=dict(maxiter=self.maxiter, disp=self.verbose, **options),
        )
```

src/pyhf/optimize/opt_scipy.py
```diff
@@ -1,53 +1,58 @@
...
-        constraints = [{'type': 'eq', 'fun': lambda v: v[indices] - values}]
-        result = minimize(
-            objective,
-            init_pars,
+        if fixed_vals:
+            constraints = [{'type': 'eq', 'fun': lambda v: v[indices] - values}]
+        else:
+            constraints = []
+
+        return self._minimizer(
```
lukasheinrich, line 50: and  then just `self._minimizer()`

src/pyhf/optimize/mixins.py
kratsg: I'm not sure we truly need a separate post-process function. If it's a matter of just making the function less complicated, we can add a `_postprocess(result, do_stitch=False, return_fitted_val=False)` to split the `def OptimizerMixin.minimize` up.

src/pyhf/optimize/opt_minuit.py
```diff
@@ -1,87 +1,115 @@
...
-            kwargs.update(**d)
-        mm = iminuit.Minuit(
-            f,
+
+        self._minimizer = iminuit.Minuit(
+            objective,
             print_level=1 if self.verbose else 0,
-            errordef=1,
+            errordef=self.errordef,
             use_array_call=True,
```
matthewfeickert, line 76: As we saw that in my tests that old versions of `iminuit` don't have this API can we set a lower bound in the `minuit` extra? Or do you want that in a different PR?

matthewfeickert: I can just make a small maintenance PR on this, as it isn't really in the scope of this PR.

src/pyhf/optimize/opt_minuit.py
```diff
@@ -1,87 +1,115 @@
...
-            constraints = {'fix_p{}'.format(index): True}
+            constraints['fix_p{}'.format(index)] = True
             initvals['p{}'.format(index)] = value
-        kwargs = {}
-        for d in [kw, constraints, initvals, step_sizes]:
-            kwargs.update(**d)
-        mm = iminuit.Minuit(
-            f,
+
+        self._minimizer = iminuit.Minuit(
```
lukasheinrich, line 72: if we go stateless route, maybe we don't need to assign this to `self`

src/pyhf/optimize/opt_scipy.py
```diff
@@ -1,53 +1,56 @@
...
         indices = [i for i, _ in fixed_vals]
         values = [v for _, v in fixed_vals]
-        constraints = [{'type': 'eq', 'fun': lambda v: v[indices] - values}]
-        result = minimize(
+        if fixed_vals:
+            constraints = [{'type': 'eq', 'fun': lambda v: v[indices] - values}]
+        else:
+            constraints = []
+
+        return self._minimizer(
```
lukasheinrich, line 48: if we go stateless route, maybe we don't need to assign this to `self`

src/pyhf/optimize/opt_scipy.py
```diff
@@ -1,53 +1,56 @@
...
+        Same signature as scipy.optimize.minimize.
+
+        This returns a callable that returns the fitresult.
 
         Returns:
-            bestfit parameters
-        
+            fitresult (`scipy.optimize.OptimizeResult`): the fit result
         """
+        self._minimizer = scipy.optimize.minimize
```
lukasheinrich, line 38: if we go stateless route, maybe we don't need to assign this to `self`

src/pyhf/optimize/mixins.py
```diff
@@ -0,0 +1,121 @@
...
+            jac = lambda pars: objective_and_grad(pars)[1]
+        else:
+            wrapped_objective = objective_and_grad
+            jac = None
+
+        minimizer_kwargs = dict(
+            method=method, jac=jac, bounds=bounds, fixed_vals=[], **kwargs
+        )
+        if not do_stitch:
+            minimizer_kwargs.update(dict(fixed_vals=fixed_vals))
```
lukasheinrich, line 87: I think it would be nice if `shim` returned `wrapped_objective `, `init`, `minimizer_kwargs` directly (and `postprocess` so we don't need access to `tv` etc directly

src/pyhf/optimize/mixins.py
lukasheinrich: see above,.. if the shim returns closures it might be quite nice 

```
wrapped_objective, minimizer_kwargs, postprocess = shim()
result = self._minimize(wrapped_objective, **minimizer_kwargs)
result = postprocess(result)
```

src/pyhf/optimize/mixins.py
```diff
@@ -0,0 +1,121 @@
...
+        if do_grad:
+            wrapped_objective = lambda pars: objective_and_grad(pars)[0]
+            jac = lambda pars: objective_and_grad(pars)[1]
+        else:
+            wrapped_objective = objective_and_grad
+            jac = None
+
+        minimizer_kwargs = dict(
+            method=method, jac=jac, bounds=bounds, fixed_vals=[], **kwargs
+        )
```
lukasheinrich, line 85: can probably include `init` here

src/pyhf/optimize/opt_minuit.py
matthewfeickert: This is now PR #975

src/pyhf/optimize/opt_minuit.py
```diff
@@ -1,87 +1,115 @@
...
+    def __init__(self, *args, **kwargs):
         """
         Create MINUIT Optimizer.
 
+        .. note::
+
+            errordef should be 1.0 for a least-squares cost function and 0.5
+            for negative log-likelihood function. See page 37 of
+            http://hep.fi.infn.it/minuit.pdf. This parameter is sometimes
+            called UP in the MINUIT docs.
```
matthewfeickert, line 22: Formatting for render on website.
```suggestion
            ``errordef`` should be 1.0 for a least-squares cost function and 0.5
            for negative log-likelihood function. See page 37 of
            http://hep.fi.infn.it/minuit.pdf. This parameter is sometimes
            called ``UP`` in the ``MINUIT`` docs.
```

src/pyhf/optimize/opt_minuit.py
kratsg: One reason for `self._minimizer` was to allow access to the underlying minimizer object. However, now that I think about it, we can get access to that via the `scipy.OptimizeResult` object if we add a `return_result_obj` argument, which provides that information.

src/pyhf/optimize/mixins.py
kratsg: `scipy.optimize` api is

```
scipy.optimize.minimize(fun, x0, args=(), method=None, jac=None, hess=None, hessp=None, bounds=None, constraints=(), tol=None, callback=None, options=None)
```

which has the objective and init as arguments.

src/pyhf/optimize/mixins.py
kratsg: done

src/pyhf/optimize/opt_minuit.py
kratsg: done. stateless.

src/pyhf/optimize/opt_scipy.py
kratsg: done, stateless.

src/pyhf/optimize/mixins.py
kratsg: done, via `stitch_pars` function.

src/pyhf/optimize/mixins.py
```diff
@@ -0,0 +1,151 @@
...
+            fixed_vals,
+            do_grad=do_grad,
+            do_stitch=do_stitch,
+        )
+
+        minimizer_kwargs = dict(
+            method=method, jac=jac, bounds=bounds, fixed_vals=fixed_vals, options=kwargs
+        )
+        if do_stitch:
+            minimizer_kwargs['fixed_vals'] = []
```
lukasheinrich, line 141: can we not have `shim` return all `minimizer_kwargs` ?

src/pyhf/optimize/opt_numpy.py
```diff
@@ -0,0 +1,30 @@
...
+        stitch_pars (`func`): callable that stitches parameters, see :func:`pyhf.optimize.common.shim`.
+        do_grad (`bool`): enable autodifferentiation mode. Default is off.
+
+    Returns:
+        objective_and_grad (`func`): tensor backend wrapped objective,gradient pair
+    """
+
+    tensorlib, _ = get_backend()
+
+    assert not do_grad, "Numpy does not support autodifferentiation"
```
lukasheinrich, line 23: should this perhaps be a pyhf exception?

src/pyhf/optimize/mixins.py
```diff
@@ -0,0 +1,147 @@
...
+        if kwargs:
+            raise KeyError(
+                f"""Unexpected keyword argument(s): '{"', '".join(kwargs.keys())}'"""
+            )
+
+    def _internal_minimize(
+        self,
+        func,
+        x0,
+        method='SLSQP',
```
lukasheinrich, line 34: this should probably be either hardcoded into the scipy one or delegated to the `opotinos` since e.g. minuit does not know about `method`

kratsg: this one is slightly tricky. I really wanted to keep the `scipy.optimize.minimize` API where possible -- but I could definitely relegate it to `options` which is not a bad idea.

src/pyhf/__init__.py
```diff
@@ -27,6 +25,12 @@ def get_backend():
     return tensorlib, optimizer
 
 
+tensorlib = tensor.numpy_backend()
```
lukasheinrich, line 28: why do we need this here and why didn't we need it before?

kratsg: `pyhf.optimizer = scipy_optimizer()` and `scipy_optimizer` from `opt_scipy` imports `OptimizerMixin` which needs `get_backend`.

src/pyhf/optimize/common.py
```diff
@@ -0,0 +1,129 @@
...
+    Returns:
+        minimizer_kwargs (`dict`): arguments to pass to a minimizer following the ``scipy.optimize.minimize`` API (see notes)
+        stitch_pars (`func`): callable that stitches fixed parameters into the unfixed parameters
+    """
+    tensorlib, _ = get_backend()
+
+    if do_stitch:
+        all_init = default_backend.astensor(init_pars)
+        all_idx = default_backend.astensor(range(pdf.config.npars), dtype='int')
+
```
lukasheinrich, line 86: why do these use default_backend and not `tensorlib`?

src/pyhf/optimize/common.py
```diff
@@ -0,0 +1,129 @@
...
+        minimizer_fixed_vals = []
+
+        tv = _TensorViewer([fixed_idx, variable_idx])
+        # NB: this is a closure, tensorlib needs to be accessed at a different point in time
+        def stitch_pars(pars, stitch_with=fixed_values):
+            tb, _ = get_backend()
+            return tv.stitch([tb.astensor(fixed_values, dtype='float'), pars])
+
+    else:
+        tv = None
```
lukasheinrich, line 104: `tv` doesn't seem to be needed (not even as `None`)

src/pyhf/optimize/common.py
```diff
@@ -0,0 +1,129 @@
...
+        variable_init = init_pars
+        variable_bounds = par_bounds
+        minimizer_fixed_vals = fixed_vals
+        stitch_pars = lambda pars, stitch_with=None: pars
+
+    objective_and_grad = _get_tensor_shim()(
+        objective, tensorlib.astensor(data), pdf, stitch_pars, do_grad=do_grad,
```
lukasheinrich, line 111: shouldn't `data` already be a tensor? do we need the cast?

src/pyhf/optimize/common.py
```diff
@@ -0,0 +1,129 @@
...
+    )
+
+    if do_grad:
+        wrapped_objective = lambda pars: objective_and_grad(pars)[0]
+        jac = lambda pars: objective_and_grad(pars)[1]
```
lukasheinrich, line 116: this requires two calls to `objective_and_grad`. But scipy can handle a algorithm that returns both in one go (`jac=True`) which would save one computation. So ideally we wouold support this and only do the callable route in  minuit.

src/pyhf/optimize/common.py
kratsg: we don't do any computations -- just slicing. Those two variables you highlighted go into here

```python
        variable_idx = [x for x in all_idx if x not in fixed_idx]
        variable_init = default_backend.tolist(all_init[variable_idx])
```

`variable_init` is passed back out, while `variable_idx` is passed into `TensorViewer`. We could use `tensorlib` here entirely and drop `default_backend`.

src/pyhf/optimize/common.py
kratsg: `data` didn't need to be a tensor. I can change it to be a tensor though. There were existing tests written that allowed non-tensor data (just a list) e.g. `[10.] + pdf.config.auxdata`.

lukasheinrich: ok

src/pyhf/__init__.py
lukasheinrich: right but why do we need to set tensorlib here?

lukasheinrich: ah sorry this was just moved down from above.. fine..

src/pyhf/optimize/common.py
kratsg: fixed it up.

src/pyhf/optimize/common.py
kratsg: #980 is opened for this.

src/pyhf/optimize/opt_jax.py
```diff
@@ -1,70 +1,76 @@
...
-            init_pars: initial parameters
-            par_bounds: parameter boundaries
-            fixed_vals: fixed parameter values
-
-        """
 
-        tensorlib, _ = get_backend()
-        all_idx = default_backend.astensor(range(pdf.config.npars), dtype='int')
-        all_init = default_backend.astensor(init_pars)
+def wrap_objective(objective, data, pdf, stitch_pars, do_grad=False, jit_pieces=None):
```
lukasheinrich, line 32: `stitch_pars` is not used anymore

kratsg: Nope, but it's used in all other shims and we keep the same signature. I can `_` it instead.

lukasheinrich: hm maybe we rather put 

```
def stitch_pars(fixed_values,fixed_idx,variable_idx)
    if len(fixed_values) > 0:
        tv = _TensorViewer([fixed_idx, variable_idx])
        constrained_pars = tv.stitch(
            [tensorlib.astensor(fixed_values, dtype='float'), pars]
        )
    else:
        constrained_pars = pars
```

into some common.py and all shims use this?

kratsg: `_internal_postprocess` also uses `stitch_pars` so it would have to be refactored too.

lukasheinrich: ok

src/pyhf/optimize/mixins.py
```diff
@@ -0,0 +1,137 @@
...
+class OptimizerMixin(object):
+    """Mixin Class to build optimizers."""
+
+    def __init__(self, **kwargs):
+        """
+        Create an optimizer.
+
+        Args:
+            maxiter (`int`): maximum number of iterations. Default is 100000.
+            verbose (`bool`): print verbose output during minimization. Default is off.
```
matthewfeickert: As these are bools we should reflect that in the docs usage.

```suggestion
            verbose (`bool`): print verbose output during minimization. Default is ``False``.
```

src/pyhf/optimize/mixins.py
```diff
@@ -0,0 +1,137 @@
...
+        Args:
+            objective (`func`): objective function
+            data (`list`): observed data
+            pdf (`pyhf.pdf.Model`): model
+            init_pars (`list`): initial parameters
+            par_bounds (`list`): parameter boundaries
+            fixed_vals (`list`): fixed parameter values
+            return_fitted_val (`bool`): return bestfit value of the objective
+            return_fit_object (`bool`): return ``scipy.optimize.OptimizeResult``
+            do_grad (`bool`): enable autodifferentiation mode. Default is off.
```
matthewfeickert, line 110:
```suggestion
            do_grad (`bool`): enable autodifferentiation mode. Default is ``False``.
```

src/pyhf/optimize/opt_jax.py
```diff
@@ -1,70 +1,80 @@
...
 
-        fixed_vals = fixed_vals or []
-        fixed_values = [x[1] for x in fixed_vals]
-        fixed_idx = [x[0] for x in fixed_vals]
+    Args:
+        objective (`func`): objective function
+        data (`list`): observed data
+        pdf (`pyhf.pdf.Model`): model
+        stitch_pars (`func`): callable that stitches parameters, see :func:`pyhf.optimize.common.shim`.
+        do_grad (`bool`): enable autodifferentiation mode. Default is off.
```
matthewfeickert, line 43:
```suggestion
        do_grad (`bool`): enable autodifferentiation mode. Default is ``False``.
```

src/pyhf/optimize/opt_minuit.py
```diff
@@ -1,87 +1,128 @@
...
         """
-        Find Function Parameters that minimize the Objective.
+        Same signature as scipy.optimize.minimize.
+
+        Note: an additional `minuit` is injected into the fitresult to get the
+        underlying minimizer.
+
+        Minimizer Options:
+            maxiter (`int`): maximum number of iterations. Default is 100000.
+            return_uncertainties (`bool`): Return uncertainties on the fitted parameters. Default is off.
```
matthewfeickert, line 87:
```suggestion
            return_uncertainties (`bool`): Return uncertainties on the fitted parameters. Default is ``False``.
```

src/pyhf/optimize/opt_numpy.py
```diff
@@ -0,0 +1,32 @@
...
+def wrap_objective(objective, data, pdf, stitch_pars, do_grad=False, jit_pieces=None):
+    """
+    Wrap the objective function for the minimization.
+
+    Args:
+        objective (`func`): objective function
+        data (`list`): observed data
+        pdf (`pyhf.pdf.Model`): model
+        stitch_pars (`func`): callable that stitches parameters, see :func:`pyhf.optimize.common.shim`.
+        do_grad (`bool`): enable autodifferentiation mode. Default is off.
```
matthewfeickert, line 16:
```suggestion
        do_grad (`bool`): enable autodifferentiation mode. Default is ``False``.
```

src/pyhf/optimize/opt_pytorch.py
```diff
@@ -1,52 +1,42 @@
...
-        self, objective, data, pdf, init_pars, par_bounds, fixed_vals=None
-    ):
-        """
-        Prepare Minimization for AutoDiff-Optimizer.
+    Args:
+        objective (`func`): objective function
+        data (`list`): observed data
+        pdf (`pyhf.pdf.Model`): model
+        stitch_pars (`func`): callable that stitches parameters, see :func:`pyhf.optimize.common.shim`.
+        do_grad (`bool`): enable autodifferentiation mode. Default is off.
```
matthewfeickert, line 16:
```suggestion
        do_grad (`bool`): enable autodifferentiation mode. Default is ``False``.
```

src/pyhf/optimize/opt_scipy.py
```diff
@@ -1,53 +1,72 @@
...
+        return_uncertainties=False,
+        options={},
     ):
         """
-        Find Function Parameters that minimize the Objective.
+        Same signature as scipy.optimize.minimize.
+
+        Minimizer Options:
+            maxiter (`int`): maximum number of iterations. Default is 100000.
+            verbose (`bool`): print verbose output during minimization. Default is off.
```
matthewfeickert, line 44:
```suggestion
            verbose (`bool`): print verbose output during minimization. Default is ``False``.
```

src/pyhf/optimize/opt_tflow.py
```diff
@@ -1,53 +1,43 @@
...
-        self, objective, data, pdf, init_pars, par_bounds, fixed_vals=None
-    ):
-        """
-        Prepare Minimization for AutoDiff-Optimizer.
+    Args:
+        objective (`func`): objective function
+        data (`list`): observed data
+        pdf (`pyhf.pdf.Model`): model
+        stitch_pars (`func`): callable that stitches parameters, see :func:`pyhf.optimize.common.shim`.
+        do_grad (`bool`): enable autodifferentiation mode. Default is off.
```
matthewfeickert, line 15:
```suggestion
        do_grad (`bool`): enable autodifferentiation mode. Default is ``False``.
```

src/pyhf/__init__.py
```diff
@@ -17,7 +15,7 @@ def get_backend():
     Example:
         >>> import pyhf
         >>> pyhf.get_backend()
-        (<pyhf.tensor.numpy_backend.numpy_backend object at 0x...>, <pyhf.optimize.opt_scipy.scipy_optimizer object at 0x...>)
+        (<pyhf.tensor.numpy_backend.numpy_backend object at 0x...>, <pyhf.optimize.scipy_optimizer object at 0x...>)
```
matthewfeickert, line 18: Probably helps to make this more explicit now (I can't edit the `pyhf.get_backend()` above so doing so here, but would need to get fixed up so I can also do this one)
```suggestion
        backend, optimizer = pyhf.get_backend()
        >>> backend
        <pyhf.tensor.numpy_backend.numpy_backend object at 0x...>
        >>> optimizer
        <pyhf.optimize.scipy_optimizer object at 0x...>
```

src/pyhf/optimize/common.py
```diff
@@ -0,0 +1,131 @@
...
+    do_grad=False,
+    do_stitch=False,
+):
+    """
+    Prepare Minimization for Optimizer.
+
+    Args:
+        objective (`func`): objective function
+        data (`list`): observed data
+        pdf (`pyhf.pdf.Model`): model
```
matthewfeickert, line 52: Want this hyperlinked

```suggestion
        pdf (~pyhf.pdf.Model): model
```

src/pyhf/optimize/mixins.py
```diff
@@ -0,0 +1,137 @@
...
+            do_stitch (`bool`): enable splicing/stitching fixed parameter.
+            kwargs: other options to pass through to underlying minimizer
+
+        Returns:
+            parameters (`tensor`): fitted parameters
+            minimum (`float`): if ``return_fitted_val`` flagged, return minimized objective value
+            result (scipy.optimize.OptimizeResult`): if ``return_fit_object`` flagged
```
matthewfeickert: This doesn't render out cleanly in Sphinx at the moment. Can we use the [style used in `hypotest`](https://github.com/scikit-hep/pyhf/blob/d35f4b487d3a03d42edb6f3a49122283fedd5d59/src/pyhf/infer/__init__.py#L48)?

src/pyhf/optimize/common.py
matthewfeickert: Note to self: This happens a lot so this should have a search and replace applies to it

src/pyhf/optimize/mixins.py
kratsg: verbose changed to `int` and reworded again.

src/pyhf/optimize/mixins.py
kratsg: default is auto-configurable, rather than `False` now.

tests/test_optim.py
```diff
@@ -27,6 +25,202 @@ def rosen(x):
...
+        if 'no_grad-minuit-pytorch-32b' in identifier:
+            # quite a large difference
+            rtol = 3e-01
+        if 'do_grad-minuit-pytorch-32b' in identifier:
+            # a small difference
+            rtol = 7e-05
+        if 'no_grad-minuit-jax-32b' in identifier:
+            rtol = 4e-02
+        if 'do_grad-minuit-jax-32b' in identifier:
+            rtol = 5e-03
```
matthewfeickert, line 122: My guess is no, but do we have any idea why this is happening for macOS?

kratsg: 🤷‍♂️ it's 32b. I'm not super concerned right now, but there's probably reasons why (e.g. processor or 32b / 64b not handled).

matthewfeickert: Okay that's fine. I'll just open an Issue and we can move on from here. :+1:

matthewfeickert: This is PR #981

PR #952, opened 2020-07-17
experimental: shim idea/concept for feat/autogradMinuit

kratsg:
Now, to disentangle this a little bit, we can do this in two ways:
- [1] we can have an optimizer "shim" for each backend that we manually pull in, and they have either ScipyOptimizer or MinuitOptimizer as a member variable to store the actual optimizer object
- [2] we can have ScipyOptimizer or MinuitOptimizer at the top-level, and they "peek" at the tensorlib to determine which backend "shim" to pull in for doing their minimization

Currently in pyhf: opt.jax.minimize->opt.autodiffmixin.minimize->opt.jax.setup_minimize->opt.autodiffmixin.minimize->scipy.optimize.minimize

For [1], opt.jax.minimize->opt.jax.setup_minimize->opt.ScipyOptimizer.minimize->opt.Optimizer.minimize->opt.ScipyOptimizer._minimize->opt.ScipyOptimizer._setup_minimizer->scipy.optimize.minimize

For [2], opt.ScipyOptimizer.minimize->opt.Optimizer._setup_minimizer->opt.Optimizer.minimize->shim_jax.setup_minimize->opt.Optimizer.minimize->opt.Optimizer._internal_minimize->opt.ScipyOptimizer._minimize->scipy.optimize.minimize
Currently, in pyhf, it's rather short, because we're not disentangling backend from optimizer -- it's a LITTLE tied in together.

[1] has the decent benefit that we keep ~same API that we used to, with every backend associated with an "optimizer" -- each capable of calling two different minimizers, depending on how things are configured

[2] has the benefit that we don't have optimizers for each backend, but decouple it into a minimizer API that's unified a bit more, and allows us to switch out backends without touching optimizer [and we might have to think if the call chain could be reduced somehow...]

-----

This PR is mostly for option [2].


kratsg:
Note: tensorflow seems currently buggy in some way with autograd=False but to be looked at beyond this PR I think. Some nice screenshots to demonstrate feasability

Screenshot 2020-07-16 23 43 18

Screenshot 2020-07-16 23 44 34

Screenshot 2020-07-16 23 44 52

Screenshot 2020-07-16 23 47 27

Screenshot 2020-07-17 00 34 12

kratsg:
Also note, this is set up to fail for numpy+autograd=True (which we want to explicitly make sure).

More work needs to be done to handle the shim imports correctly (protect against various imports) but it mostly works.

lukasheinrich:
I'm voting for [2]. We will likely run into more issues like this wherer we have some type of algorithm (here minimization) that needs to be implemented so either it's part of the tensor backend (with the backend on top)

`tb.optimize(foo,bar)`

or this shim with the algorithm on top (which I pefer somewhat). E.g. this likely comes up also for interpolation algs (#953)

kratsg:
> or this shim with the algorithm on top (which I pefer somewhat). E.g. this likely comes up also for interpolation algs (#953)

I agree.

PR #953, opened 2020-07-17
feat: Add test inversion / interval inference API

lukasheinrich:
# Description

One aspect in the inference lifecycle we don't yet provide an API for is the hypothesis test inversion stage for interval estimation.

This adds a simple API, that only depends on hypotest so this is not hifa-specific at all (keeping w/ the idea of eventually factoring out inference)

```
obs,exp, (scan,results) = pyhf.infer.intervals.upperlimit(data = data, model  = model, return_results=True, scan = np.linspace(0,5,21))
```

screenshot

But there were good reasons we didn't have this, 

* algorithms

 depending on the cost of a single hypotest, the algorithm used for the test inversion (to choose POI vals) can really make a huge difference.  E.g. brent bracketing or https://github.com/diana-hep/excursion would be better suited  (not sure what the autoscan in https://root.cern/doc/v608/classRooStats_1_1HypoTestInverter.html uses0

* backend dependence

here we use simple linearr interpolation of a fixed scan via `np.interp` (i.e. we cast into numpy and back to the active backend), but e.g. jax also has differentiable ROOT finding etc. 
but maybe having the inference stage be diffable is anyways a longer terrm issue (see neos)

* naming:

`pyhf.infer.intervals.upperlimit` looks fine to me 

before i tried

`pyhf.infer.inversion.upperlimit`
`pyhf.infer.invert.upperlimit`

but I think the thing we try to do is "interval inference" in the end. the fact that it goes through inversion of hypotests is maybe a detail.

Thoughts

@matthewfeickert @kratsg  @alexander-held @cranmer ?

ReadTheDocs build: https://pyhf.readthedocs.io/en/upperlimitapi/_generated/pyhf.infer.intervals.upperlimit.html#pyhf.infer.intervals.upperlimit

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add infer.intervals.upperlimit API for interval estimation
   - Depends only on hypotest to allow for factoring out in future 
* Add test for upperlimit
* Add infer.intervals.upperlimit API to docs

```


lukasheinrich:
another possible API

```
pyhf.infer.intervals.upperlimit(hypotester = lambda x: pyhf.infer.hypotest(x,data,model) return_results=True, scan = np.linspace(0,5,21))
```


which would maybe help avoid having to pass through hypotest kwargs/options through the API but is less readable perhaps

matthewfeickert:
@lukasheinrich I think this is a good idea, as long as we can keep this readily seperated from the core so that as you mention it can be factored out later. So this gets a :+1: from me to add in.

> `pyhf.infer.intervals.upperlimit` looks fine to me

I'm fine with this name.

I'll look at this again later once there are a few more things passing so we can discuss the API itself, but I'm glad this PR exists.

src/pyhf/infer/intervals.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def _interp(x, xp, fp):
+    import numpy as np
+
+    tb, _ = get_backend()
+    return tb.astensor(np.interp(x, xp.tolist(), fp.tolist()))
+
+
+def upperlimit(data, model, scan, return_results=False):
```
matthewfeickert, line 12: Given my comment in the PR, it would be good to set an `alpha` (or similar) API for setting the upper limit
```suggestion
def upperlimit(data, model, scan, alpha=0.05, return_results=False):
```

so the below would use

```python
limits = [_interp(alpha, resarary[i][::-1], scan[::-1]) for i in range(6)]
```

lukasheinrich: hm we als use hypotest and not hypo_test. Thoughts?

cranmer:
I haven’t reviewed all I I I of this carefully, but I think the API should make explicit if it is a CLs upper limit or a standard “CLsb” upper limit. 

It would be nice if there was a flag to switch between the two. 


Minor note, it is worth documenting it is a one-sided upper limit, not the upper end of a two-sided interval.

lukasheinrich:
the  inversion procedure though is generic, no? I.e. the switch between how the p-value is computed should go into the hypotest procedure

e.g.

```
.infer.intervals.upperlimit(data = data, model  = model, return_results=True, scan = np.linspace(0,5,21), hypotest_kwargs = dict(cls = True) )
 ```

or similar

**Comment by cranmer** [comment]
Yes, I agree with that.

File: src/pyhf/infer/intervals.py (hunk 1 above)
Comment (matthewfeickert) [reply src/pyhf/infer/intervals.py]: This is true. I've been tempted to change `hypotest`, except that I think a lot of people already are relying explicitly on that API so I'm hesitant to change it now. `hypotest` and `upperlimit` is more in the style of the [NumPy API](https://numpy.org/doc/1.18/reference/arrays.ndarray.html) choices, while `upper_limit` is more in the style of the [SciPy](https://docs.scipy.org/doc/scipy/reference/linalg.html), [Matplotlib](https://matplotlib.org/api/index.html#usage-patterns), and [PyTorch](https://pytorch.org/docs/stable/tensors.html) API style choices. I'm personally more partial to the latter, but that's just a _preference_.

File: tests/test_infer.py [hunk 2]
```diff
@@ -20,6 +21,14 @@ def check_uniform_type(in_list):
     )
 
 
+def test_upperlimit(tmpdir, hypotest_args):
+    """
+    Check that the default return structure of pyhf.infer.hypotest is as expected
+    """
+    _, data, model = hypotest_args
+    pyhf.infer.intervals.upperlimit(data, model, scan=np.linspace(0, 5, 11))
```
matthewfeickert, line 29: This still needs an actual test.

matthewfeickert (changes requested):

src/pyhf/infer/intervals.py
```diff
@@ -0,0 +1,66 @@
...
+        >>> exp_limits
+        [array(0.59576921), array(0.76169166), array(1.08504773), array(1.50170482), array(2.06654952)]
+
+    Args:
+        data (`tensor`): The observed data.
+        model (~pyhf.pdf.Model): The statistical model adhering to the schema ``model.json``.
+        scan (`Iterable`): Iterable of POI values.
+        level (`float`): The threshold value to evaluate the interpolated results at.
+        return_results (`bool`): Whether to return the per-point results.
+                                 Default is ``False``.
```
matthewfeickert, line 42:
```suggestion
```

This is visible in the signature already, so no need to repeat again.

src/pyhf/infer/intervals.py
```diff
@@ -0,0 +1,66 @@
...
+              :class:`~pyhf.infer.hypotest` results at each test POI.
+              Only returned when ``return_results`` is ``True``.
+    """
+    tb, _ = get_backend()
+    results = [
+        hypotest(mu, data, model, qtilde=True, return_expected_set=True) for mu in scan
+    ]
+    obs = tb.astensor([[r[0]] for r in results])
+    exp = tb.astensor([[r[1][shift] for shift in range(5)] for r in results])
+    result_arary = tb.concatenate([obs, exp], axis=1).T
```
matthewfeickert, line 59:
```suggestion
    result_arrary = tb.concatenate([obs, exp], axis=1).T
```

Typo on my end

matthewfeickert:
With the current API:

```python
import numpy as np
import pyhf
import pyhf.infer.intervals

pyhf.set_backend("numpy")
)
observations = [51, 48]
data = pyhf.tensorlib.astensor(observations + model.config.auxdata)
scan = np.linspace(0, 5, 21)
obs_limit, exp_limit, (scan, results) = pyhf.infer.intervals.upperlimit(
    data, model, scan, return_results=True
)

fig, ax = plt.subplots()
ax.set_xlabel(u"$\mu$")
ax.set_ylabel(u"$\mathrm{CL}_{s}$")
pyhf.contrib.viz.brazil.plot_results(ax, scan, results)
ax.vlines([obs_limit] + exp_limit, 0, 1)
fig.savefig("limits.png")
```

limits

src/pyhf/infer/intervals.py
```diff
@@ -0,0 +1,66 @@
...
+from .. import get_backend
+import numpy as np
+
+
+def _interp(x, xp, fp):
+    tb, _ = get_backend()
+    return tb.astensor(np.interp(x, xp.tolist(), fp.tolist()))
+
+
+def upperlimit(data, model, scan, level=0.05, return_results=False):
```
matthewfeickert, line 12: I think I would still go with `upper_limit` for the API, but this is something we should discuss or get feedback on.
```suggestion
def upper_limit(data, model, scan, level=0.05, return_results=False):
```

kratsg (approved):

src/pyhf/infer/intervals.py
matthewfeickert: @kratsg also prefers `hypotest` and `upperlimit` so let's go with that.

matthewfeickert:
@lukasheinrich @cranmer I think we can take care of everything else in PR #966

matthewfeickert (approved):

Issue #965, opened 2020-07-20
CLsb options


cranmer:
# Description

The Brazil band plot and the  `pyhf.infer.hypotest` API is very tailored to CLs. It’s not clear how to get expected bands for CL_sb (standard frequentist limits) or CL_b (for expected significance / discovery). Would be nice to have an option for the standard CLsb limits and to make which is being used explicit in the API.

Probably having an option like `use_CLs=True` as a default would be sufficient. Alternatively, `upper_limit_type="CLs"` could work.

## Is your feature request related to a problem? Please describe.

The astro community uses 90% CLsb upper limits mainly.

# Relevant Issues and Pull Requests

This is also relevant for the test inversion / interval API #953  

# Additional context
I also am not a huge fan of the bad physics jargon of justing “CL” instead of p-value, but that’s a different issue.

cranmer:
closing as this issue is also picked up in #966

matthewfeickert:
I'll actually reopen this, so that we can track the PRs that close it.

PR #966, opened 2020-07-20
customizable metrics in hypotest

lukasheinrich:
# Description

From a discussion with @cranmer. This allows specifying which metrics to return from a hypotest. Defaults to `['CLs']` but can be used to get e.g. obs and exp bands for CLsb only or all metrics together:

screenshot

If the API is fine w/ people I can updated docs etc.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/infer/__init__.py
+++ b/src/pyhf/infer/__init__.py
@@ -5,6 +5,26 @@ from .. import get_backend
 from .calculators import AsymptoticCalculator
 
 
+def _assemble_metrics(CLsb, CLb, metrics):
+    tensorlib, _ = get_backend()
+    CLs = CLsb / CLb
+    CLsb, CLb, CLs = (
+        tensorlib.reshape(CLsb, (1,)),
+        tensorlib.reshape(CLb, (1,)),
+        tensorlib.reshape(CLs, (1,)),
+    )
+
+    returns = []
+    for m in metrics:
+        if m == 'CLs':
+            returns.append(CLs)
+        if m == 'CLsb':
+            returns.append(CLsb)
+        if m == 'CLb':
+            returns.append(CLb)
+    return returns
+
+
 def hypotest(
     poi_test, data, pdf, init_pars=None, par_bounds=None, qtilde=False, **kwargs
 ):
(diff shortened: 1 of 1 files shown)


cranmer:
Seems nice, but I'm wondering about "metrics". It seems like it is introducing a new term or concept. 
In https://scikit-hep.org/pyhf/_generated/pyhf.infer.hypotest.html#pyhf.infer.hypotest the term `tail_probs` is used, which seems good, though CLs is a ratio. 

Options: `metrics', `tail_probs`, `limit_types`,

cranmer:
Will close #965 which is duplicative.

lukasheinrich:
yeah I thought abut "pvalues" but then that only fits for CLsb, CLb and was looking for a catch all term for values which are figures of merit for the hypothesis test.

kratsg:
Not a huge fan of `metric` - per se. I wouldn't know what "metric" refers to in this context os easily.

One thing we might need to consider is the order of the outputs. Perhaps just the order of the metrics passed in -> order of outputs returned.

lukasheinrich:
more variations (now with "pvalues")

screenshot

lukasheinrich:
(maybe also an oppoortunity to rename this to `p_sb`, `p_b` and `p_sb/pb`) instead of `CL`

lukasheinrich:
@kratsg yes it uses the order of the outputs passed in the kwarg

kratsg:
> I'm fine with considering CLs a "modified p-value"
> I'd also be fine with options being `pvalues=['p_sb', 'p_b','CLs'']` where p is used for the normal p-values and CLs is used for the modified one... the break of using a "p" prefix indicates it's unusual in some way, and it's more familiar.

Is there a way to present or phrase or word this in such a way that we're not overloading the meaning of "pvalue"? Usually people will think of it corresponding to a particular distribution, and a "modified pvalue" can be a rather new concept for some.

lukasheinrich:
the "modified" language comes from Alex Read's "MODIFIED FREQUENTIST ANALYSIS OF SEARCH RESULTS", so I think it's fine. And the purpose of dividing by p_b is explicitly to modify p_sb  to fit some physics-specific sensitivities

lukasheinrich:
We still need to add the ability for other types of tests (see https://github.com/scikit-hep/pyhf/pull/520), but even if you fix the test type you might still be interested in getting more values returned than the "headline" value. I.e. even if you do CLs-based hypotesting you might want to know CLsb,CLb, so I think having the option to return a number of values will be useful. The "default" value that will be returned might be informed by the test type /  test statistic used then (e.g. now it defaults to ["CLs"])

cranmer:
Nice.. @lukasheinrich 
Note, unlike CLs, the -2-sigma band for 95% UL is close to 0. If one wanted the -2 sigma band for the 90% UL it may "go negative" , so probably worth a check. 
image

lukasheinrich:
yeah the expected band probably requires some work in that regard. also on the other end of the spectrum. The +1,+2 sigma correspond to µ^ values that are > µ_test and the maximal. Aren't any p-values > 0.5 strictly impossible if the signal hypothesis is delta(q) + half chi2? (oiginally I coded this up mostly to match the ROOT calculation).

But these details are probably for a later PR.

cranmer:
Maybe I'm confusing myself or mis-understanding, but the +/- 1 & 2 sigma bands are fluctuations in the alternative hypothesis and there you can have small_prob*delta(0) + (1-small_prob)*chi-square

lukasheinrich:
i'm just saying for `f_s = f(q_mu|mu' = 0) = 0.5*delta(0) + 0.5* chi2` and p_values are `p_sb = 1-Phi(√qmu)`

so for any observed (or expected from the alt hypothesis) value of the test stat the maximum p_value can be either exactly 1 (if test stat == 0) or 0.5 if test stat >=0 ), or am I confusing myself?

cranmer:
Ah, for mu’=0 sure.

Need to think again how that folds into the bands (both in vertical and
horizontal directions... it’s been a while)

Maybe reasonable to use the q_mu instead of ~q_mu for bands to avoid
lumping everything in delta(0)

lukasheinrich:
leaving this from a manual calc here for me to follow up later. With proper qtilde it's looks better but unsure yet why this is not taken into account in our code but will follow up


screenshot

lukasheinrich:
In the end I think what we do is correct:

It's just that with qtilde (i.e. bunded POI) the acceptance region in the FC neyman construction shrinks close to zero and the slope is shallower than the slope of the expected test statistic value for a -2 fluctuation.

Here the same data once in the neyman construction picture (y-axis = test stat) and ones in the p-value picture for alpha 0.1

screenshot


depending on the test level this might then also hit the -1 sigma expected e.g. for 0.2

screenshot

lukasheinrich:
one last thing that might be affecting this is that this all assumes that the variance of µ^ for µ' = µ and µ' = 0  are the same, i.e. σ_b = σ_s

if σ_b = σ_s: 
screenshot
, which @cranmer and I have discussed before might not be necessarily true (but it is what the ROOT asymptotic calculator uses.

if σ_b = 0.5*σ_s: 

screenshot

matthewfeickert:
@lukasheinrich I rebased this to get it moving again, but in the event that I messed anything up there's a _pre_ rebase copy of the branch at `lukas-backup-cls_or_not`.

**REMINDER TO MATTHEW:** DELETE `lukas-backup-cls_or_not`

lukasheinrich:
based on the work in #1162 we need to decide what to do with this. If we now switch to reporting

the expected `CLs, CLsb, CLb` and quantiles of those metrics this then loses the property that

"expected CLsb / expected CLb = expected CLs"

which might be perfectly fine, but just needs to be communicated properly

Issue #981, opened 2020-07-23
Understand 32b vs. 64b precision better


matthewfeickert:
# Description

As shown in PR #951, there are significant differences in results that use `32b` vs. `64b` precision.

```python

        # handle cases where macos and ubuntu provide very different results numerical
        if 'no_grad-minuit-tensorflow-32b' in identifier:
            # not a very large difference, so we bump the relative difference down
            rtol = 3e-02
        if 'no_grad-minuit-pytorch-32b' in identifier:
            # quite a large difference
            rtol = 3e-01
        if 'do_grad-minuit-pytorch-32b' in identifier:
            # a small difference
            rtol = 7e-05
        if 'no_grad-minuit-jax-32b' in identifier:
            rtol = 4e-02
        if 'do_grad-minuit-jax-32b' in identifier:
            rtol = 5e-03
```

 As `32b` is the default precision in PyTorch this should be better researched and understood.

Maybe also worth setting the backends to all use `64b` precision by default.

matthewfeickert:
Following up good points by @kratsg on [what is the best way to validate floating point precision requested for the backends](https://github.com/scikit-hep/pyhf/pull/983#discussion_r460319453)

>  Presumably, we should have these precision checks at the backend level, rather than at the pyhf level but I am fine with this for now, since we don't necessarily have custom backends being supported and architectures likely won't be anything but 32/64.

this should also get followed up on and perhaps reexamined.

kratsg:
Idea, each backend defines `precisions` which is a list of valid precisions. Then just assert that as part of initialization likely.

PR #983, opened 2020-07-24
feat: Add precision setting string API to set_backend

### LINKED ISSUE #982: Add precision setting API to set_backend
# Description

At the moment the current way to set the precision for a backend is by passing a backend object into the `set_backend` API

```python
>>> import pyhf
>>> pyhf.set_backend("pytorch")
>>> pyhf.tensorlib.precision
'32b'
>>> pyhf.set_backend(pyhf.tensor.pytorch_backend(precision="64b"))
>>> pyhf.tensorlib.precision
'64b'
```

it would be nicer if there was the ability to also set the precision through `set_backend` through a `kwarg`

```python
>>> import pyhf
>>> pyhf.set_backend("pytorch", precision="64b")
>>> pyhf.tensorlib.precision
'64b'
```

Also, given the large discrepancies that we see with `32b` precision in the PyTorch backend maybe worth having all backends *default* to setting `64b` precision.

```python
>>> import pyhf
>>> pyhf.set_backend("pytorch")
>>> pyhf.tensorlib.precision
'64b'
```

matthewfeickert:
# Description

Resolves #982

Add the ability to set the backend precision through a `str` argument in `pyhf.set_backend`. This allows the following

```python
>>> import pyhf
>>> pyhf.set_backend("pytorch", precision="64b")  # string API
>>> pyhf.tensorlib.precision
'64b'
>>> pyhf.set_backend(pyhf.tensor.pytorch_backend(precision="32b"))  # object API 
>>> pyhf.tensorlib.precision
'32b'
>>> pyhf.set_backend("pytorch")  # no default, so defer to backend for precision
>>> pyhf.tensorlib.precision
'32b'
>>> pyhf.set_backend(pyhf.tensor.pytorch_backend(precision="32b"), precision="64b")  # set_backend string API wins if conflict
>>> pyhf.tensorlib.precision
'64b'
```

~~Additionally, set the default backend precision for _all_ backends to `64b`.~~

ReadTheDocs build: https://pyhf.readthedocs.io/en/api-add-precision-to-set_backend/

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add support for setting backend precision through str kwarg in set_backend
   - N.B.: If the kwarg is provided it will always be used even if backend is set with tensor backend object
* Adopt use of the str API internally for the CLI
```

### DIFF
--- a/src/pyhf/__init__.py
+++ b/src/pyhf/__init__.py
@@ -45,12 +45,12 @@ def set_backend(backend, custom_optimizer=None, precision=None):
         >>> pyhf.tensorlib.name
         'tensorflow'
         >>> pyhf.tensorlib.precision
-        '64b'
-        >>> pyhf.set_backend(b"pytorch", precision="32b")
+        '32b'
+        >>> pyhf.set_backend(b"pytorch", precision="64b")
         >>> pyhf.tensorlib.name
         'pytorch'
         >>> pyhf.tensorlib.precision
-        '32b'
+        '64b'
         >>> pyhf.set_backend(pyhf.tensor.numpy_backend())
         >>> pyhf.tensorlib.name
         'numpy'
@@ -60,7 +60,7 @@ def set_backend(backend, custom_optimizer=None, precision=None):
     Args:
         backend (`str` or `pyhf.tensor` backend): One of the supported pyhf backends: NumPy, TensorFlow, PyTorch, and JAX
         custom_optimizer (`pyhf.optimize` optimizer): Optional custom optimizer defined by the user
-        precision (`str`): Floating point precision to use in the backend: ``64b`` or ``32b``. Default is ``64b``.
+        precision (`str`): Floating point precision to use in the backend: ``64b`` or ``32b``. Default is backend dependent.
 
     Returns:
         None
@@ -69,20 +69,23 @@ def set_backend(backend, custom_optimizer=None, precision=None):
     global optimizer
 
     _valid_precisions = ["32b", "64b"]
-    _precision = "64b" if precision is None else precision
+    backend_kwargs = {}
 
-    if isinstance(_precision, bytes):
-        _precision = _precision.decode("utf-8")
-    _precision = _precision.lower()
+    if isinstance(precision, (str, bytes)):
+        if isinstance(precision, bytes):
+            precision = precision.decode("utf-8")
+        precision = precision.lower()
 
     if isinstance(backend, (str, bytes)):
         if isinstance(backend, bytes):
             backend = backend.decode("utf-8")
         backend = backend.lower()
+
+        if precision is not None:
+            backend_kwargs["precision"] = precision
+
         try:
-            backend = getattr(tensor, "{0:s}_backend".format(backend))(
-                precision=_precision
-            )
+            backend = getattr(tensor, "{0:s}_backend".format(backend))(**backend_kwargs)
         except TypeError:
             raise InvalidBackend(
                 "The backend provided is not supported: {0:s}. Select from one of the supported backends: numpy, tensorflow, pytorch".format(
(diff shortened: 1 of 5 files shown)


matthewfeickert:
@kratsg @lukasheinrich This would require a patch release at minimum, so should wait until `v0.5.0` is released, but if you want to give an early review to this it is welcome.

src/pyhf/__init__.py
```diff
@@ -82,6 +98,17 @@ def set_backend(backend, custom_optimizer=None):
                     backend.name, type(backend)
                 )
             )
+    if backend.precision not in _valid_precisions:
+        raise InvalidBackend(
```
matthewfeickert, line 102: I thought I could still use `InvalidBackend` as the precision is a feature of the backend.

kratsg: I'd use `exceptions.Unsupported` instead.

matthewfeickert: @kratsg Should be done now in the latest commit.

kratsg:
I don't think we should change the backend defaults as part of the PR. jax and numpy default at 64b and pytorch/tensorflow at 32b. You can look at the corresponding backend default in `__init__`.

matthewfeickert:
> I don't think we should change the backend defaults as part of the PR. jax and numpy default at 64b and pytorch/tensorflow at 32b. You can look at the corresponding backend default in `__init__`.

@kratsg This has been changed now to how you asked.

src/pyhf/__init__.py
```diff
@@ -44,44 +44,69 @@ def set_backend(backend, custom_optimizer=None):
...
+    _name_supported = getattr(tensor, f"{backend.name:s}_backend")
     if _name_supported:
         if not isinstance(backend, _name_supported):
             raise AttributeError(
-                "'{0:s}' is not a valid name attribute for backend type {1}\n                 Custom backends must have names unique from supported backends".format(
-                    backend.name, type(backend)
-                )
+                f"'{backend.name:s}' is not a valid name attribute for backend type {type(backend)}\n                 Custom backends must have names unique from supported backends"
             )
+    if backend.precision not in _valid_precisions:
```
kratsg, line 100: what is this check for? We only care about the `backend.precision` if a user is using one of our backends and tries to set it... but if it's a custom backend(?) -- then why do we care what precision they use?

src/pyhf/__init__.py
```diff
@@ -44,44 +44,69 @@ def set_backend(backend, custom_optimizer=None):
...
+        raise Unsupported(
+            f"The backend precision provided is not supported: {backend.precision:s}. Select from one of the supported precisions: {', '.join([str(v) for v in _valid_precisions])}"
+        )
+    # If "precision" arg passed, it should always win
+    # If no "precision" arg, defer to tensor backend object API if set there
+    if precision is not None:
```
kratsg, line 106: I don't understand when we get to this point. It seems like duplciated code.

src/pyhf/__init__.py
matthewfeickert: What happens if they typo the precision or try to set nonsense?

src/pyhf/__init__.py
matthewfeickert: @kratsg The example I gave in the PR body hits this:

```python
>>> import pyhf
>>> pyhf.set_backend(pyhf.tensor.pytorch_backend(precision="32b"), precision="64b")
```

src/pyhf/__init__.py
matthewfeickert: @kratsg Or are you just saying that this should get moved in a level so it only gets hit for supported backends and let them deal with it if they make their own?

kratsg: Ahh, I see.. our backends don't quite protect against this...

https://github.com/scikit-hep/pyhf/blob/161d23ac0d892e0361135326f5c1cac4c4620134/src/pyhf/tensor/numpy_backend.py#L41-L49

The thing is our `set_backend` would fail if someone created a backend that handled 128b or something. Valid precision should be up to the backends to define. We don't necessarily use that precision information anywhere except to set up the dtypemap or to uniquely identify backends.

matthewfeickert: So you would vote to just remote the precision checks entirely? Or only check for our supported backends.

kratsg: That's the key question isn't it? I am not sure what the best thing to do here is. Presumably, we should have these precision checks at the backend level, rather than at the pyhf level _but_ I am fine with this for now, since we don't necessarily have custom backends being supported and architectures likely won't be anything but 32/64...

matthewfeickert: Yeah, fair. I've added this for followup to Issue #981.

PR #988, opened 2020-07-26
feat: Provide support for returning correlations from minuit

### LINKED ISSUE #929: Expose minuit instance in fit
# Description

A fit via `pyhf.infer.mle.fit()` with minuit backend currently does not offer access to the minuit instance to for further operations, e.g. access to the correlation matrix. It would be convenient to optionally return it.

The provided `pyhf.infer.mle.fit()` can only be used for studies with limited scope, as  important information such as parameter correlations are inaccessible.

A keyword argument to optionally return the minuit instance, similar to `return_uncertainties`. Alternatively, the correlation matrix might be returned directly, though the full minuit instance might be more convenient (e.g. for Minos errors).

A user wanting more detailed control over the fit could implement it separately and not go through `infer` to bypass this limitation. This might be a good solution to prevent supporting every last use case (and will offer more detailed control over minuit, which might be desired).

#881

# Additional context
none

kratsg:
# Description

Resolves #929. ~Needs #991.~

~Do we want to support the covariance too?~ users can get it from the fit object.

ReadTheDocs build: https://pyhf.readthedocs.io/en/feat-minuitcorrelations/
- [`pyhf.optimize.mixins.OptimizerMixin`](https://pyhf.readthedocs.io/en/feat-minuitcorrelations/_generated/pyhf.optimize.mixins.OptimizerMixin.html)
- [`pyhf.optimize.opt_scipy.scipy_optimizer`](https://pyhf.readthedocs.io/en/feat-minuitcorrelations/_generated/pyhf.optimize.opt_scipy.scipy_optimizer.html)
- [`pyhf.optimize.opt_minuit.minuit_optimizer`](https://pyhf.readthedocs.io/en/feat-minuitcorrelations/_generated/pyhf.optimize.opt_minuit.minuit_optimizer.html)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Expand minuit API to return correlations if return_correlations kwarg is True
* Provide intersphinx for iminuit docs
* Add minuit options for setting tolerance and strategy
* Add tests for returning correlations
```


kratsg:
Note: this does not return a correlation matrix OUT to the user. It just simply adds in the correlation into the fitresult object... we might consider having this enabled by default (in the fitresult object) and then moving the `return_xyz` back to the `mixins` which will figure out how to return things appropriately...

src/pyhf/optimize/opt_minuit.py
```diff
@@ -84,19 +83,29 @@ def _minimize(
         underlying minimizer.
 
         Minimizer Options:
-            maxiter (`int`): maximum number of iterations. Default is 100000.
-            return_uncertainties (`bool`): Return uncertainties on the fitted parameters. Default is off.
+            maxiter (`int`): maximum number of iterations. See :attr:`~iminuit.Minuit.ncalls`. Default is ``100000``.
+            tol (`float`): tolerance for convergence. See :attr:`~iminuit.Minuit.tol`. Default is ``0.1``.
+            strategy (`int`): current minimization strategy. See :attr:`~iminuit.Minuit.strategy`. Default is ``1``.
+            return_uncertainties (`bool`): Return uncertainties on the fitted parameters. Default is off (``False``).
+            return_correlations (`bool`): Return correlations of the fitted parameters. Default is off (``False``).
```
matthewfeickert, line 90: These are nice docstrings, so it is kinda too bad they don't show up in the docs render. I'm not sure where they would show up as this is a private method, but these `kwargs` would be nice to make clear to users.

src/pyhf/optimize/opt_minuit.py
```diff
@@ -118,9 +127,16 @@ def _minimize(
         if return_uncertainties:
             unc = minimizer.np_errors()
 
+        # cov = None
+        corr = None
+        if return_correlations:
+            # cov = minimizer.hesse(maxcall=maxiter)
```
matthewfeickert, line 133: In regards to

> Do we want to support the covariance too?

I guess sure? I haven't thought about this much yet, so is there a clear reason why we might not want to? It seems that since we're already adding correlations we might as well add this in as well as I suspect that people will want it.

alexander-held:
There's a possibly slightly confusing interaction with fixed parameters. `iminuit` includes fixed parameters in its list of parameter values / uncertainties, but not in the correlation matrix. Users therefore need different labels for pulls / correlations (or maybe better, remove the fixed parameters from the list of parameters returned by `iminuit`). Not sure whether this implies any changes for this PR, but I just came across it and figured I should mention it in case it's useful.

src/pyhf/optimize/opt_minuit.py
kratsg: Migrated the last two options here up to `mixin.minimize` so it will show up in documentation.

src/pyhf/optimize/opt_minuit.py
kratsg: Users can get it from the fit object. It's not as commonly requested as the correlation (normalized covariance).

matthewfeickert:
cc @alexander-held on this again as well, given that it is ready for review.

alexander-held:
Nothing else from my side, this looks good to me and works fine in a test setup.

src/pyhf/optimize/mixins.py
```diff
@@ -105,8 +124,10 @@ def minimize(
             init_pars (:obj:`list`): initial parameters
             par_bounds (:obj:`list`): parameter boundaries
             fixed_vals (:obj:`list`): fixed parameter values
-            return_fitted_val (:obj:`bool`): return bestfit value of the objective
-            return_result_obj (:obj:`bool`): return :class:`scipy.optimize.OptimizeResult`
+            return_fitted_val (:obj:`bool`): Return bestfit value of the objective. Default is off (``False``).
+            return_result_obj (:obj:`bool`): Return :class:`scipy.optimize.OptimizeResult`. Default is off (``False``).
+            return_uncertainties (:obj:`bool`): Return uncertainties on the fitted parameters. Default is off (``False``).
+            return_correlations (`bool`): Return correlations of the fitted parameters. Default is off (``False``).
```
matthewfeickert, line 130:
```suggestion
            return_correlations (:obj:`bool`): Return correlations of the fitted parameters. Default is off (``False``).
```

small style fix

src/pyhf/optimize/opt_minuit.py
```diff
@@ -1,20 +1,20 @@
...
+from .. import exceptions
 from .mixins import OptimizerMixin
 import scipy
 import iminuit
 
 
 class minuit_optimizer(OptimizerMixin):
     """
-    Optimizer that uses iminuit.Minuit.migrad.
+    Optimizer that minimizes via :func:`iminuit.Minuit.migrad`.
```
matthewfeickert, line 10: I'm not sure why, but this isn't getting linked to https://iminuit.readthedocs.io/en/stable/reference.html#iminuit.Minuit.migrad properly

src/pyhf/optimize/opt_minuit.py
```diff
@@ -119,20 +117,20 @@ def _minimize(
             if fmin.is_above_max_edm:
                 message += " Estimated distance to minimum too large."
 
-        n = len(x0)
-        hess_inv = default_backend.ones((n, n))
+        hess_inv = None
```
matthewfeickert, line 120: Just because I'm not tracking, what's the motivation to have this be `None` by default over the previous `hess_inv = default_backend.ones((n, n))`?

matthewfeickert (commented):
@kratsg This all is looking good (also very clean :sparkles:). Just a suggestion and a few questions from me.

src/pyhf/optimize/opt_minuit.py
kratsg: Just refactored so that we don't need the `hess_inv` to be a numpy array of ones anymore. We didn't do any calculation using it. And nothing should depend on the hessian always existing.

src/pyhf/optimize/opt_minuit.py
kratsg: Unsure either.

```
$ python -m sphinx.ext.intersphinx <(curl -s https://iminuit.readthedocs.io/en/stable/objects.inv)
```

this shows me

```
py:method
	... 
	iminuit.Minuit.migrad                    reference.html#iminuit.Minuit.migrad
	... 
```

which indicates that `migrad` is there correctly as a function.

matthewfeickert: Strange. :/ Well, maybe we can make this an Issue to look more at later after this is merged?

matthewfeickert (approved):

matthewfeickert (approved):

lukasheinrich (approved):

PR #989, opened 2020-07-26
feat: Bookkeep fixed parameters

kratsg:
# Description

This will help with #846 .

The goal of this PR is to:
- [x] allow for `pdf.config.fixed_pars` to provide a list of True/False of parameters which are fixed. This can be essentially stitched with the inits provided by the user, or `pdf.config.suggested_init()` to get `fixed_vals` for the optimizer API.
  - [x] tests are added
- ~[ ] update the schema (we were idiots) to change `fixed` to be an array as it needs to be defined for each parameter~
  - ~[ ] tests are added~
  - ~we can additionally allow for a single bool *and* an array of bools. The single bool could provide backwards compatibility implying all parameters for the modifier are specified as fixed.~
- ~[ ] update `src/pyhf/readxml.py` to correctly read in `ParamSetting` fixed parameters correctly~
  - ~[ ] tests are added~
- ~[ ] update `src/pyhf/writexml.py` to write out fixed parameters correctly against the new schema~
  - ~[ ] tests are added~

This PR will not propagate fixed parameters from the `pdf` through to the optimizers. That is the goal of #846.

Help requested:
- ~we need examples of HistFactory XMLs that fixed one or two parameters of a given modifier (rather than all of them).~
  - decided for now paramsetting will fix all parameters of a given systematic

Bugs identified(?)
- ~in `src/pyhf/readxml.py`, `inits` suffers a similar bug that `fixed` does, where the `inits` for a modifier with n-parameters will be a list of one value.~
  - inits (like fixed) will be for all parameters of a systematic

Brain dump:
- ~for `readxml.py`: use `re` or `parse` to parse out parameter indices in `ParamSetting.text` values. I will need an example (see "Help requested") but I assume `gamma_stat_SR0L_IStR_cuts[2]` would be written into `ParamSetting` and we need to extract the index of the parameter that such a thing is being configured for (parameter setting) -- and if there is no indexing involved -- this is assumed for all parameters in that modifier (modifier setting).~
- ~for `readxml.py`: if we parse out parameter setting, we're going to need to figure out how many parameters the associated modifier has in total (only for: staterror, shapesys, shapefactor) so we can stitch this in appropriately~
- ~perhaps to preserve backwards compatibility, we can move the complexity out of `readxml/writexml` and handle parameter parsing inside `pdf.py`. Instead of trying to go from `gamma_stat_SR0L_IStR_cuts[2]` to `gamma_stat_SR0L_IStR_cuts` (or backwards) in read/write XML... we can just write out `gamma_stat_SR0L_IStR_cuts[2]` as literally the name of the parameter configuration which our JSON is technically for, rather than modifier configuration. As such, every parameter is single-bin -- so the complexity is moved over to trying to reinflate these parameters and inject them into the corresponding modifiers.~

ReadTheDocs build: https://pyhf.readthedocs.io/en/feat-bookkeepfixedparameters

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add functionality into _ModelConfig to bookkeep fixed systematics (all parameters for a given systematic)
* Add fixed into tests for paramsets
* Add tests for fixed parameters
```


src/pyhf/modifiers/staterror.py
```diff
@@ -19,6 +19,7 @@ def required_parset(cls, sample_data, modifier_data):
             'is_shared': True,
             'inits': (1.0,) * len(sample_data),
             'bounds': ((1e-10, 10.0),) * len(sample_data),
+            'fixed': False,
```
kratsg, line 22: For now, systematic-level application.. but we'd want to change to `(False,)*len(sample_data)` later?

src/pyhf/modifiers/shapesys.py
```diff
@@ -27,6 +27,7 @@ def required_parset(cls, sample_data, modifier_data):
             'is_shared': False,
             'inits': (1.0,) * n_parameters,
             'bounds': ((1e-10, 10.0),) * n_parameters,
+            'fixed': False,
```
kratsg, line 30: For now, systematic-level application.. but we'd want to change to `(False,)*len(sample_data)` later?

src/pyhf/modifiers/shapefactor.py
```diff
@@ -19,6 +19,7 @@ def required_parset(cls, sample_data, modifier_data):
             'is_shared': True,
             'inits': (1.0,) * len(sample_data),
             'bounds': ((0.0, 10.0),) * len(sample_data),
+            'fixed': False,
```
kratsg, line 22: For now, systematic-level application.. but we'd want to change to `(False,)*len(sample_data)` later?

src/pyhf/parameters/paramsets.py
```diff
@@ -2,25 +2,34 @@
 
 
 class paramset(object):
+    __slots__ = ['n_parameters', 'suggested_init', 'suggested_bounds', 'fixed']
+
     def __init__(self, **kwargs):
         self.n_parameters = kwargs.pop('n_parameters')
         self.suggested_init = kwargs.pop('inits')
         self.suggested_bounds = kwargs.pop('bounds')
+        self.fixed = kwargs.pop('fixed')
```
kratsg, line 11: NB: this requires all paramsets to be defined with `fixed` kwarg.

tests/test_schema.py
```diff
@@ -257,6 +257,34 @@ def test_parameters_duplicated():
...
+                        'modifiers': [
+                            {'name': 'mypoi', 'type': 'normfactor', 'data': None}
+                        ],
+                    },
+                ],
+            }
+        ],
+        'parameters': [{'name': 'mypoi', 'inits': [1], 'fixed': True}],
+    }
+    pyhf.Model(spec, poi_name='mypoi')
```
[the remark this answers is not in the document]
kratsg: we're checking to ensure there's no exception raised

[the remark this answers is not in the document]
matthewfeickert: Yes, I guess I still would have assumed we'd do an assert for existence. But as I said that's not required.

src/pyhf/parameters/paramsets.py
```diff
@@ -2,25 +2,34 @@
 
 
 class paramset(object):
+    __slots__ = ['n_parameters', 'suggested_init', 'suggested_bounds', 'fixed']
```
lukasheinrich, line 5: why do we need `__slots__`?

src/pyhf/parameters/paramsets.py
```diff
@@ -2,25 +2,34 @@
...
+
     def __init__(self, **kwargs):
         self.n_parameters = kwargs.pop('n_parameters')
         self.suggested_init = kwargs.pop('inits')
         self.suggested_bounds = kwargs.pop('bounds')
+        self.fixed = kwargs.pop('fixed')
 
 
 class unconstrained(paramset):
+    __slots__ = ['constrained']
```
lukasheinrich, line 15: why do we need `__slots__`?

src/pyhf/parameters/paramsets.py
kratsg: Only reason really is to save space and ensure we're not adding any other extra attributes to the paramsets. This also helps automatically document all the available properties on a paramset. In this case, it simply formalizes what we're already doing.

src/pyhf/parameters/paramsets.py
kratsg: Only reason really is to save space and ensure we're not adding any other extra attributes to the paramsets. This also helps automatically document all the available properties on a paramset. In this case, it simply formalizes what we're already doing.

lukasheinrich: save space?

kratsg: https://docs.python.org/3/reference/datamodel.html#slots

src/pyhf/parameters/utils.py
```diff
@@ -27,6 +28,7 @@ def reduce_paramsets_requirements(paramsets_requirements, paramsets_user_configs
         combined_paramset = {}
         for k in paramset_keys:
             for paramset_requirement in paramset_requirements:
+                # undefined: the modifier does not support configuring that property
```
lukasheinrich, line 31: what does this comment refer to?

src/pyhf/pdf.py
```diff
@@ -216,6 +217,16 @@ def _nominal_and_modifiers_from_spec(config, spec):
...
 class _ModelConfig(_ChannelSummaryMixin):
+    __slots__ = [
+        'par_map',
+        'par_order',
+        'poi_name',
+        'poi_index',
+        'auxdata',
+        'auxdata_order',
+        'modifier_settings',
+    ]
```
lukasheinrich, line 228: do we want to keep `slots` here?

matthewfeickert: @kratsg

src/pyhf/parameters/utils.py
kratsg: just a reminder of what 'undefined' meant in this context... I forgot. It'll help when going through this code again in the future.

lukasheinrich: ok

lukasheinrich:
LGTM

alexander-held:
Hi, just gave this a try. I think `pdf.config.fixed_pars` in the description should mean `pdf.config.suggested_fixed()`. I wonder whether "suggested" could be misleading, but I guess it depends on whether one views the `measurements` part of the workspace as a suggestion for how to perform a measurement, or one views the whole workspace as an instruction for doing one specific fit (I personally perceive it more like the second option).

matthewfeickert:
> Hi, just gave this a try. I think `pdf.config.fixed_pars` in the description should mean `pdf.config.suggested_fixed()`. I wonder whether "suggested" could be misleading, but I guess it depends on whether one views the `measurements` part of the workspace as a suggestion for how to perform a measurement, or one views the whole workspace as an instruction for doing one specific fit (I personally perceive it more like the second option).

@alexander-held Can you open this up as an Issue for us to discuss there (also hopefully making it easier for others to find for future reference)?

kratsg:
> Hi, just gave this a try. I think `pdf.config.fixed_pars` in the description should mean `pdf.config.suggested_fixed()`

Fixed.

> I wonder whether "suggested" could be misleading, but I guess it depends on whether one views the `measurements` part of the workspace as a suggestion for how to perform a measurement, or one views the whole workspace as an instruction for doing one specific fit (I personally perceive it more like the second option).

Part of this is to separate out a model definition from a measurement definition. The measurement definitions come into play with `suggested_xyz` accessors -- as you probably saw (inits, bounds, fixed).

alexander-held:
I'll open an issue as suggested by @matthewfeickert, but in the end it's matter of taste what to call it I think.

PR #991, opened 2020-07-26
fix: stitch_pars bug in stitching values other than fixed values

kratsg:
# Description

Unfortunately, #951 wasn't perfect. This fixes one minor bug 

https://github.com/scikit-hep/pyhf/blob/27f35e98739adc7a91d67c8a99581aa1020d407d/src/pyhf/optimize/common.py#L99-L101

where anything specified in `stitch_with` is unused, as `fixed_values` is just used directly. To also make sure this doesn't bite us in the future, I refactored the function slightly, and added tests for the expected behavior/functionality.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Fix bug in stitch_pars to use stitch_with argument
* Refactor stitch_pars to inside _make_stitch_pars function 
* Add test for stitch_pars
```

### DIFF
--- a/src/pyhf/optimize/common.py
+++ b/src/pyhf/optimize/common.py
@@ -5,16 +5,20 @@ from ..tensor.common import _TensorViewer
 
 def _make_stitch_pars(tv=None, fixed_values=None):
     """
-    Construct a callable to stitch fixed paramter values into the unfixed parameters.
+    Construct a callable to stitch fixed paramter values into the unfixed parameters. See :func:`shim`.
+
+    This is extracted out to be unit-tested for proper behavior.
+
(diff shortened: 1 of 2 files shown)


src/pyhf/optimize/common.py
```diff
@@ -3,6 +3,17 @@
 from ..tensor.common import _TensorViewer
 
 
+def _make_stitch_pars(tv=None, fixed_values=None):
```
matthewfeickert, line 6: @kratsg Can we get a short docstring here so that it is more evident in the future why this exists?

matthewfeickert:
Force pushing to try to bump CI out of problem with `jaxlib`.

PR #993, opened 2020-07-27
feat: Add kwarg for AsymptoticCalculator base distribution

### LINKED ISSUE #1159: support variations on CLs dealing w/ inclusive pvalue and clipping
# Description

enable support for computing CLs values by correctly clipping µ^/sigma to values < µ_signal. The standard ROOT asymptoticcalculator implementation does not do this.

One advantage of this is that the "kink" seen in overfluctuations is much easier to explain 

image

lukasheinrich:
# Description

Resolves #1159

In the standard brazil band plot ROOT HypotestInverter produces, there is a bit of an imprecision as the expected limits are calculated in $mu^/sigma$ space but for significant upward fluctuations it returns p-values that are not actually realizable with any data since if $expected µ^ > µ$ the test stat can at most be the one returned for $µ=µ^$, namely $q_µ = 0$, that in turn means that the p-value is capped at $CLs_obs$.

For compatibility and recognizability the default behavior is unchanged but it's now possible to get the correct asymptotic bahaviorr with `calc_kwargs = {'base_dir': 'clipped_normal'}`

image

ReadTheDocs build: https://pyhf.readthedocs.io/en/clipped_expected/api.html#inference

- [`pyhf.infer.calculators.AsymptoticTestStatDistribution`](https://pyhf.readthedocs.io/en/clipped_expected/_generated/pyhf.infer.calculators.AsymptoticTestStatDistribution.html#pyhf.infer.calculators.AsymptoticTestStatDistribution)
- [`pyhf.infer.calculators.AsymptoticCalculator`](https://pyhf.readthedocs.io/en/clipped_expected/_generated/pyhf.infer.calculators.AsymptoticCalculator.html)
- [`pyhf.infer.calculators.ToyCalculator`](https://pyhf.readthedocs.io/en/clipped_expected/_generated/pyhf.infer.calculators.ToyCalculator.html)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add calc_base_dist kwarg to AsymptoticCalculator to control cutoff for AsymptoticTestStatDistribution
* Add test for calc_base_dist behavior for 'clipped_normal'
* Add docstring example for pvalue
* Update docstring for AsymptoticCalculator and ToyCalculator

```

### DIFF
--- a/src/pyhf/infer/calculators.py
+++ b/src/pyhf/infer/calculators.py
@@ -89,7 +89,7 @@ class AsymptoticTestStatDistribution:
             >>> import pyhf
             >>> pyhf.set_backend("numpy")
             >>> bkg_dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0.0)
-            >>> bkg_dist.pvalue(0)
+            >>> bkg_dist.cdf(0.0)
             0.5
 
         Args:
@@ -119,6 +119,14 @@ class AsymptoticTestStatDistribution:
             \frac{(\mu-\mu')}{\sigma} = \sqrt{\Lambda}= \sqrt{q_{\mu,A}}
 
         given the observed test statistics :math:`q_{\mu}` and :math:`q_{\mu,A}`.
+
+        Example:
+
+            >>> import pyhf
+            >>> pyhf.set_backend("numpy")
+            >>> bkg_dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0.0)
+            >>> bkg_dist.pvalue(0.0)
+            array(0.5)
 
         Args:
             value (:obj:`float`): The test statistic value.
@@ -227,7 +235,7 @@ class AsymptoticCalculator:
             >>> _ = asymptotic_calculator.teststatistic(mu_test)
             >>> sig_plus_bkg_dist, bkg_dist = asymptotic_calculator.distributions(mu_test)
             >>> sig_plus_bkg_dist.pvalue(mu_test), bkg_dist.pvalue(mu_test)
-            (0.002192624107163899, 0.15865525393145707)
+            (array(0.00219262), array(0.15865525))
 
         Args:
             poi_test (:obj:`float` or :obj:`tensor`): The value for the parameter of interest.
@@ -351,7 +359,7 @@ class AsymptoticCalculator:
             >>> sig_plus_bkg_dist, bkg_dist = asymptotic_calculator.distributions(mu_test)
             >>> CLsb, CLb, CLs = asymptotic_calculator.pvalues(q_tilde, sig_plus_bkg_dist, bkg_dist)
             >>> CLsb, CLb, CLs
-            (0.023325019427864607, 0.4441593996111411, 0.05251497423736956)
+            (array(0.02332502), array(0.4441594), 0.05251497423736956)
 
         Args:
             teststat (:obj:`tensor`): The test statistic.
--- a/src/pyhf/infer/calculators.py
+++ b/src/pyhf/infer/calculators.py
@@ -66,7 +66,7 @@ class AsymptoticTestStatDistribution:
     the :math:`-\mu'`, where :math:`\mu'` is the true poi value of the hypothesis.
     """
 
-    def __init__(self, shift, cutoff):
+    def __init__(self, shift, cutoff=float('-inf')):
         """
         Asymptotic test statistic distribution.
 
--- a/src/pyhf/infer/calculators.py
+++ b/src/pyhf/infer/calculators.py
@@ -209,11 +209,13 @@ class AsymptoticCalculator:
             calc_base_dist (:obj:`str`): The statistical distribution, ``'normal'`` or
               ``'clipped_normal'``, to use for calculating the :math:`p`-values.
 
-              * ``'normal'``: (default) use the full Normal distribution in :math:`\mu/\sigma` space.
+              * ``'normal'``: (default) use the full Normal distribution in :math:`\hat{\mu}/\sigma`
+                space.
                 Note that expected limits may correspond to unphysical test statistics from scenarios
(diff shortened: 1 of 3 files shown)


src/pyhf/infer/calculators.py
```diff
@@ -125,8 +143,17 @@ def distributions(self, poi_test):
         """
         if self.sqrtqmuA_v is None:
             raise RuntimeError('need to call .teststatistic(poi_test) first')
-        sb_dist = AsymptoticTestStatDistribution(-self.sqrtqmuA_v)
-        b_dist = AsymptoticTestStatDistribution(0.0)
+
+        if self.base_distr == 'normal':
+            cutoff = -1e10
```
kratsg, line 148:
```suggestion
            cutoff = float('-inf')
```

src/pyhf/infer/calculators.py
```diff
@@ -125,8 +143,17 @@ def distributions(self, poi_test):
...
+        elif self.base_distr == 'clipped_normal':
+            cutoff = -self.sqrtqmuA_v
+        else:
+            raise RuntimeError(
```
kratsg, line 152: Shouldn't this error be caught at construction (rather than waiting on a call to `distributions`)?

Additionally, maybe `Unsupported` for this error as it's an unsupported base distribution configuration.

lukasheinrich: maybe `ValueError` is enuogh?

matthewfeickert:
@lukasheinrich I rebased this but have a pre-rebase version at `backup-clipped_expected` in case I borked anything too badly. :grimacing:.

We should discuss all of this again, but as we've introduced `kwargs` since this PR started I changed `calc_kwargs` to just be a `calc_base_dist` kwarg. With this rebase I'm able to have the following code snippet produce the following plots which you made earlier. 

```python
import pyhf
import pyhf.contrib.viz.brazil as barzil_band
import numpy as np

    signal_data=[20.0], bkg_data=[50.0], bkg_uncerts=[7.0]
)
data = [70.0] + model.config.auxdata

poi_vals = np.linspace(0, 2.5)
results = [
    pyhf.infer.hypotest(
        test_poi, data, model, return_expected_set=True, calc_base_dist="normal"
    )
    for test_poi in poi_vals
]

fig, ax = plt.subplots()
barzil_band.plot_results(ax, poi_vals, results)
fig.savefig("base_dist_normal.png")

results = [
    pyhf.infer.hypotest(
        test_poi, data, model, return_expected_set=True, calc_base_dist="clipped_normal"
    )
    for test_poi in poi_vals
]

fig, ax = plt.subplots()
barzil_band.plot_results(ax, poi_vals, results)
fig.savefig("base_dist_clipped_normal.png")
```

| base_dist_normal.png | base_dist_clipped_normal.png |
|-|-|
|base_dist_normal|base_dist_clipped_normal|


The tests pass also beyond having to fixup doctest, so I think this is moving in the right direction.

matthewfeickert:
This PR also addresses some of Issue #1268, which while not the goal is nice. :+1:

matthewfeickert:
> The changes look fine. Additional docstrings are needed to describe the new APIs.

In addition to the APIs being updated the docstring for `AsymptoticCalculator` should also get fixed up, as `qtilde` is no longer an arg

https://github.com/scikit-hep/pyhf/blob/9d56349868d990c66d7b9cf776be200a35fda18d/src/pyhf/infer/calculators.py#L166-L182

matthewfeickert:
@kratsg @lukasheinrich This PR still needs to address the docstring requests that @kratsg and I made, but beyond that I think is ready for review. If you both make comments I will implement anything else that needs to get taken care of. :+1:

src/pyhf/infer/calculators.py
```diff
@@ -173,13 +194,24 @@ def __init__(
...
+              (:func:`~pyhf.infer.test_statistics.qmu`), and ``'q0'`` perfoms the calculation using
+              the discovery test statistic :math:`q_{0}` (:func:`~pyhf.infer.test_statistics.q0`).
+            calc_base_dist (:obj:`str`): The statistical distribution, ``'normal'`` or
+              ``'clipped_normal'`` to use for calculating the :math:`p`-values.
+              ``'normal'`` (default) corresponds to the common misconception that :math:`p`-values
+              greater than the observed :math:`p`-value are obtainable for POI less than 1.
+              ``'clipped_normal'`` corresponds to limiting the :math:`p`-values for expected limits
+              to a maximum of the observed :math:`p`-value.
+              The choice of ``calc_base_dist`` only affects the :math:`p`-values for expected limits,
+              and the default value will be changed in a future release.
```
matthewfeickert, line 214: @lukasheinrich Please review this (c.f. the linked render on RTD in the PR body) for accuracy. I tried to keep this concise, but it is more important to be clear and correct.

matthewfeickert:
Also maybe worth discussing if the changes to the `EmpiricalDistribution` from PR #1160 should be migrated to this PR as well.

src/pyhf/infer/calculators.py
lukasheinrich: i'd say


 `normal: use full normal distribution in µ/sigma space. note: expected limits may correspond to scenarios   with  $expected µ^ > µ$`

 `clipped_normal: use clipped normal distribution in µ/sigma space. to avoid expected limits that correspond to scenarios   with  $expected µ^ > µ$`

lukasheinrich:
can't aprrovemy own PR but thanks for rebasing. lgtm.

matthewfeickert (approved):

src/pyhf/infer/calculators.py
matthewfeickert: Updated. :+1:

kratsg (approved):
Will continue discussion in #1310 to not hold up this PR any longer.

Issue #867, opened 2020-05-10
Obtaining fit parameter labels


alexander-held:
# Question

Is there a convenient way to obtain the labels corresponding to the parameters returned by `pyhf.infer.mle.fit()`? Setting up a fit is rather clean now with `infer`:
```
ws = pyhf.Workspace(spec)
model = ws.model()
data = ws.data(model)
pyhf.infer.mle.fit(data, model)
```
In comparison, the function in https://github.com/scikit-hep/pyhf/blob/master/docs/examples/notebooks/pullplot.ipynb to get the names of those parameters is difficult to parse for non-experts:
```
labels = [
        "{}[{}]".format(k, i) if model.config.param_set(k).n_parameters > 1 else k
        for k in model.config.par_order
        for i in range(model.config.param_set(k).n_parameters)
    ]
```
Is there a better way to get this list of parameter names, and if not, would it be possible to add this as a utility within `pyhf`?

# Relevant Issues and Pull Requests

not aware of any

matthewfeickert:
> Is there a convenient way to obtain the labels corresponding to the parameters returned by `pyhf.infer.mle.fit()`?

As in using the public API? Not at the moment, but this seems reasonable enough that we can look into adding this.

matthewfeickert:
@alexander-held I guess in some sense it depends on what you mean by labels. `model.config.parameters` exists

```python
>>> import pyhf
>>> model = pyhf.simplemodels.hepdata_like(
...     signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0]
... )
>>> model.config.parameters
['mu', 'uncorr_bkguncrt']
```
though it doesn't make it clear here that ` 'uncorr_bkguncrt'` contains 2 paramters:

```python
>>> model.config.npars
3
>>> model.config.param_set("mu").n_parameters
1
>>> model.config.param_set("uncorr_bkguncrt").n_parameters
2
```

kratsg:
Just throwing out another way to do it. We can certainly provide a nicer API for it.

```
    pars_pyhf = []
    for k, v in model.config.par_map.items():
        sl = v["slice"]
        npars = sl.stop - sl.start
        if npars > 1 or "staterror" in k:
            for i in range(npars):
                pars_pyhf.append("{}_{}".format(k, i))
        else:
            pars_pyhf.append(k)
```

lukasheinrich:
[...]

PR #994, opened 2020-07-27
feat: Experimental query language support for parameter indexing

kratsg:
# Description

This adds both a `pyhf.experimental` (sub)module as well as `pyhf.experimental.query` to support various ways of querying the PDFs.

Documentation is needed. ./cc @alexander-held

- [x] Tests are passing

# Before Merging

For the PR Assignees:


src/pyhf/utils.py
```diff
@@ -111,3 +112,55 @@ def digest(obj, algorithm='sha256'):
...
     return hash_alg(stringified).hexdigest()
+
+
+def parse_parameter_name(expression):
+    """
+    Convert a parameter name expression to the corresponding parameter index.
+
+    Returns:
+        par_name (`str`): parameter name
+        slice (`slice`): slice/indexing to use
```
kratsg, line 123: can return either a slice or index `int`

alexander-held:
I gave this a try and this feature is extremely useful. It helps a lot with handling the mapping between `model.config.par_order` and the actual "flattened" parameters expected by `pyhf.pdf.Model.expected_data` etc., which go into a fit.

kratsg:
> I gave this a try and this feature is extremely useful. It helps a lot with handling the mapping between `model.config.par_order` and the actual "flattened" parameters expected by `pyhf.pdf.Model.expected_data` etc., which go into a fit.

can you provide more details on the use case?

alexander-held:
The main use case I have in mind is holding groups of parameters fixed in the fit. With appropriate naming, I can get for example all indices of parameters called `ttbar_modeling_*`, which might match everything related to ttbar modeling. From the comparison of the corresponding fit (with those constant) to the nominal fit I can calculate (via quadrature subtraction) the contribution of those parameters to the POI uncertainty. 

So far I got the relevant indices by going through `model.config.par_order` and (given that each entry can correspond to more than one parameter in the fit), had to also consider `model.config.param_set(param_name).n_parameters` for each parameter.

I just looked at the code in this PR and noticed `model.config.par_slice`, which I had overlooked before. That already simplifies things quite a bit compared to my current approach.

One other thing that could also be useful is the opposite direction: given an index (within `range(model.config.npars)`), return the name of the associated parameter. I use that in a few places for logging purpose. The issue there though is that the naming of per-bin parameters like `staterror` would need to be defined (#867).

PR #1002, opened 2020-07-28
build: Begin move to uproot4

matthewfeickert:
# Description

Following PR #930 and the release of `v0.5.0`, this PR now moves to adopting `uproot4` and its API.

As of [`uproot4` `v4.0.0`](https://github.com/scikit-hep/uproot4/releases/tag/4.0.0) `uproot4` does not support the ability to write to ROOT files, so `uproot3` must still be retained to write to ROOT. This is the only remaining requirement for `uproot3` and so the move to the `uproot4` API can begin now.

As `uproot4` is also added to the HEAD of dependnecies workflow, the pattern of

```
import uproot
```

is used instead of

```
import uproot4 as uproot
```

as installing `uproot4` from GitHub installs under the `uproot` namespace. So to test the HEAD of `uproot4`, the `uproot` namespace must be used now instead of waiting to drop `uproot3`.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Require uproot4 in the xmlio extra
   - Require release compatible with uproot v4.X series 
* Adopt the uproot4 v4.0 API
   - uproot3 still needed for writing ROOT files
* Add uproot4 to the HEAD of dependencies workflow
   - import uproot namespace everywhere uproot4 is used to allow for testing of uproot4 HEAD
```

### DIFF
--- a/.github/workflows/dependencies-head.yml
+++ b/.github/workflows/dependencies-head.yml
@@ -102,7 +102,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install --ignore-installed --upgrade -q --no-cache-dir -e .[test]
-        python -m pip uninstall --yes uproot4 uproot
+        python -m pip uninstall --yes uproot
         python -m pip install --upgrade --no-cache-dir git+git://github.com/scikit-hep/uproot4.git
         python -m pip list
     - name: Test with pytest
--- a/.github/workflows/dependencies-head.yml
+++ b/.github/workflows/dependencies-head.yml
@@ -84,6 +84,31 @@ jobs:
       run: |
         python -m pytest -r sx --ignore tests/benchmarks/ --ignore tests/contrib --ignore tests/test_notebooks.py
 
+  uproot4:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: [3.8]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        python -m pip install --ignore-installed --upgrade -q --no-cache-dir -e .[test]
+        python -m pip uninstall --yes uproot4 uproot
+        python -m pip install --upgrade --no-cache-dir git+git://github.com/scikit-hep/uproot4.git
+        python -m pip list
+    - name: Test with pytest
+      run: |
+        python -m pytest -r sx --ignore tests/benchmarks/ --ignore tests/contrib --ignore tests/test_notebooks.py
+
   pytest:
 
     runs-on: ${{ matrix.os }}
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -23,7 +23,7 @@ def extract_error(hist):
     bin uncertainties are then Poisson, and so the `sqrt(entries)`.
 
     Args:
-        hist (`<dynamic>.Model_TH1F_v2`): The histogram
+        hist (:class:`uproot.behaviors.TH1.TH1`): The histogram
 
     Returns:
         list: The uncertainty for each bin in the histogram
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -6,7 +6,7 @@ from pathlib import Path
 import xml.etree.ElementTree as ET
 import numpy as np
 import tqdm
-import uproot4 as uproot
+import uproot
 import re
 
 log = logging.getLogger(__name__)
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -6,7 +6,7 @@ from pathlib import Path
 import xml.etree.ElementTree as ET
 import numpy as np
 import tqdm
-import uproot3 as uproot
+import uproot4 as uproot
 import re
 
 log = logging.getLogger(__name__)
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -232,7 +232,7 @@ def process_measurements(toplvl, other_parameter_configs=None):
     For a given XML structure, provide a parsed dictionary adhering to defs.json/#definitions/measurement.
 
     Args:
-        toplvl (:module:`xml.etree.ElementTree`): The top-level XML document to parse.
+        toplvl (:mod:`xml.etree.ElementTree`): The top-level XML document to parse.
         other_parameter_configs (:obj:`list`): A list of other parameter configurations from other non-top-level XML documents to incorporate into the resulting measurement object.
 
     Returns:
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -48,7 +48,7 @@ def import_root_histogram(rootdir, filename, path, name, filecache=None):
         f = filecache[fullpath]
     try:
         hist = f[name]
-    except KeyError or uproot.deserialization.DeserializationError:
+    except (KeyError, uproot.deserialization.DeserializationError):
         try:
             hist = f[str(Path(path).joinpath(name))]
         except KeyError:
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -52,7 +52,7 @@ def import_root_histogram(rootdir, filename, path, name, filecache=None):
         print(f"\nfile: {f}")
         print(f"file name: {name}")
         hist = f[name]
-    except KeyError or uproot4.deserialization.DeserializationError:
+    except KeyError or uproot.deserialization.DeserializationError:
         try:
             hist = f[str(Path(path).joinpath(name))]
         except KeyError:
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -14,7 +14,7 @@ log = logging.getLogger(__name__)
 __FILECACHE__ = {}
 
 
-def extract_error(h):
+def extract_error(hist):
     """
     Determine the bin uncertainties for a histogram.
 
@@ -23,12 +23,12 @@ def extract_error(h):
     bin uncertainties are then Poisson, and so the `sqrt(entries)`.
 
     Args:
-        h (uproot.rootio.TH1 object): The histogram
+        hist (uproot.rootio.TH1 object): The histogram
 
     Returns:
         list: The uncertainty for each bin in the histogram
     """
-    err = h.variances if h.variances.any() else h.numpy()[0]
+    err = hist.variances if hist.variances.any() else hist.to_numpy()[0]
     return np.sqrt(err).tolist()
 
 
@@ -46,15 +46,15 @@ def import_root_histogram(rootdir, filename, path, name, filecache=None):
     else:
         f = filecache[fullpath]
     try:
-        h = f[name]
+        hist = f[name]
     except KeyError:
         try:
-            h = f[str(Path(path).joinpath(name))]
+            hist = f[str(Path(path).joinpath(name))]
         except KeyError:
             raise KeyError(
                 f'Both {name} and {Path(path).joinpath(name)} were tried and not found in {Path(rootdir).joinpath(filename)}'
             )
-    return h.numpy()[0].tolist(), extract_error(h)
+    return hist.to_numpy()[0].tolist(), extract_error(hist)
 
 
 def process_sample(
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -29,10 +29,8 @@ def extract_error(hist):
         list: The uncertainty for each bin in the histogram
     """
 
-    sumw2 = hist.member("fSumw2", none_if_missing=True)
-    err = sumw2 if sumw2 else hist.to_numpy()[0]
-    print(f"\nerr: {err}\n")
-    return np.sqrt(err).tolist()
+    variance = hist.variances() if hist.weighted else hist.to_numpy()[0]
+    return np.sqrt(variance).tolist()
 
 
 def import_root_histogram(rootdir, filename, path, name, filecache=None):
@@ -49,8 +47,6 @@ def import_root_histogram(rootdir, filename, path, name, filecache=None):
     else:
         f = filecache[fullpath]
     try:
-        print(f"\nfile: {f}")
-        print(f"file name: {name}")
         hist = f[name]
     except KeyError or uproot.deserialization.DeserializationError:
         try:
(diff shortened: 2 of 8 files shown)


src/pyhf/readxml.py
```diff
@@ -22,12 +22,12 @@ def extract_error(h):
...
 
     Args:
-        h (uproot.rootio.TH1 object): The histogram
+        hist (uproot.rootio.TH1 object): The histogram
 
     Returns:
         list: The uncertainty for each bin in the histogram
     """
-    err = h.variances if h.variances.any() else h.numpy()[0]
+    err = hist.variances if hist.variances.any() else hist.to_numpy()[0]
```
matthewfeickert, line 30: @jpivarski This will raise the error

```
       err = hist.variances if hist.variances.any() else hist.to_numpy()[0]
E       AttributeError: 'Model_TH1F_v2' object has no attribute 'variances'
```

in tests, but doing a quick search through the `uproot4` codebase I don't see `variances`. Can you advise? I'll look again after I'm out of meetings.

matthewfeickert: I think https://github.com/scikit-hep/uproot/issues/511 might be what I was looking for. I'll read that later to see what the API decision was.

jpivarski: That is where it's being discussed. It's in flux because we want to get all the libraries to have the same interface. Currently, there's `values_errors()`, which will give you a 2-tuple of values and errors, which you'd have to square to get variances. IN this case, it looks like you're asking if `fSumw2` exists. For that, you might want to

```python
>>> f["hpx"].member("fSumw2", none_if_missing=True)
<TArrayD [] at 0x78f865ba4450>
```

because then you can

```python
>>> if f["hpx"].member("fSumw2", none_if_missing=True): print("exists")
... 
```

The big picture is that the many methods TH* used to have are being removed in favor of a slimmer interface that mostly converts, so that people use real histogram libraries, rather than the few methods that had been added to Uproot.

matthewfeickert: Thanks @jpivarski. @kratsg @lukasheinrich we should join the discussion over at https://github.com/scikit-hep/uproot/issues/511

src/pyhf/writexml.py
```diff
@@ -33,12 +33,12 @@ def _make_hist_name(channel, sample, modifier='', prefix='hist', suffix=''):
 
 
 def _export_root_histogram(histname, data):
-    h = TH1.from_numpy((np.asarray(data), np.arange(len(data) + 1)))
-    h._fName = histname
-    # NB: uproot crashes for some reason, figure out why later
+    hist = TH1((np.asarray(data), np.arange(len(data) + 1)))
```
matthewfeickert, line 36: @jpivarski similarly, can you comment on the new `TH1` API? It seems that given

```
    def _export_root_histogram(histname, data):
>       hist = TH1((np.asarray(data), np.arange(len(data) + 1)))
E       TypeError: TH1() takes no arguments
```

exporting arrays as histograms is no longer going to be supported. Given your earlier comment "so that people use real histogram libraries" does this mean that `pyhf` needs to become best friends with `boost-histogram`?

jpivarski: Basic array <--> histograms will still be supported, though you're hitting this limit because Uproot 4 can't write to ROOT files yet.

src/pyhf/readxml.py
```diff
@@ -22,12 +22,14 @@ def extract_error(h):
     bin uncertainties are then Poisson, and so the `sqrt(entries)`.
 
     Args:
-        h (uproot.rootio.TH1 object): The histogram
+        hist (uproot.rootio.TH1 object): The histogram
```
matthewfeickert, line 25: @jpivarski If I just add a

```python
print(hist, type(hist))
```

in the function below and use it on a test case I get

```python
<TH1F (version 2) at 0x7f8a9ee37b90> <class '<dynamic>.Model_TH1F_v2'>
```

Though if I search around in `uproot4` repo it seems that these are streamers that inherit from `Model`? I actually never looked into how `uproot3` did things either, but if we wanted to give a class description here then would 

```
hist(`uproot4.Model_TH1F_v2`)
```
?

jpivarski: Uproots 3 and 4 both dynamically create classes from streamers. In Uproot 4, it's more formal in that they all inherit from Model. If you want to verify that something is a histogram, you can check to see if it inherits from `uproot4.behaviors.TH1.Model_TH1` (approximate: please check!) because the behavioral mix-ins (what gives it the `.to_numpy()` method, etc.) are non-dynamic classes that you can count on existing.

jpivarski: Or if the class doesn't have any behavioral overloads, the correct way to do it would be a string comparison on its `classname` (which is a C++ class name, with templates). I don't know if there's a method that checks inheritance through the C++ class names, but there should be..

jpivarski: (in Uproot 3, the existence of `uproot.rootio.TH1` was an unintentional byproduct of the way the dynamic class was generated.)

src/pyhf/writexml.py
matthewfeickert: Ah good news. Can you point me to the source where I can learn the API then, as the above isn't it?

jpivarski: I mean, for Uproot 4, the data → TH1 → ROOT file pipeline doesn't exist yet, so the API hasn't been defined. It will probably be similar to Uproot 3, in which you take a histogram from a library like Boost or a tuple of arrays from a library like NumPy and

```python
root_file["name"] = histogram_data
```

(see similar examples at the end of the Uproot 3 GitHub README) though some explicit wrapper like `TH1` would make it possible to add metadata. To try to unify the reading and writing interfaces more, I think these classes should inherit from `Model`, but only one version of the class would be instrumented with write methods (and put in a module where it's clearly available for writing, e.g. `uproot4.write.TH1D`).

src/pyhf/writexml.py
```diff
@@ -32,12 +34,11 @@ def _make_hist_name(channel, sample, modifier='', prefix='hist', suffix=''):
...
-    h = TH1.from_numpy((np.asarray(data), np.arange(len(data) + 1)))
-    h._fName = histname
-    # NB: uproot crashes for some reason, figure out why later
-    # if histname in _ROOT_DATA_FILE:
-    #    raise KeyError(f'Duplicate key {histname} being written.')
-    _ROOT_DATA_FILE[histname] = h
+    hist = TH1.from_numpy((np.asarray(data), np.arange(len(data) + 1)))
+    hist._fName = histname
+    if histname in _ROOT_DATA_FILE:
+        raise KeyError(f"Duplicate key {histname} being written.")
```
matthewfeickert, line 40: We're [losing coverage on these two lines as they aren't being tested at the moment](https://codecov.io/gh/scikit-hep/pyhf/pull/1002/diff).

@kratsg This was added as commented out lines way back in PR #426. Do we need this? Or is this a vestige of old problems since resolved?

kratsg: We never had it. I think fundamentally, uproot3 did crash with this. I don't know what uproot4's behavior is meant to be here with duplicated names (@jpivarski ) especially since I don't think the writing functionality exists yet.

For Jim, what is the behavior of uprootX when a histogram stored under key `Foo` is written to a file with that key already existing?

src/pyhf/readxml.py
```diff
@@ -46,15 +47,15 @@ def import_root_histogram(rootdir, filename, path, name, filecache=None):
     else:
         f = filecache[fullpath]
     try:
-        h = f[name]
-    except KeyError:
+        hist = f[name]
+    except (KeyError, uproot.deserialization.DeserializationError):
         try:
-            h = f[str(Path(path).joinpath(name))]
+            hist = f[str(Path(path).joinpath(name))]
```
kratsg, line 53: I'm not sure this is the recommended way to do it? Do we guarantee that uproot supports these path semantics correctly (e.g. on windows)?

kratsg (approved):

src/pyhf/writexml.py
jpivarski: The intended behavior for UprootX is that a second object would be written to the ROOT file with a higher cycle number. I think I remember Uproot 3 doing that incorrectly, though.

(The "cycle number" is the number after the semicolon in a list of key names. It disambiguates objects with the same name.)

.github/workflows/dependencies-head.yml
```diff
@@ -84,6 +84,31 @@ jobs:
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        python -m pip install --ignore-installed --upgrade -q --no-cache-dir -e .[test]
+        python -m pip uninstall --yes uproot
+        python -m pip install --upgrade --no-cache-dir git+git://github.com/scikit-hep/uproot4.git
```
jpivarski, line 106: Is there a reason you're not getting this from pip?

```bash
python -m pip install --upgrade --no-cache-dir "uproot>=4.0.0rc1"
```

The main branch of the scikit-hep/uproot4 repo is _supposed_ to always be working, but actual releases have a stronger guarantee and unchanging version numbers (to debug a failure with `pip list` in the next line). The version number of the main branch is either the latest release (because I just released it and haven't updated the version yet) or what I expect the next release will be, but I could change my mind about what level of semantic versioning to update. Also, the code in the main branch changes with each PR merge, but the version number does not. Two runs of your CI might get different Uproot code but report the same version number in `pip list` on the next line.

The `>=4.0.0rc1` in the pip-install ensures that you get the latest release candidate, if that's what you're trying to ensure. (Although I think I'm only doing release candidates on Awkward Array, not Uproot.)

src/pyhf/readxml.py
```diff
@@ -23,13 +23,14 @@ def extract_error(h):
     bin uncertainties are then Poisson, and so the `sqrt(entries)`.
 
     Args:
-        h (uproot.rootio.TH1 object): The histogram
+        hist (`<dynamic>.Model_TH1F_v2`): The histogram
```
jpivarski, line 26: That dynamic class will be different for each ROOT file you open. However, all 1-dimensional histograms will be subclasses of `uproot.behaviors.TH1.TH1`.

src/pyhf/readxml.py
jpivarski: Uproot supports Windows pathnames by passing them to `os.path` routines. There are tests on Windows, and the `skhep_test_datapath` probably gives an absolute Windows pathname and thus it would have been tested, though I haven't checked.

Uproot also supports the new `Path` objects, but it does so by converting them into strings once you've passed them to `uproot.open`. Therefore, you can do all the `Path` gymnastics within the `uproot.open` argument list. Passing a `Path` object to `uproot.open` is also a way to avoid issues with colon (`:`) in the filename being interpreted as a separator between filesystem paths and within-ROOT file paths.

However, paths _within_ a ROOT file are not filesystem paths and `Path` is not accepted in `ReadOnlyDirectory.__getitem__`. Using `Path.joinpath` is probably a bad idea because it might get joined by backslashes on Windows, but ROOT (and Uproot) never accepts backslashes as a TDirectory separator, even on Windows. Also, there are semicolons (`;`) for cycle numbers and all that. These paths, which are actually Mapping keys, should be strings and should be built with string-manipulation, such as `"/".join`.

jpivarski (commented):
Just some comments inline!

.github/workflows/dependencies-head.yml
matthewfeickert: @jpivarski Yes, this is intentional. This is a nightly test that is supposed to grab whatever is at the HEAD of GitHub (hence the name of the workflow) to give us a **very** early heads up on API breaking behavior. We'd far prefer to be able to check out what is going on and prepare a PR months in advance to deal with breaking behavior then to get busy and miss a release candidate.

This also allows us to check that changes people are making for us do what we wanted (such as when we were testing iminuit changes Hans was making for us).

src/pyhf/readxml.py
matthewfeickert: Thanks. I'll change this.

kratsg:
One thing, can you use uproot4's sphinx referencing?

```python
intersphinx_mapping = {
    'uproot': ('https://uproot.readthedocs.io/en/latest/', None),
}
```

which should work pretty well in our case (unless we need to point to a specific version).

src/pyhf/readxml.py
kratsg: How do you handle the case when say `"/".join([path, name])` has a trailing slash in `path` ? Does `uproot` handle `//` ok or is this something we'd need to check and handle properly? In which case, we'd probably need a custom `joinpath` utility in `pyhf` just to handle this case, but then I could argue this might be a quick utility that `uproot` could have instead... but I'd rather not give you more work.

jpivarski: A quick check: Uproot handles stray slashes.

```python
>>> import skhep_testdata
>>> import uproot
>>> f = uproot.open(skhep_testdata.data_path("uproot-hepdata-example.root"))
>>> f["hpx"]
<TH1F (version 1) at 0x7f209a3cee50>
>>> f["/hpx"]
<TH1F (version 1) at 0x7f209a3cee50>
>>> f["hpx/"]
<TH1F (version 1) at 0x7f209a3cee50>
>>> f["/hpx/"]
<TH1F (version 1) at 0x7f209a3cee50>
>>> f["//hpx//"]
<TH1F (version 1) at 0x7f209a3cee50>

>>> f = uproot.open(skhep_testdata.data_path("uproot-nesteddirs.root"))
>>> f["one/two/tree"]
<TTree 'tree' (20 branches) at 0x7f209a1c3730>
>>> f["/one/two/tree/"]
<TTree 'tree' (20 branches) at 0x7f209a1c3730>
>>> f["//one//two//tree//"]
<TTree 'tree' (20 branches) at 0x7f209a1c3730>
```

src/pyhf/writexml.py
kratsg: Thanks! I've gone ahead and added a check for this. We generally (for `writexml`) do not want new cycle numbers here, so we need to check if the histogram name exists before adding it to the file as we're essentially building this ROOT file from scratch.

src/pyhf/readxml.py
```diff
@@ -23,13 +23,14 @@ def extract_error(h):
     bin uncertainties are then Poisson, and so the `sqrt(entries)`.
 
     Args:
-        h (uproot.rootio.TH1 object): The histogram
+        hist (:class:`uproot.behaviors.TH1.TH1`): The histogram
```
kratsg, line 26: @jpivarski can you confirm is this is what you'd expect users to understand (documentation-wise)?

jpivarski: Yes, that's the most useful way to annotate the type of the histogram. It's the superclass that is persistent, documented, and has all of the user-oriented methods defined.

It should point here: https://uproot.readthedocs.io/en/latest/uproot.behaviors.TH1.TH1.html (if the intersphinx stuff works as expected).

kratsg: renders as Screen Shot 2021-01-05 at 10 33 01 AM

and links to https://uproot.readthedocs.io/en/latest/uproot.behaviors.TH1.TH1.html#uproot.behaviors.TH1.TH1

PR #1010, opened 2020-07-29
feat: Add cdf method to AsymptoticTestStatDistribution

lukasheinrich:
# Description

This adds a cdf method to the asymptotic distribution. This will be useful for two-sided tests. c.f. discussion in PR #966.

ReadTheDocs build: https://pyhf.readthedocs.io/en/asymptotic_cdf

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add cumulative distribution function method to AsymptoticTestStatDistribution
* Add tests for test stat distribution calculators
```

### DIFF
--- a/src/pyhf/infer/calculators.py
+++ b/src/pyhf/infer/calculators.py
@@ -61,13 +61,13 @@ class AsymptoticTestStatDistribution(object):
 
     def cdf(self, value):
         """
-        Compute the value of the cumulative distribution funcation for a given value of the test statistic.
+        Compute the value of the cumulative distribution function for a given value of the test statistic.
 
         Args:
             value (`float`): The test statistic value.
 
         Returns:
-            Float: The integrated probability to observe a test statistic less than or equal to the observed ``value``.
+            probability (`float`): The integrated probability to observe a test statistic less than or equal to the observed ``value``.
 
         """
         tensorlib, _ = get_backend()
--- a/src/pyhf/infer/calculators.py
+++ b/src/pyhf/infer/calculators.py
@@ -67,7 +67,7 @@ class AsymptoticTestStatDistribution(object):
             value (`float`): The test statistic value.
 
         Returns:
-            Float: The integrated probability to observe  value at up to the observed one.
+            Float: The integrated probability to observe a test statistic less than or equal to the observed ``value``.
 
         """
         tensorlib, _ = get_backend()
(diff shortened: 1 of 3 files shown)


src/pyhf/infer/calculators.py
```diff
@@ -59,6 +59,20 @@ def __init__(self, shift):
         self.shift = shift
         self.sqrtqmuA_v = None
 
+    def cdf(self, value):
+        """
+        Compute the value of the cumulative distribution funcation for a given value of the test statistic.
```
kratsg, line 64:
```suggestion
        Compute the value of the cumulative distribution function for a given value of the test statistic.
```

src/pyhf/infer/calculators.py
```diff
@@ -59,6 +59,20 @@ def __init__(self, shift):
...
+
+        Args:
+            value (`float`): The test statistic value.
+
+        Returns:
+            Float: The integrated probability to observe a test statistic less than or equal to the observed ``value``.
```
kratsg, line 70:
```suggestion
            probability (`float`): The integrated probability to observe a test statistic less than or equal to the observed ``value``.
```

tests/test_calculator.py
```diff
@@ -0,0 +1,7 @@
+import pyhf
+import pyhf.infer.calculators
+
+
+def test_calc_dist():
+    asymptotic_dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0.0)
+    assert asymptotic_dist.pvalue(1) == asymptotic_dist.cdf(-1)
```
kratsg, line 7: why is this true? If AsymptoticTestStatDistribution has `0.0 = -u^/sigma` with variance 1.0, then...>?

src/pyhf/infer/calculators.py
matthewfeickert: I think that might have been my bad. Sorry.

src/pyhf/infer/calculators.py
```diff
@@ -59,6 +59,20 @@ def __init__(self, shift):
...
 
+    def cdf(self, value):
+        """
+        Compute the value of the cumulative distribution function for a given value of the test statistic.
+
+        Args:
+            value (`float`): The test statistic value.
+
+        Returns:
+            probability (`float`): The integrated probability to observe a test statistic less than or equal to the observed ``value``.
```
matthewfeickert, line 70:
```suggestion
            Float: The integrated probability to observe a test statistic less than or equal to the observed ``value``.
```

Shouldn't this be the return type? This is how it is for all the rest of our docs. e.g. https://scikit-hep.org/pyhf/_generated/pyhf.tensor.pytorch_backend.pytorch_backend.html#pyhf.tensor.pytorch_backend.pytorch_backend.astensor

PR #1013, opened 2020-07-30
feat: Add developer tools helper scripts

matthewfeickert:
# Description

Add some helper scripts for core devs to debug with

```
# doctest-on.sh has parity between source paths and imports
$ pwd
/home/feickert/Code/GitHub/pyhf
$ bash dev_tools/doctest-on.sh src/pyhf/infer/test_statistics.py 
/home/feickert/Code/GitHub/pyhf/src/pyhf/tensor/numpy_backend.py:271: RuntimeWarning: invalid value encountered in log
  return n * np.log(lam) - lam - gammaln(n + 1.0)
$ bash dev_tools/doctest-on.sh pyhf.infer.test_statistics
/home/feickert/Code/GitHub/pyhf/src/pyhf/tensor/numpy_backend.py:271: RuntimeWarning: invalid value encountered in log
  return n * np.log(lam) - lam - gammaln(n + 1.0)
```

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/.github/workflows/dev-tools.yml
+++ b/.github/workflows/dev-tools.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Install dev-tools
       run: |
         bash dev_tools/install-devtools.sh
-        ls -lhtra /usr/local/bin/
+        ls -lhtra "$(dirname $(command -v doctest-on))"
     - name: Test doctest-on
       run: |
         doctest-on src/pyhf/infer/test_statistics.py
--- a/.github/workflows/dev-tools.yml
+++ b/.github/workflows/dev-tools.yml
@@ -1,0 +1,34 @@
+name: dev-tools
+
+on:
+  pull_request:
+
+jobs:
+  test:
+
+    name: Check dev-tools
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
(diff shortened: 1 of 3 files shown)


matthewfeickert:
@kratsg @lukasheinrich If you don't like the idea of these being in the repo we can create a different project under the `pyhf` org that is something like "pyhf/developer-tools".

kratsg (approved):

lukasheinrich (approved):

matthewfeickert:
I'm just dumping Bash scripts in here now, but as @kratsg points out we could make this installable through `setup.py`. He also point to this post on how: [How To Add Custom Build Steps and Commands To `setup.py`](https://jichu4n.com/posts/how-to-add-custom-build-steps-and-commands-to-setuppy/)

matthewfeickert:
General questions:
- Do we want this to be controlled through `setup.py`? As the above article requires that you do things through `python setup.py <your command>` and running `python setup.py` is generally discouraged for anything involving installation in favor of always letting `pip install` figure it out.
- These dev tools are currently Bash scripts. Not even thinking about Windows someday, while I tried to intentionally write everything in a POSIX style and `shellcheck` is used, there can still be divergences in behavior between Linux and BSD Unix in macOS for shell behavior that is hard to catch. Should these dev tools be written to be all in Python instead?

matthewfeickert:
@lukasheinrich has pointed out that it should be possible to use the `scripts` option in `setup` to be able to add tools like this.

https://docs.python.org/3/distutils/setupscript.html#installing-scripts

Issue #1027, opened 2020-08-13
Model.expected_auxdata and expected_actualdata miss tensorlib conversion


alexander-held:
# Description

Both `Model.expected_auxdata` and `Model.expected_actualdata` do not include the same `pars = tensorlib.astensor(pars)` conversion that `Model.expected_data` uses. This results in runtime issues:

```
import pyhf

model = pyhf.simplemodels.hepdata_like(
    signal_data=[12.0, 11.0], bkg_data=[50.0, 52.0], bkg_uncerts=[3.0, 7.0]
)

model.expected_data(model.config.suggested_init())
model.expected_actualdata(model.config.suggested_init())
model.expected_auxdata(model.config.suggested_init())
```
output:
```
Traceback (most recent call last):
  File "ex.py", line 8, in <module>
    model.expected_actualdata(model.config.suggested_init())
  File "[...]/pyhf/src/pyhf/pdf.py", line 631, in expected_actualdata
    return self.make_pdf(pars)[0].expected_data()
  File "[...]/pyhf/src/pyhf/pdf.py", line 692, in make_pdf
    mainpdf = self.main_model.make_pdf(pars)
  File "[...]/pyhf/src/pyhf/pdf.py", line 456, in make_pdf
    lambdas_data = self._expected_data(pars)
  File "[...]/pyhf/src/pyhf/pdf.py", line 517, in _expected_data
    deltas, factors = self._modifications(pars)
  File "[...]/pyhf/src/pyhf/pdf.py", line 483, in _modifications
    [self.modifiers_appliers[k].apply(pars) for k in self._factor_mods],
  File "[...]/pyhf/src/pyhf/pdf.py", line 483, in <listcomp>
    [self.modifiers_appliers[k].apply(pars) for k in self._factor_mods],
  File "[...]/pyhf/src/pyhf/modifiers/shapesys.py", line 169, in apply
    shapefactors = tensorlib.gather(flat_pars, self.access_field)
  File "[...]/pyhf/src/pyhf/tensor/numpy_backend.py", line 136, in gather
    return tensor[indices]
TypeError: only integer scalar arrays can be converted to a scalar index
```
Similar for `expected_auxdata`.

I'm happy to submit a small PR to fix this, unless those functions are going to be deprecated (since at least `expected_actualdata` can alternatively be obtained via `model.expected_data, include_auxdata=False)`).

# Expected Behavior

No crash, but successful return of aux / actual data.

# Actual Behavior

code crashes

[...]

PR #1044, opened 2020-08-21
fix: Tensor conversion for expected data in _MainModel

### LINKED ISSUE #1043: Missing tensorlib conversion for Model.main_model.expected_data
# Description

This is the same issue as #1027, but for `Model.main_model.expected_data` instead. I missed it earlier due to the workaround I had been using for #1027. That function is particularly interesting for the `return_by_sample` behavior introduced by #731.

```
import pyhf

)

model.main_model.expected_data(model.config.suggested_init())
```

results in

```
  File "test.py", line 7, in <module>
    model.main_model.expected_data(model.config.suggested_init())
  File "[...]pyhf/src/pyhf/pdf.py", line 517, in expected_data
  File "[...]pyhf/src/pyhf/pdf.py", line 483, in _modifications
    [self.modifiers_appliers[k].apply(pars) for k in self._factor_mods],
  File "[...]pyhf/src/pyhf/pdf.py", line 483, in <listcomp>
    [self.modifiers_appliers[k].apply(pars) for k in self._factor_mods],
  File "[...]pyhf/src/pyhf/modifiers/shapesys.py", line 169, in apply
    shapefactors = tensorlib.gather(flat_pars, self.access_field)
  File "[...]pyhf/src/pyhf/tensor/numpy_backend.py", line 136, in gather
    return tensor[indices]
TypeError: only integer scalar arrays can be converted to a scalar index
```

# Expected Behavior

no crash, but successful return of expected data

# Actual Behavior

crash with `TypeError: only integer scalar arrays can be converted to a scalar index`

# Steps to Reproduce

`pyhf` master @ 236fbaa, see above

# Checklist

- [ ] Searched through existing Issues to confirm this is not a duplicate issue
- [ ] Filled out the Description, Expected Behavior, Actual Behavior, and Steps to Reproduce sections above or have edited/removed them in a way that fully describes the issue

alexander-held:
# Description

This resolves #1043.

By adding a tensor conversion for the parameters handed to `_MainModel.expected_data`, the behavior matches the other related functions:

- `pyhf.pdf.Model.expected_actualdata`
- `pyhf.pdf.Model.expected_auxdata`
- `pyhf.pdf.Model.expected_data`

This enables running the following:

```python
import pyhf

)

model.main_model.expected_data(model.config.suggested_init())
```

Also adding a test for the behavior that would fail on the current master branch.

Related PR: #1027, which fixed the behavior for `expected_actualdata` and `expected_auxdata`.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Ensure the parameters of pdf._MainModel.expected_data are tensors
   - Amends PR #1038 
* Add test for pdf._MainModel.expected_data acceptance of list instead of tensor
```

### DIFF
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -139,6 +139,9 @@ def test_pdf_basicapi_tests(backend):
     assert tensorlib.tolist(pdf.expected_auxdata(pars)) == pytest.approx(
         [51.020408630], 1e-08
     )
+    assert tensorlib.tolist(pdf.main_model.expected_data(pars)) == pytest.approx(
+        [60.0], 1e-08
+    )
 
     pdf = pyhf.simplemodels.hepdata_like(
         source['bindata']['sig'],


lukasheinrich:
nice to see this in. just one comment for reference: we at some point removed all internal "astensor" calls since we found out they're not free performance wise, in this case though it's probably not too bad.

matthewfeickert:
> nice to see this in. just one comment for reference: we at some point removed all internal "astensor" calls since we found out they're not free performance wise, in this case though it's probably not too bad.

Yeah, I figured that as this is all happening in the model section and not in the tensor computations of evaluating anything that small performance hits would be okay.

kratsg:
> nice to see this in. just one comment for reference: we at some point removed all internal "astensor" calls since we found out they're not free performance wise, in this case though it's probably not too bad.

One thought I had towards this end is that we split everything up as public API with astensor calls and internal API without:

```python
def expected_data(.....):
  return self._expected_data(tensorlib.astensor(....))
```

This should help us keep user-friendly interfaces while maintaining performance.

matthewfeickert:
> One thought I had towards this end is that we split everything up as public API with astensor calls and internal API without:
> 
> ```python
> def expected_data(.....):
>   return self._expected_data(tensorlib.astensor(....))
> ```
> 
> This should help us keep user-friendly interfaces while maintaining performance.

@kratsg Can you make an issue for this with a deliverable?

PR #1046, opened 2020-08-22
feat: Add API to download likelihood patchset archives

### LINKED ISSUE #1045: Add CLI API to download likelihood patchset archives from HEPData
# Description

At the moment to download a published likelihood archive from HEPData and extract the tarball into another directory you need to be a bit handy with `tar`. This is the most concise command I think you can do to download the archive for the likelihoods for [_Search for direct production of electroweakinos in final states with one lepton, missing transverse momentum and a Higgs boson decaying into two b-jets in (pp) collisions at s√=13 TeV with the ATLAS detector_](https://www.hepdata.net/record/ins1755298?version=3)

```
$ curl -sL https://www.hepdata.net/record/resource/1408476?view=true | tar -xzv --one-top-level=1Lbb-likelihoods
BkgOnly.json
patchset.json
README.md
$ cd 1Lbb-likelihoods
$ pyhf patchset verify BkgOnly.json patchset.json 
All good.
```

It might be worth considering adding a command to the CLI API to download and extract archives. Maybe something like

```
$ pyhf patchset download --help
Usage: pyhf patchset download [OPTIONS] [ARCHIVE_URL] [OUTPUT_DIRECTORY]

  Download the patchset archive from the remote URL
  and extract it in a directory at the path given.

  Returns:     None

Options:
  -h, --help  Show this message and exit.
$ pyhf patchset download https://www.hepdata.net/record/resource/1408476?view=true 1Lbb-likelihoods
$ ls 1Lbb-likelihoods/
BkgOnly.json  patchset.json  README.md
```

which would download and extract the archive at https://www.hepdata.net/record/resource/1408476?view=true to `1Lbb-likelihoods/`. The default would be silent, so similar to

```
$ curl -sL https://www.hepdata.net/record/resource/1408476?view=true | tar -xz --one-top-level=1Lbb-likelihoods
```

but if a `-v`, `--verbose` flag was passed it would similarly get passed to `tar` as

```
$ curl -sL https://www.hepdata.net/record/resource/1408476?view=true | tar -xzv --one-top-level=1Lbb-likelihoods
```

matthewfeickert:
# Description

Resolves #1045

```
$ pyhf contrib download --help
Usage: pyhf contrib download [OPTIONS] [ARCHIVE_URL] [OUTPUT_DIRECTORY]

  Download the patchset archive from the remote URL and extract it in a
  directory at the path given.

  Example:

  .. code-block:: shell

      $ pyhf contrib download --verbose
      https://www.hepdata.net/record/resource/1408476?view=true 1Lbb-
      likelihoods

          1Lbb-likelihoods/patchset.json
          1Lbb-likelihoods/README.md
          1Lbb-likelihoods/BkgOnly.json

  Raises:     :class:`~pyhf.exceptions.InvalidArchiveHost`: if the provided
  archive host name is not known to be valid

Options:
  -v, --verbose   Enables verbose mode
  -f, --force     Force download from non-approved host
  -c, --compress  Keep the archive in a compressed tar.gz form
  -h, --help      Show this message and exit.
```

CLI API Example:

```
$ pyhf contrib download https://www.hepdata.net/record/resource/1408476?view=true 1Lbb-likelihoods
$ ls 1Lbb-likelihoods/
BkgOnly.json  patchset.json  README.md
$ rm -rf 1Lbb-likelihoods && pyhf contrib download -v https://www.hepdata.net/record/resource/1408476?view=true 1Lbb-likelihoods
1Lbb-likelihoods/patchset.json
1Lbb-likelihoods/README.md
1Lbb-likelihoods/BkgOnly.json
$ pyhf contrib download --compress https://www.hepdata.net/record/resource/1408476?view=true 1Lbb-likelihoods.tar.gz
$ file 1Lbb-likelihoods.tar.gz 
1Lbb-likelihoods.tar.gz: gzip compressed data, from Unix, original size modulo 2^32 6177792
```

ReadTheDocs build:
- https://pyhf.readthedocs.io/en/feat-add-patchset-download-cli-api/_generated/pyhf.contrib.utils.download.html
- https://pyhf.readthedocs.io/en/feat-add-patchset-download-cli-api/cli.html#pyhf-contrib

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add download command to Python contrib API
* Add download command to pyhf CLI API and add contrib group
* Add pyhf custom exception if an untrusted archive host is given
* Add tests of Python and CLI API versions of contrib download
   - Thanks to Anthony Sottile (@asottile) for advice: https://youtu.be/PXu3KCMT3l4
* Add requests as dependency of contrib extra
* Add Python contrib API to docs
```

### DIFF
--- a/src/pyhf/cli/contrib.py
+++ b/src/pyhf/cli/contrib.py
@@ -15,7 +15,7 @@ log = logging.getLogger(__name__)
 
 @click.group(name="contrib")
 def cli():
-    """Contrib CLI group."""
+    """Contrib experimental operations"""
 
 
 @cli.command()
@@ -33,7 +33,7 @@ def download(archive_url, output_directory, verbose, force):
 
     .. code-block:: shell
 
-        $ pyhf download --verbose https://www.hepdata.net/record/resource/1408476?view=true 3L-likelihood
+        $ pyhf contrib download --verbose https://www.hepdata.net/record/resource/1408476?view=true 3L-likelihood
 
         \b
         3L-likelihoods/patchset.json
--- a/src/pyhf/cli/contrib.py
+++ b/src/pyhf/cli/contrib.py
@@ -16,7 +16,14 @@ log = logging.getLogger(__name__)
 
 @click.group(name="contrib")
 def cli():
-    """Contrib experimental operations"""
+    """
+    Contrib experimental operations.
+    Requires installation of the ``contrib`` ``pyhf`` extra.
+
+    .. code-block:: shell
+
+        $ python -m pip install pyhf[contrib]
+    """
 
 
 @cli.command()
@@ -28,7 +35,8 @@ def cli():
 )
 def download(archive_url, output_directory, verbose, force):
     """
-    Download the patchset archive from the remote URL and extract it in a directory at the path given.
+    Download the patchset archive from the remote URL and extract it in a
+    directory at the path given.
 
     Example:
 
--- a/src/pyhf/cli/contrib.py
+++ b/src/pyhf/cli/contrib.py
@@ -1,0 +1,54 @@
+"""CLI for functionality that will get migrated out eventually."""
+import logging
+
+import click
+import urllib.request
+import tarfile
+from io import BytesIO
+from pathlib import Path
+
+from .. import exceptions
+
+logging.basicConfig()
+log = logging.getLogger(__name__)
+
+
+@click.group(name="contrib")
+def cli():
+    """Contrib CLI group."""
+
+
+@cli.command()
+@click.argument("archive-url", default="-")
+@click.argument("output-directory", default="-")
+@click.option("-v", "--verbose", is_flag=True, help="Enables verbose mode")
+@click.option(
+    "-f", "--force", is_flag=True, help="Force download from non-approved host"
+)
+def download(archive_url, output_directory, verbose, force):
+    """
+    Download the patchset archive from the remote URL and extract it in a directory at the path given.
+
+    Raises:
+        :class:`~pyhf.exceptions.InvalidArchiveHost`: if the provided archive host name is not known to be valid
+
+    Returns:
+        None
+    """
+    if not force:
+        hostname = archive_url.split("://")[1].split("/")[0].strip("www.")
+        valid_hosts = ["hepdata.net", "doi.org"]
+        if hostname not in valid_hosts:
+            raise exceptions.InvalidArchiveHost(
+                f"{hostname} is not an approved archive host: {', '.join(str(host) for host in valid_hosts)}\n"
+                + "To download an archive from this host use the --force option."
+            )
+
+    req = urllib.request.Request(archive_url)
+    with urllib.request.urlopen(req) as response:
+        with tarfile.open(mode="r|gz", fileobj=BytesIO(response.read())) as archive:
+            archive.extractall(output_directory)
+
+    if verbose:
+        file_list = [str(file) for file in list(Path(output_directory).glob("*"))]
+        print("\n".join(file_list))
--- a/src/pyhf/cli/contrib.py
+++ b/src/pyhf/cli/contrib.py
@@ -3,10 +3,14 @@ import logging
 
 import click
 from urllib.parse import urlparse
-import requests
 import tarfile
 from io import BytesIO
 from pathlib import Path
+
+try:
+    import requests
+except ModuleNotFoundError:
+    pass
 
 from .. import exceptions
 
@@ -16,7 +20,14 @@ log = logging.getLogger(__name__)
 
 @click.group(name="contrib")
 def cli():
-    """Contrib experimental operations."""
+    """
+    Contrib experimental operations.
+    Requires installation of the ``contrib`` extra.
+
+    .. code-block:: shell
+
+        $ python -m pip install pyhf[contrib]
+    """
 
 
 @cli.command()
--- a/src/pyhf/cli/contrib.py
+++ b/src/pyhf/cli/contrib.py
@@ -28,56 +28,60 @@ def cli():
     """
 
 
-@cli.command()
-@click.argument("archive-url", default="-")
-@click.argument("output-directory", default="-")
-@click.option("-v", "--verbose", is_flag=True, help="Enables verbose mode")
-@click.option(
-    "-f", "--force", is_flag=True, help="Force download from non-approved host"
-)
-def download(archive_url, output_directory, verbose, force):
-    """
-    Download the patchset archive from the remote URL and extract it in a
-    directory at the path given.
+try:
+    import requests
 
-    Example:
+    @cli.command()
+    @click.argument("archive-url", default="-")
+    @click.argument("output-directory", default="-")
+    @click.option("-v", "--verbose", is_flag=True, help="Enables verbose mode")
+    @click.option(
+        "-f", "--force", is_flag=True, help="Force download from non-approved host"
+    )
+    def download(archive_url, output_directory, verbose, force):
+        """
+        Download the patchset archive from the remote URL and extract it in a
+        directory at the path given.
 
-    .. code-block:: shell
+        Example:
 
-        $ pyhf contrib download --verbose https://www.hepdata.net/record/resource/1408476?view=true 3L-likelihoods
+        .. code-block:: shell
 
-        \b
-        3L-likelihoods/patchset.json
-        3L-likelihoods/README.md
-        3L-likelihoods/BkgOnly.json
+            $ pyhf contrib download --verbose https://www.hepdata.net/record/resource/1408476?view=true 3L-likelihoods
 
-    Raises:
-        :class:`~pyhf.exceptions.InvalidArchiveHost`: if the provided archive host name is not known to be valid
-    """
-    try:
-        import requests
-    except ModuleNotFoundError as excep:
-        exception_info = (
-            str(excep)
-            + "\nInstallation of the contrib extra is required for: pyhf contrib download"
-            + "\nPlease install with: python -m pip install pyhf[contrib]\n"
-        )
-        print(exception_info)
-        raise
+            \b
+            3L-likelihoods/patchset.json
+            3L-likelihoods/README.md
+            3L-likelihoods/BkgOnly.json
 
-    if not force:
-        valid_hosts = ["www.hepdata.net", "doi.org"]
-        netloc = urlparse(archive_url).netloc
-        if netloc not in valid_hosts:
-            raise exceptions.InvalidArchiveHost(
-                f"{netloc} is not an approved archive host: {', '.join(str(host) for host in valid_hosts)}\n"
-                + "To download an archive from this host use the --force option."
-            )
+        Raises:
+            :class:`~pyhf.exceptions.InvalidArchiveHost`: if the provided archive host name is not known to be valid
+        """
 
-    with requests.get(archive_url) as response:
-        with tarfile.open(mode="r|gz", fileobj=BytesIO(response.content)) as archive:
-            archive.extractall(output_directory)
+        if not force:
+            valid_hosts = ["www.hepdata.net", "doi.org"]
+            netloc = urlparse(archive_url).netloc
+            if netloc not in valid_hosts:
+                raise exceptions.InvalidArchiveHost(
+                    f"{netloc} is not an approved archive host: {', '.join(str(host) for host in valid_hosts)}\n"
+                    + "To download an archive from this host use the --force option."
+                )
 
-    if verbose:
-        file_list = [str(file) for file in list(Path(output_directory).glob("*"))]
-        print("\n".join(file_list))
+        with requests.get(archive_url) as response:
+            with tarfile.open(
+                mode="r|gz", fileobj=BytesIO(response.content)
+            ) as archive:
+                archive.extractall(output_directory)
+
+        if verbose:
+            file_list = [str(file) for file in list(Path(output_directory).glob("*"))]
+            print("\n".join(file_list))
+
+
+except ModuleNotFoundError as excep:
+    exception_info = (
+        str(excep)
+        + "\nInstallation of the contrib extra is required to use the contrib CLI API"
+        + "\nPlease install with: python -m pip install pyhf[contrib]\n"
+    )
+    print(exception_info)
--- a/src/pyhf/cli/contrib.py
+++ b/src/pyhf/cli/contrib.py
@@ -29,11 +29,18 @@ def download(archive_url, output_directory, verbose, force):
     """
     Download the patchset archive from the remote URL and extract it in a directory at the path given.
 
+    Example:
+
+    .. code-block:: shell
+
+        $ pyhf download --verbose https://www.hepdata.net/record/resource/1408476?view=true 3L-likelihood
+        \b
+        3L-likelihoods/patchset.json
+        3L-likelihoods/README.md
+        3L-likelihoods/BkgOnly.json
+
     Raises:
         :class:`~pyhf.exceptions.InvalidArchiveHost`: if the provided archive host name is not known to be valid
-
-    Returns:
-        None
     """
     if not force:
         hostname = archive_url.split("://")[1].split("/")[0].strip("www.")
--- a/src/pyhf/cli/contrib.py
+++ b/src/pyhf/cli/contrib.py
@@ -2,8 +2,8 @@
 import logging
 
 import click
-import urllib.request
 from urllib.parse import urlparse
+import requests
 import tarfile
 from io import BytesIO
 from pathlib import Path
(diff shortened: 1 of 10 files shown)


src/pyhf/cli/patchset.py
```diff
@@ -16,6 +21,42 @@ def cli():
...
+        hostname = archive_url.split("://")[1].split("/")[0].strip("www.")
+        valid_hosts = ["hepdata.net", "doi.org"]
+        if hostname not in valid_hosts:
+            raise exceptions.InvalidArchiveHost(
+                f"{hostname} is not an approved archive host: {', '.join(str(host) for host in valid_hosts)}\n"
+                + "To download an archive from this host use the --force option."
+            )
+
+    req = urllib.request.Request(archive_url)
+    with urllib.request.urlopen(req) as response:
```
matthewfeickert, line 51: @kratsg @lukasheinrich I don't understand why Bandit is complaining here with the 
> Audit url open for permitted schemes. Allowing use of file:/ or custom schemes is often unexpected. (B310)
below

The [Bandit blacklist page doesn't really make things more clear either](https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html#b310-urllib-urlopen). Does this make sense to you?

kratsg: it means you need to make sure you trust the url you're opening.

src/pyhf/cli/contrib.py
```diff
@@ -0,0 +1,62 @@
...
+        \b
+        3L-likelihoods/patchset.json
+        3L-likelihoods/README.md
+        3L-likelihoods/BkgOnly.json
+
+    Raises:
+        :class:`~pyhf.exceptions.InvalidArchiveHost`: if the provided archive host name is not known to be valid
+    """
+    if not force:
+        hostname = archive_url.split("://")[1].split("/")[0].strip("www.")
```
kratsg, line 47: please use standard python libs here

```
from urllib.parse import urlparse
assert urlparse(archive_url).netloc in valid_hosts
```

src/pyhf/cli/patchset.py
matthewfeickert: Sure. Is there a way to more properly validate this then what I'm doing now?

src/pyhf/cli/contrib.py
```diff
@@ -0,0 +1,62 @@
...
+        valid_hosts = ["hepdata.net", "doi.org"]
+        if hostname not in valid_hosts:
+            raise exceptions.InvalidArchiveHost(
+                f"{hostname} is not an approved archive host: {', '.join(str(host) for host in valid_hosts)}\n"
+                + "To download an archive from this host use the --force option."
+            )
+
+    req = urllib.request.Request(archive_url)
+    with urllib.request.urlopen(req) as response:
```
kratsg, line 56: this one might necessitate a `requests` dependency. I'm not so sure what we want to do here, as this is a medium severity security issue anyway...

kratsg (approved):

src/pyhf/cli/patchset.py
kratsg: https://requests.readthedocs.io/en/master/ essentially... It's the only one not ban-listed in bandit's list. Or we add a `.bandit` config file.

src/pyhf/cli/contrib.py
matthewfeickert: Thanks. This is great.

src/pyhf/cli/contrib.py
matthewfeickert: Yeah, I'm not sure either. While I'm not against adding it, I'm hesitant to add things as core dependencies of `pyhf` especially for something in the contrib area. Should I just add it to the `contrib` extra?

matthewfeickert: Er, no this breaks the CLI now. This would require adding requests as a real dependency. Thoughts?

kratsg: We can do this in contrib as a lazy dependency. Move the `import requests` to become a

```
try:
  import requests
  # add in command
except:
  pass
```

src/pyhf/cli/contrib.py
```diff
@@ -0,0 +1,77 @@
...
+import click
+from urllib.parse import urlparse
+import tarfile
+from io import BytesIO
+from pathlib import Path
+
+try:
+    import requests
+except ModuleNotFoundError:
+    pass
```
matthewfeickert, line 13: @kratsg As we don't have any tests that run the CLI with just a `pip install pyhf` these lines are going to be uncovered (and as such I've just let this import be in the try/except block instead of the whole function). Thoughts on this? Do we just eat these two lines of coverage? Or do we add a workflow that installs just the library defaults and then see if it is possible to patch coverage reports together?

matthewfeickert: I might need an example of how you're suggesting to monkeypatch out `requests` from the environment. I haven't been able to find a good example of how to remove an entire module, but rather just attributes or items.

matthewfeickert: @kratsg https://github.com/scikit-hep/pyhf/blob/e55eea408d7c28e3109338de96252119ac63f87a/tests/test_init.py#L31-L55 obviously well enough for removing different modules from `pyhf` but it still isn't clear to me how I would implement a `ModuleNotFoundError` that a user would see if they were missing `requests` and how to delete the library from view as to even create this error. Maybe I should follow up with you on this later though.

matthewfeickert: Thanks @kratsg. Your example helped me out and made more sense after I had dinner. It is a slick solution that works very well for how simple it is. :)

tests/test_scripts.py
```diff
@@ -485,6 +486,70 @@ def test_workspace_digest(tmpdir, script_runner, algorithms, do_json):
...
+        or "certificate verify failed: Hostname mismatch, certificate is not valid for 'www.fail.org'."
+        in ret.stderr
+    )
+
+
+def test_missing_contrib_extra(datadir, script_runner):
+    module_name = "requests"
+
+    # hide
+    CACHE_MODULE, sys.modules[module_name] = sys.modules[module_name], None
```
matthewfeickert, line 528: I guess the only other thing to think about is if this should be made into a decorator.

kratsg (changes requested):
what if `pyhf contrib download` was hidden away if it didn't import correctly?

additionally, there should be a non-script/CLI test to check the python API itself.

matthewfeickert:
> additionally, there should be a non-script/CLI test to check the python API itself.

Hm. Maybe I misunderstand what you mean, but we don't have a public Python API for this, we only have the CLI API that's exposed.

kratsg:
> Hm. Maybe I misunderstand what you mean, but we don't have a public Python API for this, we only have the CLI API that's exposed.

It would generally make sense that we don't provide a CLI for anything that does not have a python API (for consistency) no? Maybe we need the python API first.

matthewfeickert:
> It would generally make sense that we don't provide a CLI for anything that does not have a python API (for consistency) no? Maybe we need the python API first.

I think I generally agree with that statement but if you're at the Python API do you really care about extracting files into directories rather than just reading them out from the in memory representation of the tarfile? I guess what you're thinking though is that if you wanted to script the download and unpacking of multiple likelihoods through the Python API you should be able to do that as well (correct?).

kratsg:
> I think I generally agree with that statement but if you're at the Python API do you really care about extracting files into directories rather than just reading them out from the in memory representation of the tarfile? I guess what you're thinking though is that if you wanted to script the download and unpacking of multiple likelihoods through the Python API you should be able to do that as well (correct?).

Yes, or likely -- someone will want to parallelize and do a signal grid... But I'm realizing that this doesn't make sense given we only really ever have two files per grid anyway (bkgonly + patchset)...

src/pyhf/cli/contrib.py
```diff
@@ -0,0 +1,87 @@
...
+                raise exceptions.InvalidArchiveHost(
+                    f"{netloc} is not an approved archive host: {', '.join(str(host) for host in valid_hosts)}\n"
+                    + "To download an archive from this host use the --force option."
+                )
+
+        with requests.get(archive_url) as response:
+            with tarfile.open(
+                mode="r|gz", fileobj=BytesIO(response.content)
+            ) as archive:
+                archive.extractall(output_directory)
```
matthewfeickert, line 74: This should also be able to handle a URL that is just a tarball put someplace in a public area.

src/pyhf/contrib/utils.py
```diff
@@ -0,0 +1,67 @@
...
+        Download the patchset archive from the remote URL and extract it in a
+        directory at the path given.
+
+        Example:
+
+            >>> from pyhf.contrib.utils import download
+            >>> download("https://www.hepdata.net/record/resource/1408476?view=true", "1Lbb-likelihoods")
+            >>> import os
+            >>> os.listdir("1Lbb-likelihoods")
+            ['patchset.json', 'README.md', 'BkgOnly.json']
```
matthewfeickert, line 26: doctest is failing as the order is random. So I need to let doctest know that it should expect a list like this, but not care about the order. Not sure if this is easy/possible.

src/pyhf/contrib/utils.py
```diff
@@ -0,0 +1,73 @@
...
+                try:
+                    with tarfile.open(
+                        mode="r|gz", fileobj=BytesIO(response.content)
+                    ) as archive:
+                        archive.extractall(output_directory)
+                except tarfile.ReadError:  # If not a gzip file
+                    with tarfile.open(
+                        mode="r|", fileobj=BytesIO(response.content)
+                    ) as archive:
+                        archive.extractall(output_directory)
```
matthewfeickert, line 64: Need a good way to test this.

tests/test_scripts.py
```diff
@@ -561,6 +563,104 @@ def test_workspace_digest(tmpdir, script_runner, algorithms, do_json):
...
+    ]
+    for module in _del_modules:
+        if module in sys.modules.keys():
+            del sys.modules[module]
+
+    import pyhf.cli
+
+    # NOTE: Only needed for pyflakes
+    # TODO: Remvoe when moving from pyflakes to flake8
+    pyhf.cli
```
matthewfeickert, line 618: `pyflakes` will complain if `import pyhf.cli` is left by itself unused.

tests/test_scripts.py
```diff
@@ -561,6 +563,104 @@ def test_workspace_digest(tmpdir, script_runner, algorithms, do_json):
...
+    # Force clean state
+    _del_modules = [
+        "pyhf.cli",
+        "pyhf.cli.contrib",
+        "pyhf.contrib",
+        "pyhf.contrib.utils",
+    ]
+    for module in _del_modules:
+        if module in sys.modules.keys():
+            del sys.modules[module]
```
matthewfeickert, line 612: This was found by trial and error. While at the moment it works, it is pretty ugly and not very satisfying. Critique here is most welcome.

matthewfeickert:
@kratsg @lukasheinrich This needs to still get fixed up on the coverage, but besides the things I've pointed out in the comments I made I think this is ready for another review.

src/pyhf/cli/contrib.py
```diff
@@ -0,0 +1,71 @@
...
+def download(archive_url, output_directory, verbose, force, compress):
+    """
+    Download the patchset archive from the remote URL and extract it in a
+    directory at the path given.
+
+    Example:
+
+    .. code-block:: shell
+
+        $ pyhf contrib download --verbose https://www.hepdata.net/record/resource/1408476?view=true 1Lbb-likelihoods
```
lukasheinrich, line 49: should we make the url be the `doi.org` one?

matthewfeickert: @lukasheinrich I would love that, but we don't have a DOI for that likelihood tarball. Do you know of one?

tests/test_scripts.py
matthewfeickert: I should probably use the `mocker` fixture here with

```python
    mocker.patch.dict(sys.modules, {"requests": None})
```
and then also mock the `del` of the modules, but I'm not sure how to get `del` working with `mocker.patch.dict`.

src/pyhf/contrib/utils.py
matthewfeickert: Dropping this for now as this is too hard to test at the moment as we don't have any archives that aren't compressed.

tests/test_scripts.py
matthewfeickert: Following this https://youtu.be/PXu3KCMT3l4 I decided to just use `mock`

src/pyhf/contrib/utils.py
```diff
@@ -0,0 +1,67 @@
...
+                    archive.extractall(output_directory)
+
+
+except ModuleNotFoundError as excep:
+    exception_info = (
+        str(excep)
+        + "\nInstallation of the contrib extra is required to use pyhf.contrib.utils.download"
+        + "\nPlease install with: python -m pip install pyhf[contrib]\n"
+    )
+    print(exception_info)
```
kratsg, line 67: log.error

src/pyhf/cli/contrib.py
```diff
@@ -0,0 +1,71 @@
...
+        if verbose:
+            file_list = [str(file) for file in list(Path(output_directory).glob("*"))]
+            print("\n".join(file_list))
+    except AttributeError as excep:
+        exception_info = (
+            str(excep)
+            + "\nInstallation of the contrib extra is required to use the contrib CLI API"
+            + "\nPlease install with: python -m pip install pyhf[contrib]\n"
+        )
+        print(exception_info)
```
kratsg, line 71: log.error

src/pyhf/contrib/utils.py
```diff
@@ -0,0 +1,67 @@
...
+            archive_url (`str`): The URL of the :class:`~pyhf.patchset.PatchSet` archive to download.
+            output_directory (`str`): Name of the directory to unpack the archive into.
+            force (`Bool`): Force download from non-approved host. Default is ``False``.
+            compress (`Bool`): Keep the archive in a compressed ``tar.gz`` form. Default is ``False``.
+
+        Raises:
+            :class:`~pyhf.exceptions.InvalidArchiveHost`: if the provided archive host name is not known to be valid
+        """
+        if not force:
+            valid_hosts = ["www.hepdata.net", "doi.org"]
```
kratsg, line 42: why do we care about valid hosts?

kratsg (changes requested):

src/pyhf/contrib/utils.py
lukasheinrich: I think in order to now have pyhf be a generic download tool .. i think constraining this for now is good, i.e. only allow "trustworthy" lhoods

lukasheinrich: there is still --force

matthewfeickert: What Lukas said.

src/pyhf/contrib/utils.py
```diff
@@ -0,0 +1,67 @@
...
+                    archive.extractall(output_directory)
+
+
+except ModuleNotFoundError as excep:
+    exception_info = (
+        str(excep)
+        + "\nInstallation of the contrib extra is required to use pyhf.contrib.utils.download"
+        + "\nPlease install with: python -m pip install pyhf[contrib]\n"
+    )
+    log.error(exception_info)
```
kratsg, line 67: any reason to have an exception info variable? And no format string?

tests/test_scripts.py
```diff
@@ -561,6 +567,100 @@ def test_workspace_digest(tmpdir, script_runner, algorithms, do_json):
...
+            download,
+            [
+                "--verbose",
+                "https://www.hepdata.net/record/resource/1408476?view=true",
+                "1Lbb-likelihoods",
+            ],
+        )
+        assert result.exit_code == 0
+
+        with caplog.at_level(logging.ERROR):
```
kratsg, line 654: this is a really strict test. Can we just check for important pieces of information? E.G. "pyhf[contrib]" and the "no attribute download..." part? I'm fine with it as it is, but the more strict we make it, the more we need to maintain it.

kratsg (changes requested):

src/pyhf/contrib/utils.py
matthewfeickert: No, can change this.

tests/test_scripts.py
matthewfeickert: I added this just to make sure that we're properly hiding `download` from users, but I can relax this.

matthewfeickert: @kratsg Oh, sorry I misread that before I pushed. You want the `module 'pyhf.contrib.utils' has no attribute 'download'` in still, correct? If so, then what is the strict part? I guess I'm unclear on what you want to have tested.

kratsg (approved):

lukasheinrich (approved):

PR #1048, opened 2020-08-26
style: Update to Black v20.8 formatting

matthewfeickert:
# Description

The final beta releases of Black bring about some changes in formatting, as we see in Black [`v20.8b1`](https://github.com/psf/black/releases/tag/20.8b1). This PR applies those style changes, and appropriatley changes the pre-commit style change so that commits don't fail with Black conflicts.

This PR also introduces the use of pinning the `rev` used in `pre-commit` as this is advocated for by `pre-commit` core dev @asottile in https://github.com/psf/black/issues/420 &mdash; thanks to @alexander-held and @henryiii for pointing this out and explaining more to us. (It seems from https://github.com/omry/omegaconf/pull/340 that others are in a similar boat as we are right now). The dev docs are also updated to reflect this.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Apply Black v20.8b1 formatting
   - Should be similar to format of first stable release of Black
* Update Black pre-commit hook to use rev v20.8b1
   - Using pinned revs is the recommended approach for pre-commit: https://github.com/psf/black/issues/420
* Update development docs to advocate for and explain the use of pre-commit autoupdate
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/psf/black
-    rev: stable
+    rev: 20.8b1
     hooks:
     - id: black
       language_version: python3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: stable
     hooks:
     - id: black
       language_version: python3
--- a/src/pyhf/parameters/paramview.py
+++ b/src/pyhf/parameters/paramview.py
@@ -9,7 +9,14 @@ def _tensorviewer_from_parmap(par_map, batch_size):
     names, slices, _ = list(
         zip(
             *sorted(
-                [(k, v['slice'], v['slice'].start,) for k, v in par_map.items()],
+                [
+                    (
+                        k,
+                        v['slice'],
+                        v['slice'].start,
+                    )
+                    for k, v in par_map.items()
+                ],
                 key=lambda x: x[2],
             )
         )
@@ -18,7 +25,9 @@ def _tensorviewer_from_parmap(par_map, batch_size):
 
 
 def extract_index_access(
-    baseviewer, subviewer, indices,
+    baseviewer,
+    subviewer,
+    indices,
 ):
     tensorlib, _ = get_backend()
 
--- a/src/pyhf/parameters/paramview.py
+++ b/src/pyhf/parameters/paramview.py
@@ -9,14 +9,7 @@ def _tensorviewer_from_parmap(par_map, batch_size):
     names, slices, _ = list(
         zip(
             *sorted(
-                [
-                    (
-                        k,
-                        v['slice'],
-                        v['slice'].start,
-                    )
-                    for k, v in par_map.items()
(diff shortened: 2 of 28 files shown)


.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
 -   repo: https://github.com/psf/black
-    rev: stable
+    rev: 20.8b1
```
matthewfeickert, line 3: Er, maybe this is a mistake. I was getting conflicts between the pre-commit version and the locally installed version once I updated, but [the Black docs say that](https://black.readthedocs.io/en/stable/version_control_integration.html)

> `stable` is a branch that tracks the latest release on PyPI

So I'm not sure why I'm getting a conflict with it then.

kratsg:
One thing might be to remove all trailing commas and see how black format handles this...

alexander-held:
With trailing commas removed it changes quite a bit, e.g.
```python
def extract_index_access(
    baseviewer, subviewer, indices,
):
```
(old black) becomes (with new black)
```python
def extract_index_access(
    baseviewer,
    subviewer,
    indices,
):
```
but when you remove the comma after indices, it becomes (still new black)
```python
def extract_index_access(baseviewer, subviewer, indices):
```

matthewfeickert:
> The trailing comma forces the split across lines no matter how long the line is I believe.

Ah this might be related to https://github.com/psf/black/issues/1629 and https://github.com/psf/black/issues/1649. Thanks for explaining that, @alexander-held. I'm going to put this back into Draft and think on this then.

matthewfeickert:
> One thing might be to remove all trailing commas and see how black format handles this...

> With trailing commas removed it changes quite a bit, e.g.
> ...
> The trailing comma forces the split across lines no matter how long the line is I believe.

Yeah, @alexander-held @kratsg I think you hit it on the head here. This PR is going to need some trailing comma removal before it can go in. Thanks to both of you for solving this for me.

src/pyhf/parameters/paramview.py
```diff
@@ -9,17 +9,15 @@ def _tensorviewer_from_parmap(par_map, batch_size):
     names, slices, _ = list(
         zip(
             *sorted(
-                [(k, v['slice'], v['slice'].start,) for k, v in par_map.items()],
+                [(k, v['slice'], v['slice'].start) for k, v in par_map.items()],
```
matthewfeickert, line 12: Note that the trailing comma is being removed here:
```python
(k, v['slice'], v['slice'].start)
```

tests/test_tensor.py
```diff
@@ -106,30 +109,39 @@ def test_zeros(backend):
 
 def test_broadcasting(backend):
     tb = pyhf.tensorlib
-    assert list(
-        map(
-            tb.tolist,
-            tb.simple_broadcast(
-                tb.astensor([1, 1, 1]), tb.astensor([2]), tb.astensor([3, 3, 3])
-            ),
+    assert (
```
matthewfeickert, line 112: These blocks seem like they're really going to stay as is for the time being given things like https://github.com/psf/black/issues/1629

matthewfeickert:
@kratsg @lukasheinrich Ready for another review. @alexander-held @henryiii your reviews are of course welcome too. :)

PR #1049, opened 2020-08-29
feat: Bump to modern pytest (6.0+)

### LINKED ISSUE #370: Fix Fixture use in pytest
# Description

In pytest `v4.0.0` the [direct call of a fixture results in an error](https://travis-ci.org/diana-hep/pyhf/jobs/455364238#L661-L669). 

```
==================================== ERRORS ====================================
__________________ ERROR collecting tests/test_validation.py ___________________
tests/test_validation.py:13: in <module>
    def spec_1bin_shapesys(source=source_1bin_example1()):
E   _pytest.warning_types.RemovedInPytest4Warning: Fixture "source_1bin_example1" called directly. Fixtures are not meant to be called directly, are created automatically when test functions request them as parameters. See https://docs.pytest.org/en/latest/fixture.html for more information.
__________________ ERROR collecting tests/test_validation.py ___________________
tests/test_validation.py:13: in <module>
    def spec_1bin_shapesys(source=source_1bin_example1()):
E   _pytest.warning_types.RemovedInPytest4Warning: Fixture "source_1bin_example1" called directly. Fixtures are not meant to be called directly, are created automatically when test functions request them as parameters. See https://docs.pytest.org/en/latest/fixture.html for more information.
```

This requires changing the way that pytest is used a bit.

This was noticed in preparation of PR #369 

# Checklist

kratsg:
# Description

Requires pytest > 6.0

Resolves #370 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Require pytest 6.0+ in the 'test' extra
* Migrate from pytest.ini to pyproject.toml
* Fix fixtures in test_validation.py
* Fix Node.get_marker calls
   - c.f. https://docs.pytest.org/en/stable/historical-notes.html#marker-revamp
* Mark pytest issue on ROADMAP complete
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ extras_require['test'] = sorted(
         + extras_require['contrib']
         + extras_require['shellcomplete']
         + [
-            'pytest>=6.0',
+            'pytest~=6.0',
             'pytest-cov>=2.5.1',
             'pytest-mock',
             'pytest-benchmark[histogram]',
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ extras_require['test'] = sorted(
         + extras_require['contrib']
         + extras_require['shellcomplete']
         + [
-            'pytest~=3.5',
+            'pytest>=6.0',
             'pytest-cov>=2.5.1',
             'pytest-mock',
             'pytest-benchmark[histogram]',
--- a/tests/test_teststats.py
+++ b/tests/test_teststats.py
@@ -16,7 +16,7 @@ def test_qmu(caplog):
         pyhf.infer.test_statistics.qmu(
             mu, data, model, init_pars, par_bounds, fixed_params
         )
-        assert "WARNING  qmu test statistic used for fit" in caplog.text
+        assert "qmu test statistic used for fit" in caplog.text
         caplog.clear()
 
 
@@ -33,7 +33,7 @@ def test_qmu_tilde(caplog):
         pyhf.infer.test_statistics.qmu_tilde(
             mu, data, model, init_pars, par_bounds, fixed_params
         )
-        assert "WARNING  qmu_tilde test statistic used for fit" in caplog.text
+        assert "qmu_tilde test statistic used for fit" in caplog.text
         caplog.clear()
 
 
@@ -49,7 +49,7 @@ def test_tmu(caplog):
         pyhf.infer.test_statistics.tmu(
             mu, data, model, init_pars, par_bounds, fixed_params
         )
-        assert "WARNING  tmu test statistic used for fit" in caplog.text
+        assert "tmu test statistic used for fit" in caplog.text
         caplog.clear()
 
 
@@ -66,7 +66,7 @@ def test_tmu_tilde(caplog):
         pyhf.infer.test_statistics.tmu_tilde(
             mu, data, model, init_pars, par_bounds, fixed_params
         )
-        assert "WARNING  tmu_tilde test statistic used for fit" in caplog.text
+        assert "tmu_tilde test statistic used for fit" in caplog.text
         caplog.clear()
 
 
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -7,14 +7,17 @@ from pathlib import Path
 import numpy as np
 
 
-@pytest.fixture(scope='module')
-def source_1bin_example1():
+def get_source_1bin_example1():
     with open('validation/data/1bin_example1.json') as read_json:
         return json.load(read_json)
 
 
 @pytest.fixture(scope='module')
-def spec_1bin_shapesys(source=source_1bin_example1()):
+def source_1bin_example1():
+    return get_source_1bin_example1()
+
+
+def get_spec_1bin_shapesys(source=get_source_1bin_example1()):
     spec = {
         'channels': [
             {
@@ -46,7 +49,11 @@ def spec_1bin_shapesys(source=source_1bin_example1()):
 
 
 @pytest.fixture(scope='module')
-def expected_result_1bin_shapesys(mu=1.0):
+def spec_1bin_shapesys():
+    return get_spec_1bin_shapesys()
+
+
+def get_expected_result_1bin_shapesys(mu=1.0):
     if mu == 1:
         expected_result = {
             "exp": [
@@ -62,11 +69,15 @@ def expected_result_1bin_shapesys(mu=1.0):
 
 
 @pytest.fixture(scope='module')
-def setup_1bin_shapesys(
-    source=source_1bin_example1(),
-    spec=spec_1bin_shapesys(source_1bin_example1()),
+def expected_result_1bin_shapesys():
+    return get_expected_result_1bin_shapesys()
+
+
+def get_setup_1bin_shapesys(
+    source=get_source_1bin_example1(),
+    spec=get_spec_1bin_shapesys(get_source_1bin_example1()),
     mu=1,
-    expected_result=expected_result_1bin_shapesys(1.0),
+    expected_result=get_expected_result_1bin_shapesys(1.0),
     config={'init_pars': 2, 'par_bounds': 2},
 ):
     return {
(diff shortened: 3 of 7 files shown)


matthewfeickert:
> Requires pytest > 6.0

So...does this mean that an epic testing PR is coming? :) Or are you just setting things up for future us?

setup.py
```diff
@@ -29,7 +29,7 @@
         + extras_require['contrib']
         + extras_require['shellcomplete']
         + [
-            'pytest~=3.5',
+            'pytest>=6.0',
```
matthewfeickert, line 32:
```suggestion
            'pytest~=6.0',
```

Maybe we should just stay at `v6.X` so that we can properly test how changes to the API whenever `v7.0` comes out affects.

tests/test_validation.py
```diff
@@ -7,14 +7,17 @@
...
-@pytest.fixture(scope='module')
-def source_1bin_example1():
+def get_source_1bin_example1():
     with open('validation/data/1bin_example1.json') as read_json:
         return json.load(read_json)
 
 
 @pytest.fixture(scope='module')
-def spec_1bin_shapesys(source=source_1bin_example1()):
+def source_1bin_example1():
```
matthewfeickert, line 16: @kratsg I think I'm missing something obvious. This fixture doesn't ever show up again

```
$ git grep "source_1bin_example1"
tests/test_validation.py:def get_source_1bin_example1():
tests/test_validation.py:def source_1bin_example1():
tests/test_validation.py:    return get_source_1bin_example1()
tests/test_validation.py:def get_spec_1bin_shapesys(source=get_source_1bin_example1()):
tests/test_validation.py:    source=get_source_1bin_example1(),
tests/test_validation.py:    spec=get_spec_1bin_shapesys(get_source_1bin_example1()),
tests/test_validation.py:    source=get_source_1bin_example1(),
```

so how does this work? I'll go re-read on how module level fixtures work but if you can briefly summarize that helps too.

matthewfeickert: Yeah, after reading the "[Scope: sharing fixtures across classes, modules, packages or session](https://docs.pytest.org/en/stable/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session)" section of the pytest docs I'm still confused as this fixture is never used.

kratsg: I'm just moving it to the appropriate fixture pattern that's suggested. This first move will eventually indicate that certain fixtures become unused (which means they never should've been fixtures in the first place).

kratsg:
@matthewfeickert you know what's up with CodeCov here?

tests/test_teststats.py
```diff
@@ -16,7 +16,7 @@ def test_qmu(caplog):
         pyhf.infer.test_statistics.qmu(
             mu, data, model, init_pars, par_bounds, fixed_params
         )
-        assert "WARNING  qmu test statistic used for fit" in caplog.text
+        assert "qmu test statistic used for fit" in caplog.text
```
matthewfeickert, line 19: Can you elaborate on why do these warnings need to get removed?

tests/test_validation.py
matthewfeickert: Why leave them in then and not just remove them in this PR? I'm a bit confused by the renaming to `source_X` to `get_source_X` everywhere

```diff
@@ -7,14 +7,17 @@ from pathlib import Path
 import numpy as np
 
 
-@pytest.fixture(scope='module')
-def source_1bin_example1():
+def get_source_1bin_example1():
     with open('validation/data/1bin_example1.json') as read_json:
         return json.load(read_json)
 
 
 @pytest.fixture(scope='module')
-def spec_1bin_shapesys(source=source_1bin_example1()):
+def source_1bin_example1():
+    return get_source_1bin_example1()
+
+
+def get_spec_1bin_shapesys(source=get_source_1bin_example1()):
     spec = {
         'channels': [
             {
@@ -46,7 +49,11 @@ def spec_1bin_shapesys(source=source_1bin_example1()):
```

Why all the renaming if it seems the only functional change is what used to be called `source_1bin_example1` is no longer a fixture?

> I'm just moving it to the appropriate fixture pattern that's suggested.

Can you link to what the suggestion is, as maybe that will explain my questions?

kratsg: https://docs.pytest.org/en/stable/deprecations.html

I’m not going to rewrite the validation tests in this PR. I’m simply updating them with the minimal amount of work that follows pytest docs and works.

tests/test_teststats.py
kratsg: Two reasons:

1) you’re already using caplog to specify warning log level so no need to check that.
2) pytest changes how logging is handled and there are most spaces and extra text between the warning and the actual message... the test should only care about the message, not depending on how the logger is formatted.

tests/test_validation.py
matthewfeickert: Okay I'll have to look at this later in more depth then. I don't understand why it is necessary to add all these `get_X` functions and keep `source_X` fixtures that are never used when the following patch would allow `setup_1bin_shapesys` to work as the fixture you need while removing a lot of code.

I suspect I'm just misunderstanding what it is that you're trying to do, so can you give me an example of why this is necessary (for a future PR or something).

```diff
diff --git a/tests/test_validation.py b/tests/test_validation.py
index ed1eb9b8..a2393173 100644
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -7,17 +7,12 @@ from pathlib import Path
 import numpy as np
 
 
-def get_source_1bin_example1():
+def source_1bin_example1():
     with open('validation/data/1bin_example1.json') as read_json:
         return json.load(read_json)
 
 
-@pytest.fixture(scope='module')
-def source_1bin_example1():
-    return get_source_1bin_example1()
-
-
-def get_spec_1bin_shapesys(source=get_source_1bin_example1()):
+def spec_1bin_shapesys(source=source_1bin_example1()):
     spec = {
         'channels': [
             {
@@ -48,12 +43,7 @@ def get_spec_1bin_shapesys(source=get_source_1bin_example1()):
     return spec
 
 
-@pytest.fixture(scope='module')
-def spec_1bin_shapesys():
-    return get_spec_1bin_shapesys()
-
-
-def get_expected_result_1bin_shapesys(mu=1.0):
+def expected_result_1bin_shapesys(mu=1.0):
     if mu == 1:
         expected_result = {
             "exp": [
@@ -68,28 +58,22 @@ def get_expected_result_1bin_shapesys(mu=1.0):
     return expected_result
 
 
-@pytest.fixture(scope='module')
-def expected_result_1bin_shapesys():
-    return get_expected_result_1bin_shapesys()
-
-
-def get_setup_1bin_shapesys(
-    source=get_source_1bin_example1(),
-    spec=get_spec_1bin_shapesys(get_source_1bin_example1()),
-    mu=1,
-    expected_result=get_expected_result_1bin_shapesys(1.0),
-    config={'init_pars': 2, 'par_bounds': 2},
-):
-    return {
-        'source': source,
-        'spec': spec,
-        'mu': mu,
-        'expected': {'result': expected_result, 'config': config},
-    }
-
-
 @pytest.fixture(scope='module')
 def setup_1bin_shapesys():
+    def get_setup_1bin_shapesys(
+        source=source_1bin_example1(),
+        spec=spec_1bin_shapesys(source_1bin_example1()),
+        mu=1,
+        expected_result=expected_result_1bin_shapesys(1.0),
+        config={'init_pars': 2, 'par_bounds': 2},
+    ):
+        return {
+            'source': source,
+            'spec': spec,
+            'mu': mu,
+            'expected': {'result': expected_result, 'config': config},
+        }
+
     return get_setup_1bin_shapesys()
 
 
@@ -158,7 +142,7 @@ def expected_result_1bin_lumi():
 
 
 def get_setup_1bin_lumi(
-    source=get_source_1bin_example1(),
+    source=source_1bin_example1(),
     spec=get_spec_1bin_lumi(),
     mu=1,
     expected_result=get_expected_result_1bin_lumi(1.0),
```

matthewfeickert (approved):

lukasheinrich (approved):

PR #1053, opened 2020-09-04
fix: Sync SciPy minimizer initialized parameter values with fixed values

### LINKED ISSUE #1052: Bug in scipy minimization: init_pars and fixed_vals are not synced
# Description

Related: #1051 (found as part of this PR).

There is a bug in `pyhf`'s codebase (since ~forever) where the `init_par` for a fixed parameter might differ from the constrained value set for that fixed parameter. See for example v0.4.4:

https://github.com/scikit-hep/pyhf/blob/6ac0f6280c56111c32f46ec9aaad0286d05d36e7/src/pyhf/infer/mle.py#L52-L90
https://github.com/scikit-hep/pyhf/blob/6ac0f6280c56111c32f46ec9aaad0286d05d36e7/src/pyhf/optimize/opt_scipy.py#L16-L53

# Expected Behavior

I would've thought the underlying optimizers (particularly scipy) would be smart enough to handle or recognize when the constraint/init_pars are at odds with each other.

# Actual Behavior

See description.

# Steps to Reproduce

Follow along with me:

call `fixed_poi_fit` with:
- `poi_val = 0.5`
- `data`
- `pdf`
- `init_pars=[1, 1, 1]`

This will call `opt.minimize` with:

- `objective=twice_nll`
- `data=data`
- `pdf=pdf`
- `init_pars=[1, 1, 1]`
- `fixed_vals=[(0, 0.5)]`

at this point, `opt.minimize` (in the case of scipy) will do

```python
        constraints = [{'type': 'eq', 'fun': lambda v: v[indices] - values}]
```

which becomes essentially

```python
        constraints = [{'type': 'eq', 'fun': lambda v: v[0] - 0.5}]
```

but still calls `scipy.optimize.minimize` with `init_pars=[1, 1, 1]`. If one changes this to `[0.5, 1, 1]` -- one observes a slightly different result... but probably more correct.

Minuit does not have the same bug here because of https://github.com/scikit-hep/pyhf/blob/6ac0f6280c56111c32f46ec9aaad0286d05d36e7/src/pyhf/optimize/opt_minuit.py#L45 which updates the initval with the fixed value `initvals['p{}'.format(index)] = value`.

I was worried that the major refactor I did #951 broke this or changed this. In fact, I managed to keep the same bug in scipy, and did not introduce this bug into minuit.

kratsg:
# Description

This fixes #1052. We will sync up the init pars for scipy when there are fixed values.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Fix bug in the SciPy optimizer where the init_pars is not synced up with the fixed values
   - Update the initial values to the fixed value for any fixed parameter
* Add tests for constraining values bounding initialization values
```


matthewfeickert (approved):
LGTM now that we have things passing CI.

matthewfeickert:
Thanks for the pathological bug catch @kratsg

PR #1051, opened 2020-09-03
feat: Support configuring fixed values automatically from Model

kratsg:
# Description

Replaces #846.

Provides support for automatically passing through the fixed parameters from the model configuration into all of the minimization/inference that we do in the `pyhf` code-base, namely:

- `pyhf.infer.hypotest`
- `pyhf.infer.calculators.generate_asimov_data`
- `pyhf.infer.calculators.AsymptoticCalculators`
- `pyhf.infer.mle.fit`
- `pyhf.infer.mle.fixed_poi_fit`
- `pyhf.infer.test_statistics._qmu_like`
- `pyhf.infer.test_statistics._tmu_like`
- `pyhf.infer.test_statistics.qmu`
- `pyhf.infer.test_statistics.qmu_tilde`
- `pyhf.infer.test_statistics.tmu`
- `pyhf.infer.test_statistics.tmu_tilde`

We will pass in `fixed_params` all the way through to the optimization which will stitch together the two if need-be.

This is a two-fold feature that needs to be refactored/be more elegant a bit later. `fixed_vals` are things that are a list of tuples like so: `[(parameter_index, parameter_value), ...]`. `fixed_params` are what comes from the `pdf.config.suggested_fixed()`.

- in cases where `fixed_params` is optional, the model fixed parameters are retrieved
- in cases where `fixed_params` is not optional, nothing is done (a pass-through)

This tries to model the behavior we have identically for `init_pars` and `par_bounds` everywhere.

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Add support for automatically passing through fixed parameters from the model configuration into all of the minimization/inference API
* Add tests for fixed values

Co-authored-by: Carlo Gottardo <carlo.gottardo91@gmail.com>
```

PR #1054, opened 2020-09-04
fix: Step sizes for fixed parameters in minuit should be 0.0

kratsg:
# Description

This resolves a comment made in #1051 where step sizes were not being zero'd out correctly.

> Suggestion: For the minuit optimizer, the initial step size is set in minuit_optimizer._get_minimizer. For a fixed Gaussian-constrained nuisance parameter with default bounds [-5, 5], this is 0.01 with the default 1000 steps. This will be the parameter uncertainty reported by minuit after the fit. Given that it does not make sense to have an uncertainty for a fixed parameter, it might be nice to set the entries of step_sizes corresponding to fixed parameters to 0 instead.

_Originally posted by @alexander-held in https://github.com/scikit-hep/pyhf/pull/1051#issuecomment-687117202_

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Step sizes are set to 0.0 for fixed parameters in a fit for minuit
   - The step size acts as the uncertainty for fixed parameters
* Add test for step sizes for fixed parameters
```

### DIFF
--- a/src/pyhf/optimize/opt_minuit.py
+++ b/src/pyhf/optimize/opt_minuit.py
@@ -38,7 +38,6 @@ class minuit_optimizer(OptimizerMixin):
     ):
 
         step_sizes = [(b[1] - b[0]) / float(self.steps) for b in init_bounds]
-
         fixed_vals = fixed_vals or []
         # Minuit wants True/False for each parameter
         fixed_bools = [False] * len(init_pars)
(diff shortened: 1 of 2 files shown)


matthewfeickert (approved):
LGTM. Thanks @kratsg. @alexander-held let us know if you have any comments here.

alexander-held:
Thanks for including this! I rebased it into #1051 to test and it worked as expected.

PR #1057, opened 2020-09-09
fix: Use negative log likelihoods as objective function

### LINKED ISSUE #1006: change to standard NLL fit objective
right now we use "twice_nll" as a fit objective and in the test statistic a simple diffence

`twice_nll_constrfit - twice_nll_globalfit`

but rather we should just to a NLL fit and in the test stat do

`2*(nll_constrfit - nll_globalfit)`

this will require updating some test reference numbers

matthewfeickert:
# Description

Resolves #1006 

Use the negative log likelihood as the objective function to give to the minimizer instead of twice the negative log likelihood, as it is the more canonical choice.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use the negative log likelihood as the objective function to give to the minimizer
* Update docstrings and testst to reflect the slightly different results
```


matthewfeickert:
Hm. Perhaps concerningly we're seeing differences large than `1e-5` in `tests/test_regression.py`.

tests/test_optim.py
```diff
@@ -66,40 +66,40 @@ def test_minimize(tensorlib, precision, optimizer, do_grad, do_stitch):
...
+            'do_grad-scipy-tensorflow-64b': [0.49999859, 1.00000122],
+            'do_grad-scipy-jax-64b': [0.49999859, 1.00000122],
             # no grad, minuit, 32b - not very consistent for pytorch
-            'no_grad-minuit-numpy-32b': [0.49622172117233276, 1.0007264614105225],
+            'no_grad-minuit-numpy-32b': [0.5021529, 0.99905187],
             #    nb: macos gives different numerics than CI
             #'no_grad-minuit-pytorch-32b': [0.7465415000915527, 0.8796938061714172],
             'no_grad-minuit-pytorch-32b': [0.9684963226318359, 0.9171305894851685],
-            'no_grad-minuit-tensorflow-32b': [0.5284154415130615, 0.9911751747131348],
+            'no_grad-minuit-tensorflow-32b': [0.5073193, 0.99639267],
```
matthewfeickert, line 86: @kratsg @lukasheinrich I expected there to be some changes, but I'm confused and a bit worried that to CI to pass I had to introduce such large changes as ones like this here. We should probably think on this a bit, but if either of you have thoughts on why I'd be very interested to hear them.

src/pyhf/infer/mle.py
```diff
@@ -30,10 +30,10 @@ def twice_nll(pars, data, pdf):
...
         >>> data = pyhf.tensorlib.astensor(observations + model.config.auxdata)
         >>> parameters = model.config.suggested_init()  # nominal parameters
-        >>> twice_nll = pyhf.infer.mle.twice_nll(parameters, data, model)
-        >>> twice_nll
-        array([30.77525435])
-        >>> -2 * model.logpdf(parameters, data) == twice_nll
+        >>> nll = pyhf.infer.mle.nll(parameters, data, model)
+        >>> nll
+        array([15.38762717])
+        >>> -1 * model.logpdf(parameters, data) == nll
```
lukasheinrich, line 36: do we need "-1" here instead of just " -model.logpdf"

matthewfeickert: Nope. I'll change that.

tests/test_optim.py
kratsg: it's 32bit. It's... ... tricky.

matthewfeickert: > it's... ... tricky.

:slightly_smiling_face: true. So what's your thoughts here?

matthewfeickert:
@HDembinski It seems that we're seeing quite substantial changes from `iminuit` when we simply change the objective function by a factor of `2`. Do you have thoughts on how we might better mitigate this with `iminuit` to make it more stable?

HDembinski:
There shouldn't be a substantial change apart from minor numerical rounding errors. The factor 2 should only affect the uncertainties and even that only if you didn't compensate by setting the errordef. If you change your function and errordef, the effects should almost perfectly cancel, giving the same result.

kratsg:
> There shouldn't be a substantial change apart from minor numerical rounding errors. The factor 2 should only affect the uncertainties and even that only if you didn't compensate by setting the errordef. If you change your function and errordef, the effects should almost perfectly cancel, giving the same result.

@HDembinski thanks! I was not aware the `errordef` played a role here... that said, it now makes some sense.

@matthewfeickert we need to update docs here (https://github.com/scikit-hep/pyhf/blob/cab1d8d4392bb76ddd16553f90ab81a6dfd0494a/src/pyhf/optimize/opt_minuit.py#L21-L24) to state that the default is `0.5` and then change the line here: https://github.com/scikit-hep/pyhf/blob/cab1d8d4392bb76ddd16553f90ab81a6dfd0494a/src/pyhf/optimize/opt_minuit.py#L32

tests/test_optim.py
```diff
@@ -66,18 +66,18 @@ def test_minimize(tensorlib, precision, optimizer, do_grad, do_stitch):
...
             'no_grad-scipy-jax-32b': [1.0, 1.0],
             # no grad, scipy, 64b
-            'no_grad-scipy-numpy-64b': [0.49998815367220306, 0.9999696999038924],
-            'no_grad-scipy-pytorch-64b': [0.49998815367220306, 0.9999696999038924],
-            'no_grad-scipy-tensorflow-64b': [0.49998865164653106, 0.9999696533705097],
-            'no_grad-scipy-jax-64b': [0.4999880886490433, 0.9999696971774877],
+            'no_grad-scipy-numpy-64b': [0.4999985, 1.00000122],
+            'no_grad-scipy-pytorch-64b': [0.4999985, 1.00000122],
+            'no_grad-scipy-tensorflow-64b': [0.49999869, 1.00000121],
+            'no_grad-scipy-jax-64b': [0.49999894, 1.00000114],
```
matthewfeickert, line 72: @lukasheinrich @kratsg While these changes are relatively small (mostly on the O(e-6) to O(e-7) range) I'm not sure if it makes sense to me that we're still seeing these, or the JAX change that is quite a bit larger.

src/pyhf/infer/test_statistics.py
```diff
@@ -43,7 +43,7 @@ def _tmu_like(
     )
     log_likelihood_ratio = fixed_poi_fit_lhood_val - unconstrained_fit_lhood_val
     tmu_like_stat = tensorlib.astensor(
-        tensorlib.clip(log_likelihood_ratio, 0.0, max_value=None)
+        tensorlib.clip(2.0 * log_likelihood_ratio, 0.0, max_value=None)
```
kratsg, line 46: why is this changed?

tests/test_optim.py
kratsg: these are ok, and self-consistent and close to [0.5,1.0]

tests/test_optim.py
```diff
@@ -66,18 +66,18 @@ def test_minimize(tensorlib, precision, optimizer, do_grad, do_stitch):
...
             # do grad, scipy, 32b
-            'do_grad-scipy-pytorch-32b': [0.49993881583213806, 1.0001085996627808],
-            'do_grad-scipy-tensorflow-32b': [0.4999384582042694, 1.0001084804534912],
-            'do_grad-scipy-jax-32b': [0.4999389052391052, 1.0001085996627808],
+            'do_grad-scipy-pytorch-32b': [0.49999850, 1.00000131],
+            'do_grad-scipy-tensorflow-32b': [0.49999896, 1.0000011],
+            'do_grad-scipy-jax-32b': [0.5002691, 1.0001281],
```
kratsg, line 76: this is the worst one by far.

tests/test_optim.py
```diff
@@ -117,6 +117,8 @@ def test_minimize(tensorlib, precision, optimizer, do_grad, do_stitch):
             rtol = 7e-05
         if 'no_grad-minuit-jax-32b' in identifier:
             rtol = 4e-02
+        if 'do_grad-scipy-jax-32b' in identifier:
```
kratsg, line 120: is this needed?

tests/test_regression.py
```diff
@@ -43,17 +43,17 @@ def test_sbottom_regionA_1300_205_60(
...
         np.isclose(
             np.array(CLs_exp),
             np.array(
                 [
-                    0.09022509053507759,
-                    0.1937839194960632,
-                    0.38432344933992,
-                    0.6557757334303531,
-                    0.8910420971601081,
+                    np.array(0.09023417),
```
kratsg, line 52: why are these wrapped in nested arrays?

tests/test_regression.py
```diff
@@ -73,17 +73,17 @@ def test_sbottom_regionA_1400_950_60(
...
         np.isclose(
             np.array(CLs_exp),
             np.array(
                 [
-                    0.0026445531093281147,
-                    0.013976126501170727,
-                    0.06497105816950004,
-                    0.23644030478043676,
-                    0.5744785776763938,
+                    np.array(0.00264459),
```
kratsg, line 82: why are these wrapped in nested arrays?

tests/test_regression.py
```diff
@@ -103,17 +103,17 @@ def test_sbottom_regionA_1500_850_60(
...
         np.isclose(
             np.array(CLs_exp),
             np.array(
                 [
-                    0.00598431785676406,
-                    0.026102240062850574,
-                    0.10093641492218848,
-                    0.31019245951964736,
-                    0.6553623337518385,
+                    np.array(0.00598365),
```
kratsg, line 112: why are these wrapped in nested arrays?

tests/test_regression.py
```diff
@@ -133,17 +133,17 @@ def test_sbottom_regionB_1400_550_60(
...
         np.isclose(
             np.array(CLs_exp),
             np.array(
                 [
-                    0.9338879894557114,
-                    0.9569045303300702,
-                    0.9771296335437559,
-                    0.9916370124133669,
-                    0.9983701133999316,
+                    np.array(0.9939882),
```
kratsg, line 142: why are these wrapped in nested arrays?

tests/test_regression.py
```diff
@@ -169,11 +169,11 @@ def test_sbottom_regionC_1600_850_60(
             np.array(CLs_exp),
             np.array(
                 [
-                    0.2955492909588046,
-                    0.4446885457298284,
-                    0.6371473864200973,
-                    0.8336149623750603,
-                    0.9585901381554178,
+                    np.array(0.29555621),
```
kratsg, line 172: why are these wrapped in nested arrays?

tests/test_validation.py
```diff
@@ -848,11 +848,11 @@ def validate_hypotest(pdf, data, mu_test, expected_result, tolerance=1e-6):
...
         ('setup_1bin_shapesys', 1e-6),
         ('setup_1bin_lumi', 4e-6),
-        ('setup_1bin_normsys', 2e-9),
+        ('setup_1bin_normsys', 6e-6),
         ('setup_2bin_histosys', 8e-5),
-        ('setup_2bin_2channel', 1e-6),
+        ('setup_2bin_2channel', 2e-5),
         ('setup_2bin_2channel_couplednorm', 1e-6),
-        ('setup_2bin_2channel_coupledhistosys', 1e-6),
+        ('setup_2bin_2channel_coupledhistosys', 5e-5),
```
kratsg, line 855: this shouldn't get changed, should it?

kratsg (changes requested):
Overall, you're truncating a lot of the values based on the numpy output -- can you not truncate the floats when you update them? I'm still concerned that some values are larger than expected differences.

src/pyhf/infer/test_statistics.py
matthewfeickert: Because the test statistic that we use should still be based on `-2*Lambda`, but the `2` is no longer automatically included in the `fixed_poi_fit_lhood_val` and `unconstrained_fit_lhood_val` above.

tests/test_regression.py
matthewfeickert: We made changes in the return structure a while back so that `pyhf.infer.hypotest(..., return_expected_set=True)` returns a list of tensors. The goal he was to actually match that return structure, and as these are all being run with the NumPy backend only I thought just use `np.array`.

```
>>> import pyhf
>>> test_mu = 1.0
>>> CLs_obs, CLs_exp = pyhf.infer.hypotest(test_mu, data, model, qtilde=True, return_expected_set=True)
>>> CLs_obs
array(0.0525151)
>>> CLs_exp
[array(0.0026063), array(0.01382019), array(0.06445367), array(0.2352575), array(0.5730375)]
```

This change is not strictly needed, but it seemed like it should be dealt with here if anywhere.

tests/test_validation.py
matthewfeickert: Yeah that's a good point. As these tolerances are coming into play with

```
assert abs(CLs_obs - expected_result['obs']) / expected_result['obs'] < tolerance
```

the expected values should be updated if anything.

HDembinski:
Note: the accuracy of MINUIT is typically way lower than machine precision. Deviations of 1e-6 can be expected, also see https://iminuit.readthedocs.io/en/stable/benchmark.html.

kratsg:
> Note: the accuracy of MINUIT is typically way lower than machine precision. Deviations of 1e-6 can be expected, also see https://iminuit.readthedocs.io/en/stable/benchmark.html.

Interesting, I didn't see this page before! @matthewfeickert we should definitely link to it in our documentation.

HDembinski:
Yes, please do that. However, the benchmark does not tell the whole story. I also did an informal test of a case with limits on parameters and there Minuit performed very well.

HDembinski:
Another note: the fact that Minuit stops early, once the accuracy is well below 1 sigma, is an advantage. It is wasteful for a statistical fitter to continue iterating until the parameter values have machine precision.

tests/test_regression.py
```diff
@@ -133,17 +135,18 @@ def test_sbottom_regionB_1400_550_60(
...
-                    0.9338879894557114,
-                    0.9569045303300702,
-                    0.9771296335437559,
-                    0.9916370124133669,
-                    0.9983701133999316,
+                    0.9939882032779436,
+                    0.9961316425804301,
+                    0.9979736724671936,
+                    0.9992686756878328,
+                    0.9998593342137401,
```
alexander-held, line 149: These changes look much larger than the changes in the surrounding regions, what could cause this?

tests/test_optim.py
alexander-held: This change seems to currently not be included in the PR anymore, is the change no longer needed? It seemed rather large, especially compared to everything else changing.

matthewfeickert: @alexander-held there is still a large relative tolerance under `if "minuit-tensorflow-32b" in identifier:` in `tests/test_optim.py` in the PR, but this specific changing of the expected hasn't been set again since I rebased it.

tests/test_regression.py
matthewfeickert: Yeah...I'll need to look at this again once I rebase it after we get another PR #1068 resolved.

tests/test_optim.py
matthewfeickert: Resolved by removal in PR #1504

tests/test_optim.py
matthewfeickert: Resolved by removal in PR #1504

PR #1060, opened 2020-09-11
feat: Add fit to infer CLI API

### LINKED ISSUE #784: provide "pyhf fit" cli
# Description

to just do a max lhood fit one should be able to do `pyhf fit lhood.json` 

this is demo code (from debugging a likelihood w/ @danikam) this is functional but could be cleaned up and integrated into cli/infer.py
```
import json
import sys
import pyhf


w = pyhf.Workspace(json.load(open(sys.argv[1])))
m = w.model(        modifier_settings={
            'normsys': {'interpcode': 'code4'},
            'histosys': {'interpcode': 'code4p'},
        },
)
d = w.data(m)

bestfit = pyhf.infer.mle.fit(d,m)


parameters = {}
for k,v in m.config.par_map.items():
    sl = v['slice']
    npars = sl.stop-sl.start
    value = bestfit[sl]
    parameters[k] = value.tolist()


print(json.dumps(parameters))
```

### LINKED ISSUE #828: docs: List options for optimizers in CLI API docs and add short aliases
# Description

In the [Command Line API docs for `cls`](https://scikit-hep.org/pyhf/cli.html#pyhf-cls) there are available options shown for `--teststat` and `--backend` but not for `--optimizer`.

cls_API_docs

The available options should also be shown for `--optimizer`.

In addition, the options all [currently require](https://github.com/scikit-hep/pyhf/blob/eb9944c043f7f8a75b9cfe6f8bd781bd1dd7014b/src/pyhf/optimize/__init__.py) that the  word "_optimizer" be added to the end (`scipy_optimizer`, `minuit_optimizer`). It would be better to additionally alias these to be just the name as someone at the CLI is already specifying the `--optimizer` flag so having to write `_optimizer` again seems strange. It is nicer to write

```
pyhf cls --patch patch.json --optimizer minuit BkgOnly.json
```

than

```
pyhf cls --patch patch.json --optimizer minuit_optimizer BkgOnly.json
```

though we should keep the `X_optimizer` name available of course.

matthewfeickert:
# Description

* Resolves #784 
* Resolves #828

Adds the following functionality to the CLI API:

```
$ pyhf fit --help
Usage: pyhf fit [OPTIONS] [WORKSPACE]

  Perform a maximum likelihood fit for a given pyhf workspace.

  Example:

  .. code-block:: shell

      $ curl -sL https://git.io/JJYDE | pyhf fit --value

          {
              "mle_parameters": {
                  "mu": [
                      0.00017298628839781602
                  ],
                  "uncorr_bkguncrt": [
                      1.0000015671710816,
                      0.9999665895859197
                  ]
              },
              "twice_nll": 23.19636590468879
          }

Options:
  --output-file TEXT              The location of the output json file. If not
                                  specified, prints to screen.

  --measurement TEXT
  -p, --patch TEXT
  --value                         Flag for returning the fitted value of the
                                  objective function.

  --backend [numpy|pytorch|tensorflow|jax|np|torch|tf]
                                  The tensor backend used for the calculation.
  --optimizer [scipy|minuit]      The optimizer used for the calculation.
  --optconf EQUAL-DELIMITED OPTION
  -h, --help                      Show this message and exit.
```

ReadTheDocs build: https://pyhf.readthedocs.io/en/feat-add-fit-cli-api/cli.html#pyhf-fit

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add fit command to CLI API
   - Performs pyhf.infer.mle.fit
* Add tests for functionality
* Add optimizer options to pyhf cls CLI API docs
```

### DIFF
--- a/src/pyhf/cli/infer.py
+++ b/src/pyhf/cli/infer.py
@@ -34,7 +34,12 @@ def cli():
     help="The tensor backend used for the calculation.",
     default="numpy",
 )
-@click.option("--optimizer")
+@click.option(
+    "--optimizer",
+    type=click.Choice(["scipy", "minuit"]),
+    help="The optimizer used for the calculation.",
+    default="scipy",
+)
 @click.option("--optconf", type=EqDelimStringParamType(), multiple=True)
 def fit(
     workspace,
@@ -138,7 +143,12 @@ def fit(
     help='The tensor backend used for the calculation.',
     default='numpy',
 )
-@click.option('--optimizer')
+@click.option(
+    "--optimizer",
+    type=click.Choice(["scipy", "minuit"]),
+    help="The optimizer used for the calculation.",
+    default="scipy",
+)
 @click.option('--optconf', type=EqDelimStringParamType(), multiple=True)
 def cls(
     workspace,
--- a/src/pyhf/cli/infer.py
+++ b/src/pyhf/cli/infer.py
@@ -69,9 +69,6 @@ def fit(
             "twice_nll": 23.19636590468879
         }
     """
-    with click.open_file(workspace, "r") as specstream:
-        spec = json.load(specstream)
-
     # set the backend if not NumPy
     if backend in ["pytorch", "torch"]:
         set_backend("pytorch", precision="64b")
@@ -90,8 +87,11 @@ def fit(
         )
         set_backend(tensorlib, new_optimizer(**optconf))
 
+    with click.open_file(workspace, "r") as specstream:
+        spec = json.load(specstream)
     ws = Workspace(spec)
     patches = [json.loads(click.open_file(pfile, "r").read()) for pfile in patch]
+
     model = ws.model(
         measurement_name=measurement,
         patches=patches,
--- a/src/pyhf/cli/infer.py
+++ b/src/pyhf/cli/infer.py
@@ -104,19 +104,14 @@ def fit(
 
     fit_result = mle_fit(data, model, return_fitted_val=value)
 
-    bestfit_pars = fit_result
-    if value:
-        bestfit_pars = fit_result[0]
-        bestfit_value = fit_result[-1]
-
-    parameters = {
-        k: tensorlib.tolist(bestfit_pars[v["slice"]])
-        for k, v in model.config.par_map.items()
+    _pars = fit_result if not value else fit_result[0]
+    bestfit_pars = {
+        k: tensorlib.tolist(_pars[v["slice"]]) for k, v in model.config.par_map.items()
     }
 
-    result = {"mle_parameters": parameters}
+    result = {"mle_parameters": bestfit_pars}
     if value:
-        result["twice_nll"] = tensorlib.tolist(bestfit_value)
+        result["twice_nll"] = tensorlib.tolist(fit_result[-1])
 
     if output_file is None:
         click.echo(json.dumps(result, indent=4, sort_keys=True))
--- a/src/pyhf/cli/infer.py
+++ b/src/pyhf/cli/infer.py
@@ -114,9 +114,7 @@ def fit(
         for k, v in model.config.par_map.items()
     }
 
-    result = {
-        "mle_parameters": parameters,
-    }
+    result = {"mle_parameters": parameters}
     if value:
         result["twice_nll"] = tensorlib.tolist(bestfit_value)
 
--- a/src/pyhf/cli/infer.py
+++ b/src/pyhf/cli/infer.py
@@ -110,7 +110,8 @@ def fit(
         bestfit_value = fit_result[-1]
 
     parameters = {
-        k: bestfit_pars[v["slice"]].tolist() for k, v in model.config.par_map.items()
+        k: tensorlib.tolist(bestfit_pars[v["slice"]])
+        for k, v in model.config.par_map.items()
     }
 
     result = {
--- a/src/pyhf/cli/infer.py
+++ b/src/pyhf/cli/infer.py
@@ -26,7 +26,12 @@ def cli():
     default=None,
 )
 @click.option("--measurement", default=None)
-@click.option("--value", default=False, is_flag=True)
+@click.option(
+    "--value",
+    help="Flag for returning the fitted value of the objective function.",
+    default=False,
+    is_flag=True,
+)
 @click.option("-p", "--patch", multiple=True)
 @click.option(
     "--backend",
--- a/src/pyhf/cli/infer.py
+++ b/src/pyhf/cli/infer.py
@@ -6,6 +6,7 @@ import json
 
 from ..utils import EqDelimStringParamType
 from ..infer import hypotest
+from ..infer.mle import fit as mle_fit  # Avoid namespace collision
 from ..workspace import Workspace
 from .. import get_backend, set_backend, optimize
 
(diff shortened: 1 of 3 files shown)


src/pyhf/cli/infer.py
```diff
@@ -17,6 +18,114 @@ def cli():
...
+@click.option("--measurement", default=None)
+@click.option("--value", default=False, is_flag=True)
+@click.option("-p", "--patch", multiple=True)
+@click.option(
+    "--backend",
+    type=click.Choice(["numpy", "pytorch", "tensorflow", "jax", "np", "torch", "tf"]),
+    help="The tensor backend used for the calculation.",
+    default="numpy",
+)
+@click.option("--optimizer")
```
matthewfeickert, line 37: Should add the optimizer options like we do for `backend`

src/pyhf/cli/infer.py
```diff
@@ -6,6 +6,7 @@
 
 from ..utils import EqDelimStringParamType
 from ..infer import hypotest
+from ..infer.mle import fit as mle_fit  # Avoid namespace collision
```
matthewfeickert, line 9: Maybe smarter way to do this.

src/pyhf/cli/infer.py
```diff
@@ -17,6 +18,114 @@ def cli():
...
+@click.option("-p", "--patch", multiple=True)
+@click.option(
+    "--backend",
+    type=click.Choice(["numpy", "pytorch", "tensorflow", "jax", "np", "torch", "tf"]),
+    help="The tensor backend used for the calculation.",
+    default="numpy",
+)
+@click.option("--optimizer")
+@click.option("--optconf", type=EqDelimStringParamType(), multiple=True)
+def fit(
```
matthewfeickert, line 39: Need to add tests

src/pyhf/cli/infer.py
```diff
@@ -17,6 +18,114 @@ def cli():
...
+    data = ws.data(model)
+
+    fit_result = mle_fit(data, model, return_fitted_val=value)
+
+    _pars = fit_result if not value else fit_result[0]
+    bestfit_pars = {
+        k: tensorlib.tolist(_pars[v["slice"]]) for k, v in model.config.par_map.items()
+    }
+
+    result = {"mle_parameters": bestfit_pars}
```
matthewfeickert, line 117: Open to renaming this something other than "mle_parameters"

src/pyhf/cli/infer.py
```diff
@@ -17,6 +18,114 @@ def cli():
...
+    if value:
+        result["twice_nll"] = tensorlib.tolist(fit_result[-1])
```
matthewfeickert, line 119: "twice_nll" would change to "nll" in the other PR I have open, but should still leave it for now.

src/pyhf/cli/infer.py
```diff
@@ -17,6 +18,119 @@ def cli():
...
+    elif backend in ["tensorflow", "tf"]:
+        set_backend("tensorflow", precision="64b")
+    elif backend in ["jax"]:
+        set_backend("jax")
+    tensorlib, _ = get_backend()
+
+    optconf = {k: v for item in optconf for k, v in item.items()}
+
+    # set the new optimizer
+    if optimizer:
```
matthewfeickert, line 94: Codecov is reporting only partial coverage for this line. I have no idea why, so feedback is useful here.

src/pyhf/cli/infer.py
```diff
@@ -17,6 +18,119 @@ def cli():
...
+            "mle_parameters": {
+                "mu": [
+                    0.00017298628839781602
+                ],
+                "uncorr_bkguncrt": [
+                    1.0000015671710816,
+                    0.9999665895859197
+                ]
+            },
+            "twice_nll": 23.19636590468879
```
lukasheinrich, line 79: should we return here NLL given #1057 ?

src/pyhf/cli/infer.py
lukasheinrich: ah ok

src/pyhf/cli/infer.py
lukasheinrich: ok see the other comment..

matthewfeickert: To be clear, I'm fine with getting PR #1057 in first and then this one. I'm just waiting on further feedback on it.

lukasheinrich (approved):

lukasheinrich:
I like this a lot! Probably at some point we want more features to the fit cli (e.g. fixed poi) but we can add those step by step.

kratsg (approved):

matthewfeickert:
> Probably at some point we want more features to the fit cli (e.g. fixed poi) but we can add those step by step.

@lukasheinrich @kratsg If you have thoughts on the API I'm happy to add in fixed parameters now. It seems though that it doesn't make much sense to add a `--fixed` option to the CLI API without also adding an `--init` option. These probably should be something more than a list

```
pyhf fit --fixed [False, True, True]
```

makes it unclear what is actually happening. Though if this discussion is going to get too far down into design we can merge this in and then make a new issue for the API.

kratsg:
> Though if this discussion is going to get too far down into design

main issue here is we'll start getting into maybe making the CLI do too much, and then design a fit config JSON that can be passed in alongside all `infer`-related APIs from the CLI. For now, this is currently fine, as an nonconstrained fit is useful.

src/pyhf/cli/infer.py
kratsg: is this still an issue?

src/pyhf/cli/infer.py
kratsg: can we rename maybe to `postfit_parameters` ?

tests/test_scripts.py
```diff
@@ -96,6 +139,23 @@ def test_import_prepHistFactory_and_cls(tmpdir, script_runner):
...
+    )
+    ret = script_runner.run(*shlex.split(command))
+
+    command = "pyhf fit --backend {0:s} {1:s}".format(backend, temp.strpath)
+    ret = script_runner.run(*shlex.split(command))
+
+    assert ret.success
+    ret_json = json.loads(ret.stdout)
+    assert ret_json
+    assert "mle_parameters" in ret_json
```
kratsg, line 156: we'll also need a slightly fleshed out test to assert the particular output structure to make sure we don't break things in the future for users who rely on the output of this CLI

lukasheinrich:
yeah I agree with @kratsg probably `pyhf fit wspace.json --fitconfig cfg.json` is sufficient... not sure, maybe some general infrastructure for CLIs is usefule where we can override values from the cmd line.. 

`--foo inputjson --opt hello.world=1.0` 

but yeah it's a slippery slope

matthewfeickert:
The discussion RE: the extension of the CLI API has been moved to Issue #1064. The feedback is great so far! :)

tests/test_scripts.py
kratsg: @matthewfeickert this coverage is needed.

matthewfeickert: Okay. Can you change your review status in the future so if I miss these after you've given your review it shows up in the check at the end?

src/pyhf/cli/infer.py
matthewfeickert: Yes, this also got missed.

PR #1068, opened 2020-09-16
fix: Use qtilde as test stat for validation tests

### LINKED ISSUE #1067: Use qtilde for validate_hypotest in validation tests
# Description

In PR #442 `validate_hypotest` set `qtilde` to `False`

https://github.com/scikit-hep/pyhf/blob/8a6ee36da4f566d8a37df01e20201098aa1f8a54/tests/test_validation.py#L572-L580

However, as the parameter bounds are all going to be positive given 

https://github.com/scikit-hep/pyhf/blob/8a6ee36da4f566d8a37df01e20201098aa1f8a54/tests/test_validation.py#L569-L570

one will get the warnings

```
```

@kratsg's memory of this is that

> we validated against `qmu` initially with histfitter, but later down the line figured out a way to get qmu/qtilde automatically determined. So this was less of an issue.

So `qtilde=True` should be used instead.

matthewfeickert:
# Description

Resolves #1067 

Set `qtilde=True` in `validate_hypotest` as changes in the validation process now make `qtilde` the correct test stat.

https://github.com/scikit-hep/pyhf/blob/8a6ee36da4f566d8a37df01e20201098aa1f8a54/tests/test_validation.py#L572-L580

See Issue #1067 for some historical notes and motivation.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
- Use qtilde as the test statistic for validate_hypotest in validation tests
   - Change from PR #442 
   - Eliminates warnings on test stat choice
- Correct expected result for 1bin_lumi test
```


kratsg (approved):

lukasheinrich:
Should we maybe have this test in addition to the old test instead of replacing it?

kratsg:
> Should we maybe have this test in addition to the old test instead of replacing it?

We probably need the old test, but change the bounds to allow the POI to go negative so as to not raise warnings during the test itself.

matthewfeickert:
> > Should we maybe have this test in addition to the old test instead of replacing it?
> 
> We probably need the old test, but change the bounds to allow the POI to go negative so as to not raise warnings during the test itself.

Is this actually necessary? There wasn't anything in particular related to the validation that required the POI to be negative.

matthewfeickert:
@kratsg @lukasheinrich it has been a long time so can you help remind me of the plan? Was the idea to make a new `expected_result_1bin_lumi` fixture and model to support `q` having lower bounds that are negative?

kratsg:
> @kratsg @lukasheinrich it has been a long time so can you help remind me of the plan? Was the idea to make a new `expected_result_1bin_lumi` fixture and model to support `q` having lower bounds that are negative?

Basically just have a fixture similar to one of them, but using the qtilde test stat instead of the q one -- with the values, etc... we need to refactor all of this later, but for now, just copy/paste and make the test_validation even longer.

matthewfeickert:
> Basically just have a fixture similar to one of them, but using the qtilde test stat instead of the q one -- with the values, etc... we need to refactor all of this later, but for now, just copy/paste and make the test_validation even longer.

This will require more than that given the setup is all name based

https://github.com/scikit-hep/pyhf/blob/5c05e6ea90157576bc3c3b2277d6bc9861ed6e7a/tests/test_validation.py#L755-L759

but we can't reuse the spec from `qtilde` for `qmu` as the bounds will necessarily need to be changed. As the name has to be the same for the source and the spec this is going to require a rename of all `source` and `spec` fixtures to be able to differentiate between `qtilde` and `q`.

tests/test_validation.py
```diff
@@ -689,67 +693,74 @@ def validate_hypotest(
             1e-6,
             "toybased",
         ),
-        ('1bin_lumi', {'init_pars': 2, 'par_bounds': 2}, 1.0, "q", 4e-6, "asymptotics"),
         (
-            '1bin_normsys',
+            '1bin_lumi_qtilde',
+            {'init_pars': 2, 'par_bounds': 2},
+            1.0,
+            "qtilde",
```
matthewfeickert, line 700: Note to self: Even if the naming change can be reverted this change from `q` -> `qtilde` needs to stay.

kratsg:
> without renaming everything and creating a new spec?

no, hence why the need for refactor later. #1241.

kratsg:
Oh, I misunderstood. The names don't need to change. Only need to copy out the spec for one of the existing ones, add "qmu" to the name instead -- and change the bounds to `[-10, 10]` so that the rest don't need to get renamed, and are switched to qtilde test stat.

matthewfeickert:
> Only need to copy out the spec for one of the existing ones, add "qmu" to the name instead -- and change the bounds to `[-10, 10]` so that the rest don't need to get renamed, and are switched to qtilde test stat.

Hm, so maybe I'm misunderstanding the original request, but you only want to test a single spec with `q` and `qtilde` and not all? If so, then probably just do `1bin_shapesys_mu1` so that we'd go from 

https://github.com/scikit-hep/pyhf/blob/5c05e6ea90157576bc3c3b2277d6bc9861ed6e7a/tests/test_validation.py#L743-L745

to 

```python
         '1bin_shapesys_qtilde_mu1', 
         '1bin_shapesys_q_mu1', 
         '1bin_shapesys_q0_mu1', 
         '1bin_shapesys_q0_mu1_toys', 
```

?

In the last commit that I had made here I was thinking that we needed to test all the specs with both `qtilde` and `q`, but if we just want to test and validate one then that's much easier.

Issue #1069, opened 2020-09-21
iminuit v1.5.0 breaks optimization tests


matthewfeickert:
# Description

With the release of [`iminuit` `v1.5.0`](https://github.com/scikit-hep/iminuit/releases/tag/v1.5.0) on 2020-09-17 the nightly tests are failing in `test_optim.py`. Specifically

https://github.com/scikit-hep/pyhf/blob/8a6ee36da4f566d8a37df01e20201098aa1f8a54/tests/test_optim.py#L47

is failing with errors of 

```pytb
        try:
            assert result.success
        except AssertionError:
            log.error(result)
>           raise exceptions.FailedMinimization(result)
E           pyhf.exceptions.FailedMinimization: Optimization failed. Estimated distance to minimum too large.

src/pyhf/optimize/mixins.py:52: FailedMinimization
------------------------------ Captured log call -------------------------------
ERROR    pyhf.optimize.mixins:mixins.py:51       fun: 15.5887451171875
 hess_inv: array([[1., 1.],
       [1., 1.]])
  message: 'Optimization failed. Estimated distance to minimum too large.'
   minuit: <iminuit._libiminuit.Minuit object at 0x5619c82f90a0>
     nfev: 110
     njev: 0
  success: False
      unc: None
        x: array([0.97325551, 0.91712703])
```

where the `pyhf.exceptions.FailedMinimization` being raised comes from the `raise exceptions.FailedMinimization(result)` in

https://github.com/scikit-hep/pyhf/blob/8a6ee36da4f566d8a37df01e20201098aa1f8a54/src/pyhf/optimize/mixins.py#L31-L53

which are of course coming from

https://github.com/scikit-hep/pyhf/blob/8a6ee36da4f566d8a37df01e20201098aa1f8a54/src/pyhf/optimize/opt_minuit.py#L122-L132

in

https://github.com/scikit-hep/pyhf/blob/8a6ee36da4f566d8a37df01e20201098aa1f8a54/src/pyhf/optimize/opt_minuit.py#L69

# Steps to Reproduce

Run the tests using current master.

![iminuit_breaks](https://user-images.githubusercontent.com/5142394/93731644-8ba78f80-fb93-11ea-8060-9e05dc61b102.png)

To show that this is definitley an issue with `iminuit` `v1.5.0+`

```
$ python -m pip install --upgrade "iminuit<1.5.0"
$ pip list | grep iminuit
iminuit                       1.4.9
[...]

PR #1071, opened 2020-09-21
ci: Add nightly tests using HEAD of dependencies

matthewfeickert:
# Description

In light of Issue #1069, this PR adds nightly tests that run at the `HEAD` of dependencies that are the most reasonable for us to check and interact with. If we detect breaking changes in these dependencies at their `HEAD` then we should have time to either proactively address this in `pyhf` or work with the dependency dev team to mitigate them. They can also be run on demand with workflow dispatch.

Note that these tests only run on CRON jobs and on demand, but never on pushes of PRs.

Example workflow: https://github.com/scikit-hep/pyhf/actions/runs/265261547

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add nightly tests to run test suite at HEAD of dependencies
   - Use as early alert for breaking changes in dependencies
   - Run for SciPy, iminuit, and uproot
```

### DIFF
--- a/.github/workflows/dependencies-head.yml
+++ b/.github/workflows/dependencies-head.yml
@@ -25,6 +25,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install --ignore-installed -U -q --no-cache-dir -e .[test]
+        python -m pip install --upgrade --no-cache-dir cython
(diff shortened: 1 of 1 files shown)


matthewfeickert:
I didn't include PyTorch, TensorFlow, or JAX as in addition to being huge and taking forever, their `master` branches are also not very stable to begin with.

kratsg (approved):

PR #1088, opened 2020-10-02
feat: Merge channel functionality in pyhf combine

kratsg:
# Description

This adds functionality to get channel merging as part of a "combination". Can only be done with `outer`, `left outer`, or `right outer`:

```
pyhf combine --join "left outer" \
             --merge-channels \
             <(pyhf prune --sample C1N2_Wh_hbb_800_250 pyhf_likelihood.json)\
             test.json
```

Additionally, clean up the logic in `pyhf.workspace._join_items` to fix up coverage.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Extend combine to allow for merging channels from two workspaces when performing an outer join
  - This is generally useful when you want to combine a bkg-only workspace with a signal-only workspace with the same set of channels, but need to merge the channels together (extend bkg-only samples with signal-only samples for each channel)
* Improve the coverage of pyhf/workspace.py by cleaning up the pyhf.workspace._join_items logic
```

### DIFF
--- a/src/pyhf/workspace.py
+++ b/src/pyhf/workspace.py
@@ -17,7 +17,7 @@ from .mixins import _ChannelSummaryMixin
 log = logging.getLogger(__name__)
 
 
-def _join_items(join, left_items, right_items, key='name'):
+def _join_items(join, left_items, right_items, key='name', deep_merge_key=''):
     """
     Join two lists of dictionaries along the given key.
 
@@ -27,6 +27,7 @@ def _join_items(join, left_items, right_items, key='name'):
         join (`str`): The join operation to apply. See ~pyhf.workspace.Workspace for valid join operations.
         left_items (`list`): A list of dictionaries to join on the left
         right_items (`list`): A list of dictionaries to join on the right
+        deep_merge_key (`str`): A key on which to deeply merge items if set.
 
     Returns:
         :obj:`list`: A joined list of dictionaries.
@@ -39,14 +40,23 @@ def _join_items(join, left_items, right_items, key='name'):
     joined_items = copy.deepcopy(primary_items)
     for secondary_item in secondary_items:
         # outer join: merge primary and secondary, matching where possible
+        keys = [item[key] for item in joined_items]
         if join == 'outer' and secondary_item in primary_items:
             continue
         # left/right outer join: only add secondary if existing item (by key value) is not in primary
         # NB: this will be slow for large numbers of items
-        elif join in ['left outer', 'right outer'] and secondary_item[key] in [
-            item[key] for item in joined_items
-        ]:
+        elif join in ['left outer', 'right outer'] and secondary_item[key] in keys:
+            # Deeply merge a sublist as well, if we need to
+            if deep_merge_key:
+                _deep_left_items = joined_items[keys.index(secondary_item[key])][
+                    deep_merge_key
+                ]
+                _deep_right_items = secondary_item[deep_merge_key]
+                joined_items[keys.index(secondary_item[key])][
+                    deep_merge_key
+                ] = _join_items('left outer', _deep_left_items, _deep_right_items)
             continue
+
         joined_items.append(copy.deepcopy(secondary_item))
     return joined_items
 
@@ -74,7 +84,7 @@ def _join_versions(join, left_version, right_version):
     return left_version
 
 
-def _join_channels(join, left_channels, right_channels):
+def _join_channels(join, left_channels, right_channels, merge=False):
     """
     Join two workspace channel specifications.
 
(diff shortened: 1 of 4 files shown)


src/pyhf/workspace.py
```diff
@@ -685,6 +698,7 @@ def combine(cls, left, right, join='none'):
             left (~pyhf.workspace.Workspace): A workspace
             right (~pyhf.workspace.Workspace): Another workspace
             join (:obj:`str`): How to join the two workspaces. Pick from "none", "outer", "left outer", or "right outer".
+            merge_channels (:obj:`bool`): Whether or not to merge channels when performing the combine. This is only done with "left outer" or "right outer" options.
```
matthewfeickert, line 701: Above you have ```merge (`bool`)``` and here ``` merge_channels (:obj:`bool`)```. It would be good if we can just stick to one style.

tests/test_scripts.py
```diff
@@ -529,6 +529,23 @@ def test_combine_outfile(tmpdir, script_runner):
...
+    temp_1 = tmpdir.join("parsed_output.json")
+    temp_2 = tmpdir.join("renamed_output.json")
+    command = f'pyhf xml2json validation/xmlimport_input/config/example.xml --basedir validation/xmlimport_input/ --output-file {temp_1.strpath} --hide-progress'
+    ret = script_runner.run(*shlex.split(command))
+
+    command = (
+        f'pyhf prune {temp_1.strpath} --sample signal --output-file {temp_2.strpath}'
+    )
+
+    ret = script_runner.run(*shlex.split(command))
```
matthewfeickert, line 542:
```suggestion
    ret = script_runner.run(*shlex.split(command))
    assert ret.success

    command = (
        f'pyhf prune {temp_1.strpath} --sample signal --output-file {temp_2.strpath}'
    )

    ret = script_runner.run(*shlex.split(command))
    assert ret.success
```

For clarity we should probably not not assign if things are never going to be used. Can we either remove the assignment to `ret` or add in asserts?

src/pyhf/workspace.py
```diff
@@ -17,7 +17,7 @@
 log = logging.getLogger(__name__)
 
 
-def _join_items(join, left_items, right_items, key='name'):
+def _join_items(join, left_items, right_items, key='name', deep_merge_key=''):
```
matthewfeickert, line 20:
```suggestion
def _join_items(join, left_items, right_items, key='name', deep_merge_key=None):
```

Probably keep this as `None` instead of empty string?

src/pyhf/workspace.py
```diff
@@ -39,14 +40,23 @@ def _join_items(join, left_items, right_items, key='name'):
...
         if join == 'outer' and secondary_item in primary_items:
             continue
         # left/right outer join: only add secondary if existing item (by key value) is not in primary
         # NB: this will be slow for large numbers of items
-        elif join in ['left outer', 'right outer'] and secondary_item[key] in [
-            item[key] for item in joined_items
-        ]:
+        elif join in ['left outer', 'right outer'] and secondary_item[key] in keys:
+            # Deeply merge a sublist as well, if we need to
+            if deep_merge_key:
```
matthewfeickert, line 50:
```suggestion
            if deep_merge_key is not None:
```

In keeping with my `None` suggestion above.

src/pyhf/workspace.py
```diff
@@ -85,13 +95,16 @@ def _join_channels(join, left_channels, right_channels):
...
+        merge (`bool`): Whether to deeply merge channels or not.
 
     Returns:
         :obj:`list`: A joined list of channels. Each channel follows the :obj:`defs.json#/definitions/channel` `schema <https://scikit-hep.org/pyhf/likelihood.html#channel>`__
 
     """
 
-    joined_channels = _join_items(join, left_channels, right_channels)
+    joined_channels = _join_items(
+        join, left_channels, right_channels, deep_merge_key='samples' if merge else ''
```
matthewfeickert, line 106:
```suggestion
        join, left_channels, right_channels, deep_merge_key='samples' if merge else None
```

tests/test_workspace.py
```diff
@@ -300,14 +300,21 @@ def test_rename_measurement(workspace_factory):
...
+        {'name': 'right', 'key': 'value', 'deep': [{'name': 2}]},
+        {'name': 'common', 'key': 'right', 'deep': [{'name': 2}]},
+    ]
     return (left, right)
 
 
 def test_join_items_none(join_items):
     left_items, right_items = join_items
     joined = pyhf.workspace._join_items('none', left_items, right_items, key='name')
+    breakpoint()
```
matthewfeickert, line 317:
```suggestion
```

matthewfeickert (changes requested):

src/pyhf/workspace.py
kratsg: they're different functions. one is `_join_channels` and this one is the public-facing API.

matthewfeickert: I mean the syntax of ````bool```` vs ``` :obj:`bool` ```

kratsg: #1116

matthewfeickert: This is now https://github.com/scikit-hep/pyhf/issues/1116

matthewfeickert (approved):

src/pyhf/workspace.py
```diff
@@ -39,14 +40,23 @@ def _join_items(join, left_items, right_items, key='name'):
...
         if join == 'outer' and secondary_item in primary_items:
             continue
         # left/right outer join: only add secondary if existing item (by key value) is not in primary
         # NB: this will be slow for large numbers of items
-        elif join in ['left outer', 'right outer'] and secondary_item[key] in [
-            item[key] for item in joined_items
-        ]:
+        elif join in ['left outer', 'right outer'] and secondary_item[key] in keys:
+            # Deeply merge a sublist as well, if we need to
+            if deep_merge_key is not None:
```
matthewfeickert, line 50: I'm not sure why Codecov is still complaining, but given that it is a partial hit instead of just no coverage I think we can let it slide, though not ideal. I'm confused why it thinks it isn't getting full coverage though.

lukasheinrich (approved):

PR #1101, opened 2020-10-13
feat: Add pyhf.workspace.build

lukasheinrich:
# Description

```
)
workspace = pyhf.workspace.build(model, data)
```

this is the inverse of 

```
w = pyf.Workspace(foo)
model = w.model()
data = w.data(model)
```

maybe interesting for @alexander-held, too

ReadTheDocs build: https://pyhf.readthedocs.io/en/workspace_maker/_generated/pyhf.workspace.Workspace.html

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add workspace class method to build workspaces from a model and data
* Add test to validate build behavior
```


src/pyhf/workspace.py
```diff
@@ -743,3 +743,27 @@ def sorted(cls, workspace):
...
+        model (tensor): A array holding observations to store into a workspace
+
+    Returns:
+        ~pyhf.workspace.Workspace: A new  workspace object
+
+    """
+    workspace = copy.deepcopy(model.spec)
+    workspace['version'] = '1.0.0'
+    workspace['measurements'] = [
+        {'name': name, 'config': {'poi': model.config.poi_name, 'parameters': []}}
```
lukasheinrich, line 763: this ignores fixed parameters. @kratsg maybe we add it in a follow-up PR?

lukasheinrich:
black seems happy on my machine, but not in  CI, @matthewfeickert any idea?

matthewfeickert:
> black seems happy on my machine, but not in CI, @matthewfeickert any idea?

Where is `black` unhappy? It isn't complaining in the CI as the Lint workflow is passing.

src/pyhf/workspace.py
```diff
@@ -743,3 +743,27 @@ def sorted(cls, workspace):
         newspec['observations'].sort(key=lambda e: e['name'])
 
         return cls(newspec)
+
+
+def build(model, data, name='measurement'):
```
kratsg, line 748: maybe make this a `@classmethod` so we can return a `cls(....)` ?

src/pyhf/workspace.py
```diff
@@ -743,3 +743,27 @@ def sorted(cls, workspace):
...
+
+    Args:
+        model (~pyhf.pdf.Model): A model to store into a workspace
+        model (tensor): A array holding observations to store into a workspace
+
+    Returns:
+        ~pyhf.workspace.Workspace: A new  workspace object
+
+    """
+    workspace = copy.deepcopy(model.spec)
```
kratsg, line 760: i think we should be a bit explicit here, no?
```suggestion
    workspace = dict(channels=copy.deepcopy(model.spec['channels']))
```

src/pyhf/workspace.py
```diff
@@ -743,3 +743,27 @@ def sorted(cls, workspace):
...
+    workspace['version'] = '1.0.0'
```
kratsg, line 761:
```suggestion
    workspace['version'] = pyhf.utils.SCHEMA_VERSION
```

tests/test_workspace.py
```diff
@@ -721,3 +722,19 @@ def test_sorted(workspace_factory):
...
+    data = [51, 48]
+    one = pyhf.infer.hypotest(1.0, data + model.config.auxdata, model)
+
+    workspace = pyhf.workspace.build(model, data)
+    wspace = pyhf.Workspace(json.loads(json.dumps(workspace)))
+    newmodel = wspace.model()
+    newdata = wspace.data(newmodel)
+    two = pyhf.infer.hypotest(1.0, newdata, newmodel)
+
+    assert one == two
```
kratsg, line 740: perhaps we can just compute the digest of the two workspaces and assert the equivalence?

kratsg (changes requested):

tests/test_workspace.py
lukasheinrich: there is no "previous" workspace,.. this is a method to create a wspace from model + data

src/pyhf/workspace.py
lukasheinrich: this returns pure JSON, not a workspace object, should we change it?

src/pyhf/workspace.py
matthewfeickert: @lukasheinrich I haven't looked at this much yet, but how much work would it take to get this all in one?

kratsg: something like this

```
param_configs = [{'name': param_name, 'inits': model.config.suggested_init()[model.config.param_set(param_name)['slice']], 'bounds': model.config.suggested_bounds()[model.config.param_set(param_name)['slice']), 'fixed': True} for param_name in model.config.parameters if model.config.suggested_fixed()[.....]]
```

tests/test_workspace.py
kratsg: right, but we should check a "roundtrip" of sorts with simpler workspaces... more complex ones won't be so easy to round-trip.

src/pyhf/workspace.py
kratsg: fix the docstring then? it wasn't clear to me. Although for a python API - it makes sense to return the workspace object. This is dumpable as JSON for printing to the screen if needed, etc.

matthewfeickert (commented):
This is looking good @lukasheinrich. :+1: I rebased it and fixed a few small things that got dropped in the merges and added in some things missing to the docstring.

This is basically done, but I agree with @kratsg thought that
> but we should check a "roundtrip" of sorts with simpler workspaces

src/pyhf/workspace.py
```diff
@@ -743,3 +743,42 @@ def sorted(cls, workspace):
...
+                'name': name,
+                'config': {
+                    'poi': model.config.poi_name,
+                    'parameters': [
+                        {
+                            "bounds": [list(x) for x in v['paramset'].suggested_bounds],
+                            "inits": v['paramset'].suggested_init,
+                            "fixed": v['paramset'].fixed,
+                            "name": k,
+                        }
```
lukasheinrich, line 774: this works, but i'm not sure whether we can avoid being very verbose here. In standard workspaces from xml2json we only add stuff here that is "non-default' while this will dump all parameter configurations even the default ones

lukasheinrich: @kratsg

src/pyhf/workspace.py
lukasheinrich: done

tests/test_workspace.py
lukasheinrich: done

src/pyhf/workspace.py
```diff
@@ -743,3 +743,42 @@ def sorted(cls, workspace):
...
+            model (~pyhf.pdf.Model): A model to store into a workspace
+            data (tensor): A array holding observations to store into a workspace
+            name (str): The name of the workspace measurement
+
+        Returns:
+            ~pyhf.workspace.Workspace: A new workspace object
+
+        """
+        workspace = copy.deepcopy(dict(channels=model.spec['channels']))
+        workspace['version'] = '1.0.0'
```
matthewfeickert, line 762: @lukasheinrich This keeps on getting broken in the merge. This should use @kratsg's suggestion.
```suggestion
        workspace['version'] = utils.SCHEMA_VERSION
```

matthewfeickert (changes requested):

src/pyhf/workspace.py
lukasheinrich: alternatively one could use `model.spec['parameters']` but a) i don't know why parameters is part of the model spec in the first place and b) we strip out the names of the parameter sets.

kratsg: Can we compare against the default and only add in non-default settings? This should be possible based on the code you have

```
parameters = []
for par_name, par_config in model.config.par_map.items():
  _par_config = {}
  bounds = par_config['paramset'].suggested_bounds
  inits = par_config['paramset'].suggested_init
  fixed = par_config['paramset'].fixed
  default_parset = getattr(pyhf.modifiers, dict(model.config.modifiers)[par_name]).required_parset([0]*len(inits), [0]*len(inits))
  if bounds != list(default_parset['bounds']):
    _par_config['bounds'] = bounds
  if inits != list(default_parset['inits']):
    _par_config['inits'] = inits
  if fixed != _par_config['fixed']:
    _par_config['fixed'] = fixed
  parameters.append(_par_config)
```

src/pyhf/exceptions/__init__.py
```diff
@@ -33,8 +34,8 @@ def __init__(self, ValidationError):
                 self.path += '.{}'.format(item)
         self.path = self.path.lstrip('.')
         self.instance = ValidationError.instance
-        message = '{0}.\n\tPath: {1}\n\tInstance: {2}'.format(
-            ValidationError.message, self.path, self.instance
+        message = '{0}.\n\tPath: {1}\n\tInstance: {2} Schema: {3}'.format(
+            ValidationError.message, self.path, self.instance, self.schema
```
matthewfeickert, line 38: It would be great if these were f-strings instead
```suggestion
        message = f"{ValidationError.message}.\n\tPath: {self.path}\n\tInstance: {self.instance} Schema: {self.schema}"
```

matthewfeickert (commented):
I just have a formatting suggestion and then I agree with @kratsg's suggestion to only add in non-default settings.

matthewfeickert (approved):
Thanks for this @lukasheinrich. :+1:

PR #1102, opened 2020-10-13
JSON example

### LINKED ISSUE #1100: docs: add JSON workspace example to READMe
# Description

people often ask what the JSON looks like. we should give them an easy way to see.

lukasheinrich:
# Description

resolves #1100

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,6 +45,50 @@ Hello World
    >>> print(f"Observed: {CLs_obs}, Expected: {CLs_exp}")
    Observed: 0.05251497423736956, Expected: 0.06445320535890459
 
+
+.. code:: bash
+
+   cat << EOF  | pyhf cls
+   {
+      "channels": [
+         {
+            "name": "singlechannel",
+            "samples": [
+               {
+                  "name": "signal", "data": [12,11],
+                  "modifiers": [ {"name": "mu","type": "normfactor","data": null } ]
+               },
+               {
+                  "name": "background", "data": [ 50, 52 ],
+                  "modifiers": [ { "name": "uncorr_bkguncrt", "type": "shapesys", "data": [3,7]} ]
+               }
+            ]
+         }
+      ],
+      "version": "1.0.0",
+      "measurements": [
+         {  "name": "measurement", "config": { "poi": "mu", "parameters": [] } }
+      ],
+      "observations": [
+            {
+               "name": "singlechannel",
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -45,6 +45,50 @@ Hello World
...
+      "measurements": [
+         {  "name": "measurement", "config": { "poi": "mu", "parameters": [] } }
+      ],
+      "observations": [
+            {
+               "name": "singlechannel",
+               "data": [ 51, 48 ]
+            }
+      ]
+   }
```
matthewfeickert, line 78: It would be good if we could have consistent formatting between the example above this and here. Can we generate this from Blacked Python code and then dump it as JSON?

README.rst
```diff
@@ -45,6 +45,50 @@ Hello World
    >>> print(f"Observed: {CLs_obs}, Expected: {CLs_exp}")
    Observed: 0.05251497423736956, Expected: 0.06445320535890459
 
+
+.. code:: bash
+
+   cat << EOF  | pyhf cls
```
matthewfeickert, line 51:
```suggestion
   $ cat << EOF  | pyhf cls
```

Maybe add a shell indicator?

README.rst
```diff
@@ -45,6 +45,50 @@ Hello World
    >>> print(f"Observed: {CLs_obs}, Expected: {CLs_exp}")
    Observed: 0.05251497423736956, Expected: 0.06445320535890459
 
+
```
matthewfeickert, line 48: Can there be some additional text explaining what is happening so that it isn't just two code blocks one after another?

kratsg (approved):

README.rst
lukasheinrich: done

README.rst
lukasheinrich: doone

README.rst
lukasheinrich: I'm not sure I follow: want to run black on JSON?

matthewfeickert: No, I'm just saying that we don't have consistent formatting in this example. I just mean that it would be nice to have the formatting of https://github.com/scikit-hep/pyhf/blob/master/docs/examples/json/2-bin_1-channel.json but with the numbers used here. I can fix this though.

matthewfeickert (changes requested):

Issue #1004, opened 2020-07-29
Explicit setting of the default backend


phinate:
It may be desirable in some cases (e.g. #882) to have one consistent backend for all cases, which would enable the propagation of gradients throughout pyhf model construction, for instance. Backends will hit a wall when trying to take the gradient with respect to something that involves pure numpy.

An example implementation of this could be as simple as adding the argument `set_default=False` to `set_backend`, followed by something like

```
def set_backend(..., set_default=False)
...
backend = tensorlib
if set_default:
        global default_backend
        default_backend = tensorlib
...
tensorlib._setup()
if set_default:
    default_backend._setup()
```

It seems like this would be fine at first glance, since the backend checks are done in the lines before this. I managed to get the `logpdf` of `hepdata_like` with no issues, but I was unsure what would be a more robust test for changing the default backend. Ideally, one would want to run the whole test suite I guess, to make sure this doesn't secretly change some functionality.

Hope this is useful!

kratsg:
The default backend can be set w/o changing the rest of the code. It'll help with the diff'able case for sure.

PR #1118, opened 2020-10-15
feat: Add ability to set default backend

phinate:
Adresses https://github.com/scikit-hep/pyhf/issues/1004.

Adds a `get_default_backend` function that is called instead of the global `default_backend`, meaning that one can change the default backend with `pyhf.default_backend = pyhf.tensor.tensorflow_backend(precision='64b')` or similar.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,64 +46,6 @@ jobs:
       run: |
         python -m pytest -r sx --benchmark-sort=mean tests/benchmarks/test_benchmark.py
 
-  docs:
-
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: [3.8]
-
-    steps:
-    - uses: actions/checkout@v2
-      with:
-        fetch-depth: 0
-    - name: Set up Python 3.8
-      uses: actions/setup-python@v2
-      with:
-        python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip setuptools wheel
-        python -m pip install --ignore-installed -U -q --no-cache-dir -e .[docs,test]
-        python -m pip list
-        sudo apt-get update
-        sudo apt-get -qq install pandoc
-    - name: Check docstrings
-      run: |
-        # Group 1 is related to docstrings
-        pydocstyle --select D1 src/pyhf/pdf.py \
-                               src/pyhf/workspace.py \
-                               src/pyhf/probability.py \
-                               src/pyhf/patchset.py \
-                               src/pyhf/interpolators \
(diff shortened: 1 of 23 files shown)


matthewfeickert:
@phinate Thanks for stepping up to give this a crack. :) We're all obviously happy to help out on this should you have questions.

A heads up that this is going to be a bit tricky to cleanly keep up to date as this was made from your fork's `master` branch. If we rebase this of `pyhf`'s `master` at any point GitHub will get confused and think that they will be at the same point with no version history and automatically close the PR. c.f. https://github.com/scikit-hep/pylhe/pull/27 where I learned this the hard way.

phinate:
> @phinate Thanks for stepping up to give this a crack. :) We're all obviously happy to help out on this should you have questions.
> 
> A heads up that this is going to be a bit tricky to cleanly keep up to date as this was made from your fork's `master` branch. If we rebase this off `pyhf`'s `master` at any point GitHub will get confused and think that they will be at the same point with no version history and automatically close the PR. c.f. [scikit-hep/pylhe#27](https://github.com/scikit-hep/pylhe/pull/27) where I learned this the hard way.

Ah, this is my mistake! I was trying to port changes from a different branch to master, so my brain defaulted that of course I must keep these changes in master itself...

Is it cleaner to make a different PR from a new branch, or just to try and get tests etc passing quickly?

matthewfeickert:
> Is it cleaner to make a different PR from a new branch, or just to try and get tests etc passing quickly?

@phinate Truly up to you. It might be easier, but I don't think it is going to break your ability to work here.

phinate:
@matthewfeickert do you know why codefactor fails for doing `global` statements? I just copied the syntax for the normal backend, and it doesn’t flag that?

kratsg (changes requested):
I have a hard time reviewing since this isn't quite so up to date with pyhf master. That said, I would like to suggest a slight improvement to the API to reduce the public namespace `pyhf` exposes:

```python
tensorlib, optimizer = pyhf.get_backend(....) # default=False by.. default.. pun intended
default_backend, default_optimizer = pyhf.get_backend(...., default=True)
pyhf.set_backend(....)
pyhf.set_backend(...., default=True)
```

This way, we can keep using the same functions that have existed for so long, continue to do the caching as normal, but simply add a shortcircuit in those functions to change the default if needed. That is, `set_backend(default=True)` will set the backend as expected, but also additionally change the default backends.

phinate:
Not sure what went wrong with reviewing, but I'll start this again from the current `master` and a separate branch to make life easier @kratsg @matthewfeickert :)

PR #1121, opened 2020-10-16
Feat: change default backend

phinate:
Adresses https://github.com/scikit-hep/pyhf/issues/1004.

Following suggestions from @kratsg, this adds the kwarg `default=False` to `set_backend()` and `get_backend()`. 

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
```


phinate:
One comment ahead of time: setting the default backend to `jax`, for instance, breaks things. That's great though, because it allows autodiff-related roadblocks to be identified before they happen :)

Will investigate more and follow up with another issue.

src/pyhf/__init__.py
```diff
@@ -138,17 +133,44 @@ def set_backend(backend, custom_optimizer=None, precision=None):
...
-    if tensorlib_changed:
-        events.trigger("tensorlib_changed")()
-    if optimizer_changed:
-        events.trigger("optimizer_changed")()
-    # set up any other globals for backend
-    tensorlib._setup()
+    if default:
+        global default_backend
+        global default_optimizer
+        # need to determine if the default tensorlib changed or the default optimizer changed for events
```
kratsg, line 139: noting for now -- but a problem here is that we can't re-run the `__init__` of many things with events trigger. Likely, a bunch of the other code will need to be refactored so that we can change the default like this. We might need to consider having the default backend being set up at `import` time instead of being configurable like this. Might be a more sane option, otherwise, we end up needing to account for a significant more logic than expected.

phinate: hm I see, how would a user configure that ahead of time before `import pyhf`?

lukasheinrich: One thing to ponder is whether we might want to just drop support for changing backcends.. and just requirre people to choose ahead of time (like matplotlib has backends etc)

kratsg: > drop support for changing backcends

👀

I think at the present state, it's very helpful for users to be able to change quickly (and helps us gain more trust with them). I would not be opposed in the long term, but at the moment... ehh.

@phinate -- an idea would be to do something more like what matplotlib does

`pyhf.use('numpy')` to set the default tensor backend at the very start after importing pyhf.

kratsg:
Closing in favor of #1646.

PR #1128, opened 2020-10-19
fix: Logic bug for outer joins of merge channels

kratsg:
# Description

This technically had the right tests in place, but simply forgot to add `outer` join as a join method to check the logic of the merge channels for. This PR will add it in, but then also fix the logic bug (reverse the conditionals).

See #1088 .

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Fix logic bug for merge channels of combine for outer join functionality
* Amends PR #1088
```


matthewfeickert (approved):
Thanks @kratsg

PR #1134, opened 2020-10-20
fix: Use JAX v0.2.4+ API

matthewfeickert:
# Description

With release [`v0.2.4` of JAX](https://github.com/google/jax/releases/tag/jax-v0.2.4) the underlying stats API has changed as there is no longer a `jax.scipy.stats.norm.osp_stats` or `jax.scipy.stats.poisson.osp_stats` API, but instead [JAX now just uses the SciPy API directly with](https://github.com/google/jax/blob/4b8334ab0b8d9d3f1c676866d177f3e3e6b4c250/jax/_src/scipy/stats/poisson.py#L16)

```
import scipy.stats as osp_stats
```

As such, this requires a minimum version bump of `jax` and `jaxlib` and then also requires changing the imports to use SciPy directly. Given that the continuous approximation to the Poisson is used there is actually no need to import `jax.scipy.stats.poisson` anymore either.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use JAX v0.2.4 API with regards to using scipy.stats directly
* Update to jax v0.2.4+ and jaxlib v0.1.56+
   - Restrict to jax v0.2.X and jaxlib v0.1.X to ensure stability
* Add tests for Poisson and Normal sample shape
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ extras_require = {
         'tensorflow-probability~=0.10.0',
     ],
     'torch': ['torch~=1.2'],
-    'jax': ['jax~=0.1,>0.1.51', 'jaxlib~=0.1,>0.1.33'],
+    'jax': ['jax~=0.2,>=0.2.4', 'jaxlib~=0.1,>=0.1.56'],
     'xmlio': ['uproot~=3.6'],  # Future proof against uproot4 API changes
     'minuit': ['iminuit~=1.4.3'],  # v1.5.0 breaks pyhf for 32b TensorFlow and PyTorch
 }
--- a/setup.py
+++ b/setup.py
@@ -7,7 +7,7 @@ extras_require = {
         'tensorflow-probability~=0.10.0',
     ],
     'torch': ['torch~=1.2'],
-    'jax': ['jax~=0.2,>=0.2.4', 'jaxlib~=0.1,>=0.1.56'],
+    'jax': ['jax~=0.2.4', 'jaxlib~=0.1.56'],
     'xmlio': ['uproot~=3.6'],  # Future proof against uproot4 API changes
     'minuit': ['iminuit~=1.4.3'],  # v1.5.0 breaks pyhf for 32b TensorFlow and PyTorch
 }
--- a/src/pyhf/tensor/jax_backend.py
+++ b/src/pyhf/tensor/jax_backend.py
@@ -7,6 +7,7 @@ from jax.scipy.special import gammaln
 from jax.scipy import special
 from jax.scipy.stats import norm, poisson
 import numpy as onp
+import scipy.stats as osp_stats
 import logging
 
 log = logging.getLogger(__name__)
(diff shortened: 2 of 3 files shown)


src/pyhf/tensor/jax_backend.py
```diff
@@ -32,7 +31,7 @@ def __init__(self, loc, scale):
         self.scale = scale
 
     def sample(self, sample_shape):
-        return norm.osp_stats.norm(self.loc, self.scale).rvs(
+        return osp_stats.norm(self.loc, self.scale).rvs(
```
matthewfeickert, line 34: For reasons I don't understand, Codecov is now trying to say that this line is uncovered. :? https://codecov.io/gh/scikit-hep/pyhf/pull/1134/diff#D2-34

matthewfeickert:
I also opened up an Issue on JAX RE: the docs of `jax.scipy.stats.poisson`: https://github.com/google/jax/issues/4648

matthewfeickert:
@alexander-held heads up on this.

setup.py
```diff
@@ -7,7 +7,7 @@
         'tensorflow-probability~=0.10.0',
     ],
     'torch': ['torch~=1.2'],
-    'jax': ['jax~=0.1,>0.1.51', 'jaxlib~=0.1,>0.1.33'],
+    'jax': ['jax~=0.2.4', 'jaxlib~=0.1.56'],
```
matthewfeickert, line 10: I think we should pin down to the minor version here as it seems that JAX is still new enough that there are breaking API changes on minor releases (which is totally fair given that it is a `v0.X` library still).

lukasheinrich (approved):

src/pyhf/tensor/jax_backend.py
kratsg: it might not be covered since we don't have toys implemented to do the sampling yet.

kratsg (approved):

src/pyhf/tensor/jax_backend.py
```diff
@@ -17,9 +18,7 @@ def __init__(self, rate):
         self.rate = rate
 
     def sample(self, sample_shape):
-        return poisson.osp_stats.poisson(self.rate).rvs(
-            size=sample_shape + self.rate.shape
-        )
+        return osp_stats.poisson(self.rate).rvs(size=sample_shape + self.rate.shape)
```
lukasheinrich, line 21: hm do we ever check that this returns jax tensors?

src/pyhf/tensor/jax_backend.py
lukasheinrich: yeah same with above: should also be cast as jax tensors

src/pyhf/tensor/jax_backend.py
```diff
@@ -17,8 +18,9 @@ def __init__(self, rate):
         self.rate = rate
 
     def sample(self, sample_shape):
-        return poisson.osp_stats.poisson(self.rate).rvs(
-            size=sample_shape + self.rate.shape
+        tensorlib = jax_backend()
```
matthewfeickert, line 21: Getting the `tensorlib` this way is how we've done thing previously inhere, but not sure if there's a smarter way.

src/pyhf/tensor/jax_backend.py
matthewfeickert: This should be resolved now.

src/pyhf/tensor/jax_backend.py
kratsg: this isn't the best, since they might have some configurations on the jax backend that are different than the default you're relying on here.

kratsg (approved):

PR #1135, opened 2020-10-20
style: Adopt new JAX style of np and jnp

matthewfeickert:
# Description

As a follow up to PR #1134, I noticed in [JAX PR 3760](https://github.com/google/jax/pull/3760) that JAX has changed style from

```python
import numpy as onp
import jax.numpy as np
```

to now using

```python
import numpy as np
import jax.numpy as jnp
```

To try to make the code as clear as possible for comparison purposes we should also adopt this style change.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Adopt JAX v0.2 API style change of numpy and jax.numpy imports:
   - import numpy as np
   - import jax.numpy as jnp
* c.f. https://github.com/google/jax/pull/3760
```

### DIFF
--- a/src/pyhf/tensor/jax_backend.py
+++ b/src/pyhf/tensor/jax_backend.py
@@ -6,7 +6,7 @@ import jax.numpy as jnp
 from jax.scipy.special import gammaln
 from jax.scipy import special
 from jax.scipy.stats import norm
-import numpy as onp
+import numpy as np
 import scipy.stats as osp_stats
 import logging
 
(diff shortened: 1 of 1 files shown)


matthewfeickert:
The output of `git grep "np\." src/pyhf/tensor/jax_backend.py` makes me think I got everything too.

src/pyhf/tensor/jax_backend.py
```diff
@@ -179,14 +179,14 @@ def conditional(self, predicate, true_callable, false_callable):
 
     def tolist(self, tensor_in):
         try:
-            return onp.asarray(tensor_in).tolist()
+            return np.asarray(tensor_in).tolist()
```
matthewfeickert, line 182: `jax.numpy` now supports `tolist` so this could instead be

```suggestion
            return jnp.asarray(tensor_in).tolist()
```

but this should probably be left to another follow up PR as this is a style PR where no real code is supposed to change.

kratsg: #1137

kratsg (approved):

PR #1138, opened 2020-10-20
refactor: Use jax.numpy for JAX backend tensorlib.tolist

### LINKED ISSUE #1137: Remove use of NumPy from JAX backend
`jax.numpy` now supports `tolist` so this

https://github.com/scikit-hep/pyhf/blob/0ed1b518bf0ea137b7d676dac52575a8473950de/src/pyhf/tensor/jax_backend.py#L180-L182

 could instead be

```suggestion
            return jnp.asarray(tensor_in).tolist()
```

but this should probably be left to another follow up PR as this is a style PR where no real code is supposed to change.

_Originally posted by @matthewfeickert in https://github.com/scikit-hep/pyhf/pull/1135#discussion_r508719046_

matthewfeickert:
# Description

Resolves #1137 

In more recent releases of JAX the `jax.numpy` module is more filled out and it now supports `jax.numpy.tolist`, which was the only thing NumPy was used for previously.

NumPy must still be kept as JAX relies on `numpy` itself to convert from JAX DeviceArray to NumPy

https://github.com/scikit-hep/pyhf/blob/2b2b28181a038b25bd4634808cdfec869b30d834/src/pyhf/tensor/jax_backend.py#L568-L593

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use jax.numpy.tolist to provide the tolist method for the JAX backend
   - Note that NumPy dependency can never be removed as JAX depends on
     NumPy for JAX to NumPy conversion
```

### DIFF
--- a/src/pyhf/tensor/jax_backend.py
+++ b/src/pyhf/tensor/jax_backend.py
@@ -180,7 +180,7 @@ class jax_backend:
     def tolist(self, tensor_in):
         try:
             return jnp.asarray(tensor_in).tolist()
-        except AttributeError:
+        except TypeError:
             if isinstance(tensor_in, list):
                 return tensor_in
             raise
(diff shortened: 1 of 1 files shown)


matthewfeickert:
I was hoping to remove NumPy entirely, but turns out that's not possible with JAX as JAX needs NumPy to convert to NumPy.

kratsg (approved):

PR #1160, opened 2020-10-29
clip & incl

### LINKED ISSUE #1159: support variations on CLs dealing w/ inclusive pvalue and clipping

lukasheinrich:
# Description

resolves #1159

### DIFF
--- a/src/pyhf/contrib/viz/brazil.py
+++ b/src/pyhf/contrib/viz/brazil.py
@@ -2,7 +2,7 @@
 import numpy as np
 
 
-def plot_results(ax, mutests, tests, test_size=0.05):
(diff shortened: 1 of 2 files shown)


kratsg:
#993 seems to do the same thing.

matthewfeickert:
> #993 seems to do the same thing.

Yeah, agreed. @lukasheinrich if there isn't further motivation for this PR beyond PR #993 then maybe this PR could be revised to just contribute the notebook instead.

matthewfeickert:
In my mind as soon as this is done we should get `v0.6.4` out and then move to working on features for `v0.7.0`.

Issue #815, opened 2020-04-03
Add a percentile function to tensorlib


matthewfeickert:
# Description

First brought up in PR #790, there should be a percentile function added to the tensor backends that adheres to [NumPy's `percentile`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.percentile.html). For reference for the other backnds:

- [`tfp.stats.percentile`](https://www.tensorflow.org/probability/api_docs/python/tfp/stats/percentile)
- PyTorch doesn't have anything yet, but it is part of the NumPy functionality request rollup ([PyTorch Issue 38349](https://github.com/pytorch/pytorch/issues/38349)) and should be implemented in [PyTorch PR 39417](https://github.com/pytorch/pytorch/pull/39417).
- [`jax.numpy.percentile`](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.percentile.html)

Currently PyTorch doesn't support anything that matches these other `percentile` implimentations. The [`torch.distributions.distribution.icdf`](https://pytorch.org/docs/stable/distributions.html#torch.distributions.distribution.Distribution.icdf) gives the

> the inverse cumulative density/mass function evaluated at value

which is what is desired as [`np.percentile`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.percentile.html) 

> compute[s] the q-th percentile of the data along the specified axis

but the `icdf` is a distributions method, and not a tensor method.

matthewfeickert:
There is a [Gist of a function](https://gist.github.com/spezold/42a451682422beb42bc43ad0c0967a30) that can give something similar to `np.percentile` for PyTorch, however we want to make sure that we have the ability to have interpolation schemes like the other implementations.

matthewfeickert:
It is also worth noting here that [TensorFlow](https://www.tensorflow.org/probability/api_docs/python/tfp/stats/percentile) has different behavior than [NumPy](https://docs.scipy.org/doc/numpy/reference/generated/numpy.percentile.html) and JAX when it comes to the "nearest" interpolation scheme.

```python
>>> import pyhf
>>> import numpy as np
>>> import tensorflow_probability as tfp
>>> pyhf.set_backend("numpy")
>>> a = pyhf.tensorlib.astensor([[10, 7, 4], [3, 2, 1]])
>>> np.percentile(a, 50, interpolation="nearest")
3.0
>>> pyhf.set_backend("tensorflow")
[...]

PR #1162, opened 2020-10-30
fix: Report expected p-values based on their quantiles under background hypotheses

### LINKED ISSUE #1161: Toys giving mysterious results
Hi pyhf friends!

I'm playing with the newly supported toys in pyhf v0.5.4.dev3 - thank you so much for introducing them. I'm doing a mu scan to obtain an upper limit in a single bin region with 0 observed events. An example script is given at the bottom.

If I run my scan using asymptotics (see line 51 of the script),
`val = pyhf.infer.hypotest(mu_seed[0], data, model, qtilde=True, return_expected_set=True,return_tail_probs=True)`
I get this plot of the result:
result_mu

Which, other than being a better-than-zero result, is fine. If I run using toys (line 53 of the script),
`val = pyhf.infer.hypotest(mu_seed[0], data, model, calctype='toybased', ntoys=int(sys.argv[1]),return_expected_set=True,return_tail_probs=True)`
I get this output instead:

result_mu

Which ...... yeah. 😬 

I had a poke around to see what might be going on and the underlying s-plus-b and b-only distributions look super weird. For that run of toys (10k), here's the two distributions, with the CLb and CLs+b values that were calculated from them, for one tested mu value of 0.552:

bonly_0p552

splusb_0p552

There's a ton of repetition of the exact same values and what looks like some kind of default upper value in the b-only distribution, but what it is changes with mu.

Any idea what might be going on here? I've filed this as a bug but it could also be that I'm just using this wrong. 

Many thanks for your help, and again for adding the toys!

Cheers,
Kate

==================

Example script:

```
#!/usr/env/python

from multiprocessing import Pool

import sys
import pyhf

nSig = 4.166929245
errSig = 0
nBkg = 0.11
errBkgUp = 0.20
errBkgDown = 0.11

model_json = {
    "channels": [
        { "name": "SR_combined",
          "samples": [
            { "name": "signal",
              "data": [nSig],
              "modifiers": [ { "name": "mu", "type": "normfactor", "data": None},
              ]
            },
            { "name": "background",
              "data": [nBkg],
              "modifiers": [ {"name": "uncorr_bkguncrt", "type": "normsys", "data": {"hi": errBkgUp, "lo": errBkgDown}} ]
            }
          ]
        }
    ],
    "observations": [
        { "name": "SR_combined", "data": [0.0] }
    ],
    "measurements": [
        { "name": "Measurement", "config": {"poi": "mu", "parameters": []} }
    ],
    "version": "1.0.0"
}

ws = pyhf.Workspace(model_json)
model = ws.model()
data = ws.data(model)

def getCLS(mu_seed):
    
    # Asymptotics
    #val = pyhf.infer.hypotest(mu_seed[0], data, model, qtilde=True, return_expected_set=True,return_tail_probs=True) 
    # Toys
    val = pyhf.infer.hypotest(mu_seed[0], data, model, calctype='toybased', ntoys=int(sys.argv[1]),return_expected_set=True,return_tail_probs=True)
    return val

pool = Pool(processes=40)

final_norms = [1.7,1.8,1.9,2.0,2.1,2.2,2.3,2.4,2.5,2.6,2.7,2.8,2.9,3.0,3.2,3.4,3.6]
muInput = [i/nSig for i in final_norms]
nSeeds = 1

results = {}
mu_seedList = []
for seed in range(nSeeds):
    for mu in muInput:
        mu_seedList.append( [mu, seed] )

print("Seed; mu value; nSignal events; CLs")
results = pool.map(getCLS, mu_seedList)
result_plottable = []
CLsb_list = []
CLb_list = []
for result in results :
  plottable = ((result[0],result[-1]))
  result_plottable.append(plottable)
  CLsb_list.append(result[1][0])
  CLb_list.append(result[1][1])

fig, ax = plt.subplots()
fig.set_size_inches(5, 5)
ax.set_xlabel(r"$\mu$ (POI)")
ax.set_ylabel(r"$\mathrm{CL}_{s}$")
pyhf.contrib.viz.brazil.plot_results(ax, muInput, result_plottable)
plt.plot(muInput, CLsb_list, 'bo-')
plt.plot(muInput, CLb_list, 'ro-')
plt.savefig("result_mu.eps")
```

### LINKED ISSUE #1264: Update ToDo on readme to indicate that we've added support for toys
See title.

lukasheinrich:
# Description

* Resolves #1161 
* Resolves #1264

Fixes issue reported by @kpachal

This changes the way we report expected limits

we used to
* find quantiles in test stat distribution [q1,q2,....q5]
* report CLs/p-values for those test stats [p1,p2,...p5]

however this can lead to orderings in which the p-values are not monotonic as the relationship between CLs and test statistic value is not necessarily monotonic..

we now
* compute a sample of Cls values based on test statistics sampled from toys
* report exected CLs/p-values  as quantiles of that those test stats [p1,p2,...p5]

This is also what ROOT does. It might present some future questions regarding #966 but is ok for now.

```
import pyhf
import json
import numpy as np
import pyhf.contrib.viz.brazil as brazil

w = pyhf.Workspace(json.loads('''
{
    "channels": [
        {
            "name": "SR_combined",
            "samples": [
                {
                    "data": [4],
                    "modifiers": [
                        {"data": null,"name": "lumi","type": "lumi"},
                        {"data": null,"name": "mu","type": "normfactor"}
                    ],
                    "name": "signal"
                },
                {
                    "data": [0.1],
                    "modifiers": [
                        {"data": null,"name": "lumi","type": "lumi"},
                        {"data": {"hi": 2.0,"lo": 0.5},"name": "uncorr_bkguncrt","type": "normsys"}
                    ],
                    "name": "background"
                }
            ]
        }
    ],
    "measurements": [
        {
            "config": {
                "parameters": [
                    {"auxdata": [1.0],"bounds": [[0.5,1.5]],"inits": [1.0],"name": "lumi","sigmas": [0.01]},
                    {"bounds": [[0.0,10.0]],"inits": [1.0],"name": "mu"}
                ],
                "poi": "mu"
            },
            "name": "Measurement"
        }
    ],
    "observations": [{"data": [0],"name": "SR_combined"}
    ],
    "version": "1.0.0"
}
'''))
m = w.model()
d = w.data(m)

mutests = np.linspace(0.5,1.0,4)
results = [pyhf.infer.hypotest(mu,d,m,return_expected_set=True, calctype='toybased', ntoys = 15000) for mu in mutests]

f,ax = plt.subplots()
brazil.plot_results(ax,mutests,results)
plt.semilogy()
plt.ylim(1e-3,1)
plt.xlim(0,1.35)
```

produces this plot

image

ReadTheDocs build: https://pyhf.readthedocs.io/en/fix_for_kate/api.html#inference

For the PR Assignees:

```
* Add two ROOT validation scripts for toys
* Use same initialization parameters for b and s+b toys
* Leave (expected) p-value calculation to calculators
   - Add pvalue and expected_pvalues to calculator API
* Toy-based calculator returns expected values based on CLs distribution quantiles

```


src/pyhf/infer/calculators.py
```diff
@@ -543,7 +543,7 @@ def distributions(self, poi_test, track_progress=None):
                     poi_test,
                     sample,
                     self.pdf,
-                    signal_pars,
+                    self.init_pars,
```
matthewfeickert, line 546: @lukasheinrich I think these and the `bkg_pars` should instead be set above with something like

```python
signal_pars = init_pars or self.pdf.config.suggested_init()
```

matthewfeickert (changes requested):

src/pyhf/infer/calculators.py
lukasheinrich: that already happends here no

        self.init_pars = init_pars or pdf.config.suggested_init()
 in the init of the calc

lukasheinrich:
I added a validation script StandardHypoTestDemo.py which roughly follows StandardHypoTestDemo.C except that it configures it for exclusion instead of discovery.

With this we can xcheck toys against ROOT.

for the workspace 'xmlimport_input_bkg.json' (which si xmlimport_input_bkg excecpt with data put on bkg expectation) it looks like this


image


https://root.cern.ch/doc/v614/StandardHypoTestDemo_8C.html

lukasheinrich:
ok some more important points

* in this regime lumi uncertainty becomes important so one must
  * add a lumi modifier to each
  * add a lumi sigma into the parameter config 


image

once this is done, the delta peak gets broadened  (left ROOT right toys)


image


the brazil band doesn't look too great still, but it's looking better - trying w/ more toys

image


This is the workspace


```
{
    "channels": [
        {
            "name": "SR_combined",
            "samples": [
                {
                    "data": [
                        1
                    ],
                    "modifiers": [
                        {
                            "data": null,
                            "name": "lumi",
                            "type": "lumi"
                        },
                        {
                            "data": null,
                            "name": "mu",
                            "type": "normfactor"
                        }
                    ],
                    "name": "signal"
                },
                {
                    "data": [
                        0.1
                    ],
                    "modifiers": [
                        {
                            "data": null,
                            "name": "lumi",
                            "type": "lumi"
                        },
                        {
                            "data": {
                                "hi": 1.1,
                                "lo": 0.9
                            },
                            "name": "uncorr_bkguncrt",
                            "type": "normsys"
                        }
                    ],
                    "name": "background"
                }
            ]
        }
    ],
    "measurements": [
        {
            "config": {
                "parameters": [
                    {
                        "auxdata": [
                            1.0
                        ],
                        "bounds": [
                            [
                                0.5,
                                1.5
                            ]
                        ],
                        "inits": [
                            1.0
                        ],
                        "name": "lumi",
                        "sigmas": [
                            0.1
                        ]
                    },
                    {
                        "bounds": [
                            [
                                0.0,
                                10.0
                            ]
                        ],
                        "inits": [
                            1.0
                        ],
                        "name": "mu"
                    }
                ],
                "poi": "mu"
            },
            "name": "Measurement"
        }
    ],
    "observations": [
        {
            "data": [
                0
            ],
            "name": "SR_combined"
        }
    ],
    "version": "1.0.0"
}
```

src/pyhf/infer/calculators.py
```diff
@@ -566,6 +584,32 @@ def distributions(self, poi_test, track_progress=None):
...
+            [
+                self.pvalues(
+                    tb.astensor(x), sig_plus_bkg_distribution, b_only_distribution
+                )[-1]
+                for x in b_only_distribution.samples
+            ]
+        )
+        import numpy as np
+
+        CLs_exp = np.percentile(
```
lukasheinrich, line 607: also needs to be changed once we have #815

src/pyhf/infer/calculators.py
```diff
@@ -303,6 +303,20 @@ def _false_case():
             )
         return teststat
 
+    def pvalues(self, teststat, sig_plus_bkg_distribution, b_only_distribution):
+        '''calculate pvalues.'''
```
matthewfeickert, line 307: I'll review this thoroughly tomorrow morning, but just to hit on the docs related things, as these methods are in the public API we should have full docstrings with examples for them (for both `pvalues` and `expected_pvalues` both here and in the `EmpiricalDistribution` ).

validation/StandardHypoTestDemo.py
```diff
@@ -0,0 +1,74 @@
...
+    var.setVal(oldval)
+
+    var = sbModel.GetParametersOfInterest().first()
+    sbModel.SetSnapshot(ROOT.RooArgSet(var))
+
+    profll = ROOT.RooStats.ProfileLikelihoodTestStat(bModel.GetPdf())
+    profll.SetOneSidedDiscovery(False)
+    profll.SetOneSided(True)
+    hypoCalc = ROOT.RooStats.FrequentistCalculator(data, bModel, sbModel)
+    # profll.SetPrintLevel(2)
```
matthewfeickert, line 33: This is fine, but at this level why not make it a `Click` CLI script?

src/pyhf/infer/calculators.py
matthewfeickert: I should add that this PR is super nice to have! :)

validation/StandardHypoTestDemo.py
lukasheinrich: b/c my ROOT image doesn't have click installed. my memory is a bit vague: did we bulid a pyhf-root validation image?

lukasheinrich:
this leaving this for later

this is now the plot w/ ROOT

image

and this w/ pyhf

image

which should be equivalent within the toy sample variance (20k)

validation/StandardHypoTestDemo.py
matthewfeickert: Yup. https://hub.docker.com/r/pyhf/pyhf-validation/tags But it needs to be updated.

src/pyhf/infer/calculators.py
```diff
@@ -566,6 +657,32 @@ def distributions(self, poi_test, track_progress=None):
...
+                for x in b_only_distribution.samples
+            ]
+        )
+        import numpy as np
+
+        CLs_exp = np.percentile(
+            tb.tolist(pvalues),
+            [2.27501319, 15.86552539, 50.0, 84.13447461, 97.72498681],
+        )
+        return tb.astensor(CLs_exp)
```
matthewfeickert, line 684: Does this need to return a tensor here?

src/pyhf/infer/calculators.py
```diff
@@ -566,6 +657,108 @@ def distributions(self, poi_test, track_progress=None):
...
+            ]
+        )
+        # TODO: Add percentile to tensorlib
+        # c.f. Issue #815, PR #817
+        import numpy as np
+
+        CLs_exp_band = np.percentile(
+            tb.tolist(pvalues),
+            [2.27501319, 15.86552539, 50.0, 84.13447461, 97.72498681],
+        ).tolist()
```
matthewfeickert, line 759: This was originally a tensor, but I don't think it needs to be so for consistency with the asymptotic distribution methods I'm having it return a `list` for now.

docs/api.rst
```diff
@@ -119,6 +119,7 @@ Inference
 
 .. autosummary::
    :toctree: _generated/
+   :template: modifierclass.rst
```
matthewfeickert, line 122: @kratsg Feelings on this? I can remove this to a separate PR, but I added it accidentally so thought I'd get your feelings in the RTD preview.

validation/StandardHypoTestDemo.py
```diff
@@ -0,0 +1,121 @@
...
+        label=r"$f(\tilde{q}|0)$ Background",
+    )
+    ax.hist(
+        sig_plus_bkg_dist.samples / 2.0,
+        alpha=0.2,
+        bins=bins,
+        density=True,
+        label=r"$f(\tilde{q}|1)$ Signal",
+    )
+    ax.axvline(q_tilde / 2, color="black", label="Observed test statistic")
```
matthewfeickert, line 109: I assume that if I look at this again later it will make sense, but why is there the need to divide the samples and test stat by 2 to get a plot that agrees with the ROOT plot?

lukasheinrich: because we use the definition of teststat as `2*log lhood ratio`, while root uses just `log lhood ratio`

docs/api.rst
matthewfeickert: It seems we agree this is okay for now and we can fix things up later.

matthewfeickert:
> this leaving this for later this is now the plot w/ ROOT

@lukasheinrich Not super important to getting this in, but which model are you using for these mu scan plots?

If I do

```
$ mkdir -p xmlimport_input_bkg
$ pyhf json2xml --output-dir xmlimport_input_bkg validation/xmlimport_input_bkg.json
$ hist2workspace xmlimport_input_bkg/FitConfig.xml
$ python validation/run_toys.py xmlimport_input_bkg/config/FitConfig_combined_GaussExample_model.root 20000
```
I get the following `poi_scan_ROOT.png`

poi_scan

matthewfeickert:
> Not super important to getting this in, but which model are you using for these mu scan plots?

For context these 20k plots were made using [the workspace in the PR body](https://github.com/scikit-hep/pyhf/pull/1162#issue-513111182). It might be important to choose one representative workspace to have added in this PR.

matthewfeickert:
Now that PR #1232 is in, this can be rebased, have conflicts resolved, and then with a few outstanding fixes should be ready to go in. :+1:

kratsg (approved):

src/pyhf/infer/calculators.py
```diff
@@ -574,6 +675,110 @@ def distributions(self, poi_test, track_progress=None):
...
+                    b_only_distribution,
+                )
+                for test_stat in b_only_distribution.samples
+            ]
+        )
+        # TODO: Add percentile to tensorlib
+        # c.f. Issue #815, PR #817
+        import numpy as np
+
+        _percentiles = [2.27501319, 15.86552539, 50.0, 84.13447461, 97.72498681]
```
matthewfeickert, line 774: I think we're going to need a comment here or some better description in the docstring as the nature of these percentiles as it is late and I don't remember what exactly these are &mdash; which mens there will be a time in the future when it will also be late and I will want to know what these are.

matthewfeickert (changes requested):
Thanks for the nice rebasing work @kratsg.

So a concern that I have here, that I myself contributed to when working on this weeks ago, is that we're introducing inconsistent return types for CLs-like values now. Previously, all CLs-like / p-value-like values were `0`-d tensors

https://github.com/scikit-hep/pyhf/blob/0e71f2f40153cf4f3ae3c62fedf23e4281b240d9/src/pyhf/infer/__init__.py#L149

However, now we have things like `hypotest` returning `0`-d tensors and things like `pvalues` and `expected_pvalues` returning floats only for NumPy and for all other backends returning `0`-d tensors, yet in all the examples we're showing CLs-like / p-value-like values being returned.

Example: Run the docstring example from this PR for `expected_pvalues` for various backends and note the return type of `CLs_exp_band[0]`.

```python
import pyhf

for backend in ["numpy", "pytorch", "jax"]:
    print(f"\nbackend: {backend}")
    pyhf.set_backend(backend)
    )
    observations = [51, 48]
    data = observations + model.config.auxdata
    mu_test = 1.0
    asymptotic_calculator = pyhf.infer.calculators.AsymptoticCalculator(
        data, model, test_stat="qtilde"
    )
    _ = asymptotic_calculator.teststatistic(mu_test)
    sig_plus_bkg_dist, bkg_dist = asymptotic_calculator.distributions(mu_test)
    CLsb_exp_band, CLb_exp_band, CLs_exp_band = asymptotic_calculator.expected_pvalues(
        sig_plus_bkg_dist, bkg_dist
    )
    print(f"CLs expected band: {CLs_exp_band}")
    print(
        f"of type: {type(CLs_exp_band[0])} and shape {pyhf.tensorlib.shape(CLs_exp_band[0])}"
    )
```
gives
```

backend: numpy
CLs expected band: [0.0026062609501074576, 0.01382005356161206, 0.06445320535890459, 0.23525643861460702, 0.573036205919389]
of type: <class 'numpy.float64'> and shape ()

backend: pytorch
CLs expected band: [tensor(0.0026), tensor(0.0138), tensor(0.0645), tensor(0.2353), tensor(0.5730)]
of type: <class 'torch.Tensor'> and shape ()

backend: jax
CLs expected band: [DeviceArray(0.00260626, dtype=float64), DeviceArray(0.01382005, dtype=float64), DeviceArray(0.0644532, dtype=float64), DeviceArray(0.23525643, dtype=float64), DeviceArray(0.57303619, dtype=float64)]
of type: <class 'jax.interpreters.xla._DeviceArray'> and shape ()
```

We should try to come to a consensus on what the return type for a CLs-like / p-value-like value should be. From looking back at PR #944 and Issue #714, I think that the CLs-like values should be `0`-d tensor as it allows for us to sidestep this difference in backend behavior (I believe this is the motivation for the current behavior). While conceptually it makes sense to have a p-value-like just be a float to emphasize the scalar nature, having it be a `0`-d tensor make it appear to a user as still having scalar like behavior.

The validation notebooks here are fine for now and we can revise them when we address Issue #1241.

@kratsg all the notebook stuff is great though. :+1:

src/pyhf/infer/calculators.py
matthewfeickert: Oh right, as this is for the background only hypothesis then these are Normal percentiles.

matthewfeickert (approved):
The CLs-like / p-value-like return type problem is now moved to Issue #1268 and will be resolved in a follow up PR.

docs/api.rst
kratsg: we're ok with this for now. can change it later.

src/pyhf/infer/__init__.py
```diff
@@ -141,57 +141,35 @@ def hypotest(
...
-        tensorlib.astensor(CLs),
+    tb, _ = get_backend()
+    CLsb_obs, CLb_obs, CLs_obs = tuple(
+        tb.astensor(pvalue)
+        for pvalue in calc.pvalues(
+            teststat, sig_plus_bkg_distribution, b_only_distribution
+        )
+    )
+    CLsb_exp, CLb_exp, CLs_exp = calc.expected_pvalues(
+        sig_plus_bkg_distribution, b_only_distribution
```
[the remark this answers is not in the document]
kratsg: #1268

PR #1167, opened 2020-11-08
fix: Pin additional dependencies in pre-commit hooks

MarcoGorelli:
Hi,

Thanks for trying out nbqa - our default hooks use the latest available version of each tool, but for best reproducibility it's best to pin them. Admittedly, that's something we'd neglected to document :flushed:  but [it's there now](https://github.com/nbQA-dev/nbQA#pre-commit)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Set dependency versions of black and pyupgrade for nbQA to use
* Remove language_version from Black pre-commit hook as already specified
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,6 @@ repos:
     rev: 20.8b1
     hooks:
     - id: black
-      language_version: python3
 -   repo: https://github.com/asottile/pyupgrade
     rev: v2.7.3
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,4 +13,6 @@ repos:
     rev: 0.4.0
     hooks:
     - id: nbqa-black
+      additional_dependencies: [black==20.8b1]
     - id: nbqa-pyupgrade
+      additional_dependencies: [pyupgrade==2.7.3]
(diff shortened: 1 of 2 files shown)


matthewfeickert:
> our default hooks use the latest available version of each tool, but for best reproducibility it's best to pin them. 

Thanks very much for this PR @MarcoGorelli. Just to make sure I understand the behavior, without these dependency versions being specified, the versions of the tools already specified in `.pre-commit-config.yaml`

https://github.com/scikit-hep/pyhf/blob/e4011ffa9a27c527ade7a927800d76a80ad5af9b/.pre-commit-config.yaml#L2-L11

are ignored and the latest versions are installed? Do I have this right?

Also, would you mind throwing your name on the [`contributors.rst` list](https://github.com/scikit-hep/pyhf/blob/master/docs/contributors.rst)? :)

MarcoGorelli:
> Do I have this right?

Yup, exactly!

> Also, would you mind throwing your name on the contributors.rst list? :)

Sure, done, thanks!

Also, I removed `language_version` from your `black` hooks, as [it's already specified upstream](https://github.com/psf/black/blob/master/.pre-commit-hooks.yaml)

matthewfeickert (approved):
LGTM. Thanks very much @MarcoGorelli

PR #1179, opened 2020-11-18
fix: Ensure float-like values used in division in writexml

### LINKED ISSUE #1177: json2xml breaks for a simple JSON
# Description

```
{
  "channels": [
    {
      "name": "singlechannel",
      "samples": [
        {
          "name": "signal",
          "data": [
            5
          ],
          "modifiers": [
            {
              "name": "mu",
              "type": "normfactor",
              "data": null
            }
          ]
        },
        {
          "name": "background",
          "data": [
            50
          ],
          "modifiers": [
            {
              "name": "uncorr_bkguncrt",
              "type": "shapesys",
              "data": [
                6
              ]
            }
          ]
        }
      ]
    }
  ],
  "version": "1.0.0",
  "measurements": [
    {
      "name": "measurement",
      "config": {
        "poi": "mu",
        "parameters": [
          {
            "bounds": [
              [
                0,
                10
              ]
            ],
            "inits": [
              1
            ],
            "fixed": false,
            "name": "mu"
          },
          {
            "bounds": [
              [
                1e-10,
                10
              ]
            ],
            "inits": [
              1
            ],
            "fixed": false,
            "name": "uncorr_bkguncrt"
          }
        ]
      }
    }
  ],
  "observations": [
    {
      "name": "singlechannel",
      "data": [
        50
      ]
    }
  ]
}
```

matthewfeickert:
# Description

Resolves #1177

Ensure the necessary float-like values for division carried out in `writexml` by using NumPy's `float` dtype.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Ensure float-like values are used in division by using float dtype
* Add test for float-like values using integer spec data
```


tests/test_export.py
```diff
@@ -172,6 +172,57 @@ def spec_shapefactor():
     return spec
 
 
+def spec_integer_data():
```
matthewfeickert, line 175: @kratsg There is probably a more concise / intelligent way to write this spec + test, but I just grabbed the example failing spec that @lukasheinrich provided and went with it tonight. So revision requests welcome!

alexander-held: An alternative way to do this is to put the function in `conftest.py` and decorate it with `@pytest.fixture`, and then you can make it available in `test_integer_data` by adding it to the signature:
```
def test_integer_data(mocker, spec_integer_data):
    channel_spec = spec_integer_data["channels"][0]
```

kratsg: Easiest way is to use the fixture I created for local test data. Move the spec into `tests/test_export/workspace_integer_data.json` and then change the test (dropping the `spec_integer_data` completely)

```
def test_integer_data(datadir, mocker):
    """
    Test that a spec with only integer data will be written correctly
    """
    spec = json.load(open(datadir.join('workspace_integer_data.json')))
    channel_spec = spec["channels"][0]
    mocker.patch('pyhf.writexml._ROOT_DATA_FILE')

    channel = pyhf.writexml.build_channel(spec, channel_spec, {})
    assert channel
```

matthewfeickert: @kratsg This is done and ready for review then.

kratsg (approved):

PR #1183, opened 2020-11-18
feat: Set minuit strategy automatically for gradient/non-gradient mode

### LINKED ISSUE #1172: use strategy 0 for user-provided gradients in minuit
# Description

since we have exact gradient we can disable the checks minuit does

cc @alexander-held

kratsg:
# Pull Request Description

Teach pyhf to set the minuit strategy correctly depending on whether user provides gradient or not. Additionally allow for this to be configurable via `strategy` kwarg.

Resolves #1172.

ReadTheDocs build: https://pyhf.readthedocs.io/en/feat-userprovidedgradientstominuit/_generated/pyhf.optimize.opt_minuit.minuit_optimizer.html

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Set MINUIT strategy to 0/1 automatically if using gradients/no-gradient
* Allow for MINUIT strategy to be configurable
* Add test for strategy selection
```

### DIFF
--- a/src/pyhf/optimize/opt_minuit.py
+++ b/src/pyhf/optimize/opt_minuit.py
@@ -87,7 +87,7 @@ class minuit_optimizer(OptimizerMixin):
         Minimizer Options:
             maxiter (:obj:`int`): maximum number of iterations. Default is 100000.
             return_uncertainties (:obj:`bool`): Return uncertainties on the fitted parameters. Default is off.
-            strategy: (:obj:`int`): See :attr:`iminuit.Minuit.strategy`. Default is to configure based on `do_grad`.
+            strategy: (:obj:`int`): See :attr:`iminuit.Minuit.strategy`. Default is to configure in response to `do_grad`.
 
         Returns:
             fitresult (scipy.optimize.OptimizeResult): the fit result
--- a/src/pyhf/optimize/opt_minuit.py
+++ b/src/pyhf/optimize/opt_minuit.py
@@ -10,7 +10,7 @@ class minuit_optimizer(OptimizerMixin):
     Optimizer that uses iminuit.Minuit.migrad.
     """
 
-    __slots__ = ['name', 'errordef', 'steps']
+    __slots__ = ['name', 'errordef', 'steps', 'strategy']
 
     def __init__(self, *args, **kwargs):
         """
--- a/src/pyhf/optimize/opt_minuit.py
+++ b/src/pyhf/optimize/opt_minuit.py
@@ -96,7 +96,7 @@ class minuit_optimizer(OptimizerMixin):
         return_uncertainties = options.pop('return_uncertainties', False)
         # 0: Fast, user-provided gradient
         # 1: Default, no user-provided gradient
-        strategy = options.pop('strategy', not(do_grad))
+        strategy = options.pop('strategy', not (do_grad))
         if options:
             raise exceptions.Unsupported(
                 f"Unsupported options were passed in: {list(options.keys())}."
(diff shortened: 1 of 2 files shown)


src/pyhf/optimize/opt_minuit.py
```diff
@@ -87,17 +87,22 @@ def _minimize(
         Minimizer Options:
             maxiter (:obj:`int`): maximum number of iterations. Default is 100000.
             return_uncertainties (:obj:`bool`): Return uncertainties on the fitted parameters. Default is off.
+            strategy: (:obj:`int`): See :attr:`iminuit.Minuit.strategy`. Default is to configure based on `do_grad`.
```
matthewfeickert, line 90:
```suggestion
            strategy: (:obj:`int`): See :attr:`iminuit.Minuit.strategy`. Default is to configure in response to `do_grad`.
```

Maybe slightly revise here? For whatever reason "based on" makes my mind think that this means "True -> True, False -> False" even though that isn't what that means. But just that this wasn't 100% clear to me right away makes me think it might be worth tweaking.

tests/test_optim.py
```diff
@@ -93,7 +93,7 @@ def test_minimize(tensorlib, precision, optimizer, do_grad, do_stitch):
             'no_grad-minuit-jax-64b': [0.5000493563528641, 1.0000043833614634],
             # do grad, minuit, 32b
             'do_grad-minuit-pytorch-32b': [0.5017611384391785, 0.9997190237045288],
-            'do_grad-minuit-tensorflow-32b': [0.501288652420044, 1.0000219345092773],
+            'do_grad-minuit-tensorflow-32b': [0.5012885928153992, 1.0000673532485962],
```
matthewfeickert, line 96: Do we know why this is happening? Seems strange that this is affecting anything. :?

kratsg: it's so tiny. I don't know. But this is just minuit being minuit 🤷‍♂️. It most likely stops optimizing at some point... and for 32b, minuit is pretty junk about.

src/pyhf/optimize/opt_minuit.py
```diff
@@ -87,17 +87,22 @@ def _minimize(
...
 
         Returns:
             fitresult (scipy.optimize.OptimizeResult): the fit result
         """
         maxiter = options.pop('maxiter', self.maxiter)
         return_uncertainties = options.pop('return_uncertainties', False)
+        # 0: Fast, user-provided gradient
+        # 1: Default, no user-provided gradient
+        strategy = options.pop('strategy', not (do_grad))
```
matthewfeickert, line 99: Why is the `()` around `do_grad` needed? That is why not just do

```python
strategy = options.pop('strategy', not do_grad)
```
?

tests/test_optim.py
matthewfeickert: unsatisfying, but fair.

matthewfeickert (approved):
LGTM. Thanks @kratsg.

matthewfeickert (changes requested):
Woops. 

```pytb
ImportError while loading conftest '/home/runner/work/pyhf/pyhf/tests/conftest.py'.
tests/conftest.py:87: in <module>
src/pyhf/optimize/opt_minuit.py:35: in __init__
    self.strategy = kwargs.pop('strategy', None)
E   AttributeError: 'minuit_optimizer' object has no attribute 'strategy'
```
I'll check again in a bit.

matthewfeickert (approved):
The CI is still running, but I'll approve this so that it can be merged in once everything in CI hits green.

Issue #850, opened 2020-05-02
Application of POIs using formulas


rafaellopesdesa:
Hi,

It seems that, right now, the modifiers are limited to addition of multiplication, with the POIs as normfactors being defined by simple modification POI*pdf. That's ok for most models, but some models require more complex likelihoods. The case that comes to my mind (need?) is EFT analyses which have an interference term and a squared term and, therefore, would need something like sqrt(POI)*inter + POI*square.

I don't know exactly what is the best way to implement that. I think that a simple way would be if one could define multiple POIs and if one could define the value of a modifier as a function of other modifiers. Even simple functions like polynomials would already encompass most of the case while keeping all operations very fast and efficient.

Let me know if there is anything I can do to help.

Thanks for considering.

lukasheinrich:
hi @rafaellopesdesa 

apologies for the slow response.

the way to implement this is to add a new `modifier`. We could call it a `formulafactor`, with the rule being that it can only reuse existing parameters and does not introduce new ones

```
{
"type": "formulasys", 
"data": {
  "formula" : "sqrt(localvar)", dependencies: {"localvar": "mu"}
}
}
```

I think in the constrained language of numexpr https://github.com/pydata/numexpr and the with this constraint that no *new* parameters are introduced this would be quite doable.

cc @alexanderheld who has also raised this issue

rafaellopesdesa:
Hi @lukasheinrich ,

Yes, that would be great for more complex likelihood models.

I will give it a try this weekend and, if I manage to implement your idea, will come back here to report.

Thanks again!

lukasheinrich:
I would suggest you orient yourself along the lines of the normfactor code 

https://scikit-hep.org/pyhf/_modules/pyhf/modifiers/normfactor.html#normfactor

the output shape of a modifier is 

```
(n_modifiers, n_global_samples, n_alphas, n_global_bin)
```

here's the meaning of the output shape

* different "formulas" to be applied
* the affect of that formula for each sample  (if the sample doesn't declare/is unaffected by the formulafactor, it should just be multiplied by 1)
* the batch size of parameters.. this is probably just 1
* the global bin  number (across all channels)

so if you  need to compute the affect of 3 formulas and in total have 10 sample (which might  or might not be affected by the formula(s) and want to evaluate the formula for 5 parameter settings and the total number of bins is 20 you would have

an output tensor of 
```
(3,10,5,20)
```

lukasheinrich:
[...]

PR #1188, opened 2020-11-19
first example

lukasheinrich:
# Description

This is a first example on how to e.g. have a "gaussian" bump on top of a simple model 

cc @alexander-held 

```
#%%
import pyhf
import numpy as np
# %%
class custom_modifier(object):
    def apply(self,pars):
        #shape is  (n_modifiers, n-global_samples, n_alphas, n_global_bin)
        base = np.zeros((1,len(m.config.samples),1,sum(m.config.channel_nbins.values())))

        import scipy.stats
        bins = np.linspace(-5,5,11)
        yields = 20*(scipy.stats.norm.cdf(bins[1:]) - scipy.stats.norm.cdf(bins[:-1]))

        base[0,self.config.samples.index('signal'),0,:] = yields
        return base



m = pyhf.Model(
{'channels': [{'name': 'singlechannel',
   'samples': [{'name': 'signal',
     'data': [0]*10,
     'modifiers': [{'name': 'mu', 'type': 'normfactor', 'data': None}]},
    {'name': 'background',
     'data': [10]*10,
     'modifiers': [{'name': 'uncorr_bkguncrt',
       'type': 'shapesys',
       'data': [5]*10}]}]}]},
       custom_modifiers=[custom_modifier()]
)

bp = pyhf.tensorlib.astensor(m.config.suggested_init())
bp[m.config.poi_index] = 0.2
plt.bar(range(10),m.expected_actualdata(bp))
# %%

# %%
```

image


alexander-held:
A big question for solving #850 is what kind of language to use to specify custom expressions. I think for many use cases the expressions are simple enough that the language does not matter much. And it might be possible to add the language to the workspace in such a way that in the future new languages could be added?

lukasheinrich:
yeah I think we need to solve 2 issues:

* just technically allow the specification of new modifiers at the python level
* how to codify these into JSON this is in the direction of HiFav2

a big question is also in what order these modifications happen, i.e. here we have the custom modifier an additive modifier where the total rate in pyhf is computed as

`lambda = (nominal + additive.....) * multiplicative`

but you could also imagine expressions that expect to operate on yields onec all other modifiers have been applied

alexander-held:
I think this is technically still in the scope of normal HistFactory https://root.cern/doc/v620/classRooStats_1_1HistFactory_1_1Measurement.html#ade80531b4588d22ad5e7339ce74e7027. The version implemented there just acts as a normfactor, so this question of order is avoided. Once you allow more custom variations the problem becomes more complicated indeed.

lukasheinrich:
this now supports declaring additional parameters in the modifiers.. here e.g. a mean of the gaussian

```
class custom_modifier(object):
    def __init__(self,name = 'name'):
        self.required_parsets = {
            'mean': {
                'paramset_type': pyhf.parameters.paramsets.unconstrained,
                'n_parameters': 1,
                'modifier': 'normfactor',
                'is_constrained': False,
                'is_shared': True,
                'inits': (0.0,),
                'bounds': ((-5, 5),),
                'fixed': False
        }}
    def apply(self,pars):
        base = np.zeros((1,len(m.config.samples),1,sum(m.config.channel_nbins.values())))

        bins = np.linspace(-5,5,11)
        mean = pars[self.config.par_slice('mean')][0]
        yields = 20*(scipy.stats.norm(loc = mean).cdf(bins[1:]) - scipy.stats.norm(loc = mean).cdf(bins[:-1]))
        base[0,self.config.samples.index('signal'),0,:] = yields
        return base
```

tests/test_optim.py
```diff
@@ -407,7 +407,7 @@ def test_optim_uncerts(backend, source, spec, mu):
         return_uncertainties=True,
     )
     assert result.shape[1] == 2
-    assert pytest.approx([0.0, 0.26418431]) == pyhf.tensorlib.tolist(result[:, 1])
+    assert pytest.approx([0.26418431, 0.0]) == pyhf.tensorlib.tolist(result[:, 1])
```
lukasheinrich, line 410: this just changes the order

tests/test_modifiers.py
```diff
@@ -14,138 +11,30 @@
...
-
-    with pytest.raises(KeyError):
-
-        class myCustomModifier:
-            @classmethod
-            def required_parset(cls, sample_data, modifier_data):
-                pass
-
-        pyhf.modifiers.add_to_registry(myCustomModifier, 'histosys')
+def test_shapefactor_build():
```
kratsg, line 14: coverage is lost here as we're only testing that we can build shapefactor and none of the others? This would be great if we could parametrize it for tests.

tests/test_optim.py
kratsg: i guess can we add one more assert here on the order of the parameters in the model, just so we can keep track moving forward as well? This will future-proof in case it somehow changes again.

tests/test_pdf.py
```diff
@@ -557,7 +557,7 @@ def test_invalid_modifier():
         ]
     }
     with pytest.raises(pyhf.exceptions.InvalidModifier):
-        pyhf.pdf._ModelConfig(spec)
+        pyhf.pdf.Model(spec, validate=False)  # don't validate to delay exception
```
kratsg, line 560:
```suggestion
        pyhf.pdf.Model(spec, validate=False)  # skip validation exception
```

tests/test_pdf.py
```diff
@@ -817,8 +817,7 @@ def test_model_integration_fixed_parameters():
         'parameters': [{'name': 'mypoi', 'inits': [1], 'fixed': True}],
     }
     model = pyhf.Model(spec, poi_name='mypoi')
-    assert model.config.suggested_fixed() == [False, True]
-    assert model.config.poi_index == 1
+    assert model.config.suggested_fixed()[model.config.par_slice('mypoi')] == [True]
```
kratsg, line 820: the order changed again didn't it? why not just swap it around instead of combining the two? We won't catch a bug if both parameters are somehow "fixed" moving forward (so we lose coverage)

tests/test_pdf.py
```diff
@@ -849,5 +848,8 @@ def test_model_integration_fixed_parameters_shapesys():
     }
     model = pyhf.Model(spec, poi_name='mypoi')
     assert len(model.config.suggested_fixed()) == 5
-    assert model.config.suggested_fixed() == [False, True, True, True, False]
-    assert model.config.poi_index == 4
+    assert model.config.suggested_fixed()[model.config.par_slice('uncorr')] == [
```
kratsg, line 851: same comment here about catching the bug.

tests/test_regression.py
```diff
@@ -56,7 +56,7 @@ def test_sbottom_regionA_1300_205_60(
                     0.8910420971601081,
                 ]
             ),
-            rtol=1e-5,
+            rtol=2e-5,
```
kratsg, line 59: why did this get relaxed (if it's a refactor)?

kratsg (changes requested):
Overall, I think I do like a bulk of the refactoring. One thing that struck me is that we still kept around the op_code stuff -- which I suppose is fine. I guess a future iteration would somehow add a `modify_rate()` method that would know how to modify the rate that it gets passed in... and then at some point, we'd have to deal with priority/preference in some way.. but this is a rabbit hole anyway.

One thing I appreciate a lot, that was not so clear from the body of the PR, is that you've elided a lot of the logic from the megasamples back down to each of the individual modifier modules -- so that `pdf.py` isn't as scary as it used to be.

balunas:
It looks like it's been quite some time since any progress was made here. It's been bumped back for a few releases, is there some technical limitation, or has it just not been prioritized? There are some analysis use cases this could be really helpful for, and we've been having to use some really inefficient work-arounds instead. Is there a way we could help?

IzaV10:
Hi,

I tried running the simple example with the "gaussian" bump on top of a simple mode. The branch was installed as `pip install git+https://github.com/scikit-hep/pyhf.git@custom_modifiers` but when running the code abovel (from the very first entry of the conversation) I get an error.

```
File "/<PATH TO VIRTUAL ENVIRONMENT>/pyhf-env/lib/python3.7/site-packages/pyhf/pdf.py", line 497, in __init__
    self.config = _ModelConfig(self.spec, **config_kwargs)
  File "/<PATH TO VIRTUAL ENVIRONMENT>/pyhf-env/lib/python3.7/site-packages/pyhf/pdf.py", line 171, in __init__
    f"Unsupported options were passed in: {list(config_kwargs.keys())}."
pyhf.exceptions.Unsupported: Unsupported options were passed in: ['custom_modifiers'].
```
I suspect that the way the `class custom_modifier()`  is defined needs to be changed. Is there any documentation available on how to use this option?

Thank you for your help and time.

Cheers,
Iza

PR #1189, opened 2020-11-19
refactor: Drop 'modifier' from parameter set requirements

kratsg:
# Pull Request Description

@lukasheinrich identified this. It seems that #323 and #285 are the culprit here. Primarily, as it was a large PR, the refactoring wasn't so perfect and we left some cruft in at the end. This just cleans it up.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Drop the modifier from required_parset()
   - Parameter set requirements do not need to know modifier
```


matthewfeickert (approved):
Thanks @kratsg and thanks for the catch @lukasheinrich

PR #1191, opened 2020-11-24
chore: Use flake8 for linting

### LINKED ISSUE #863: Investigate switching from pyflakes to flake8
# Description

After a [strange incident](https://github.com/scikit-hep/pyhf/pull/862#issuecomment-625642362) with [`pyflakes`](https://github.com/PyCQA/pyflakes) and the branch name(?) it seems quite hard to figure out how to ignore certain paths if needed with `pyflakes`. It might be worth checking out [`flake8`](https://pypi.org/project/flake8/) instead.

@alexander-held is using `flake8` for [`cabinetry`](https://github.com/alexander-held/cabinetry) so he might have thoughts here.

matthewfeickert:
# Description

Resolves #863

Switch from `pyflakes` to the more configurable and modern `flake8` for linting.

In `setup.cfg` ignore
- [E203: whitespace before ':'](https://www.flake8rules.com/rules/E203.html) (avoid things like `text[len(prefix) :]`)
- [E402: module level import not at top of file](https://www.flake8rules.com/rules/E402.html) (sometimes needed)
- [E501: line too long](https://www.flake8rules.com/rules/E501.html) (not worth changing)

and fix:
- `src/pyhf/readxml.py:43:8:` [E713 test for membership should be 'not in'](https://www.flake8rules.com/rules/E713.html)
- `src/pyhf/infer/utils.py`: [E231 missing whitespace after ','](https://www.flake8rules.com/rules/E23.html)

and inline ignore `# noqa` (special cases):
- [F401: Module imported but unused](https://www.flake8rules.com/rules/F401.html)
- [E731: Do not assign a lambda expression, use a def](https://www.flake8rules.com/rules/E731.html)
- [E722: Do not use bare except, specify exception instead](https://www.flake8rules.com/rules/E722.html) (this could be removed, but would need to go back and identify all the exception cases)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Switch from pyflakes to flake8 for linting in pre-commit hooks and CI
   - Use flake8 in 'lint' extra
* Add flake8 config options to setup.cfg (as opposed to .flake8)
* Remove code that existed only to placate pyflakes
```

### DIFF
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -41,7 +41,7 @@ def test_missing_backends(isolate_modules, param):
 
     try:
         delattr(pyhf.tensor, module_name)
-    except:
+    except:  # noqa: E722
         pass
 
     with expectation:
@@ -80,7 +80,7 @@ def test_missing_optimizer(isolate_modules, param):
     )
     try:
         delattr(pyhf.optimize, module_name)
-    except:
+    except:  # noqa: E722
         pass
 
     with expectation:
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,5 @@
+[flake8]
+ignore = E203, E266, E501, W503
+max-line-length = 88
+max-complexity = 18
+select = B,C,E,F,W,T4,B9
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,4 @@
 [flake8]
 ignore = E203, E266, E402, E501, W503
 max-line-length = 88
-max-complexity = 18
 select = B,C,E,F,W,T4,B9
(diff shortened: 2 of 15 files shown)


tests/test_init.py
```diff
@@ -41,7 +41,7 @@ def test_missing_backends(isolate_modules, param):
 
     try:
         delattr(pyhf.tensor, module_name)
-    except:
+    except:  # noqa: E722
```
matthewfeickert, line 44: We might want to deal with these instead of the `noqa`

matthewfeickert:
This needs PR #1192 to go in first.

kratsg (approved):

tests/test_init.py
kratsg: these can be dealt with in another PR. It's just a matter of switching to `except Exception as err` and adding a `log.error(err)` to be safe.

PR #1192, opened 2020-11-24
fix: Use uproot3 namespace

matthewfeickert:
# Description

Avoid errors seen in CI with by switching over to using the new [`uproot3`](https://pypi.org/project/uproot3/) namespace. It is seen that some of the tests like

https://github.com/scikit-hep/pyhf/blob/a33baddf6620df371338c6d1b98fdbab1b9da5c4/tests/test_scripts.py#L169

 find an 

```
ModuleNotFoundError: No module named 'uproot3'
```

without the namespace switch. cc-ing @jpivarski just in case this is of interest.

I think this is because in `uproot_methods` instances of 'uproot' were renamed to 'uproot3': https://github.com/scikit-hep/uproot-methods/commit/02f9911b4437ae895a742a173eb88094e27eb65c

We could require a max on `uproot_methods` of `v0.9` so that we never get `v0.9.1` in which this takes place, but it seems easier to just update to `uproot3` and we'll probably be able to switch over to `uproot4` as well soon.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use uproot3 library namespace explicitly to avoid errors from uproot-methods v0.9.1
   - Use uproot3 in 'xmlio' extra
* Use pattern: import uproot3 as uproot
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,6 @@ extras_require['test'] = sorted(
             'papermill~=2.0',
             'nteract-scrapbook~=0.2',
             'jupyter',
-            'uproot~=3.3',
             'graphviz',
             'jsonpatch',
         ]
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ extras_require = {
     ],
     'torch': ['torch~=1.2'],
     'jax': ['jax~=0.2.4', 'jaxlib~=0.1.56'],
-    'xmlio': ['uproot~=3.6'],  # Future proof against uproot4 API changes
+    'xmlio': ['uproot3~=3.14'],  # Future proof against uproot4 API changes
     'minuit': ['iminuit~=1.4.3'],  # v1.5.0 breaks pyhf for 32b TensorFlow and PyTorch
 }
 extras_require['backends'] = sorted(
--- a/src/pyhf/cli/rootio.py
+++ b/src/pyhf/cli/rootio.py
@@ -32,7 +32,7 @@ def cli():
 def xml2json(entrypoint_xml, basedir, output_file, track_progress):
     """Entrypoint XML: The top-level XML file for the PDF definition."""
     try:
-        import uproot
+        import uproot3 as uproot
 
         assert uproot
     except ImportError:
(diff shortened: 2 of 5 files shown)


matthewfeickert:
We should probably wait for feedback from @jpivarski before merging this btw.

jpivarski (approved):
That ought to be right. I don't know why it's saying that the module is not found—it's on PyPI.

matthewfeickert:
> That ought to be right. I don't know why it's saying that the module is not found—it's on PyPI.

@jpivarski This is happening when we just try to install _just_ `uproot`, which in a fresh virtual environment gets

```
(uproot-test) feickert@thinkpad:~$ pip list
Package        Version
-------------- -------
awkward        0.14.0
awkward0       0.15.0
cachetools     4.1.1
numpy          1.19.4
pip            20.2.4
pkg-resources  0.0.0
setuptools     50.3.2
uproot         3.13.0
uproot-methods 0.9.1
wheel          0.35.1
```

So I think(?) that as `uproot-methods` `v0.9.1` uses `uproot3` instead of `uproot` the problem lies there.

matthewfeickert:
> So I think(?) that as `uproot-methods` `v0.9.1` uses `uproot3` instead of uproot the problem lies there.

Yeah, it seems that has to be the issue as applying the following patch to `master` gets the CI to pass:

```diff
diff --git a/setup.py b/setup.py
index 68d5bb16..b3e8fb9e 100644
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@
     ],
     'torch': ['torch~=1.2'],
     'jax': ['jax~=0.2.4', 'jaxlib~=0.1.56'],
-    'xmlio': ['uproot~=3.6'],  # Future proof against uproot4 API changes
+    'xmlio': ['uproot~=3.6', 'uproot3'],  # Future proof against uproot4 API changes
     'minuit': ['iminuit~=1.4.3'],  # v1.5.0 breaks pyhf for 32b TensorFlow and PyTorch
 }
 extras_require['backends'] = sorted(
@@ -40,7 +40,6 @@
             'papermill~=2.0',
             'nteract-scrapbook~=0.2',
             'jupyter',
-            'uproot~=3.3',
             'graphviz',
             'jsonpatch',
         ]
```

Though it might be worth it to still keep what I have in this PR now rather than have to require both `uproot3` and `uproot`. Thoughts @kratsg?

lukasheinrich (approved):

kratsg (approved):

PR #1194, opened 2020-11-24
chore: Specify exceptions to remove inline ignores of E722

matthewfeickert:
# Description

As a follow up to PR #1191, this PR specify some exception for all exception cases, allowing for the `#noqa`s for [E722: Do not use bare except, specify exception instead](https://www.flake8rules.com/rules/E722.html) to be removed.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Specify exception for all exception cases
* Remove all inline noqa: E722
   - E722: Do not use bare except, specify exception instead
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -731,9 +731,10 @@ class Model:
             ):  # force to be not scalar, should we changed with #522
                 return tensorlib.reshape(result, (1,))
             return result
-        except:  # noqa: E722
+        except Exception as excep:
             log.error(
-                f'eval failed for data {tensorlib.tolist(data)} pars: {tensorlib.tolist(pars)}'
+                str(excep)
+                + f'\nEval failed for data {tensorlib.tolist(data)} pars: {tensorlib.tolist(pars)}'
             )
             raise
 
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -733,8 +733,8 @@ class Model:
             return result
         except Exception as excep:
             log.error(
-                repr(excep)
-                + f'\nEval failed for data {tensorlib.tolist(data)} pars: {tensorlib.tolist(pars)}'
+                f"Eval failed for data {tensorlib.tolist(data)} pars: {tensorlib.tolist(pars)}",
+                exc_info=True,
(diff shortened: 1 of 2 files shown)


src/pyhf/pdf.py
```diff
@@ -731,9 +731,10 @@ def logpdf(self, pars, data):
             ):  # force to be not scalar, should we changed with #522
                 return tensorlib.reshape(result, (1,))
             return result
-        except:  # noqa: E722
+        except Exception as excep:
             log.error(
-                f'eval failed for data {tensorlib.tolist(data)} pars: {tensorlib.tolist(pars)}'
+                str(excep)
```
MarcoGorelli, line 736: quick tip - `repr(excep)` will show the error as well as the message
```
In [1]: try:
   ...:     1/0
   ...: except Exception as e:
   ...:     print(repr(e))
   ...: 
ZeroDivisionError('division by zero')

In [2]: try:
   ...:     1/0
   ...: except Exception as e:
   ...:     print(str(e))
   ...: 
division by zero
```

matthewfeickert: Nice. Thanks for the good tip, @MarcoGorelli!

src/pyhf/pdf.py
```diff
@@ -731,9 +731,10 @@ def logpdf(self, pars, data):
             ):  # force to be not scalar, should we changed with #522
                 return tensorlib.reshape(result, (1,))
             return result
-        except:  # noqa: E722
+        except Exception as excep:
             log.error(
-                f'eval failed for data {tensorlib.tolist(data)} pars: {tensorlib.tolist(pars)}'
+                repr(excep)
+                + f'\nEval failed for data {tensorlib.tolist(data)} pars: {tensorlib.tolist(pars)}'
```
matthewfeickert, line 737: @kratsg has pointed out that this can also just be done with [`log.error(..., exc_info=True)`](https://docs.python.org/3/library/logging.html#logging.Logger.error) which gets info from [`sys.exc_info`](https://docs.python.org/3/library/sys.html#sys.exc_info).

kratsg (approved):

PR #1197, opened 2020-11-25
test: Verify NLL values and Minuit parameter uncertainties

### LINKED ISSUE #1193: Expand test for parameter uncertainties
# Question

I was looking through #1183 and noticed that there does not seem to be a test to check the values of parameter uncertainties returned by minuit, which I would expect to be done here: https://github.com/scikit-hep/pyhf/blob/a33baddf6620df371338c6d1b98fdbab1b9da5c4/tests/test_optim.py#L384-L408

I am happy to prepare a PR to add such a test, but wanted to check first whether this is of interest. I noticed it since that PR adds a HESSE call, which should slightly change the reported uncertainties, but did not find a corresponding change in the test.

#1183

alexander-held:
# Description

Resolves #1193.

This adds a check for the NLL value returned by minimization with various backends, which is compared against a reference.

It also adds a check for the parameter uncertainties returned by Minuit, which was motivated by #1183 (the addition of the HESSE call would have resulted in a slight change in the uncertainties returned here). A test for the parameter values could also be added here, but since that behavior is tested in `test_minimize` already I did not add it again here.

I noticed that three functions set parameters constant without using the keyword argument `fixed_vals`, which I expect is not on purpose. I added another commit in here that changes that.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add a test for the NLL value returned by minimization
* Explicitly specify the fixed_vals kwarg for function calls passing in fixed parameter values
```

### DIFF
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -406,7 +406,7 @@ def test_optim_uncerts(backend, source, spec, mu):
         return_uncertainties=True,
     )
     assert result.shape[1] == 2
-    assert pyhf.tensorlib.tolist(result)
+    assert pytest.approx([0.0, 0.26418431]) == pyhf.tensorlib.tolist(result[:, 1])
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


kratsg (approved):

matthewfeickert (approved):
Thanks very much for the PR, @alexander-held! LGTM. :shipit:

PR #1208, opened 2020-12-08
feat: Use iminuit v2.0 API

### LINKED ISSUE #1166: Participate in iminuit v2.0 beta?
Dear pyhf team,

I am about to finish a major rewrite of iminuit, version 2.0, that replaces Cython as the tool to wrap C++ Minuit2 with pybind11, which is going to solve several issues that the legacy code had. All the good things that this will bring are listed on top of this PR:
scikit-hep/iminuit#502

Switching to the new version of iminuit should be completely transparent to you, since the new version passes the comprehensive suite of unit tests of iminuit-v1.x. However, I would like to use this opportunity to finally remove interface that has been successively marked as deprecated in versions 1.3 to 1.5.

Therefore my two question to you:

 * Did you take note of the deprecation warnings in iminuit and did you keep up with the interface changes so far?
 * Are you interested in trying out a Beta release of v2.0 to work out any possible bugs in the new version before the release?

Best regards,
Hans, iminuit maintainer

matthewfeickert:
# Description

Resolves #1166 

Adopt the new `iminuit` `v2.0` API.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Require iminuit release compatible with v2.1
* Migrate to iminuit v2.0 API
```

### DIFF
--- a/src/pyhf/optimize/opt_minuit.py
+++ b/src/pyhf/optimize/opt_minuit.py
@@ -141,7 +141,7 @@ class minuit_optimizer(OptimizerMixin):
             fun=minimizer.fval,
             hess_inv=hess_inv,
             message=message,
-            nfev=minimizer.ncalls_total,
-            njev=minimizer.ngrads_total,
+            nfev=minimizer.nfcn,
+            njev=minimizer.ngrads,
             minuit=minimizer,
         )
--- a/src/pyhf/optimize/opt_minuit.py
+++ b/src/pyhf/optimize/opt_minuit.py
@@ -58,9 +58,6 @@ class minuit_optimizer(OptimizerMixin):
             wrapped_objective = objective_and_grad
             jac = None
 
-        # FAILING:
-        # * tests/test_optim.py
-        # * tests/test_validation.py
         minuit = iminuit.Minuit(wrapped_objective, init_pars, grad=jac)
         minuit.errors = step_sizes
         minuit.limits = init_bounds
@@ -112,7 +109,7 @@ class minuit_optimizer(OptimizerMixin):
         minimizer.tol = tolerance
         minimizer.migrad(ncall=maxiter)
         # Following lines below come from:
-        # https://github.com/scikit-hep/iminuit/blob/f12cd519aa08e244ad71f56fe7a97d42117196e4/src/iminuit/_minimize.py#L111-L130
+        # https://github.com/scikit-hep/iminuit/blob/23bad7697e39d363f259ca8349684df939b1b2e6/src/iminuit/_minimize.py#L111-L130
         message = "Optimization terminated successfully."
         if not minimizer.valid:
             message = "Optimization failed."
--- a/src/pyhf/optimize/opt_minuit.py
+++ b/src/pyhf/optimize/opt_minuit.py
@@ -143,14 +143,14 @@ class minuit_optimizer(OptimizerMixin):
         if minimizer.valid:
             # Extra call to hesse() after migrad() is always needed for good error estimates. If you pass a user-provided gradient to MINUIT, convergence is faster.
             minimizer.hesse()
-            hess_inv = minimizer.covariance()
+            hess_inv = minimizer.covariance
 
         unc = None
         if return_uncertainties:
             unc = minimizer.np_errors()
 
         return scipy.optimize.OptimizeResult(
-            x=minimizer.np_values(),
+            x=minimizer.values,
             unc=unc,
             success=minimizer.valid,
             fun=minimizer.fval,
--- a/src/pyhf/optimize/opt_minuit.py
+++ b/src/pyhf/optimize/opt_minuit.py
@@ -58,6 +58,9 @@ class minuit_optimizer(OptimizerMixin):
             wrapped_objective = objective_and_grad
             jac = None
 
+        # FAILING:
+        # * tests/test_optim.py
+        # * tests/test_validation.py
         minuit = iminuit.Minuit(wrapped_objective, init_pars, grad=jac)
         minuit.errors = step_sizes
         minuit.limits = init_bounds
--- a/src/pyhf/optimize/opt_minuit.py
+++ b/src/pyhf/optimize/opt_minuit.py
@@ -58,23 +58,24 @@ class minuit_optimizer(OptimizerMixin):
             wrapped_objective = objective_and_grad
             jac = None
 
-        # kwargs = dict(
-        #     fcn=wrapped_objective,
-        #     grad=jac,
-        #     start=init_pars,
-        #     error=step_sizes,
-        #     limit=init_bounds,
-        #     fix=fixed_bools,
-        #     print_level=self.verbose,
-        #     errordef=self.errordef,
-        # )
-        minuit = iminuit.Minuit(fcn=wrapped_objective, grad=jac)
-        minuit.start = init_pars
-        minuit.error = step_sizes
-        minuit.limit = init_bounds
-        minuit.fix = fixed_bools
-        minuit.print_level = self.verbose
-        minuit.errordef = self.errordef
+        kwargs = dict(
+            fcn=wrapped_objective,
+            grad=jac,
+            start=init_pars,
+            error=step_sizes,
+            limit=init_bounds,
+            fix=fixed_bools,
+            print_level=self.verbose,
+            errordef=self.errordef,
+        )
+        minuit = iminuit.Minuit(**kwargs)
+        # minuit = iminuit.Minuit(fcn=wrapped_objective, grad=jac)
+        # minuit.start = init_pars
+        # minuit.error = step_sizes
+        # minuit.limit = init_bounds
+        # minuit.fix = fixed_bools
+        # minuit.print_level = self.verbose
+        # minuit.errordef = self.errordef
         return minuit
 
     def _minimize(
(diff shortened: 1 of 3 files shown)


src/pyhf/optimize/opt_minuit.py
```diff
@@ -58,17 +58,32 @@ def _get_minimizer(
...
-            start=init_pars,
             error=step_sizes,
             limit=init_bounds,
             fix=fixed_bools,
             print_level=self.verbose,
             errordef=self.errordef,
         )
-        return iminuit.Minuit.from_array_func(**kwargs)
+        # minuit = iminuit.Minuit(**kwargs)
+        minuit = iminuit.Minuit(wrapped_objective, init_pars, grad=jac, **kwargs)
```
matthewfeickert, line 79: This needs to get fixed.

matthewfeickert:
@kratsg We're also going to need to revise these tests:

https://github.com/scikit-hep/pyhf/blob/924b86c5933ba47bf35af20718bdac1cd050c18f/tests/test_optim.py#L496-L517

tests/test_optim.py
```diff
@@ -496,22 +496,34 @@ def test_init_pars_sync_fixed_values_scipy(mocker):
...
+    # @kratsg, What is the need for mocker here?
+    # # patch all we need
+    # from pyhf.optimize import opt_minuit
+    #
+    # minimizer = mocker.patch.object(opt_minuit, 'iminuit')
+    minuit = opt._get_minimizer(f, [9, 9, 9], [(0, 10)] * 3, fixed_vals=[(0, 1)])
+    # assert minimizer.Minuit.from_array_func.call_args[1]['start'] == [1, 9, 9]
+    # assert minimizer.Minuit.from_array_func.call_args[1]['fix'] == [True, False, False]
+    assert list(minuit.values) == [1, 9, 9]
+    assert list(minuit.fixed) == [True, False, False]
```
matthewfeickert, line 512: @kratsg Maybe a dumb question, but why did we need `mocker` here in the first place? Am I missing something that `mocker` does, or if I clean this up is using this `f` I've defined fine in your view?

HDembinski:
Does not look trivial, but I have very little time over the weekend to look into this. I can have a look next week.

HDembinski:
You can try what happens when you call Minuit.fcn(parameter_values). If you get the error, then that would narrow it down.

matthewfeickert:
> I can have a look next week.

Thanks @HDembinski. This shouldn't be a high priority for you, but when you do get time for it please let us know if we can help. :+1:

HDembinski:
I want this to work in pyhf, I am eager to help, but I am very short on time right now. Thanks for adopting so quickly!

matthewfeickert:
> I want this to work in pyhf, I am eager to help, but I am very short on time right now. 

We fully understand and appreciate this :slightly_smiling_face:  &mdash; I think we're all rather impressed with how much development work you're getting done right now when everyone is so busy at the end of the year.

HDembinski:
> I think we're all rather impressed with how much development work you're getting done right now when everyone is so busy at the end of the year.

It is how I procrastinate from other work.

HDembinski:
I checked out this branch locally. `pip install -e .` seemed to work. I cannot run tests yet, it complains that pytorch is missing, which I am currently installing (800MB download, seriously?!)

HDembinski:
Ok, got pytorch, now when trying to run tests/test_optim.py, it complains that tensorflow is not installed. If these are optional backends, shouldn't the tests just skip when they are not installed?

HDembinski:
- It is only happening when analytical gradients are used.
- I compiled iminuit locally in DEBUG mode (just run make in the iminuit repo) and that gives more information:
`E           RuntimeError: Unable to cast Python instance of type <class 'torch.Tensor'> to C++ type 'std::vector<double, std::allocator<double> >'`

matthewfeickert:
> I checked out this branch locally. `pip install -e .` seemed to work

Okay great. I'll rebase this against `master` so that this branch also knows about the fix from PR #1220.

> * It is only happening when analytical gradients are used.
> * I compiled iminuit locally in DEBUG mode (just run make in the iminuit repo) and that gives more information:
>   `E           RuntimeError: Unable to cast Python instance of type <class 'torch.Tensor'> to C++ type 'std::vector<double, std::allocator<double> >'`
> 
> I let pybind11 do the conversion. I am not sure how exactly the cast works, but I guess it expects the type to support the PySequence protocol, which torch.Tensor perhaps does not. I can try to work around that.

Okay let us know if there is any information that we can provide.

> If these are optional backends, shouldn't the tests just skip when they are not installed?

Perhaps, but as we don't ship the tests with the code we assume that the dev team is going to be running the full test suite locally and we want to know that all the backends agree. So I guess I would say that the backends are optional for runtime but not optional for the dev team.

HDembinski:
Indeed, torch.Tensor does register as a Sequence. `isinstance(some_tensor, collections.abc.Sequence)` is False. The sequence protocol is required by pybind11. Now, there are three ways to fix this.
 * in iminuit: I make a custom caster that also accepts torch.Tensor or rather everything that supports the buffer protocol.
 * in pybind11: you could argue that their caster does not try hard enough (there are collections which are not sequences and which can still be converted to `std::vector`)
 * in pyhf: you could convert the tensor to a numpy array in your gradient function wrapper.

matthewfeickert:
> Indeed, torch.Tensor does **not** register as a Sequence. `isinstance(some_tensor, collections.abc.Sequence)` is False. The sequence protocol is required by pybind11. Now, there are three ways to fix this.
> 
> * in iminuit: I make a custom caster that also accepts torch.Tensor or rather everything that supports the buffer protocol.
> * in pybind11: you could argue that their caster does not try hard enough (there are collections which are not sequences and which can still be converted to `std::vector`)
> * in pyhf: you could convert the tensor to a numpy array in your gradient function wrapper.

Thanks for this diagnosis @HDembinski. I think we should avoid you doing any extra work if possible and I _believe_ that we can just case to NumPy here without any issue. I'll let you know if we end up running into additional issues, but I'll work on implementing the

> * in pyhf: you could convert the tensor to a numpy array in your gradient function wrapper.

option.

HDembinski:
I am working on the implementation for iminuit. Having a custom caster allows me to improve performance a bit.

HDembinski:
@matthewfeickert I checked that my patch https://github.com/scikit-hep/iminuit/pull/574 fixes the issue.

HDembinski:
It is now fixed in the iminuit develop branch.

matthewfeickert:
> It is now fixed in the iminuit develop branch.

Fantastic @HDembinski! Thanks so much. I've double verified that in our [HEAD of dependencies workflow that indeed this branch is passing now](https://github.com/scikit-hep/pyhf/runs/1554863374?check_suite_focus=true). So after the next patch release of `iminuit` we're good to go here on `pyhf` adopting `v2.0`.

src/pyhf/optimize/opt_minuit.py
```diff
@@ -58,17 +58,16 @@ def _get_minimizer(
...
-            error=step_sizes,
-            limit=init_bounds,
-            fix=fixed_bools,
-            print_level=self.verbose,
-            errordef=self.errordef,
-        )
-        return iminuit.Minuit.from_array_func(**kwargs)
+        # FAILING:
+        # * tests/test_optim.py
+        # * tests/test_validation.py
```
matthewfeickert, line 63: These are passing on the HEAD of `iminuit` so the next patch release this will be good to go. :+1:

lukasheinrich (approved):

tests/test_optim.py
kratsg: This was done so that we didn't need to do a proper function at all (e.g. if you mock out Minuit and just check what it gets called with -- without actually checking that it returns a proper object). The goal of this test was to test the behavior of our code, rather than Minuit's code.

kratsg (approved):
All looks good generally. Note that we mock out minuit since we don't need to use it to check for behavior from it. We're checking that the functions we've implemented handle the fixed values/init values correspondingly.

kratsg (changes requested):
Just realized we do want to mock out iminuit since we need to check functionality of our code, rather than of minuit's code in these tests.

tests/test_optim.py
```diff
@@ -499,10 +499,12 @@ def test_init_pars_sync_fixed_values_minuit(mocker):
...
-    minimizer = mocker.patch.object(opt_minuit, 'iminuit')
-    opt._get_minimizer(None, [9, 9, 9], [(0, 10)] * 3, fixed_vals=[(0, 1)])
-    assert minimizer.Minuit.from_array_func.call_args[1]['start'] == [1, 9, 9]
-    assert minimizer.Minuit.from_array_func.call_args[1]['fix'] == [True, False, False]
+    minuit = mocker.patch.object(getattr(opt_minuit, 'iminuit'), 'Minuit')
+    minimizer = opt._get_minimizer(None, [9, 9, 9], [(0, 10)] * 3, fixed_vals=[(0, 1)])
+
+    assert minuit.called
+    assert minuit.call_args.args[1] == [1, 9, 9]
+    assert minimizer.fixed == [True, False, False]
```
matthewfeickert, line 507: @kratsg These changes work on Python 3.8, like you saw locally, but they are failing in CI for Python 3.6.

kratsg (approved):

PR #1209, opened 2020-12-09
chore: Use exc_info in log.error, re-raise last exceptions

### LINKED ISSUE #1195: chore: Use exc_info in log.error to show error as well as the message
quick tip - `repr(excep)` will show the error as well as the message
```
In [1]: try:
   ...:     1/0
   ...: except Exception as e:
   ...:     print(repr(e))
   ...: 
ZeroDivisionError('division by zero')

In [2]: try:
   ...:     1/0
   ...: except Exception as e:
   ...:     print(str(e))
   ...: 
division by zero
```

_Originally posted by @MarcoGorelli in https://github.com/scikit-hep/pyhf/pull/1194#discussion_r529733090_

This happens in two places in the codebase:

```
$ git grep "str(excep)"
src/pyhf/contrib/cli.py:            str(excep)
src/pyhf/contrib/utils.py:        str(excep)
```

and so would be a good first issue for someone who wanted to make a small PR.

kratsg:
# Pull Request Description

This will resolve #1195.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use exc_info in log.error to show error + message
* Use raise without arguments to re-raise last raised exception
```

### DIFF
--- a/src/pyhf/cli/rootio.py
+++ b/src/pyhf/cli/rootio.py
@@ -38,7 +38,8 @@ def xml2json(entrypoint_xml, basedir, output_file, track_progress):
     except ImportError:
         log.error(
             "xml2json requires uproot, please install pyhf using the "
-            "xmlio extra: python -m pip install pyhf[xmlio]"
+            "xmlio extra: python -m pip install pyhf[xmlio]",
+            exc_info=True,
         )
     from .. import readxml
 
@@ -67,7 +68,8 @@ def json2xml(workspace, output_dir, specroot, dataroot, resultprefix, patch):
     except ImportError:
         log.error(
             "json2xml requires uproot, please install pyhf using the "
-            "xmlio extra: python -m pip install pyhf[xmlio]"
+            "xmlio extra: python -m pip install pyhf[xmlio]",
+            exc_info=True,
         )
     from .. import writexml
 
--- a/src/pyhf/contrib/cli.py
+++ b/src/pyhf/contrib/cli.py
@@ -63,10 +63,9 @@ def download(archive_url, output_directory, verbose, force, compress):
         if verbose:
             file_list = [str(file) for file in list(Path(output_directory).glob("*"))]
             print("\n".join(file_list))
-    except AttributeError as excep:
-        exception_info = (
-            str(excep)
-            + "\nInstallation of the contrib extra is required to use the contrib CLI API"
-            + "\nPlease install with: python -m pip install pyhf[contrib]\n"
+    except AttributeError:
+        log.error(
+            "\nInstallation of the contrib extra is required to use the contrib CLI API"
+            + "\nPlease install with: python -m pip install pyhf[contrib]\n",
+            exc_info=True,
         )
-        log.error(exception_info)
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -58,9 +58,9 @@ try:
                     archive.extractall(output_directory)
 
 
-except ModuleNotFoundError as excep:
+except ModuleNotFoundError:
     log.error(
-        str(excep)
-        + "\nInstallation of the contrib extra is required to use pyhf.contrib.utils.download"
-        + "\nPlease install with: python -m pip install pyhf[contrib]\n"
+        "\nInstallation of the contrib extra is required to use pyhf.contrib.utils.download"
+        + "\nPlease install with: python -m pip install pyhf[contrib]\n",
+        exc_info=True,
     )
--- a/src/pyhf/optimize/mixins.py
+++ b/src/pyhf/optimize/mixins.py
@@ -48,7 +48,7 @@ class OptimizerMixin:
         try:
             assert result.success
         except AssertionError:
-            log.error(result)
+            log.error(result, exc_info=True)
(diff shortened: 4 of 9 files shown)


matthewfeickert:
@kratsg This PR idea is fine, but why does this need to be a patch release and not just part of the upcoming `v0.6.0` minor release?

matthewfeickert:
So the motivation for the patch release if I understand this correctly is that while we properly pinned things in `v0.5.3` thanks to @jpivarski's advice as to not break `uproot` in the future and that works 

```
# pip install --quiet pyhf[xmlio]
# pip list
Package        Version
-------------- -------
attrs          20.3.0
awkward        0.14.0
cachetools     4.1.1
click          7.1.2
jsonpatch      1.28
jsonpointer    2.0
jsonschema     3.2.0
numpy          1.19.4
pip            20.3.1
pyhf           0.5.3
pyrsistent     0.17.3
PyYAML         5.3.1
scipy          1.5.4
setuptools     50.3.2
six            1.15.0
tqdm           4.54.1
uproot         3.13.1
uproot-methods 0.8.0
wheel          0.36.0
```

However, it seems that @alexander-held has found issues where this is breaking the dev cycle of `cabinetry` that uses `uproot4` as

```
# pip install pyhf[xmlio]
# pip list | grep uproot
uproot         3.13.1
uproot-methods 0.8.0
# pip install uproot4
# pip list | grep uproot
uproot         4.0.0
uproot-methods 0.8.0
uproot4        4.0.0
```

So this means that anyone who wants to try to use `pyhf` and `uproot` `v4.x` in the same environment is in for some trouble, but a patch would fix this:

```
# python -m pip install --quiet --extra-index-url https://test.pypi.org/simple/ --pre pyhf[xmlio]
# pip list
Package         Version
--------------- -----------
attrs           20.3.0
awkward0        0.15.1
cachetools      4.1.1
click           8.0.0a1
jsonpatch       1.28
jsonpointer     2.0
jsonschema      3.2.0
numpy           1.20.0rc1
pip             20.3.1
pyhf            0.5.4.dev27
pyrsistent      0.17.3
PyYAML          5.3.1
scipy           1.5.4
setuptools      50.3.2
six             1.15.0
tqdm            4.54.1
uproot3         3.14.1
uproot3-methods 0.10.0
wheel           0.36.0
# pip list | grep uproot
uproot          4.0.0
uproot3         3.14.1
uproot3-methods 0.10.0
```

While I was originally against this idea I think that this might be necessary to ensure that `pyhf` `v3.x` release work in the future.

alexander-held:
There may be one other solution I did not think of before - pinning `uproot4==0.1.2` should keep the `uproot` namespace clean. That however does not help users who want the latest `uproot` at some point in the future, whether they install it as `uproot` or `uproot4`.

matthewfeickert:
> That however does not help users who want the latest `uproot` at some point in the future, whether they install it as `uproot` or `uproot4`.

Yeah, what changed my mind on if this should be a patch release vs. going into `v0.6.0` is that a user in the future trying to use `pyhf` `v0.5.x` would have `uproot` broken for them unless they magically knew to pin to this version.

matthewfeickert (approved):
Thanks for the PR @kratsg. I'll obviously approve it, and if we're still in agreement on the idea of a patch release then we can keep the label on there too. :+1:

jpivarski:
Be sure to only pin (a subset of) `uproot3`, `uproot>=4.0.0`, `awkward0`, `awkward>=1.0.0`. In particular, don't pin `uproot-methods`.

I had to add an `uproot3-methods` to avoid incompatibilities with old `uproot-methods` packages that did not explicitly require `awkward<1.0.0`. The package named `uproot-methods` should be dropped entirely.

matthewfeickert:
> Be sure to only pin (a subset of) `uproot3`, `uproot>=4.0.0`, `awkward0`, `awkward>=1.0.0`. In particular, don't pin `uproot-methods`.

Yup. We just use compatible releases for `uproot3`

https://github.com/scikit-hep/pyhf/blob/0370dbd5d728b2be90bab95c020ca85b384657b8/setup.py#L11

jpivarski:
After talking with @alexander-held and @matthewfeickert in the Analysis Systems meeting, I recommended updating pyhf to the new Uproot/Awkward names (including `uproot3` and `awkward0` as "new names"), rather than back-porting Cabinetry. It's true that if users have old names, they'll have conflicts, but users need to update to new names, too.

The reason I introduced new names for the old interfaces is to allow scripts and analyses to continue with minimal disruption while also "living at head" with the latest versions. Technically, `uproot3` and `awkward0` are the "latest versions" of these old projects.

In particular, I think it's a better idea to release a patch version of pyhf that only changes the Uproot/Awkward dependencies without stealthily adding 0.6.0 features. Even though there isn't a branch for 0.5.x, it's not hard to make one: (1) git checkout the latest 0.5.x tag (0.5.3), (2) create a branch for this detached HEAD state, (3) spot-change the dependencies, and (4) push to GitHub, where you can make a tag to deploy from this 0.5.x branch.

matthewfeickert:
For the time being I'm going to remove the patch-release label from this PR, until we finalize the discussion #1211.

PR #1221, opened 2020-12-13
docs: Use sphinxcontrib-bibtex v2.1 and bibtex_bibfiles setting

matthewfeickert:
# Description

~~Require `sphinxcontrib-bibtex` releases compatible with `v2.0`, which requires the definition of the `bibtex_bibfiles` in the Sphinx `conf.py`. With `v2.0` [Sphinx must be run twice to generate the bibliography](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html#running-sphinx):~~

> ~~It is good to be aware that the extension stores all citation information in a `bibtex.json` file in the source folder. If it does not exist, the file will be created on your first sphinx build, and you will have to rerun the build to make use of it. The file is automatically kept up to date, with a warning whenever you need to rerun the build (i.e. whenever your citations change). The file can be stored in version control if you do not want your users to have to run sphinx twice when they clone your project, or when you need your documentation to be built on an external service which only runs sphinx once.~~

Require `sphinxcontrib-bibtex` releases compatible with `v2.1`, which requires the definition of the `bibtex_bibfiles` in the Sphinx `conf.py`, thanks to https://github.com/mcmtroffaes/sphinxcontrib-bibtex/pull/219 making the `sphinxcontrib-bibtex` `v2.1` release series able to run with a single pass through a Sphinx build. :tada: 

Requires PR #1220 to go in first.

ReadTheDocs build: https://pyhf.readthedocs.io/en/docs-use-sphinxcontrib-bibtex-v2.0/

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Require sphinxcontrib-bibtex releases compatible with v2.1
* Add bibtex_bibfiles list to Sphinx conf.py
* Update RTD to Python 3.8 runtime
```

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -43,9 +43,10 @@ jobs:
     - name: Test and build docs
       run: |
         python -m doctest README.rst
-        python setup.py build_sphinx
+        cd docs
+        make html
+        make html # Second time for sphinxcontrib-bibtex
         touch docs/_build/html/.nojekyll
-        python setup.py build_sphinx
     - name: Check schemas are copied over
       run: |
         # is a directory
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -45,6 +45,7 @@ jobs:
         python -m doctest README.rst
         python setup.py build_sphinx
         touch docs/_build/html/.nojekyll
+        python setup.py build_sphinx
     - name: Check schemas are copied over
       run: |
         # is a directory
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -43,9 +43,7 @@ jobs:
     - name: Test and build docs
       run: |
         python -m doctest README.rst
-        pushd docs
-        make html
-        popd
+        python setup.py build_sphinx
         touch docs/_build/html/.nojekyll
     - name: Check schemas are copied over
       run: |
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -45,7 +45,6 @@ jobs:
         python -m doctest README.rst
         cd docs
         make html
-        make html # Second time for sphinxcontrib-bibtex
         cd ..
         touch docs/_build/html/.nojekyll
     - name: Check schemas are copied over
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -46,6 +46,7 @@ jobs:
         cd docs
         make html
         make html # Second time for sphinxcontrib-bibtex
+        cd ..
         touch docs/_build/html/.nojekyll
     - name: Check schemas are copied over
       run: |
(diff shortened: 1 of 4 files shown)


matthewfeickert:
> With `v2.0` Sphinx must be run twice to generate the bibliography

I don't know how to get this to work on the RTD tests.

.github/workflows/docs.yml
```diff
@@ -43,7 +43,10 @@ jobs:
     - name: Test and build docs
       run: |
         python -m doctest README.rst
-        python setup.py build_sphinx
+        cd docs
+        make html
+        make html # Second time for sphinxcontrib-bibtex
```
matthewfeickert, line 48: How to do this for RTD builds?

matthewfeickert:
> > With `v2.0` Sphinx must be run twice to generate the bibliography
> 
> I don't know how to get this to work on the RTD tests.

All issues are resolved with `sphinxcontrib-bibtex` `v2.1`.

lukasheinrich (approved):

.github/workflows/docs.yml
```diff
@@ -43,7 +43,9 @@ jobs:
     - name: Test and build docs
       run: |
         python -m doctest README.rst
-        python setup.py build_sphinx
+        pushd docs
+        make html
```
kratsg, line 47: kinda confused on this change. the makefile just invokes the `setup.py build_sphinx` for us in the same way.

kratsg (approved):

.github/workflows/docs.yml
matthewfeickert: @kratsg similar, but not exact, no? 

https://github.com/scikit-hep/pyhf/blob/1f152585232b919075189ea80eb42d531347609c/docs/Makefile#L54-L58

kratsg: Still not sure why we want to change to explicitly using `make html`. https://www.sphinx-doc.org/en/master/usage/advanced/setuptools.html seems to be supported more nicely  -- so I'm not sure.

matthewfeickert: Let me revert this change here so that we can discuss this more in an issue. I could be wrong, and will need to go hunting for references that support this idea, that using `setuptools` to do anything explicitly is more and more being looked at as "bad" by the Python community and that APIs through more dedicated tooling is preferred. 

Off the top of my head I'm thinking of https://github.com/tox-dev/tox/issues/1100 as being a starting place for some of these discussions.

I have been wanting to move away from using `setuptools` for a while, and I would normally suggest using [`sphinx-build` directly](https://www.sphinx-doc.org/en/1.8/man/sphinx-build.html) but 

https://github.com/scikit-hep/pyhf/blob/1f152585232b919075189ea80eb42d531347609c/docs/Makefile#L4-L9
https://github.com/scikit-hep/pyhf/blob/1f152585232b919075189ea80eb42d531347609c/docs/Makefile#L54-L58

are just using `sphinx-build` anyway.

I'll make an Issue from this comment, but maybe @henryiii has additional thoughts here as well.

PR #1226, opened 2020-12-14
feat: Add to_numpy API to tensorlib

### LINKED ISSUE #1225: Add .to_numpy() API to tensorlib
It would be beneficial to add a ~`.numpy()`~ `.to_numpy()` API to all the tensor backends to make the transition needed for PR #1208 easier:

>> Indeed, torch.Tensor does **not** register as a Sequence. `isinstance(some_tensor, collections.abc.Sequence)` is False. The sequence protocol is required by pybind11. Now, there are three ways to fix this.
>> 
>> * in iminuit: I make a custom caster that also accepts torch.Tensor or rather everything that supports the buffer protocol.
>> * in pybind11: you could argue that their caster does not try hard enough (there are collections which are not sequences and which can still be converted to `std::vector`)
>> * in pyhf: you could convert the tensor to a numpy array in your gradient function wrapper.
>
>Thanks for this diagnosis @HDembinski. I think we should avoid you doing any extra work if possible and I _believe_ that we can just case to NumPy here without any issue. I'll let you know if we end up running into additional issues, but I'll work on implementing the
>
>> * in pyhf: you could convert the tensor to a numpy array in your gradient function wrapper.
>
>option.

_Originally posted by @matthewfeickert in https://github.com/scikit-hep/pyhf/issues/1208#issuecomment-744729277_

matthewfeickert:
# Description

Resolves #1225 

Add `.to_numpy()` API that returns the tensor converted to a [NumPy `ndarray`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html#numpy.ndarray).

Example:

```python
>>> import pyhf
>>> pyhf.set_backend("jax")
>>> tensor = pyhf.tensorlib.astensor([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]])
>>> tensor
DeviceArray([[1., 2., 3.],
             [4., 5., 6.]], dtype=float64)
>>> numpy_ndarray = pyhf.tensorlib.to_numpy(tensor)
>>> numpy_ndarray
array([[1., 2., 3.],
       [4., 5., 6.]])
>>> type(numpy_ndarray)
<class 'numpy.ndarray'>
```

`.to_numpy()` is used rather than `.numpy()` as this is being called from tensorlib (rather than by a tensor object) and the grammar seems to make more sense this way. c.f. [Awkward's API.](https://awkward-array.readthedocs.io/en/latest/_auto/ak.to_numpy.html)

ReadTheDocs build: https://pyhf.readthedocs.io/en/feat-add-tensor-to-numpy-api/api.html#backends
* [NumPy backend](https://pyhf.readthedocs.io/en/feat-add-tensor-to-numpy-api/_generated/pyhf.tensor.numpy_backend.numpy_backend.html#pyhf.tensor.numpy_backend.numpy_backend.to_numpy)
* [TensorFlow backend](https://pyhf.readthedocs.io/en/feat-add-tensor-to-numpy-api/_generated/pyhf.tensor.tensorflow_backend.tensorflow_backend.html#pyhf.tensor.tensorflow_backend.tensorflow_backend.to_numpy)
* [JAX backend](https://pyhf.readthedocs.io/en/feat-add-tensor-to-numpy-api/_generated/pyhf.tensor.jax_backend.jax_backend.html#pyhf.tensor.jax_backend.jax_backend.to_numpy)
* [PyTorch backend](https://pyhf.readthedocs.io/en/feat-add-tensor-to-numpy-api/_generated/pyhf.tensor.pytorch_backend.pytorch_backend.html#pyhf.tensor.pytorch_backend.pytorch_backend.to_numpy)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add pyhf.tensorlib.to_numpy API to convert tensor to NumPy ndarray
* Add intersphinx mapping to NumPy docs
* Add tests for to_numpy
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -61,6 +61,7 @@ xref_links = {"arXiv:1007.1727": ("[1007.1727]", "https://arxiv.org/abs/1007.172
 
 intersphinx_mapping = {
     'python': ('https://docs.python.org/3', None),
+    'numpy': ('http://docs.scipy.org/doc/numpy/', None),
     'scipy': ('https://docs.scipy.org/doc/scipy/reference/', None),
 }
 
--- a/src/pyhf/tensor/jax_backend.py
+++ b/src/pyhf/tensor/jax_backend.py
@@ -490,3 +490,30 @@ class jax_backend:
 
         """
         return _BasicNormal(mu, sigma)
+
+    def to_numpy(self, tensor_in):
+        """
+        Convert the TensorFlow tensor to a :class:`numpy.ndarray`.
+
+        Example:
+            >>> import pyhf
+            >>> pyhf.set_backend("jax")
+            >>> tensor = pyhf.tensorlib.astensor([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]])
+            >>> tensor
+            DeviceArray([[1., 2., 3.],
+                         [4., 5., 6.]], dtype=float64)
+            >>> numpy_ndarray = pyhf.tensorlib.to_numpy(tensor)
+            >>> numpy_ndarray
+            array([[1., 2., 3.],
+                   [4., 5., 6.]])
+            >>> type(numpy_ndarray)
+            <class 'numpy.ndarray'>
+
+        Args:
+            tensor_in (:obj:`tensor`): The input tensor object.
+
+        Returns:
+            :class:`numpy.ndarray`: The tensor conversion to a NumPy ``ndarray``.
+
+        """
+        return np.asarray(tensor_in, dtype=tensor_in.dtype)
--- a/src/pyhf/tensor/jax_backend.py
+++ b/src/pyhf/tensor/jax_backend.py
@@ -513,7 +513,7 @@ class jax_backend:
             tensor_in (:obj:`tensor`): The input tensor object.
 
         Returns:
-            :class:`numpy.ndarray`: The tensor conversion to a NumPy ``ndarray``.
+            :class:`numpy.ndarray`: The tensor converted to a NumPy ``ndarray``.
 
         """
         return np.asarray(tensor_in, dtype=tensor_in.dtype)
--- a/src/pyhf/tensor/numpy_backend.py
+++ b/src/pyhf/tensor/numpy_backend.py
@@ -472,3 +472,31 @@ class numpy_backend:
 
         """
         return _BasicNormal(mu, sigma)
+
+    def to_numpy(self, tensor_in):
+        """
+        Return the input tensor as it already is a :class:`numpy.ndarray`.
+        This API exists only for ``pyhf.tensorlib`` compatibility.
+
+        Example:
+            >>> import pyhf
+            >>> pyhf.set_backend("numpy")
+            >>> tensor = pyhf.tensorlib.astensor([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]])
+            >>> tensor
+            array([[1., 2., 3.],
+                   [4., 5., 6.]])
+            >>> numpy_ndarray = pyhf.tensorlib.to_numpy(tensor)
+            >>> numpy_ndarray
+            array([[1., 2., 3.],
+                   [4., 5., 6.]])
+            >>> type(numpy_ndarray)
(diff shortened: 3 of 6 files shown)


kratsg (commented):
I'm thinking we should maintain consistent API naming so `tolist` and `tonumpy` or `to_list` and `to_numpy` but maybe don't mix/match?

matthewfeickert:
> I'm thinking we should maintain consistent API naming so `tolist` and `tonumpy` or `to_list` and `to_numpy` but maybe don't mix/match?

I would argue for `to_list` and `to_numpy` but I also admit this is difficult.

Libraries like [`numpy`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.tolist.html) and [`torch`](https://pytorch.org/docs/stable/tensors.html#torch.Tensor.tolist)  adopted APIs like `tolist()` for their tensor **objects** and likewise adopted `numpy()` for their tensor -> ndarray API.

However, Awkward doesn't have these operations be methods of the Array API but of the library API. E.g.

- [`ak.to_list`](https://awkward-array.readthedocs.io/en/latest/_auto/ak.to_list.html)
- [`ak.to_numpy`](https://awkward-array.readthedocs.io/en/latest/_auto/ak.to_numpy.html)

In my mind this is the more clear choice for our API as `tensorlib` manipulated tensor objects but we do not have a generic `tensorlib.tensor`. So to me

```python
pyhf.tensorlib.to_numpy(tensor)
pyhf.tensorlib.to_list(tensor)
``` 

makes more sense than

```
pyhf.tensorlib.numpy(tensor)
pyhf.tensorlib.tolist(tensor)
```

In general I think it is good for us to match the tensor backends as much as possible, but I think here when there is a **conversion** of the tensor it makes sense to have a bit of an exception.

Maybe @jpivarski can comment more on the API design choice here.

jpivarski:
**Some comments on the use of underscores in Awkward Array:**

Many standard library Python names are compounds of short words without camel-case or underscores, and NumPy's `tolist` method is such an example. However, that was probably a bad convention for Python to have started.

In scikit-hep/awkward-0.x#240, I was convinced to change all multi-word names to include underscores and I've been using that convention since. Always having an underscore relieves you of having to decide on a case-by-case basis: "Is this word long enough that it needs an underscore? Is the criterion number of letters or syllables?" Not only do you not need to come up with a rule, but your users don't have to know the rule.

That said, I turned the method name `to_list` _back into_ `tolist` because we want to use it frequently and quickly, and because this is how NumPy spells it. Each method on `ak.Array` is a rule-breaking concession for convenience anyway: `x.tolist()` can be `ak.to_list(x)`, `x.type` can be `ak.type(x)`. In Awkward Array, free-standing functions are strongly preferred (to leave the `ak.Array` namespace free for `__getattr__` column access and for custom methods defined by `ak.behavior`), but turning a small array into a list of dicts to see more of its contents and printing its type are such frequent operations that we want to just up-arrow and add the suffix without jumping to the beginning of the line and balancing parentheses, and while we're at it, the no-underscore `tolist` is quicker to type.

So that's why I made an exception only for `tolist` as a method. I wouldn't have done it if not for the frequency of this operation (ten times more than anything else, I think, when working interactively), the precedent set by NumPy, and the fact that it's a method, which makes it an exception anyway—less shocking that a different rule applies.

(There are also internal names in Awkward Array's C++ layer that don't follow these rules, but only the high-level Python was subjected to this scrutiny.)

matthewfeickert:
Thanks for this nice summary @jpivarski! I think from what you've said this make sense, and it somewhat reinforces my thoughts on the idea of `pyhf.tensorlib.to_list()` and `pyhf.tensorlib.to_numpy()` given that we don't provide array objects but rather a module for array manipulation.

@kratsg @lukasheinrich what are your thoughts? I think that if you agree we should keep `to_numpy` then I would just have this PR move forward as is and then if you agree about changing the API to `to_list` than we should do that in a separate PR (as that's changing the API while here we're adding a new API).

kratsg (approved):

lukasheinrich (approved):

PR #1227, opened 2020-12-15
ci: Uninstall dependencies before installed HEAD versions

matthewfeickert:
# Description

After doing some testing on PR #1208 I've found that if `iminuit` (installed from PyPI) is not uninstalled before it is installed from HEAD from GitHub that it is not guaranteed to properly pick up changes.

This was found when Hans reported the PyTorch fix that allowed tests to pass for `pyhf` locally for `iminuit` `v2.0` was merged into the `develop` branch (which is `iminuit`'s default branch) and yet the HEAD of dependencies workflow as failing still. After I added in a 

```yaml
python -m pip uninstall --yes iminuit
```

to the workflow things were passing. This behavior should then be applied to all HEAD of dependencies tests to make sure that the source code on GitHub is being properly tested.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* pip uninstall the dependency that is being tested before installing from the HEAD of version control
   - Has been shown for iminuit to be required to properly test the HEAD of the default repo branch
```

### DIFF
--- a/.github/workflows/dependencies-head.yml
+++ b/.github/workflows/dependencies-head.yml
@@ -25,6 +25,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip setuptools wheel
         python -m pip install --ignore-installed --upgrade -q --no-cache-dir -e .[test]
+        python -m pip uninstall --yes scipy
         python -m pip install --upgrade --no-cache-dir cython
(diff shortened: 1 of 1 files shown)


matthewfeickert:
I'm going to take an executive action here to merge this without review so that PR #1208 can move forward.

PR #1232, opened 2020-12-19
feat: Add discovery p0 test statistic

kratsg:
# Description

Supersedes #520. Particularly it's just a full reimplementation of @nikoladze 's work but on top of more recent `pyhf` master. I initially tried to rebase, but there's been a lot of changes in the `infer` API that I felt it was better to pick out all the pieces, and re-implement it accordingly.

To-Do:
- [ ] ~how to handle `poi_test` gracefully? Should `q0` simply warn and set `poi_test` or should we fail harder?~ -- see #1247
  - `hypotest` should handle more gracefully, but the calculators should not
- [ ] ~how to handle `asimov_mu` gracefully? Should `hypotest` or `AsymptoticCalculator` check if `asimov_mu` is consistent with the test statistic being used? Should the test statistic functions have a `func.default_asimov_mu` attribute?~ -- see #1247
- [ ] ~what to name all the various "p-values" returned via hypotest? (can't reasonably call it CLs or w/e) [related discussion: #966]~ -- see #1247
- [ ] ~should `AsymptoticCalculator` return a `b_only_distribution` for `q0` (what is this? we're very stuck on signal/background namings here)~ -- see #1247

Idea: change `poi_test` to `alternative_mu` (`alt_mu`) and `asimov_mu` to `null_mu`. This perhaps clarifies the meaning of the testing at least. Then we need to have toys to support these distributions. Then maybe change from `signal_plus_b` to `alternative_distribution` and `b_only` to `null_distribution` ?

ReadTheDocs build: https://pyhf.readthedocs.io/en/feat-discoveryteststat/_generated/pyhf.infer.test_statistics.q0.html

Closes #520

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add discovery test statistic q0
* Teach the calculators to switch the asimov mu depending on the test statistic (to be refactored in later PR)
* Add tests for discovery test stat
* Add validation ROOT files and tests

Co-authored-by: Nikolai Hartmann <nikoladze@posteo.de>
```


kratsg:
Idea: "change poi_test" to "alternative_mu" and "asimov_mu" to "null_mu". This perhaps clarifies the meaning of the testing at least. Then we need to have toys to support these distributions. Then maybe change from `signal_plus_b` to `alternative_distribution` and `b_only` to `null_distribution` ?

validation/run_single_q0.py
```diff
@@ -0,0 +1,34 @@
+import ROOT
```
matthewfeickert, line 1: Can we make this follow the normal guidelines of using `if __name__ == "__main__"` and defining imports at the top?

matthewfeickert (commented):
@kratsg overall this looks really good but let me think a bit more about the questions you raise in the PR body. Also for adding new binaries like `validation/multibin_histfactory_p0/data/data.root` for validation, do we want to add more binaires to the repo? Or should we have these added to [scikit-hep-testdata](https://github.com/scikit-hep/scikit-hep-testdata)?

validation/run_single_q0.py
kratsg: Since we don't do this for `validation/run_single.py` just yet -- can we get both of those in a separate PR?

kratsg: Added in #1241.

kratsg:
> @kratsg overall this looks really good but let me think a bit more about the questions you raise in the PR body. Also for adding new binaries like `validation/multibin_histfactory_p0/data/data.root` for validation, do we want to add more binaires to the repo? Or should we have these added to [scikit-hep-testdata](https://github.com/scikit-hep/scikit-hep-testdata)?

I think we should add them to this repo for now. If we then decide to migrate later -- we should. But we should start considering using `json2xml` functionality to generate on-the-fly binaries in the future... I would be ok with removing it here, and doing an on-the-fly generation for tests.

src/pyhf/infer/test_statistics.py
```diff
@@ -290,3 +290,67 @@ def tmu_tilde(mu, data, pdf, init_pars, par_bounds, fixed_params):
...
+        >>> test_mu = 0.0
+        >>> init_pars = model.config.suggested_init()
+        >>> par_bounds = model.config.suggested_bounds()
+        >>> fixed_params = model.config.suggested_fixed()
+        >>> pyhf.infer.test_statistics.q0(test_mu, data, model, init_pars, par_bounds, fixed_params)
+        array(2.98339447)
+
+    Args:
+        mu (Number or Tensor): The signal strength parameter (must be set to zero)
+
```
matthewfeickert, line 327: Seems that there's an extra space here.
```suggestion
```

validation/run_single_q0.py
```diff
@@ -0,0 +1,36 @@
+if __name__ == "__main__":
```
matthewfeickert, line 1: Since `validation/run_single_q0.py` never get run in the CI can we get this cleaned up a bit with some comments on how this is expected to be used in validation?

src/pyhf/infer/__init__.py
```diff
@@ -153,18 +153,27 @@ def hypotest(
...
         CLs_exp = []
         for n_sigma in [2, 1, 0, -1, -2]:
 
             expected_bonly_teststat = b_only_distribution.expected_value(n_sigma)
 
-            CLs = sig_plus_bkg_distribution.pvalue(
-                expected_bonly_teststat
-            ) / b_only_distribution.pvalue(expected_bonly_teststat)
+            if is_q0:
+                # despite the name in this case this is the discovery p value
```
matthewfeickert, line 171: We should probably make things more clear. If we're trying to read the code and it isn't clear what p-values are what that's going to be be problematic in the future.

matthewfeickert (commented):
@kratsg Thanks a lot for this pretty critical PR. :+1:

Technically everything looks good I think. Though I agree that we should probably take this opportunity to consider the terminology and the APIs to try and make things as clear as possible. So tagging @lukasheinrich and @alexander-held here for additional comments and thoughts.

matthewfeickert:
> I think we should add them to this repo for now. If we then decide to migrate later -- we should. But we should start considering using `json2xml` functionality to generate on-the-fly binaries in the future... I would be ok with removing it here, and doing an on-the-fly generation for tests.

Yeah they're all told only a few kB so I think it is fine. Though at the same time I wouldn't be against doing on-the-fly generation here.

alexander-held (commented):
Thanks for all the work on this! I tried out the API yesterday, it was straightforward to adopt and in agreement with a ROOT reference for the examples tested.

I do not have any immediate suggestions for naming, but think that the naming inside `hypotest` may in the long term be difficult to read. @kratsg's idea of alternative/null sounds like it could work (maybe with extra comments calling things by their name in the respective branches in the function?).

src/pyhf/infer/test_statistics.py
```diff
@@ -290,3 +290,67 @@ def tmu_tilde(mu, data, pdf, init_pars, par_bounds, fixed_params):
...
+    if pdf.config.poi_index is None:
+        raise UnspecifiedPOI(
+            'No POI is defined. A POI is required for profile likelihood based test statistics.'
+        )
+    if mu != 0.0:
+        log.warning(
+            'q0 test statistic only used for fit configuration with POI set to zero.'
+        )
+        mu = 0.0
```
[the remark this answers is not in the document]
kratsg: The docstrings do contain this. I added slightly more information into the warning.

src/pyhf/infer/utils.py
```diff
@@ -72,6 +79,7 @@ def get_test_stat(name):
         callable: The test statistic function
     """
     _mapping = {
+        "q0": q0,
         "q": qmu,
         "qtilde": qmu_tilde,
```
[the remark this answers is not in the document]
kratsg: We can add multiple aliases in a different PR. I'd say it's out of scope for this one.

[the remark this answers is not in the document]
kratsg: #1246

validation/run_single_q0.py
kratsg: I'd say this falls under #1241 for now -- since we don't have `validation/run_single.py` tested either -- since this requires ROOT.

src/pyhf/infer/__init__.py
kratsg: I opened #1247 to perhaps iterate on the internal names.

matthewfeickert: Okay, I would have preferred to get this in 1 go, but as you already have this follow up PR ready to go then we can just focus on that right after this.

validation/run_single_q0.py
matthewfeickert: Okay good enough

matthewfeickert (approved):
@kratsg Once https://github.com/google/jax/issues/5374 is resolved and @lukasheinrich reviews then I think this can go in, given that you've made dedicated followup PRs or Issues for most of my revision requests. Can you take care of the

> Summarize commit messages into a comprehensive review of the PR

though?

tests/test_validation.py
```diff
@@ -576,34 +659,95 @@ def validate_hypotest(pdf, data, mu_test, expected_result, tolerance=1e-6):
...
+            "asymptotics",
+        ),
+        (
+            '1bin_shapesys_q0_toys',
+            {'init_pars': 2, 'par_bounds': 2},
+            0.0,
+            "q0",
+            1e-6,
+            "toybased",
+        ),
```
matthewfeickert, line 696: This test is causing 

```python
assert abs(CLs_obs - expected_result['obs']) / expected_result['obs'] < tolerance
```

to fail in the `validate_hypotest` in `test_validation[1bin_shapesys_q0_mu1_toys]` given that

```
test_stat: q0                                                                                                                                                                                              
CLs_obs: 0.0035
obs expected results: 0.003
tolerance: 1e-06
abs difference: 0.0005
ratio: 0.16666666666666666
```

This is not related to the other JAX issues, as the results above comes from

```
$ pip list | grep jax
jax                           0.2.7
jaxlib                        0.1.57
```

@kratsg Can you set it up so that we can have the tests also accept a number of toys?

matthewfeickert (changes requested):

matthewfeickert:
The complexity of the fixtures in `test_validation` is getting increasingly more difficult to come back to and make sense of, and it is now over 1000 lines long. In another PR we need to either revise all of it or include some sort of walkthrough of how things work.

kratsg:
> The complexity of the fixtures in `test_validation` is getting increasingly more difficult to come back to and make sense of, and it is now over 1000 lines long. In another PR we need to either revise all of it or include some sort of walkthrough of how things work.

I have ideas on how to simplify it. Validation is very tricky to do, but I have a good idea of how to refactor it into something slightly more maintainble.

Issue #764, opened 2020-02-05
for later reference: how to compare to ROOT NLL


lukasheinrich:
this is a short script that compares our NLL to ROOT's NLL numerically. This will be useful for comparing apples to apples in terms of fit result.

```
import pyhf
import json


pyhf.Workspace(json.load(open('./shapesys.json')))                    
w = pyhf.Workspace(json.load(open('./shapesys.json')))   
m = w.model(measurement_name = 'GaussExample')   
d = w.data(m,)              


def root_fcn(d,m,pars):
    bin_width  = 1.0
    main,aux = m.fullpdf_tv.split(pyhf.tensorlib.astensor(d))  
    norm_aux, pois_aux = m.constraint_model.constraints_tv.split(aux)

    # poisson constraint
    const_pois = pyhf.tensorlib.sum(pyhf.tensorlib.log(pois_aux))     # is constant
    pois_comp  = m.constraint_model.constraints_poisson.make_pdf(pars).log_prob(pois_aux) + const_pois

    # normal constraint
    normal_comp = m.constraint_model.constraints_gaussian.make_pdf(pars).log_prob(norm_aux)

    #term from shape of main model
    main_rates = m.main_model.make_pdf(pars)._pdf.rate
    total_exp = pyhf.tensorlib.sum(main_rates)
    total_obs = pyhf.tensorlib.sum(main)
    main_const = - pyhf.tensorlib.sum(main*pyhf.tensorlib.log(bin_width))    # is constant
    main_comp = pyhf.tensorlib.sum(main*pyhf.tensorlib.log(main_rates/total_exp)) + main_const

    #extended term of main model
    extended_comp =  total_obs * pyhf.tensorlib.log(total_exp) - total_exp

    logl =  extended_comp + main_comp + normal_comp + pois_comp
    nll  = -logl

    print(logl, m.logpdf(pars,d)[0])

    return nll

pars = pyhf.tensorlib.astensor(m.config.suggested_init())
# pars[m.config.par_slice('bkg1Shape')] = [0.3,1.2]
nll = root_fcn(d,m,pars)
print('nll is',nll)
```

compared to 

```
import ROOT
import sys

f = ROOT.TFile.Open(sys.argv[1])
w = f.Get('combined')
p  = w.pdf('simPdf')
d  = w.data('obsData')

#w.var('gamma_bkg1Shape_bin_0').setVal(0.3)
#w.var('gamma_bkg1Shape_bin_1').setVal(1.2)
[...]

PR #1233, opened 2020-12-21
[WIP] learn nb

lukasheinrich:
# Description

spent some more time with MINUIT code to see how it does the covariance calculation and putting the lessons into a notebook

### DIFF
--- a/docs/examples/notebooks/learn/Hessians.ipynb
+++ b/docs/examples/notebooks/learn/Hessians.ipynb
@@ -1,0 +1,49 @@
+{
+ "metadata": {
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.7.2-final"
+  },
+  "orig_nbformat": 2,
+  "kernelspec": {
+   "name": "python3",
+   "display_name": "Python 3",
+   "language": "python"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 2,
+ "cells": [
+  {
+   "source": [
+    "# Hessian/Covariance Calculation in Minuit (with internal transforms)\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


lukasheinrich:
Added a second notebook on errors (the calc in #764 ) and we see compatible errors with various versions of how to calculate them. 

maybe a question to @alexander-held: is it important to your use-case to have MINUIT-like uncertainties or would uncertainties based on the standard covariance matrix (the blue dashed line) suffice? Separately from this we can implement MINOS-like uncert by doing a line search.

image

lukasheinrich:
fwiw, this reproduces the errors in the 1Lbb lhood

image

alexander-held:
In the `minuit_errors.ipynb` notebook, the objective is 2 NLL, but `errordef=0.5` - I think `errordef=1` is needed if the objective is not the NLL.

alexander-held:
I did not understand the last entry in the legend in the plots above, which seemed to have a completely different behavior. For me all the implementations look much more similar when using this notebook (and fixing index in the plotting, so the second parameter is shown):
uncs

I thought the discrepancy might be due to boundary effects, since the POI is limited to [0, 10] and hits both ends of this boundary (at 50 and 70 observed events).

I tried out 
```python
pdf = pyhf.simplemodels.hepdata_like([10.0], [500.0], [5.0])
scan = jnp.linspace(515, 545, 10)
```

which is further away from boundaries, but am surprised there is still a substantial difference:
uncs_2

Increasing all yields by a factor 10 (with constant background uncertainty) makes the discrepancy smaller, and with a factor 100 it slowly disappears.

Is the 1Lbb statistically limited? Maybe this is some systematics effect that causes differences? In any case I need to understand that other notebook better, since I did not expect such large differences.

lukasheinrich:
the last entry (brown) is using minuit but supplying the AD gradient. This uses `strategy` one, so perhaps  it's stoppinng eariler and therefore the underlying hesssian is different.

alexander-held:
Thanks, the strategy was the difference between our setups. I was running with 0.5.4 which does not yet include the switch to strategy 0 if gradients are provided. I am surprised that the difference is this large. I thought the HESSE call would help, but maybe the convergence fails completely? Strange since this likelihood is relatively simple.

PR #1257, opened 2021-01-14
docs: Build docs with sphinx-build via Sphinx Makefile

### LINKED ISSUE #1256: Use sphinx-build or other tools over setuptools for building docs
With regards to changing building the docs with `setuptools`

https://github.com/scikit-hep/pyhf/blob/1f152585232b919075189ea80eb42d531347609c/.github/workflows/docs.yml#L46

to using something like

```bash
pushd docs
make html
popd
```

@kratsg had pointed out that Sphinx does have defined tooling for using `setuptools`, so the change might not be needed

> Still not sure why we want to change to explicitly using `make html`. https://www.sphinx-doc.org/en/master/usage/advanced/setuptools.html seems to be supported more nicely  -- so I'm not sure.

I could be wrong, and will need to go hunting for references that support this idea, that using `setuptools` to do anything explicitly is more and more being looked at as "bad" by the Python community and that APIs through more dedicated tooling is preferred. 

Off the top of my head I'm thinking of https://github.com/tox-dev/tox/issues/1100 as being a starting place for some of these discussions.

I have been wanting to move away from using `setuptools` for a while, and I would normally suggest using [`sphinx-build` directly](https://www.sphinx-doc.org/en/1.8/man/sphinx-build.html) but 

https://github.com/scikit-hep/pyhf/blob/1f152585232b919075189ea80eb42d531347609c/docs/Makefile#L4-L9
https://github.com/scikit-hep/pyhf/blob/1f152585232b919075189ea80eb42d531347609c/docs/Makefile#L54-L58

are just using `sphinx-build` anyway.

Maybe @henryiii has additional thoughts here as well.

_Originally posted by @matthewfeickert in https://github.com/scikit-hep/pyhf/pull/1221#discussion_r557032023_

matthewfeickert:
# Description

Resolves #1256

Use Sphinx's Makefile to build the docs using `sphinx-build` via `make html`. 

https://github.com/scikit-hep/pyhf/blob/fd8fdf3c280604c173f6229837baab3616abe8c3/docs/Makefile#L54-L58

This deprecates the previously used method of using [Sphinx's `setuptools` integration](https://www.sphinx-doc.org/en/master/usage/advanced/setuptools.html).

This also more closely resembles the [ReadTheDocs procedure](https://docs.readthedocs.io/en/stable/builds.html)

> When we build your Sphinx documentation, we run `sphinx-build -b <format> . _build/<format>`. We also create pdf’s and ePub’s automatically based on your project.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Build the docs using the html Make target from Sphinx's Makefile
   - Sphinx Make targets all invoke sphinx-build
   - Deprecate pattern `python setup.py <command>`
* Remove build_sphinx info for setuptools from setup.cfg as deprecated
```

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -43,7 +43,9 @@ jobs:
     - name: Test and build docs
       run: |
         python -m doctest README.rst
-        python setup.py build_sphinx
+        pushd docs
+        make html
+        popd
         touch docs/_build/html/.nojekyll
     - name: Check schemas are copied over
       run: |
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,13 +52,6 @@ where = src
 console_scripts =
     pyhf = pyhf.cli:cli
 
-[build_sphinx]
-project = pyhf
-source-dir = docs
-build-dir = docs/_build
-all-files = 1
-warning-is-error = 1
-
 [flake8]
 # E203: whitespace before ':'
 # E402: module level import not at top of file


kratsg:
Kinda confused as supposedly they both do exactly the same thing

- https://github.com/sphinx-doc/sphinx/blob/7ecf6b88aa5ddaed552527d2ef60f1bd35e98ddc/sphinx/setup_command.py#L169-L191
- https://github.com/sphinx-doc/sphinx/blob/7ecf6b88aa5ddaed552527d2ef60f1bd35e98ddc/sphinx/cmd/build.py#L272-L284

However I suspect the reason we have to do it this way is b/c of setuptools_scm (https://github.com/pypa/setuptools_scm#usage-from-sphinx) so we should probably check the versioning to see if it's still working via `make html` which I suspect it wasn't given the change in #676.

kratsg (approved):

kratsg:
> we should probably check the versioning to see if it's still working

probably if we haven't noticed anything wrong on RTD then it's probably ok.

matthewfeickert:
> Kinda confused as supposedly they both do exactly the same thing

That's just comparing the core build part, no? Isn't most of the community concern over `python setup.py <command>` the additional side effects? I think that the tools that build around `setuptools` might still be using parts of the same underlying components, but in a more restricted manner. Herny can again probably comment more or we can go GitHub Issue diving.


> However I suspect the reason we have to do it this way is b/c of setuptools_scm (https://github.com/pypa/setuptools_scm#usage-from-sphinx) so we should probably check the versioning to see if it's still working via `make html` which I suspect it wasn't given the change in #676.

Sorry, can you spell this out more for me? The versioning of what? The docs? The schema?

As a related aside, how do you and @lukasheinrich build the docs locally when you want to see what they look like when you're working on something? I just have always done 

```
cd docs
make clean && make html
```

Have you been running 

```
python setup.py build_sphinx
```

locally ?

kratsg:
> Sorry, can you spell this out more for me? The versioning of what? The docs? The schema?

setuptools_scm versioning handles the "devXX" part of the versions which I thought would be picked up by the docs we built. Looking at this, I realized they aren't, so perhaps its a moot point.



> How do you build the docs locally when you want to see what they look like when

I run `python setup.py build_sphinx` myself. But I'm fine switching to a makefile. I'm just surprised there's many ways to do this and not a clear recommendation... or maybe I'm looking at the wrong stuff.

matthewfeickert:
> I'm just surprised there's many ways to do this and not a clear recommendation... or maybe I'm looking at the wrong stuff.

No, sadly I don't think you are looking at the wrong stuff. I think that this is an example of there being multiple ways to do something and then the Python core dev community starting to form a preference without the message getting well distributed. Maybe not too unlike `src/` dirs for projects.

PR #1274, opened 2021-01-21
feat: Add autoscan for upper limit using TOMS Algorithm 748

### LINKED ISSUE #1265: Auto scan for hypothesis test inversion (Implemented)
# Description

An auto scan feature for hypothesis test inversion can provide faster and more precise upper limits. One good way I've found to do this is to use the [TOMS748](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.toms748.html#scipy.optimize.toms748) root finding algorithm.

An `upperlimit_auto` (or some other appropriate name) function that runs an auto scan instead of a linear scan.

### Implementation
An implementation is below. I can make a PR including this in `pyhf.infer.intevals` if this issue has support.

```python
from scipy.optimize import toms748 as _toms748

def upperlimit_auto(data, model, low, high, level=0.05, atol=2e-12, rtol=1e-15):
    """
    Calculate an upper limit interval ``(0, poi_up)`` for a single
    Parameter of Interest (POI) using an automatic scan through
    POI-space, using the TOMS748 algorithm.
    
    ..., mostly copied from upperlimit docstring.
    """

    def f_all(mu):
        return hypotest(mu, data, model, test_stat="qtilde", return_expected_set=True)

    def f(mu, limit=0):
        # Use integers for limit so we don't need a string comparison
        if limit == 0:
            # Obs
            return f_all(mu)[0] - level
        else:
            # Exp
            # (These are in the order -2, -1, 0, 1, 2 sigma)
            return f_all(mu)[1][limit - 1] - level

    tb, _ = get_backend()
    obs = tb.astensor(_toms748(f, low, high, args=(0), k=2, xtol=atol, rtol=rtol))
    exp = [
        tb.astensor(_toms748(f, low, high, args=(i), k=2, xtol=atol, rtol=rtol))
        for i in range(1, 6)
    ]
    return obs, exp
```

beojan:
# Description

Resolves #1265

This PR uses SciPy's TOMS748 implementation to autoscan for the upper limit. @alexander-held's suggestions about memoizing the objective function and determining the best brackets for each band have been implemented.

ReadTheDocs build: https://pyhf--1274.org.readthedocs.build/en/1274/api.html#confidence-intervals

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add pyhf.infer.intervals and pyhf.infer.intervals.upper_limits modules.
* Migrate pyhf.infer.intervals.upperlimit to pyhf.infer.intervals.upper_limits.upper_limit.
* Add pyhf.infer.intervals.upper_limits.toms748_scan to API which implements the SciPy TOMS748
  root finding method to scan for the upper limit.
* Add pyhf.infer.intervals.upper_limits.linear_grid_scan to preserve the linear scan of the
  old pyhf.infer.intervals.upperlimit API in a specific API.
* Use pyhf.infer.intervals.upper_limits.toms748_scan as the default method of scanning for
  and upper limit in pyhf.infer.intervals.upper_limits.upper_limit.
* Update lower bound of the supported scipy versions to v1.2.0 to use scipy.optimize.toms748.
* Add tests for the pyhf.infer.intervals.upper_limits.toms748_scan API and the behavior 
  changes in pyhf.infer.intervals.upper_limits.upper_limit.
* Add 'Confidence Intervals' section to the API docs and move the pyhf.infer.intervals.upper_limits
  APIs there.
* Add a deprecation warning to pyhf.infer.intervals.upperlimit and note it is scheduled
  for removal in pyhf v0.9.0.
* Add Beojan Stanislaus to contributors list.

```


src/pyhf/infer/intervals.py
```diff
@@ -2,13 +2,96 @@
...
 import numpy as np
+from scipy.optimize import toms748 as _toms748
 
 
 def _interp(x, xp, fp):
     tb, _ = get_backend()
     return tb.astensor(np.interp(x, xp.tolist(), fp.tolist()))
 
 
+def upperlimit_auto(data, model, low, high, level=0.05, atol=2e-12, rtol=1e-15):
```
lukasheinrich, line 13: I think it would be great if we can integrate this into `pyhf.infer.upperlimit` via e.g. a `method` arg

`pyhf.infer.upperlimit( ... method = 'toms')` and 
`pyhf.infer.upperlimit( ... method = 'gridscan')`

src/pyhf/infer/intervals.py
```diff
@@ -2,13 +2,96 @@
...
+    """
+    Calculate an upper limit interval ``(0, poi_up)`` for a single
+    Parameter of Interest (POI) using an automatic scan through
+    POI-space, using the TOMS748 algorithm.
```
kratsg, line 17:
```suggestion
    POI-space, using the :func:`~scipy.optimize.toms748` algorithm.
```

tests/test_infer.py
```diff
@@ -20,6 +20,38 @@ def check_uniform_type(in_list):
     )
 
 
+def test_upperlimit_auto(tmpdir, hypotest_args):
```
kratsg, line 23: we should add a test checking that the scan returns close to the other upperlimit methods.

src/pyhf/infer/intervals.py
alexander-held: I think `toms748` would work better as a name for the method, since that is easily searchable and will point to the algorithm without needing to read through docstrings (for people who prefer google over that).

tests/test_infer.py
beojan: The grid scan could be tested the same way the auto-scan is tested (with the `pytest.approx` tolerance turned right down because a grid scan needs a large number of points to reach the same precision as the auto-scan).

alexander-held:
A few more ideas that might be out of scope for this implementation:
- optional return of the mu values and obs/exp CLs would be convenient for visualizations, e.g. with `pyhf.contrib.viz.brazil`
- it might be useful to allow calculating the limit for obs/exp only, or obs + exp (without bands), which is much faster and sufficient for analysis optimizations
- some kind of status update (e.g. via `logging`) could be useful for complex workspaces, where the whole scan could take a very long time and the user may wonder whether the code is stuck
- it would be useful to protect against exceptions (e.g. when signs of left/right do not differ), since otherwise all results may be lost

src/pyhf/infer/intervals.py
alexander-held, line 13: This is personal taste, but I think the default tolerance (I think it is tolerance in mu and not the CLs, but could be wrong) could be much larger, e.g. `0.01`.

beojan: I've set the default tolerances to the minimum possible. It might make sense to have no default relative tolerance, and make it a required parameter instead.

beojan: OK, we now emit a warning if the user doesn't provide an `rtol` explaining that it should be set as high as is acceptable.

tests/test_infer.py
beojan: I've added a test comparing the automatic limits to the linear scan. The comparison is done within an absolute tolerance of 0.25, which is the granularity of the scan I compare to.

beojan:
I think this should be ready for formal review now.

tests/test_infer.py
matthewfeickert: > The comparison is done within an absolute tolerance of 0.25, which is the granularity of the scan I compare to.

Just a quick note that relative tolerances are preferred (c.f. the `help(pytest.approx)` for why).

matthewfeickert:
> I think this should be ready for formal review now.

Thanks very much @beojan. We have a few other PRs that we're trying to prioritize for the `v0.6.0` release, but we'll try to get to this one early next week as well.

tests/test_infer.py
beojan: I used an absolute tolerance here because the linear scan is conducted with an absolute step size so naturally the result would have a guaranteed absolute precision rather than a guaranteed relative precision.

matthewfeickert:
@beojan believe it or not, we haven't forgotten about this PR :grimacing:. We're just trying to make sure that we have some particular pieces in for a patch release before we start merging in new features for a minor release. But this will get taken care of in due course.

src/pyhf/infer/intervals.py
```diff
@@ -2,13 +2,125 @@
...
+    tb, _ = get_backend()
+    obs = tb.astensor(_toms748(f, low, high, args=(0), k=2, xtol=atol, rtol=rtol))
+    exp = [
+        tb.astensor(_toms748(f, *best_bracket(i), args=(i), k=2, xtol=atol, rtol=rtol))
+        for i in range(1, 6)
+    ]
+    return obs, exp
+
+
 def upperlimit(data, model, scan, level=0.05, return_results=False):
```
lukasheinrich, line 124: hi @beojan couldn't we do `scan = 'auto'` and trigger TOMS here?

i.e if scan is `scan = np.linspace(0,...1)` or any other kind of array honor that ... if `scan='auto'` run TOMS

lukasheinrich: if this works, we'd probably need to return `obs,exp,mus` (where mus=scan in the non-TOMS case) in order to communicate back where the evluation happened

beojan: Added.

src/pyhf/infer/intervals.py
```diff
@@ -54,6 +169,19 @@ def upperlimit(data, model, scan, level=0.05, return_results=False):
...
+            bounds = model.config.suggested_bounds()[
+                model.config.par_slice(model.config.poi_name).start
+            ]
+            obs_limit, exp_limit, results = upperlimit_auto(
+                data, model, bounds[0], bounds[1], rtol=1e-3, from_upperlimit_fn=True
+            )
+            if return_results:
+                return obs_limit, exp_limit, results
+            return obs_limit, exp_limit
+
```
lukasheinrich, line 184: thanks a lot @beojan -- could I ask you to factor the below out into `upperlimit_fixedscan` and then we have

```
    if isinstance(scan, str) and scan.lower() == 'auto':
            obs_limit, exp_limit, results = upperlimit_auto(
                data, model, bounds[0], bounds[1], rtol=1e-3, from_upperlimit_fn=True
            )
   else:
            obs_limit, exp_limit, results = upperlimit_scan(
                data, model, scan
            )            
            ...
```


to get   a bit more symmetry ?

matthewfeickert: No worries. We (or at least I) need to get back to this and give it another look over and review. I was hoping to figure out a way to have improved our release setup to make feature contributions like yours easier to fit in with patch vs. minor release cycles, but I don't think it is worth it here and we should just get your contribution in once it is ready and reviewed. :+1: I don't want to make you wait a huge amount of time (more than the time you've already waited).

gollumben:
Hi, 
sorry to bother, I was wondering whether there has been any further advance on this issue. The linear scan is showing to be either slow or inaccurate for our ensemble of signals and we'd be curious to try out this alternative. We could of course go back to cabinetry and use what is there, but personally I believe that having this in `pyhf` would be beneficial for the package.
Cheers,
Ben

matthewfeickert:
> I was wondering whether there has been any further advance on this issue.

This will be going into `v0.7.0`, but because we wanted this to go into a minor release and not a patch release we let it linger for a long time. I need to take some time to basically rebuild this on another branch and then merge it back here or get @beojan's approval to close this PR for another more updated one with him as co-author. The fact that this PR is so old is a good example of us needing to find a new way of doing releases (e.g. using release branches and not releasing of `master`/`main`).

> The linear scan is showing to be either slow or inaccurate for our ensemble of signals and we'd be curious to try out this alternative.

@gollumben Do you have examples you can share in a GitHub Discussion?

beojan:
I could try and rebase this PR later this week.

matthewfeickert:
> I could try and rebase this PR later this week.

You're welcome to do that (and I usually like rebasing PRs of my own to keep things clear on my local dev branch for myself) but my guess is that it will be way easier to resolve the merge conflict if we just attempt to merge in `master` to this PR's branch and then fix it there once.

gollumben:
Hi @matthewfeickert, thank you so much! I will share my example during the next week. Cheers, Ben

src/pyhf/infer/intervals.py
```diff
@@ -15,7 +17,132 @@ def _interp(x, xp, fp):
...
+    if rtol is None:
+        _warn(
+            "upperlimit_auto: rtol not provided, defaulting to 1e-15. "
+            "For optimal performance rtol should be set to the highest acceptable relative tolerance."
+        )
+        rtol = 1e-15
+
+    cache = {}
+
+    def f_all(mu):
```
kratsg, line 85: this should just use the `lru_cache` in python

src/pyhf/infer/intervals.py
```diff
@@ -15,7 +22,132 @@ def _interp(x, xp, fp):
...
+    data,
+    model,
+    low,
+    high,
+    level=0.05,
+    atol=2e-12,
+    rtol=None,
+    calctype="asymptotics",
+    test_stat="qtilde",
+    from_upperlimit_fn=False,
```
matthewfeickert, line 35: These should probably be taking the hypotest_kwargs like `upperlimit_fixedscan` does at the moment. Might want to do so renaming of things as well.

src/pyhf/infer/intervals.py
```diff
@@ -73,3 +205,55 @@ def upperlimit(data, model, scan, level=0.05, return_results=False, **hypotest_k
     if return_results:
         return obs_limit, exp_limits, (scan, results)
     return obs_limit, exp_limits
+
+
+def upperlimit(data, model, scan, level=0.05, return_results=False):
```
matthewfeickert, line 210: Same comment about taking kwargs as above.

src/pyhf/infer/intervals.py
kratsg: can't use `lru_cache` here easily. gotta roll our own to access all values that `f_all` has been called with.

src/pyhf/infer/intervals.py
```diff
@@ -64,12 +200,75 @@ def upperlimit(data, model, scan, level=0.05, return_results=False, **hypotest_k
...
+        data,
+        model,
+        bounds[0],
+        bounds[1],
+        rtol=1e-3,
+        from_upperlimit_fn=True,
+        **hypotest_kwargs,
+    )
+    if return_results:
+        return obs_limit, exp_limit, results
```
matthewfeickert, line 273: Need to still cover this line. Will get it later tonight.

matthewfeickert (changes requested):

matthewfeickert (approved):
@lukasheinrich, @kratsg, @alexander-held This is ready for review now given that I have full coverage and everything passing CI.

There's a bit of some API juggling going on, so I would invite you to think about how changes here as I've implimented them might make things worse. The most noticeable being that `intervals.upperlimit` now has `scan=None` by default which will then fall back to `intervals.upperlimit_auto_scan`. Thoughts on if having this be a default is better than having people explicitly ask for something like `scan="auto"`?

src/pyhf/infer/intervals.py
```diff
@@ -64,12 +200,77 @@ def upperlimit(data, model, scan, level=0.05, return_results=False, **hypotest_k
...
+    """
+    if scan is not None:
+        return upperlimit_fixed_scan(
+            data, model, scan, level, return_results, **hypotest_kwargs
+        )
+    # else:
+    bounds = model.config.suggested_bounds()[
+        model.config.par_slice(model.config.poi_name).start
+    ]
+    relative_tolerance = hypotest_kwargs.pop("rtol", 1e-8)
```
matthewfeickert, line 264: @alexander-held this is essentially the default relative tolerance now. Is this too tight in your mind?

kratsg (changes requested):
I think we should renamed autoscan and fixedscan to better names. autoscan should use toms748 or similar, while fixed scan should probably be more like linspace scan or range scan or similar. Secondly, I think we need to refactor it so that we have an `intervals` module/folder which cleans up some of the code. I suspect `intervals.py` is going to get very large in the future, but we can hold off on the refactor until then or do it now (doing it now makes the history easier to follow but adds more work to the PR).

tests/test_infer.py
```diff
@@ -466,16 +591,17 @@ def test_toy_calculator(tmpdir, hypotest_args):
...
     """
 
-    _, _, pdf = hypotest_args
-    pdf.config.param_set('mu').suggested_fixed = [True]
+    test_poi, data, pdf = hypotest_args
+    pdf = deepcopy(pdf)  # deepcopy to avoid altering hypotest_args fixture
+    pdf.config.param_set("mu").suggested_fixed = [True]
     with pytest.raises(pyhf.exceptions.InvalidModel):
-        pyhf.infer.hypotest(*hypotest_args)
+        pyhf.infer.hypotest(test_poi, data, pdf)
```
matthewfeickert, line 604: This is PR #2011, and it should go in by itself and then get merged in here.

src/pyhf/exceptions/__init__.py
```diff
@@ -175,3 +176,15 @@ def __init__(self, result):
...
+# Deprecated APIs
+def _deprecated_api_warning(
+    deprecated_api, new_api, deprecated_release, remove_release
+):
+    warn(
+        f"{deprecated_api} is deprecated in favor of {new_api} as of pyhf v{deprecated_release} and will be removed in pyhf {remove_release}."
+        + f" Please use {new_api}.",
+        DeprecationWarning,
+        stacklevel=3,  # Raise to user level
+    )
```
matthewfeickert, line 190: This is in PR #2012 and should go in there instead of here.

src/pyhf/infer/intervals/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+def upperlimit(
+    data, model, scan=None, level=0.05, return_results=False, **hypotest_kwargs
+):
+    """
+    .. note:: Deprecated API: Use :func:`~pyhf.infer.intervals.upper_limits.upper_limit`
+     instead.
+    .. warning:: :func:`~pyhf.infer.intervals.upperlimit` will be removed in
+     ``pyhf`` ``v0.9.0``.
+    """
+    from pyhf.exceptions import _deprecated_api_warning
```
matthewfeickert, line 20: This needs PR #2012 to go in first.

kratsg (changes requested):
Can we add some #1910 related sphinx directives in the docstrings? That's the last piece and it's all good to go.

src/pyhf/infer/intervals/upper_limits.py
```diff
@@ -0,0 +1,282 @@
...
+        hypotest_kwargs (:obj:`string`): Kwargs for the calls to
+         :class:`~pyhf.infer.hypotest` to configure the fits.
+
+    Returns:
+        Tuple of Tensors:
+
+            - Tensor: The observed upper limit on the POI.
+            - Tensor: The expected upper limits on the POI.
+
+    .. versionadded:: 0.7.0
```
matthewfeickert, line 79: I put the `.. versionadded` at the bottom to match how the CPython docs use it. c.f. https://docs.python.org/3/library/shutil.html#shutil.copystat

src/pyhf/infer/intervals.py
alexander-held: Yes, see this thread https://github.com/scikit-hep/pyhf/pull/1274#pullrequestreview-573310845. I would personally prefer a larger tolerance but am also fine with the resolution achieved there.

alexander-held (commented):
> The most noticeable being that intervals.upperlimit now has scan=None by default which will then fall back to intervals.upperlimit_auto_scan. Thoughts on if having this be a default is better than having people explicitly ask for something like scan="auto"?

I think it makes perfect sense to use this new version as default. I imagine that most users just care about the crossing and this in general should be a good way to obtain that.

kratsg (changes requested):
`intervals.upper_limits.upper_limit` and `intervals.upperlimit` is very confusing. Is there a way to clean this API up slightly? Otherwise, the rest looks good to me. I'd typically prefer if we don't have `upper_limit`.

matthewfeickert:
> `intervals.upper_limits.upper_limit` and `intervals.upperlimit` is very confusing. Is there a way to clean this API up slightly? Otherwise, the rest looks good to me. I'd typically prefer if we don't have `upper_limit`.

`intervals.upperlimit` is being deprecated so it is being left in place until we remove it in `v0.9.0` so that we don't break a bunch of people hard. Can you suggest the API signature that you'd like to use?

kratsg (approved):
I’m that case, all good from my side!

tests/test_infer.py
```diff
@@ -20,27 +23,124 @@ def check_uniform_type(in_list):
...
+    assert expected_cls == pytest.approx(0.05)
+
+
+def test_toms748_scan_rtol_warning(hypotest_args):
+    """
+    Test the UserWarning is raised if no rtol is given.
+    """
+    _, data, model = hypotest_args
+    with pytest.raises(UserWarning):
+        pyhf.infer.intervals.upper_limits.toms748_scan(data, model, 0, 5)
```
matthewfeickert, line 66: This will require an update to 

https://github.com/scikit-hep/pyhf/blob/f7e974f7c65524e5240c871154c00ba8e954f0bd/.github/workflows/lower-bound-requirements.yml#L35-L42

given that the current workflow ignores all warnings, and so this fails for the Minimum supported dependencies workflow. Do this in another PR.

```console
$ PYTHONWARNINGS='default' pytest tests/test_infer.py
```

works, but I don't think that will work for the entire test suite.

src/pyhf/infer/intervals.py
matthewfeickert: This has now been backed off to `1e-4`.

tests/test_infer.py
matthewfeickert: Just get rid of the UserWarning and set a default. Henry has warned us before that `UserWarnings` might as well be useless.

matthewfeickert:
I'm unsure as to what is causing Issue #2015, as I'm not able to reproduce it at all locally, but I can reproduce it in CI, but only on branches that contain the commits in this PR. :? @kratsg @lukasheinrich any thoughts here are welcome.

src/pyhf/infer/intervals/upper_limits.py
```diff
@@ -0,0 +1,273 @@
...
+        >>> data = pyhf.tensorlib.astensor(observations + model.config.auxdata)
+        >>> scan = np.linspace(0, 5, 21)
+        >>> obs_limit, exp_limits, (scan, results) = pyhf.infer.intervals.upper_limits.upper_limit(
+        ...     data, model, scan, return_results=True
+        ... )
+        >>> obs_limit
+        array(1.01764089)
+        >>> exp_limits
+        [array(0.59576921), array(0.76169166), array(1.08504773), array(1.50170482), array(2.06654952)]
+        >>> pyhf.set_backend("jax")
```
matthewfeickert, line 234: @kratsg I'm not sure on this, but I think that because simply calling `pyhf.set_backend("jax")` here fixes the problem seen in Issue #2015 in CI makes me think that there is a bug somewhere in how the backends are getting reset in the tensor manager.

What do you think? Should we allow this somewhat ridiculous hack to get this PR in now? Or should we bug hunt before we let this go ahead?

matthewfeickert:
> I'm not sure on this, but I think that because simply calling `pyhf.set_backend("jax")` here fixes the problem seen in Issue #2015 in CI makes me think that there is a bug somewhere in how the backends are getting reset in the tensor manager.

PR #2017 will fix this, so after it is merged in and the `master` branch is merged back into this PR the tests will pass. I've verified this on a test branch already.

src/pyhf/infer/intervals/upper_limits.py
kratsg: fixed in #2017.

src/pyhf/infer/intervals/upper_limits.py
```diff
@@ -0,0 +1,269 @@
...
+):
+    """
+    Calculate an upper limit interval ``(0, poi_up)`` for a single
+    Parameter of Interest (POI) using an automatic scan through
+    POI-space, using the :func:`~scipy.optimize.toms748` algorithm.
+
+    Example:
+        >>> import numpy as np
+        >>> import pyhf
+        >>> pyhf.set_backend("numpy")
```
kratsg, line 39:
```suggestion
```

src/pyhf/infer/intervals/upper_limits.py
```diff
@@ -0,0 +1,269 @@
...
+    data, model, scan, level=0.05, return_results=False, **hypotest_kwargs
+):
+    """
+    Calculate an upper limit interval ``(0, poi_up)`` for a single
+    Parameter of Interest (POI) using a linear scan through POI-space.
+
+    Example:
+        >>> import numpy as np
+        >>> import pyhf
+        >>> pyhf.set_backend("numpy")
```
kratsg, line 153:
```suggestion
```

src/pyhf/infer/intervals/upper_limits.py
```diff
@@ -0,0 +1,269 @@
...
+    data, model, scan=None, level=0.05, return_results=False, **hypotest_kwargs
+):
+    """
+    Calculate an upper limit interval ``(0, poi_up)`` for a single Parameter of
+    Interest (POI) using root-finding or a linear scan through POI-space.
+
+    Example:
+        >>> import numpy as np
+        >>> import pyhf
+        >>> pyhf.set_backend("numpy")
```
kratsg, line 217:
```suggestion
```

kratsg (changes requested):
Perhaps dropping the explicit `set_backend` calls?

matthewfeickert:
> Perhaps dropping the explicit `set_backend` calls?

We can, though is the idea here that this avoids people focusing too much on NumPy? The main reason I've been putting those into examples is that I wanted the output tensor format to be made explicitly clear in advance. Though would you recommend not doing that now that we have the default backend machinery? (I'm going to go offline now, so I'll defer to you here.)

kratsg:
> Though would you recommend not doing that now that we have the default backend machinery? (I'm going to go offline now, so I'll defer to you here.)

It's cleaner since we might change default backend later. Or if we drop numpy later (doubt it) but I'm fine with it as it is. I don't think we set backend explicitly in most other tests.

kratsg (approved):

PR #1287, opened 2021-02-03
docs: Make docs badge link to latest stable version on ReadTheDocs

matthewfeickert:
# Description

Resolves #1286 

Make the default badge on the `README` link to the docs for the latest stable release of `pyhf` hosted on ReadTheDocs.

Example for `v0.5.4` (latest at the time of PR):

[RTD latest stable release](https://pyhf.readthedocs.io/)

```markdown
[![RTD latest stable release](https://img.shields.io/badge/docs-latest-blue.svg)](https://pyhf.readthedocs.io/)
```

ReadTheDocs build: https://pyhf.readthedocs.io/en/docs-make-default-docs-rtd/

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Add badge to README for latest versioned docs on ReadTheDocs
   - version number in badge display URL is updated through bump2version
* Change PyPI documentation URL to also point to ReadTheDocs
* Remove LGTM status badge as service is no longer used
```

Issue #1286, opened 2021-02-03
Make default docs link point to latest stable release on ReadTheDocs


matthewfeickert:
# Description

We have enough users and enough versions out that we should prioritize having the docs be versioned cleanly along with the releases. At the moment our docs links point to the docs build from `master` but we should probably have the default view of the docs correspond to the latest stable release on PyPI &mdash; so at the time of writing that would be the `v0.5.4` API.

We already have versioned docs on ReadTheDocs, so that might just mean adding a link to the README of https://pyhf.readthedocs.io/

[![RTD_docs_versioned](https://user-images.githubusercontent.com/5142394/106814195-a60c7200-6637-11eb-9fa2-f57c441ad79d.png)](https://pyhf.readthedocs.io/)

(Note: This has the [`v0.6.0`](https://github.com/scikit-hep/pyhf/projects/10) label so that this will go in and get picked up for the upcoming PyPI release.)

matthewfeickert:
I'm not sure if @jpivarski and @henryiii have any ideas or suggestions on the best way for redirect of https://scikit-hep.org/pyhf/ (I also need to reducate myself on how and where this happens).

Also looking at @henryiii's projects it seems that just https://pyhf.readthedocs.io/ is sufficient as it will resolve to `latest` by default.

kratsg:
Yeah, so one annoying thing is that we've got it set up on RTD to always build latest tags -- but scikit-hep is built from master. We could think about how to improve this more -- but versioning isn't so easy with gh-pages -- so the current way we have it makes sense.. and we should... ... i'm not sure. scikit-hep should just have a banner at top saying "latest dev! for latest release, see <link>"

matthewfeickert:
> We could think about how to improve this more -- but versioning isn't so easy with gh-pages

Exactly. I think if it was easy to do most projects would self host their versioned docs, but it is hard and so everyone uses RTD. I agree that I _want_ to keep things on our gh-pages, but I think the need of having versioned docs for users needs to go first and then we can always figure out ways to change it up later. But I think we **need** a badge on README that is like [![RTD latest](https://img.shields.io/badge/docs-latest-blue.svg)](https://pyhf.readthedocs.io/en/latest/)


> scikit-hep should just have a banner at top saying "latest dev! for latest release, see "

@kratsg I like this idea...but how do we implement this without having this eventually end up with RTD versions saying "latest dev"? Sorry if I'm missing the obvious.

kratsg:
[...]

PR #1288, opened 2021-02-04
docs: Add warning banners for dev versions on docs

kratsg:
# Pull Request Description

See #1286 and #1287. This adds banners at the top of the docs if the docs point to development docs or old versions.

For development docs, this checks to see if we're on `pyhf.readthedocs.io` or not. If we are not, then we're looking at latest (unstable) docs [e.g. on scikit-hep].

For old docs, if we are on `pyhf.readthedocs.io`, this will then make an asynchronous request to determine latest version of the docs and see if that matches what we're currently on.

Rendered pyhf.readthedocs.io docs here: https://pyhf.readthedocs.io/en/docs-versionwarnings/ (expect it to say "not a stable version" because it's on RTD and it's not the latest that RTD redirects to). This will also look different when pushed to scikit-hep as it will mention that this is the development version instead.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add banner to top to warn when docs being shown at scikit-hep.org/pyhf are development
   - Add custom CSS and JavaScript for Sphinx to use in docs/conf.py
```


kratsg:
Screen Shot 2021-02-03 at 8 46 39 PM

docs/index.rst
```diff
@@ -25,6 +25,13 @@
 
    <a class="github-fork-ribbon right-top fixed" href="https://github.com/scikit-hep/pyhf/" data-ribbon="View me on GitHub" title="View me on GitHub">View me on GitHub</a>
 
+
+.. raw:: html
+
+   <p id="dev-version"><strong>Warning:</strong> This is a development version. The latest stable version is at <a href="https://pyhf.readthedocs.io/">ReadTheDocs</a>.</p>
+
+   <p id="old-version"><strong>Warning:</strong> This is not a stable version. The latest stable version is <a id="latest-version-link" href="https://pyhf.readthedocs.io/">v?.?.?</a>.</p>
```
matthewfeickert, line 33:
```suggestion
   <p id="old-version"><strong>Warning:</strong> This is not the latest stable version. The latest stable version is <a id="latest-version-link" href="https://pyhf.readthedocs.io/">v?.?.?</a>.</p>
```

I think I'd still call old releases stable, but we just want to encourage using newer releases. So maybe just mention it is not the latest stable?

matthewfeickert (commented):
@kratsg this is _really_ nice! Thank you! I just have one small change.

Also, did you write this all from scratch or did you find templating on another project? If there was another project that you were able to look at can you mention that in the PR body for reference?

matthewfeickert:
@kratsg I turned on [Version warning in ReadTheDocs](https://docs.readthedocs.io/en/stable/versions.html#version-warning)

show_version_warning

which now show the following on older versions for RTD:

[warning_old_version](https://pyhf.readthedocs.io/en/v0.5.3/)

Do we still need the RTD logic in the `docs/_static/js/custom.js`? It seems no, and we could reduce this to just having to deal with `scikit-hep.org/pyhf`. Thoughts?

docs/_static/js/custom.js
```diff
@@ -0,0 +1,29 @@
...
+            url: "https://pyhf.readthedocs.io/",
+            success: function(data, textStatus, resp){
+                console.log(resp.getAllResponseHeaders());
+                const version = resp.getResponseHeader("x-rtd-version") || 'here';
+                if(window.location.href.indexOf(version) === -1){
+                    document.getElementById("latest-version-link").text = version;
+                    old_version.classList.add("version-warning");
+                }
+            }
+        });
```
matthewfeickert, line 27: Do we still need this logic given https://github.com/scikit-hep/pyhf/pull/1288#issuecomment-773022449?

kratsg: nope. removed.

kratsg:
> Also, did you write this all from scratch or did you find templating on another project? If there was another project that you were able to look at can you mention that in the PR body for reference?

most of it is from scratch. I grabbed some styles based on the pallet theme, but it's pretty much all custom.

matthewfeickert:
> most of it is from scratch. I grabbed some styles based on the pallet theme, but it's pretty much all custom.

:sunglasses:

matthewfeickert (approved):
@kratsg LGTM! This is an awesome PR. :smile:

PR #1295, opened 2021-02-05
fix: Make wheel naming indicate wheels are not Python 2 compatible

henryiii:
The universal setting means that it gets a py2.py3 tag, which is not true. This should already get the correct, `py3` only tag without any extra settings. (the Requires-Python metadata setting is set correctly, so no harm, just not needed)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Remove universal wheel setting as pyhf is Python 3 only
   - Fixes py2.py3 tag naming scheme for wheels
* Add Henry Schreiner to list of contributors   
```

### DIFF
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -21,3 +21,4 @@ Contributors include:
 - Marco Gorelli
 - Pradyumna Rahul K
 - Eric Schanet
+- Henry Schreiner
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,9 +25,6 @@ classifiers =
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
 
-[bdist_wheel]
-universal = 1
-
 [options]
 setup_requires =
     setuptools_scm>=1.15.0


kratsg (approved):

kratsg:
Thanks! I was a little surprised that the universal setting was the culprit... Good news is we will have 0.6.0 out soon (our next minor release) so this can go in just in time :)

matthewfeickert (approved):
Awesome. Thanks @henryiii!

matthewfeickert:
For posterity (and self referencE), in the [PyPA's "Packaging and distributing projects" guide the section on wheels](https://packaging.python.org/guides/distributing-packages-using-setuptools/#universal-wheels) notes that

> Only use the `--universal setting`, if:
>
> 1. Your project runs on Python 2 and 3 with no changes (i.e. it does not require 2to3).
> 2. Your project does not have any C extensions.

matthewfeickert:
And TestPyPI shows that things are looking much better now :+1: 

[TestPyPI-non-universal](https://test.pypi.org/project/pyhf/0.5.4.dev71/#files)

PR #1300, opened 2021-02-09
docs: Add syntax highlighting for bash snippets

henryiii:
# Description

Add a little syntax highlighting to make snippets prettier :)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add bash syntax highlighting to command line examples in CONTRIBUTING.md 
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,13 +52,13 @@ We recommend first reading the "[Developing](https://scikit-hep.org/pyhf/develop
 
 You can install the development environment (which includes a number of extra) libraries and all others needed to run the tests via `pip`:
 
-```
+```bash
 python -m pip install --ignore-installed -U -e .[complete]
 ```
 
 To make the PR process much smoother we also strongly recommend that you setup the Git pre-commit hook for [Black](https://github.com/psf/black) by running
 
-```
+```bash
 pre-commit install
 ```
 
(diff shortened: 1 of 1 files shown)


matthewfeickert:
Thanks for the PR @henryiii. Out of curiosity what render does this give highlighting for you? I've never put these in for `bash` as I've never really gotten GitHub to show any syntax highlighting that is Shell related.

with_without

matthewfeickert (approved):

henryiii:
Well, mostly because it is correct, the quality of the render isn't as important as the fact it _is_ bash. Plain code fences should be for output or text, not code. But, anyway, it can make a difference: 

```bash
mkdir cmake && wget -qO- "https://cmake.org/files/v3.16/cmake-3.16.3-Linux-x86_64.tar.gz" | tar --strip-components=1 -xz -C cmake
export PATH=`pwd`/cmake/bin:$PATH
```

Your unspecified code fences were copied into mplhep which is why I noticed it. Always assume everything you write can be copied as a good example. :)

matthewfeickert:
> Well, mostly because it is correct, the quality of the render isn't as important as the fact it _is_ bash. Plain code fences should be for output or text, not code.

Sure, I've just noticed that different web frameworks can get confused about syntax highlighting for shell input so unless there is clearly helpful syntax highlighting then I have defulated to letting the renderer decide what code should look like, as it definitely isn't wrong.

PR #1338, opened 2021-02-23
docs: Drop modifierclass and improve autosummary

### LINKED ISSUE #761: Update modifierclass.rst template to accept functions that aren't methods of a class
# Description

First noticed in PR #750, while trying to get the docs to show the full signatures and docstrings:

```rst
.. autosummary::
   :toctree: _generated/
   :nosignatures:
   :template: modifierclass.rst
```

I'm getting the following warnings (which we treat as errors) when I try to build the docs:
```
WARNING: error while formatting arguments for pyhf.infer.calculators.generate_asimov_data: 'function' object has no attribute '__mro__'
WARNING: error while formatting arguments for pyhf.infer.hypotest: 'function' object has no attribute '__mro__'
WARNING: error while formatting arguments for pyhf.infer.mle.fit: 'function' object has no attribute '__mro__'
WARNING: error while formatting arguments for pyhf.infer.mle.fixed_poi_fit: 'function' object has no attribute '__mro__'
WARNING: error while formatting arguments for pyhf.infer.mle.twice_nll: 'function' object has no attribute '__mro__'
WARNING: error while formatting arguments for pyhf.infer.test_statistics.qmu: 'function' object has no attribute '__mro__'
```

which I believe is happening as `__mro__` only exists on the class, and these functions exist in the source code outside of a class definition.

This means that the [`modifierclass.rst` template](https://github.com/scikit-hep/pyhf/blob/1ee6e38d42d9551220f20de483e0049b28c848b0/docs/_templates/modifierclass.rst) will need to get updated to deal with functions that aren't methods of a class.

### LINKED ISSUE #1339: Fix up docstring for _minimize() functions in the optimizers
# Description

From https://github.com/scikit-hep/pyhf/pull/1338#pullrequestreview-596818258

> I'm not sure if it can be fixed here, but the "Minimizer Options" aren't being displayed correctly for the optimizer _minimize methods.

_minimize

> we've never been able to see the `_minimize` methods before, so it isn't surprising they might not look perfect.

kratsg:
# Pull Request Description

We've been a bit held back by the terrible `modifierclass.rst` I wrote a long time ago and it's been making some of the documentation a bit worse for the wear in general. This PR will migrate some portion of this into the autosummary extension instead under `class.rst` to only impact the `autoclass` parts.

Fixes #761, fixes #1339.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Improve autosummary documentation for classes, modules, functions and drop modifierclass template
* Split up inference into sub-sub-sections for clearer organization
* Remove method/function signatures from everything on the top-level API page
```


kratsg:
Note to self: need to understand what the `:nosignatures:` is actually doing here, but we likely don't need it...

docs/api.rst
```diff
@@ -119,27 +114,47 @@ Inference
...
    test_statistics.qmu
    test_statistics.qmu_tilde
    test_statistics.tmu
    test_statistics.tmu_tilde
-   mle.twice_nll
-   mle.fit
-   mle.fixed_poi_fit
-   hypotest
-   intervals.upperlimit
+   utils.get_test_stat
```
matthewfeickert, line 130: I get the idea of trying to group the `utils` bits together with their area of use, but then it seems a bit confusing to have the API "path" be different and still be under the same heading. What are your thoughts on making a small "Utils" section that would go under "Fits and Tests" that would just have `utils.get_test_stat` and `utils.create_calculator`?

matthewfeickert (commented):
@kratsg overall I **really** like these changes. They make things much nicer IMO and I really like in general how the Infer docs are sectioned out now.

I'm not sure if it can be fixed here, but the "Minimizer Options" aren't being displayed correctly for the optimizer `_minimize` methods.

https://pyhf--1338.org.readthedocs.build/en/1338/_generated/pyhf.optimize.opt_scipy.scipy_optimizer.html

_minimize


we've never been able to see the `_minimize` methods before, so it isn't surprising they might not look perfect.

matthewfeickert:
> - [ ] ~fix up github links in a smart way if possible~

What is the problem with the GitHub links? I think I missed this.

kratsg:
> What is the problem with the GitHub links? I think I missed this.

the `Edit on GitHub` etc have been broken for a while -- and it seems like nobody really uses them, so I'm dropping them by default.

docs/api.rst
kratsg: Well, then we'd do the same thing for "Fits and Tests" since those are different API paths too... All of this is under `pyhf.infer` and I'm not sure we need to split by anything further in the API for subsection.

matthewfeickert:
> the `Edit on GitHub` etc have been broken for a while -- and it seems like nobody really uses them, so I'm dropping them by default.

I had to go look and see what these buttons even were...so I think that's a very safe bet.

docs/api.rst
matthewfeickert: fair.

matthewfeickert (approved):
@kratsg I'll approve this now and I'll let you decide if we worry about the "Minimizer Options" render here or make a new Issue for it for another later PR.

kratsg:
> @kratsg I'll approve this now and I'll let you decide if we worry about the "Minimizer Options" render here or make a new Issue for it for another later PR.

Do you know how to fix the render? Particularly, it should just be specifying the additional keyword arguments that go into `options` in that signature.

kratsg:
> I'm not sure if it can be fixed here, but the "Minimizer Options" aren't being displayed correctly for the optimizer `_minimize` methods.

#1339 will handle the fix of the docstring.

matthewfeickert:
> Do you know how to fix the render? Particularly, it should just be specifying the additional keyword arguments that go into `options` in that signature.

@kratsg Can you take a look at what I have now? I turned them into bulleted lists.

matthewfeickert:
> #1339 will handle the fix of the docstring.

Oh. Seems we just missed each other. Do you want me to revert this and make a new PR?

kratsg:
> Oh. Seems we just missed each other. Do you want me to revert this and make a new PR?

Looks good. We'll close both issues in this PR and move on.

matthewfeickert:
> Looks good. We'll close both issues in this PR and move on.

okay sounds good. Let me push one more style fix.

PR #1344, opened 2021-03-01
docs: Clarify and harmonize docstrings for fit parameters

kratsg:
# Pull Request Description

This clarifies a lot of the docstrings and makes them slightly more verbose and appropriate for the corresponding function. This was initially raised due to confusino about the `fixed_params` docstrings which wasn't explicit about what that list or tensor should/could be.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Clarify docstrings for fit parameters: init_pars, par_bounds, fixed_params, and fixed_vals
```

### DIFF
--- a/src/pyhf/infer/__init__.py
+++ b/src/pyhf/infer/__init__.py
@@ -44,7 +44,8 @@ def hypotest(
         data (Number or Tensor): The data considered
         pdf (~pyhf.pdf.Model): The statistical model adhering to the schema ``model.json``
         init_pars (:obj:`tensor` of :obj:`float`): The starting values of the model parameters for minimization.
-        par_bounds (:obj:`tensor`): The parameter value bounds to be used for minimization
+        par_bounds (:obj:`tensor` of shape Nx2): The extrema of values the model parameters
+            are allowed to reach in the fit.
         fixed_params (:obj:`tensor` of :obj:`bool`): The flag to set a parameter constant to the value
             ``init_pars`` value during minimization.
         calctype (:obj:`str`): The calculator to create. Choose either 'asymptotics' (default) or 'toybased'.
--- a/src/pyhf/infer/__init__.py
+++ b/src/pyhf/infer/__init__.py
@@ -43,7 +43,7 @@ def hypotest(
         poi_test (Number or Tensor): The value of the parameter of interest (POI)
         data (Number or Tensor): The data considered
         pdf (~pyhf.pdf.Model): The statistical model adhering to the schema ``model.json``
-        init_pars (:obj:`tensor`): The initial parameter values to be used for minimization
+        init_pars (:obj:`tensor` of :obj:`float`): The starting values of the model parameters for minimization.
         par_bounds (:obj:`tensor`): The parameter value bounds to be used for minimization
         fixed_params (:obj:`tensor` of :obj:`bool`): The flag to set a parameter constant to the value
             ``init_pars`` value during minimization.
--- a/src/pyhf/infer/__init__.py
+++ b/src/pyhf/infer/__init__.py
@@ -44,8 +44,9 @@ def hypotest(
         data (Number or Tensor): The data considered
         pdf (~pyhf.pdf.Model): The statistical model adhering to the schema ``model.json``
         init_pars (:obj:`tensor` of :obj:`float`): The starting values of the model parameters for minimization.
-        par_bounds (:obj:`tensor` of shape Nx2): The extrema of values the model parameters
+        par_bounds (:obj:`tensor`): The extrema of values the model parameters
             are allowed to reach in the fit.
+            The shape should be ``(n, 2)`` for ``n`` model parameters.
         fixed_params (:obj:`tensor` of :obj:`bool`): The flag to set a parameter constant to its starting
             value during minimization.
         calctype (:obj:`str`): The calculator to create. Choose either 'asymptotics' (default) or 'toybased'.
--- a/src/pyhf/infer/__init__.py
+++ b/src/pyhf/infer/__init__.py
@@ -46,8 +46,8 @@ def hypotest(
         init_pars (:obj:`tensor` of :obj:`float`): The starting values of the model parameters for minimization.
         par_bounds (:obj:`tensor` of shape Nx2): The extrema of values the model parameters
             are allowed to reach in the fit.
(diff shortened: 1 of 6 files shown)


src/pyhf/infer/__init__.py
```diff
@@ -43,9 +43,11 @@ def hypotest(
...
         data (Number or Tensor): The data considered
         pdf (~pyhf.pdf.Model): The statistical model adhering to the schema ``model.json``
-        init_pars (:obj:`tensor`): The initial parameter values to be used for minimization
-        par_bounds (:obj:`tensor`): The parameter value bounds to be used for minimization
-        fixed_params (:obj:`tensor`): Whether to fix the parameter to the init_pars value during minimization
+        init_pars (:obj:`tensor` of :obj:`float`): The starting values of the model parameters for minimization.
+        par_bounds (:obj:`tensor` of shape Nx2): The extrema of values the model parameters
+            are allowed to reach in the fit.
+        fixed_params (:obj:`tensor` of :obj:`bool`): The flag to set a parameter constant to its starting
+            value during minimization.
```
matthewfeickert, line 50:
```suggestion
         value during minimization.
```

Not really sure if this is worth changing, but if we're trying to keep the docstings width to around 80-ish charecters then we might want to minimize our whitespace at the front of continued lines. It seems that (at least in this situation) Sphinx needs only 1 space as an offset to understand it is a continued line and not a new line.

src/pyhf/infer/calculators.py
```diff
@@ -38,9 +38,11 @@ def generate_asimov_data(asimov_mu, data, pdf, init_pars, par_bounds, fixed_para
         asimov_mu (:obj:`float`): The value for the parameter of interest to be used.
         data (:obj:`tensor`): The observed data.
         pdf (~pyhf.pdf.Model): The statistical model adhering to the schema ``model.json``.
-        init_pars (:obj:`tensor`): The initial parameter values to be used for fitting.
-        par_bounds (:obj:`tensor`): The parameter value bounds to be used for fitting.
-        fixed_params (:obj:`tensor`): Parameters to be held constant in the fit.
+        init_pars (:obj:`tensor` of :obj:`float`): The starting values of the model parameters for minimization.
+        par_bounds (:obj:`tensor` of shape Nx2): The extrema of values the model parameters
+            are allowed to reach in the fit.
```
matthewfeickert, line 43:
```suggestion
        par_bounds (:obj:`tensor`): The extrema of values the model parameters
         are allowed to reach in the fit.
         The shape should be ``(n, 2)`` for ``n`` model parameters.
```

src/pyhf/infer/calculators.py
```diff
@@ -192,10 +194,11 @@ def __init__(
         Args:
             data (:obj:`tensor`): The observed data.
             pdf (~pyhf.pdf.Model): The statistical model adhering to the schema ``model.json``.
-            init_pars (:obj:`tensor`): The initial parameter values to be used for fitting.
-            par_bounds (:obj:`tensor`): The parameter value bounds to be used for fitting.
-            fixed_params (:obj:`tensor`): Whether to fix the parameter to the init_pars value
-              during minimization.
+            init_pars (:obj:`tensor` of :obj:`float`): The starting values of the model parameters for minimization.
+            par_bounds (:obj:`tensor` of shape Nx2): The extrema of values the model parameters
+                are allowed to reach in the fit.
```
matthewfeickert, line 199:
```suggestion
        par_bounds (:obj:`tensor`): The extrema of values the model parameters
         are allowed to reach in the fit.
         The shape should be ``(n, 2)`` for ``n`` model parameters.
```

src/pyhf/infer/calculators.py
```diff
@@ -618,10 +621,11 @@ def __init__(
...
```
matthewfeickert, line 626:
```suggestion
        par_bounds (:obj:`tensor`): The extrema of values the model parameters
         are allowed to reach in the fit.
         The shape should be ``(n, 2)`` for ``n`` model parameters.
```

src/pyhf/infer/mle.py
```diff
@@ -96,9 +96,11 @@ def fit(data, pdf, init_pars=None, par_bounds=None, fixed_params=None, **kwargs)
     Args:
         data (:obj:`tensor`): The data
         pdf (~pyhf.pdf.Model): The statistical model adhering to the schema model.json
-        init_pars (:obj:`list`): Values to initialize the model parameters at for the fit
-        par_bounds (:obj:`list` of :obj:`list`\s or :obj:`tuple`\s): The extrema of values the model parameters are allowed to reach in the fit
-        fixed_params (:obj:`list`): Parameters to be held constant in the fit.
+        init_pars (:obj:`list` of :obj:`float`): The starting values of the model parameters for minimization.
+        par_bounds (:obj:`list` of :obj:`list`/:obj:`tuple`): The extrema of values the model parameters
+            are allowed to reach in the fit.
```
matthewfeickert, line 101: Should we talk about expected shape here too?

matthewfeickert (changes requested):
Thanks @kratsg. This is nice to have. I have some suggestions on the shape description part, but let me know your thoughts.

src/pyhf/infer/__init__.py
kratsg: if we use 4-space tabs everywhere, why not also in docstrings for consistency?

src/pyhf/infer/calculators.py
kratsg: can't add this, the spacing/indenting is wrong.

src/pyhf/infer/calculators.py
kratsg: can't add this, the spacing/indenting is wrong.

src/pyhf/infer/__init__.py
matthewfeickert: That's fine. Though I don't think we universally apply them. Maybe something to look at later.

src/pyhf/infer/mle.py
matthewfeickert: I was asking here because it is a bit harder to talk about shape given that it is a list of objects, but I think this is fine. People will probably complain if things are unclear.

matthewfeickert (approved):
LGTM

PR #1355, opened 2021-03-07
test: Test lower bounds of dependencies

### LINKED ISSUE #1350: Validate lower bounds of dependencies in CI
# Description

To ensure that the lower bound of all dependencies are still sufficient for the API used, add a test that installs from a `lower-bound-requirements.txt` that might look something like

```
# core
scipy==1.4.0
click==7.0.0
tqdm==4.56.0
jsonschema==3.2.0
jsonpatch==1.23.0
pyyaml==5.1.0
# xmlio
uproot3==3.14.0
uproot==4.0.0
# minuit
iminuit==2.1.0
# ...
```

and then run the test suite.

matthewfeickert:
# Description

Resolves #1350

To ensure that the lower bound of all dependencies are still sufficient for the API used, add a test that installs from a `lower-bound-requirements.txt`.

- [x] Tests are passing (CI failing for known reasons)

# Before Merging

For the PR Assignees:

```
* Test lowest version of dependencies specified in setup.cfg and setup.py
   - Add lower-bound-requirements.txt with lowest versions pinned
   - Add lower-bound-requirements.yml GHA nightly workflow to test lower-bound-requirements.txt versions on test suite
* Update lower bounds of extra dependencies in setup.py to reflect their actual lowest bounds
```

### DIFF
--- a/.github/workflows/lower-bound-requirements.yml
+++ b/.github/workflows/lower-bound-requirements.yml
@@ -1,0 +1,34 @@
+name: Minimum supported dependencies
+
+on:
+  push:
+  # Run daily at 0:01 UTC
+  schedule:
+  - cron:  '1 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  test:
+
+    runs-on: ${{ matrix.os }}
+    os: [ubuntu-latest]
+    python-version: [3.8]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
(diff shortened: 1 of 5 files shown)


kratsg (approved):
This seems ok, but also seems like a lot of extra maintenance for very little benefit? What situation are we trying to catch here with respect to users?

matthewfeickert:
> This seems ok, but also seems like a lot of extra maintenance for very little benefit? What situation are we trying to catch here with respect to users?

Hopefully not extra maintenance as this should be have to get updated very infrequently (also why I don't have it run on PRs). You'll notice that I adjusted some of the lower bounds in `setup.py` as well, and this was with the goal of making sure that we don't say "we can go as low as `X.Y.0`" when in reality the lowest we can go is `X.Y.1`.

PR #1377, opened 2021-03-25
feat: Add CLs component plotting kwargs to contrib.viz.brazil.plot_results

### LINKED ISSUE #1376: Add contrib.viz.brazil.cls_components to visualize CLs+b and CLb
# Description

Add an API to `pyhf.contrib.viz.brazil` that accepts the observed CLs+b and CLb values from `hypotest`'s `return_tail_probs` arg and then plots them along with the CLs. This could call `pyhf.contrib.viz.brazil.plot_results` to do most of the work.

The goal is something similar &mdash; in results not API &mdash; to RooStat's [`HypoTestInverterPlot::MakePlot`](https://root.cern.ch/doc/master/classRooStats_1_1HypoTestInverterPlot.html#af49ecd3827b8faecb9c22194213707a3).

matthewfeickert:
# Description

Resolves #1376 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add BrazilBandCollection class to hold artists drawn in Brazil band plots
* Add components kwargs to pyhf.contrib.viz.brazil.plot_results API
   - Plots the CLs+b and CLb components of the CLs ratio ontop of output of plot_results
   - Add kwargs for line color to plot_results
* Try to be more like matplotlib and return lists of artists drawn on axis and move ax arg to end of args
   - Create an axis object if ax is None
* Add tests, test data, and reference images for components kwarg
* Update README example plots
* Update intersphinx URL for numpy
   - intersphinx inventory has moved: http://docs.scipy.org/doc/numpy/objects.inv -> https://numpy.org/doc/stable/objects.inv
* Add intersphinx mapping for matplotlib
* Update usage behavior in notebooks
```


src/pyhf/contrib/viz/brazil.py
```diff
@@ -48,3 +48,93 @@ def plot_results(ax, mutests, tests, test_size=0.05):
 
     ax.set_xlabel(r"$\mu$ (POI)")
     ax.set_ylabel(r"$\mathrm{CL}_{s}$")
+
+
+def plot_cls_components(
```
matthewfeickert, line 53: @kratsg @lukasheinrich This is still WIP, but early feedback on API is very welcomed.

matthewfeickert:
If I remember correctly, the reason why the diff coverage is failing even though there are tests covering the control flow, is because the test coverage is reported _before_ the contrib tests are run

https://github.com/scikit-hep/pyhf/blob/8a6529513329639a44398283434d5530bb9dac94/.github/workflows/ci.yml#L34-L46

matthewfeickert:
Follow up reading on trying to get contrib covered as well: https://docs.codecov.io/docs/flags

matthewfeickert:
I'm factoring the coverage uploading trick out to PR #1380, and then I'll rebase this PR on it.

src/pyhf/contrib/viz/brazil.py
```diff
@@ -48,3 +48,94 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+
+    # plot CLs_obs and CLs_expected set
+    plot_results(ax, mutests, CLs_results, test_size)
+
+    CLsb_obs = np.array([tail_prob[0] for tail_prob in tail_probs])
+    CLb_obs = np.array([tail_prob[1] for tail_prob in tail_probs])
+
+    linewidth = kwargs.pop("linewidth", 2)
+    if not clb_only:
+        CLsb_color = kwargs.pop("clsb_color", "red")
```
matthewfeickert, line 124: Feedback on default color choices is welcomed. Also on thinking about how to get better and more colorblind friendly colorwheels in use.

From Giordon: https://seaborn.pydata.org/tutorial/color_palettes.html

kratsg (approved):

matthewfeickert:
For color context, at the moment the docstring example produces this:
debug

src/pyhf/contrib/viz/brazil.py
```diff
@@ -48,3 +48,94 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+    # split into components
+    CLs_obs = np.array([test[0] for test in tests])
+    tail_probs = np.array([test[1] for test in tests])
+    CLs_exp_set = np.array([test[2] for test in tests])
+
+    # zip CLs_obs and CLs_exp_set back into format for plot_results
+    CLs_results = [(obs, exp_set) for obs, exp_set in zip(CLs_obs, CLs_exp_set)]
+
+    # plot CLs_obs and CLs_expected set
+    plot_results(ax, mutests, CLs_results, test_size)
```
matthewfeickert, line 117: Maybe take a page from ROOT here and allow for setting the y-axis label to `p-value` for the components plot.

src/pyhf/contrib/viz/brazil.py
kratsg: there should be an internal function call `plot_pvalue()` which is just an elided version of `ax.plot` for the particular CL value being plotted -- since you're essentially doing the same thing for CLs, CLs+b, CLb.

matthewfeickert: Can you elaborate further? I'm not really sure what this gets us as these are all just calls to `ax.plot(x, y)` with different values of `y`.

src/pyhf/contrib/viz/brazil.py
matthewfeickert: I've made this its own Issue in Issue #1386

alexander-held: Besides picking suitable colors, it can also be useful to have different linestyles for all relevant parts of the figure. `dashdot` and `dotted` seem to not be used here yet, and may make this more readable no matter the colors.

src/pyhf/contrib/viz/brazil.py
```diff
@@ -27,24 +27,124 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+    if not clsb_only:
+        CLb_color = kwargs.pop("clb_color", "blue")
+        ax.plot(
+            mutests,
+            CLb_obs,
+            color=CLb_color,
+            linewidth=linewidth,
+            label=r"$\mathrm{CL}_{b}$",
+        )
+    ax.legend(loc="best")
```
alexander-held, line 150: If you add a legend, I would suggest also including the expected CLs values and uncertainties in there, and possibly even the CLs=0.05 line.

src/pyhf/contrib/viz/brazil.py
alexander-held: The line widths differ, but possibly the color being the same as `test_size_color` could also be confusing.

src/pyhf/contrib/viz/brazil.py
```diff
@@ -27,24 +27,124 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+    # split into components
+    CLs_obs = np.array([test[0] for test in tests])
+    tail_probs = np.array([test[1] for test in tests])
+    CLs_exp_set = np.array([test[2] for test in tests])
+
+    # zip CLs_obs and CLs_exp_set back into format for plot_results
+    CLs_results = [(obs, exp_set) for obs, exp_set in zip(CLs_obs, CLs_exp_set)]
+
+    # plot CLs_obs and CLs_expected set
+    y_label = kwargs.pop("ylabel", r"$p\,$-value")
```
alexander-held, line 124: Very nit-picky, but maybe some phrasing to acknowledge that CLs is not a p-value is useful. `(modified) p-value` maybe?

src/pyhf/contrib/viz/brazil.py
```diff
@@ -27,24 +27,124 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+          hypothesis tests were performed.
+        tests (:obj:`list` or :obj:`array`): The collection of :math:`p`-values
+          from the hypothesis tests.
+          ``tests`` is required to have the same structure as
+          :func:`pyhf.infer.hypotest`'s return when using ``return_expected_set=True``
+          and ``return_tail_probs=True``: a tuple of :math:`\mathrm{CL}_{s}`,
+          :math:`\left[\mathrm{CL}_{s+b}, \mathrm{CL}_{b}\right]`,
+          :math:`\mathrm{CL}_{s,\mathrm{exp}}` band.
+        test_size (:obj:`float`): The size, :math:`\alpha`, of the test.
+        clb_only (:obj:`bool`): Bool for plotting only the :math:`\mathrm{CL}_{b}`
```
alexander-held, line 103: I believe this also plots CLs? "Only" sounds to me like it would only plot CLb, instead what the options does looks to be more like "skip CLsb".

src/pyhf/contrib/viz/brazil.py
matthewfeickert: Hm. I think the only concern that I have here is that the `CL_s+b` and the `CL_b` _are_ `p-values` and it is just the `CL_s` that is modified as it the ratio. This is a problem with trying to plot both the components _and_ the CLs values all at once &mdash; they're different things.

src/pyhf/contrib/viz/brazil.py
matthewfeickert: Yeah, at the moment all of these plot the `CLs` along with either both or one of the `CL_s+b` or `CL_b`. You're right though that there should be an option to plot only the components and not the `CLs`. Maybe I should change these from `only_X` to `no_X` args.

src/pyhf/contrib/viz/brazil.py
matthewfeickert: Good point. Time to bust out the `matplotlib` docs on how patches work again. :P

src/pyhf/contrib/viz/brazil.py
matthewfeickert: This is part of the reason that @lukasheinrich eventually suggested that we go with [`return_tail_probs` for `hypotest`](https://pyhf.readthedocs.io/en/v0.6.1/_generated/pyhf.infer.hypotest.html) as we had been going back and forth for days about how to describe the returned objects. I'm not sure how to name things better here though.

matthewfeickert:
test_plot_cls_components


Using `linestyle="dashdot"` as a default for the test alpha hopefully makes things more distinguished.

matthewfeickert:
@kratsg @lukasheinrich there isn't a huge rush on this, but as I think there was some final review that was still needed here it would be cool if we could review this and get it in before next week. Other work should take precedence though.

tests/contrib/test_viz.py
```diff
@@ -1,12 +1,78 @@
...
+
+    fig = Figure()
+    ax = fig.subplots()
+    brazil.plot_cls_components(ax, data["testmus"], data["results"], test_size=0.05)
+    return fig
+
+
+@pytest.mark.mpl_image_compare
+def test_plot_cls_components_no_clb():
+    data = json.load(open("tests/contrib/tail_probs_hypotest_results.json"))
```
kratsg, line 35: you should use the `testdata` fixture in this case. You'd just need a `tests/contrib/test_viz` folder with the data, and it will point at the right path for you and load it up correctly.

kratsg (changes requested):

tests/contrib/test_viz.py
matthewfeickert: Good point. Fixed it.

src/pyhf/contrib/viz/brazil.py
```diff
@@ -24,27 +24,205 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
 
-    ax.set_xlabel(r"$\mu$ (POI)")
-    ax.set_ylabel(r"$\mathrm{CL}_{s}$")
+    x_label = kwargs.pop("xlabel", r"$\mu$ (POI)")
+    y_label = kwargs.pop("ylabel", r"$\mathrm{CL}_{s}$")
+    ax.set_xlabel(x_label)
+    ax.set_ylabel(y_label)
+
+    axis_artists.extend((one_sigma_band, two_sigma_band, test_size_line))
+    return axis_artists
```
matthewfeickert, line 89: After learning more about how to be matplotlib like from a PR on `hist` I changed this so that it now returns a list of the artists like a normal `maplotlib` axis function would.

matthewfeickert:
Okay now I'm actually done messing about with the docs. :)

src/pyhf/contrib/viz/brazil.py
```diff
@@ -24,27 +24,210 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+    y_label = kwargs.pop("ylabel", r"$\mathrm{CL}_{s}$")
+    ax.set_xlabel(x_label)
+    ax.set_ylabel(y_label)
+
+    return [
+        cls_obs_line_artist,
+        cls_exp_line_artists,
+        [one_sigma_band, two_sigma_band],
+        test_size_line,
+    ]
```
lukasheinrich, line 94: should this perhaps be a object / nameduple such a la ? (see discusion in 

`return BrazilBandPlot(cls_obs = cls_obs_line_artist, ...)`


https://github.com/alexander-held/cabinetry/issues/113
so  that you can do 

```
artists = pyhf.contrib.viz.brazil.plot_results(ax, poi_vals, results)
artists.cls_obs.set_linecolor('k') 
```

matthewfeickert: Very much like this. :)

src/pyhf/contrib/viz/brazil.py
```diff
@@ -1,9 +1,25 @@
...
 import numpy as np
+from collections import namedtuple
 
+BrazilBandArtists = namedtuple(
+    "BrazilBandArtists",
+    ["cls_obs", "cls_exp", "cls_exp_band", "test_size"],
+    defaults=[None, [None] * 5, [None, None], None],
+)
+r"""
+:obj:`collections.namedtuple` of :obj:`matplotlib.artist` in the Brazil band plot.
```
matthewfeickert, line 11: For reasons I do not understand, Sphinx is having a fit and failing because of `namedtuple`.

```pytb
WARNING: Failed to get a method signature for pyhf.contrib.viz.brazil.BrazilBandArtists._field_defaults: {'cls_obs': None, 'cls_exp': [None, None, None, None, None], 'cls_exp_band': [None, None], 'test_size': None} is not a callable object
/home/feickert/Code/GitHub/pyhf/src/pyhf/contrib/viz/brazil.py:docstring of pyhf.contrib.viz.brazil.BrazilBandArtists._field_defaults:3: WARNING: Unexpected indentation.
/home/feickert/Code/GitHub/pyhf/src/pyhf/contrib/viz/brazil.py:docstring of pyhf.contrib.viz.brazil.BrazilBandArtists._field_defaults:4: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/feickert/Code/GitHub/pyhf/src/pyhf/contrib/viz/brazil.py:docstring of pyhf.contrib.viz.brazil.BrazilBandArtists._field_defaults:7: WARNING: Unexpected indentation.
/home/feickert/Code/GitHub/pyhf/src/pyhf/contrib/viz/brazil.py:docstring of pyhf.contrib.viz.brazil.BrazilBandArtists._field_defaults:8: WARNING: Inline strong start-string without end-string.
WARNING: Failed to get a method signature for pyhf.contrib.viz.brazil.BrazilBandArtists._fields: ('cls_obs', 'cls_exp', 'cls_exp_band', 'test_size') is not a callable object
WARNING: Failed to get a method signature for pyhf.contrib.viz.brazil.BrazilBandArtists._fields_defaults: {'cls_obs': None, 'cls_exp': [None, None, None, None, None], 'cls_exp_band': [None, None], 'test_size': None} is not a callable object
/home/feickert/Code/GitHub/pyhf/src/pyhf/contrib/viz/brazil.py:docstring of pyhf.contrib.viz.brazil.BrazilBandArtists._fields_defaults:3: WARNING: Unexpected indentation.
/home/feickert/Code/GitHub/pyhf/src/pyhf/contrib/viz/brazil.py:docstring of pyhf.contrib.viz.brazil.BrazilBandArtists._fields_defaults:4: WARNING: Block quote ends without a blank line; unexpected unindent.
/home/feickert/Code/GitHub/pyhf/src/pyhf/contrib/viz/brazil.py:docstring of pyhf.contrib.viz.brazil.BrazilBandArtists._fields_defaults:7: WARNING: Unexpected indentation.
/home/feickert/Code/GitHub/pyhf/src/pyhf/contrib/viz/brazil.py:docstring of pyhf.contrib.viz.brazil.BrazilBandArtists._fields_defaults:8: WARNING: Inline strong start-string without end-string.
/home/feickert/Code/GitHub/pyhf/docs/examples/notebooks/sbottom-pallet/README.md:27: WARNING: Inline literal start-string without end-string.
/home/feickert/Code/GitHub/pyhf/docs/examples/notebooks/sbottom-pallet/README.md:32: WARNING: Inline literal start-string without end-string.
/home/feickert/Code/GitHub/pyhf/docs/examples/notebooks/sbottom-pallet/README.md:32: WARNING: Inline interpreted text or phrase reference start-string without end-string.
```

I haven't figured out how to get it to keep from balking, but if people have ideas that would be great @lukasheinrich @kratsg

alexander-held: I do not know what causes this, but have been using `typing.NamedTuple` instead of `collections. namedtuple` myself and not seen such an issue.

matthewfeickert: Hm, okay good to know. I tried this with `typing.NameTuple` already in https://github.com/scikit-hep/pyhf/pull/1377/commits/b08a42760e83dc81241ddff112b438d6febcb93f, but that failed as well.

lukasheinrich: maybe actually it's better to make this a subclass of `matplotlib.container.Container` which "are classes that collect semantically related Artists" 

https://matplotlib.org/stable/api/container_api.html#matplotlib.container.Container

tests/contrib/test_viz/tail_probs_hypotest_results.json
```diff
@@ -0,0 +1 @@
...0306e-12, 8.277717406724008e-10, 2.1871024270051275e-07]], [1.71782565371694e-13, [7.565008939636483e-14, 0.4403828131957232], [5.623198801534755e-19, 5.611123291797592e-16, 4.620721835800917e-13, 2.667563671013248e-10, 8.395178386244305e-08]], [4.4264538053488135e-14, [1.9494767682506594e-14, 0.4404150261084757], [1.0456498253680896e-19, 1.2445134378447365e-16, 1.2217287655073372e-13, 8.401713598966383e-11, 3.1462347681656477e-08]], [1.1169446644370348e-14, [4.9196235733322765e-15, 0.4404536527162585], [1.9080250006767082e-20, 2.7053322683717073e-17, 3.162291370596417e-14, 2.587595613009092e-11, 1.1518009445150847e-08]], [2.7613663488989174e-15, [1.2163770933924125e-15, 0.44049826777871665], [3.418096559152543e-21, 5.766686417109634e-18, 8.016951593246653e-15, 7.796887738082306e-12, 4.121059344016396e-09]], [6.691824089371058e-16, [2.9480728583174304e-16, 0.4405484691386306], [6.014343036613147e-22, 1.2059350559181132e-18, 1.991631729977912e-15, 2.299628789191756e-12, 1.4417855671818827e-09]], [1.590370785130139e-16, [7.007235252659966e-17, 0.44060387163654846], [1.0398976286066765e-22, 2.4752233266368097e-19, 4.850732788241502e-16, 6.642253139501716e-13, 4.934728037238966e-10]]]}
```
lukasheinrich, line 1: this I guess is not yet the format @alexander-held designed for µ-scans?

src/pyhf/contrib/viz/brazil.py
matthewfeickert: > maybe actually it's better to make this a subclass of `matplotlib.container.Container` which "are classes that collect semantically related Artists"
> 
> https://matplotlib.org/stable/api/container_api.html#matplotlib.container.Container

This worked really nicely. Thanks for the great suggestion @lukasheinrich. :+1: 

> and I wonder whether it's a good idea to have an `.axes` property - though e.g. `BarContainer` doesn't have it. But forcing a user to index into some artist such to get the axes seems unsatisfactory.

I think if we want to keep to the `matplotlib` API to make it easy for people to know how to interact with these objects we should avoid it. Also with the current API this isn't too bad

```python
fig = Figure()
ax = fig.subplots()
artists = pyhf.contrib.viz.brazil.plot_results(ax, poi_vals, results)
print(artists.cls_obs.axes)  # AxesSubplot(0.125,0.11;0.775x0.77)
```

tests/contrib/test_viz/tail_probs_hypotest_results.json
matthewfeickert: > this I guess is not yet the format/schema @alexander-held designed for µ-scans?

Sorry, can you link me (again maybe?)? I think I missed this.

src/pyhf/contrib/viz/brazil.py
```diff
@@ -24,27 +69,212 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+        >>> data = observations + model.config.auxdata
+        >>> poi_vals = np.linspace(0, 5, 41)
+        >>> results = [
+        ...     pyhf.infer.hypotest(
+        ...         test_poi, data, model, return_expected_set=True, return_tail_probs=True
+        ...     )
+        ...     for test_poi in poi_vals
+        ... ]
+        >>> fig, ax = plt.subplots()
+        >>> artists = pyhf.contrib.viz.brazil.plot_cls_components(ax, poi_vals, results)
```
lukasheinrich, line 180: why does this return multiple artists, oculdn't we add the components to the `BrazilBandContainer?`

matthewfeickert: As `BrazilBandContainer` acts as a `namedtuple` then would the idea be to have `plot_results` have something like

```python
    return BrazilBandContainer(
        (
            cls_obs_line,
            cls_exp_lines,
            one_sigma_band,
            two_sigma_band,
            test_size_line,
            None,
            None
        ),
    )
```

then? That seems a little strange as now `plot_results` returns an object that has public attributes that will never get used by it. Not sure what's the best approach here.

src/pyhf/contrib/viz/brazil.py
```diff
@@ -1,9 +1,67 @@
 """Brazil Band Plots."""
 import numpy as np
+from matplotlib.container import Container
 
 
-def plot_results(ax, mutests, tests, test_size=0.05):
+class BrazilBandContainer(Container):
+    r"""
+    Container for the :obj:`matplotlib.artist` s created in a
```
matthewfeickert, line 8: Not sure how to fix this space between the end of `matplotlib.artist` and the start of the `s`. If I remove it Sphinx freaks out. Thoughts @kratsg? Or just leave it.

src/pyhf/contrib/viz/brazil.py
lukasheinrich: I think of it like default ctor args that are also often `None`

src/pyhf/contrib/viz/brazil.py
```diff
@@ -1,9 +1,67 @@
...
+         band.
+
+        test_size (:class:`matplotlib.lines.Line2D`): The artist of the test size
+         line.
     """
+
+    def __init__(self, brazil_band_artists, **kwargs):
+        r"""
+        Args:
+            brazil_band_artists (:obj:`tuple`): Tuple of
```
matthewfeickert, line 36: I don't know why this is failing now. :( 

```pytb
/home/runner/work/pyhf/pyhf/src/pyhf/contrib/viz/brazil.py:docstring of pyhf.contrib.viz.brazil.BrazilBandContainer.__init__:3: WARNING: Unexpected indentation.
```

src/pyhf/contrib/viz/brazil.py
lukasheinrich: "Container for the :obj:`matplotlib.artist` objects created in a "..

src/pyhf/contrib/viz/brazil.py
lukasheinrich: this does not work?

```
    return BrazilBandContainer(
        (
            cls_obs_line,
            cls_exp_lines,
            one_sigma_band,
            two_sigma_band,
            test_size_line,
        ),
    )
```

and when we have it just pass a larger tuple?

```
    return BrazilBandContainer(
        (
            cls_obs_line,
            cls_exp_lines,
            one_sigma_band,
            two_sigma_band,
            test_size_line,
            pb_line,
            psb_line,
        ),
    )
```

src/pyhf/contrib/viz/brazil.py
matthewfeickert: This was somehow related to not having spaces between bullet points.

src/pyhf/contrib/viz/brazil.py
```diff
@@ -1,9 +1,72 @@
...
+
+              * ``test_size`` contains the :class:`matplotlib.lines.Line2D` of the test size.
+        """
+        (
+            cls_obs,
+            cls_exp,
+            one_sigma_band,
+            two_sigma_band,
+            test_size,
+        ) = brazil_band_artists
```
lukasheinrich, line 59: e.g. here you could unpacck 

```
default_comps,*optional_comp = brazil_band_artists
if optional_comp:
   pb_line, psb_line = optional_comp
```

src/pyhf/contrib/viz/brazil.py
matthewfeickert: > this does not work?
> 
> ```
>     return BrazilBandContainer(
>         (
>             cls_obs_line,
>             cls_exp_lines,
>             one_sigma_band,
>             two_sigma_band,
>             test_size_line,
>         ),
>     )
> ```
> 
> and when we have it just pass a larger tuple?
> 
> ```
>     return BrazilBandContainer(
>         (
>             cls_obs_line,
>             cls_exp_lines,
>             one_sigma_band,
>             two_sigma_band,
>             test_size_line,
>             pb_line,
>             psb_line,
>         ),
>     )
> ```

I'll have to check later, but I'm pretty sure this doesn't work as `matplotlib.container.Container` should behave like `collections.namedtuple` (c.f. [this comment in `matplotlib.container.StemContainer`](https://matplotlib.org/stable/_modules/matplotlib/container.html#StemContainer)) and 

```python
>>> from collections import namedtuple
>>> Point = namedtuple('Point', ['x', 'y'])
>>> Point(11, 12, z=5)
TypeError: __new__() got an unexpected keyword argument 'z'
>>> Point(11, 12, 5)
TypeError: __new__() takes 3 positional arguments but 4 were given
```

but I might be wrong.

src/pyhf/contrib/viz/brazil.py
henryiii: Might be a good idea for mplhep / hist! :)

lukasheinrich: @henryiii do you have thoughts on the `.axes` property on the container. @matthewfeickert I agree `artists.cls_obs.axes` isn't bad but it's not clear to me why the user needs to think about the sub-artiists if all they want is the axes. I could imagine the reason `.axes` is not part of the  `Container` base class is thatt in princicple a container could collect artists attached to various axes instances

matthewfeickert: Added for now in https://github.com/scikit-hep/pyhf/pull/1377/commits/a3e8ec6206e376df5e6e8566b4f3312f22f3a14b

lukasheinrich (approved):

lukasheinrich:
This LGTM! Thanks @matthewfeickert

src/pyhf/contrib/viz/brazil.py
```diff
@@ -24,27 +91,213 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+            cls_obs_line,
+            cls_exp_lines,
+            one_sigma_band,
+            two_sigma_band,
+            test_size_line,
+        )
+    )
+
+
+def plot_cls_components(
```
lukasheinrich, line 167: any reason we could't combine this witth `plot_results` (i.e. just as kwargs whether or not to add the component arttists?)

matthewfeickert: @lukasheinrich @kratsg I think this comes down to what we want the API to be like. For `plot_cls_components` as it is now it requires that the `results` returned by `hypotest` had the `return_expected_set=True, return_tail_probs=True` kwargs so that the CL_s+b and CL_b components can be drawn. If we integrate the `plot_cls_components` functionality into `plot_results` via kwargs then we need to make it clear to users that the kwargs they select also affect the arguments that they're providing.

I can see the benefits of separating out APIs that have different argument requirements and have the kwargs only affect those specific arguments. Though I do understand what you're saying RE: the convenience of not having to change your API to add more context to your visualization, but rather just throw on some kwargs.

I'll take some time today/tonight to play with this a bit and see what the revision looks like.

src/pyhf/contrib/viz/brazil.py
```diff
@@ -1,9 +1,76 @@
...
+            one_sigma_band,
+            two_sigma_band,
+            test_size,
+        ) = brazil_band_artists
+        self.cls_obs = cls_obs
+        self.cls_exp = cls_exp
+        self.one_sigma_band = one_sigma_band
+        self.two_sigma_band = two_sigma_band
+        self.test_size = test_size
+        self.axes = cls_obs.axes
```
henryiii, line 68: I don't recommend this. It's providing a "cute" shortcut but why not just teach users to use `self.cls_obs.axes`? That will work throughout all matplotlib, not just here. If you really need it, or if you can find an example in matplotlib where this is done this way, then it should be a property on BrazilBandContainer, and not stored here.

lukasheinrich: the reason I don't like `artist.cls_obs.axes` is that it provides many ways to do the same thing without a clear canonical choice

e.g. why `artist.cls_obs.axes` and not `artist. one_sigma_band.axes` or `artist.cls_exp.axes`.. providing a canonical way to me seems cleaner and doesn't require knowing/exposing yourself/depending on the details of `BrazilBandContainer`

consider the simple case where I want to to a log scale which has nothinng to do with cls_obs at all

```
artists = make_plot(...)
artists.cls_obs.axes.set_yscale('log')
```

vs

```
artists = make_plot(...)
artists.axes.set_yscale('log')
```

henryiii: Well, if you want more than the basics, I'd probably write:

```python
fig, ax = plt.subplots()
artists = make_plot(..., axes=ax)
ax.set_yscale('log')
```

Now you have access to the figure and the axes, and you can change the size and such when you construct it.

If you are guaranteed that all the Artists are plotted on one Axes only, then I'd at least make this a property.

src/pyhf/contrib/viz/brazil.py
matthewfeickert: @lukasheinrich I think I'm convinced of your idea as adopting this would also make `plot_results` more robust as it could handle more information in the `tests` argument to `plot_results`.

src/pyhf/contrib/viz/brazil.py
lukasheinrich: yeah having it as a immutable @property makes sense to me (indeed it's guaranteed for all of those to be onn the same Axes object). we already have the `ax` argument pattern so maybe this is superfluous.. i'll let @matthewfeickert make the call

matthewfeickert: > I don't recommend this. It's providing a "cute" shortcut but why not just teach users to use `self.cls_obs.axes`?

This was my first thought as well. https://github.com/scikit-hep/pyhf/pull/1377#discussion_r612685478 Someone should correct me if I'm wrong, but this (going back to how it was pre adding `axis` as an attribute) also seems in line with the idea that plotting things on an axis in `maptlotlib` returns the artists and you probably are going to be wanting to be explicitly manipulating an `axis` object that already existed instead of the `axis` object that you can get from the artists. 

That is (hopefully more clear in code) this use of the API

```python
fig, ax = plt.subplots()
artists = pyhf.contrib.viz.brazil.plot_results(ax, poi_vals, results)
ax.set_title("API demonstration")
```

is more "natural" in `matplotlib` thinking than the following

```python
fig, ax = plt.subplots()
artists = pyhf.contrib.viz.brazil.plot_results(ax, poi_vals, results)
artists.cls_obs.axes.set_title("API demonstration")  # just use the axis object you already have available to you
```

I'll revert the changes I made in https://github.com/scikit-hep/pyhf/commit/a3e8ec6206e376df5e6e8566b4f3312f22f3a14b once I finish cleaning up the the API changes and tests to move all of the component plotting into kwargs for `plot_results`.

src/pyhf/contrib/viz/brazil.py
matthewfeickert: @henryiii what are your views on having a `matplotlib.container.Container` that has `None` for some of the artists attributes, like described in the above?

src/pyhf/contrib/viz/brazil.py
henryiii: That's weird, the ax is a required (non-keyword) argument, in fact the first one? Usually, it's an optional keyword-only argument. Such as in pandas, for example. If you don't have one, it pulls from `plt.gca`, which creates an axis and figure as needed. Not providing `ax=` is generally seen as a shortcut, so "easy" access to the axes is not provided, only Artists.

matthewfeickert:
@kratsg @lukasheinrich @henryiii As this PR is getting very long and so hard to follow, I'll make sure that when I request a final review that I'll add an example script that demos the API features that should hopefully make it easier to visually take in the effects as well.

src/pyhf/contrib/viz/brazil.py
lukasheinrich: These  seems natural to me

```
fig, ax = plt.subplots()
container = pyhf.contrib.viz.brazil.plot_results(poi_vals, results, ax = ax) #hypoothetical optional kwarg
ax.set_title("API demonstration")
```

```
fig, ax = plt.subplots()
container = pyhf.contrib.viz.brazil.plot_results(poi_vals, results,) #pull from plt.gca()
container.axes.set_title("API demonstration")
```

```
container = pyhf.contrib.viz.brazil.plot_results(poi_vals, results,) #pull from plt.gca()
plt.gca().set_title("API demonstration")
```

while this seems to depend on unnecessary detail

```
fig, ax = plt.subplots()
container = pyhf.contrib.viz.brazil.plot_results(poi_vals, results,) #pull from plt.gca()
container.cls_obs.axes.set_title("API demonstration")
```

but I guess the point then is to generally not access axes from artists.. but either use `plt.gca()` or a pre-created `ax` you anyway have control over

lukasheinrich: the reason `ax` is a non-defaullt argument is that I have a distaste for importing `plt` / relying on global variables like `plt.gcf()`/`plt.gca()` .. but that's probably just prejudice.. (having `ax` kwarg of course important to be able to plot on subplots)

henryiii: 1 and 3 are natural. Why do you expect the thing that `pyhf.contrib.viz.brazil.plot_results` returns to have a `.axes` property? Many matplotlib functions return lists of Axis, so calling `.axes` on the return value is not something you can always do - it depends on the return type. While the other two will always work. You should mostly consider this a return tuple of axes, that happens to be "nicer" and provide named access. If you provide `.axes`, why not `.figure`? Where does it stop?

And, currently, you _have_ to pass in axis, as the first argument (?), so you can't do anything besides the first example (and not written that way).

src/pyhf/contrib/viz/brazil.py
matthewfeickert: My current plan is to just start factorizing the plotting into smaller functions and then make more container classes: `BrazilBandContainer`, `ClsComponentsContainer`, `ResultsPlotContainer`

src/pyhf/contrib/viz/brazil.py
henryiii: (I'm not horribly against providing the property if you are sure they will always be plotted on the same axis, just arguing from a purity/API simplicity/teaching standpoint)

lukasheinrich: yeah i think this PR does double duty in us rethinking how we want to do plotting, so part of that would be to allow a default `ax` value.   The reason I find it natural to be able to call `.axes`on the return value of a complex plotting function like the above  is that it often works for simplpe plotting functions


```
a = plt.scatter([1,2],[2,3])
a.axes.set_yscale('log')
```

```
a = plt.plot([1,2],[2,3])
a[0].axes.set_yscale('log')
```

so extending from "simple plot" to "complex plot" it seems nice to keep that pattern

matthewfeickert:
Okay, here's a demo of the API at the moment:

```python
import matplotlib
import numpy as np
from matplotlib.figure import Figure

import pyhf

if __name__ == "__main__":
    pyhf.set_backend("numpy")
    )
    observations = [51, 48]
    data = observations + model.config.auxdata
    poi_vals = np.linspace(0, 5, 41)
    results = [
        pyhf.infer.hypotest(
            test_poi, data, model, return_expected_set=True, return_tail_probs=True
        )
        for test_poi in poi_vals
    ]

    _fig_size = matplotlib.rcParams["figure.figsize"]
    scale_factor = 2.2
    fig = Figure(figsize=(scale_factor * _fig_size[0], scale_factor * _fig_size[1]))
    ((ax1, ax2), (ax3, ax4)) = fig.subplots(2, 2)

    pyhf.contrib.viz.brazil.plot_results(poi_vals, results, ax=ax1)
    ax1.set_title("Brazil band")

    pyhf.contrib.viz.brazil.plot_results(poi_vals, results, ax=ax2, components=True)
    ax2.set_title("Brazil band with components")

    pyhf.contrib.viz.brazil.plot_results(
        poi_vals, results, ax=ax3, components=True, no_cls=True
    )
    ax3.set_title("components only")

    ax4.set_yscale("log")
    pyhf.contrib.viz.brazil.plot_results(poi_vals, results, ax=ax4, components=True)
    ax4.set_title("Brazil band with components log y")

    fig.tight_layout()
    fig.savefig("api_example.png")
```

giving

api_example


I think it is also worth reviewing the docs as well: https://pyhf--1377.org.readthedocs.build/en/1377/_generated/pyhf.contrib.viz.brazil.html

README.rst
```diff
@@ -156,7 +156,7 @@ A one bin example
    import pyhf
```
lukasheinrich, line 156: should the images in the README also be updated to the new ones (w/ legends)

matthewfeickert: yup! They are.

```console
$ git diff origin/master --name-only | grep static
docs/_static/img/README_1bin_example.png
docs/_static/img/README_2bin_example.png
```

c.f. the "rich diff" view on the images in the "Files changed" tab to see the difference rendered out.

lukasheinrich: yeah I see thoses, but looking at

https://github.com/scikit-hep/pyhf/blob/7ec11ae45ca1053b27ca23ac2069ec36a65d3ba2/README.rst

I don't see it

matthewfeickert: Oh yeah, that's because we have it pinned to the `master` branch files in the README so that it shows up properly other places on the web (I think that's the reason at least).

```console
$ grep README_1bin_example README.rst 
.. image:: https://raw.githubusercontent.com/scikit-hep/pyhf/master/docs/_static/img/README_1bin_example.png
```

src/pyhf/contrib/viz/brazil.py
matthewfeickert: After looking at the docs I _think_ the API argument for no `axes` attribute on containers is the following:

[`scatter()`](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.scatter.html) returns a [`matplotlib.collections.PathCollection`](https://matplotlib.org/stable/api/collections_api.html#matplotlib-collections) 

```python
>>> import matplotlib.pyplot as plt
>>> fig, ax = plt.subplots()
>>> _scatter = ax.scatter([1, 2], [2, 3])
>>> type(_scatter)
<class 'matplotlib.collections.PathCollection'>
```

which I see from the inheritance diagram is descended from `Artist`.

[inheritance](https://matplotlib.org/stable/api/collections_api.html#matplotlib-collections)

So as all `artists` have `axes` attributes, then so does `collection`. [`plot()`](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.plot.html) returns a list of [`matplotlib.lines.Line2D`](https://matplotlib.org/stable/api/_as_gen/matplotlib.lines.Line2D.html#matplotlib.lines.Line2D) artists.

```python
>>> import matplotlib.pyplot as plt
>>> fig, ax = plt.subplots()
>>> _plot = ax.plot([1, 2], [2, 3])
>>> _plot
[<matplotlib.lines.Line2D object at 0x7f8dcd04dbe0>]
```

I don't understand _why_ `scatter` doesn't return a collection, but that's I guess a different story.

I guess my point from all that rambling is that both the returns of `scatter` and `plot` don't return [`matplotlib.container`s](https://matplotlib.org/stable/api/container_api.html), which don't have `axes` as they are groupings of `artists` that have `axes`.

src/pyhf/contrib/viz/brazil.py
matthewfeickert: To summarize some problems we've had with trying to reduce things down to fewer containers as Lukas was hoping for:

I don't think we want to do something like

```python
class BrazilBandContainer(Container):
    def __init__(self, brazil_band_artists, clsb=None, clb=None, **kwargs):
        (
            cls_obs,
            cls_exp,
            one_sigma_band,
            two_sigma_band,
            test_size,
        ) = brazil_band_artists
        self.cls_obs = cls_obs
        self.cls_exp = cls_exp
        self.one_sigma_band = one_sigma_band
        self.two_sigma_band = two_sigma_band
        self.test_size = test_size
        self.clsb = clsb
        self.clb = clb
        super().__init__(brazil_band_artists, **kwargs)
```

as since `clsb` and `clb` do not end up in the `namedtuple`-like container (but just as attributes) and so you have them as strangely tacked on artists that are kept separate from the rest of the artists. I also don't think that you want to go with

```python
class BrazilBandContainer(Container):
    def __init__(self, brazil_band_artists **kwargs):
        (
            cls_obs,
            cls_exp,
            one_sigma_band,
            two_sigma_band,
            test_size,
            clsb,
            clb
        ) = brazil_band_artists
        self.cls_obs = cls_obs
        self.cls_exp = cls_exp
        self.one_sigma_band = one_sigma_band
        self.two_sigma_band = two_sigma_band
        self.test_size = test_size
        self.clsb = clsb
        self.clb = clb
        super().__init__(brazil_band_artists, **kwargs)
```

as then in `plot_brazil_band()` you're returning

```python
    return BrazilBandContainer(
        (
            cls_obs_line,
            cls_exp_lines,
            one_sigma_band,
            two_sigma_band,
            test_size_line,
            None,
            None,
        )
    )
```

which is a bit awkward to be forced to tack on the `None`s, and then you need to unpack and repack in `plot_results()` to combine this with a `ClsComponentsContainer` if you want to return a `BrazilBandContainer`, which feels like forcing things into a container that don't need to be.

I'm not sure what's best to do, but It feels like this is trying to combine too many things into a single container, as the [`container` classes in `matplotlib`](https://matplotlib.org/stable/api/container_api.html) all only are used for one group of artists plotted on a subplot.

matthewfeickert:
@lukasheinrich @kratsg for easier reference, the comment from the container discussion is here: https://github.com/scikit-hep/pyhf/pull/1377#discussion_r617942205

matthewfeickert:
> @lukasheinrich @kratsg for easier reference, the comment from the container discussion is here: [#1377 (comment)](https://github.com/scikit-hep/pyhf/pull/1377#discussion_r617942205)

Following up a discussion with Thomas Caswell, this is his advice:

> Thomas: 
> I would not sub-class `matplotlib.container.Container`. If you want to go down the custom composite object route (a-la grave https://github.com/networkx/grave/blob/19aec1eacb2bd3577fd2660f6199dcab6f8d97c5/grave/grave.py#L221-L242) the go that route, implement draw that passes down to it's children. You may want to add `remove` on your named tuple as a helper. One reason you might want to sub-class `Container` is to toss your containers into `Axes.containers` which mean it will be offered to the legend machinery as a complex and if you custom handlers there you can get nice very custom legend entries. Duck-typing to `Container` is probably enough.
>
> @lukasheinrich: 
> yeah `Container` seemed correct as it's advertised as a "semantic set of artists", but we can leave it out.
>
> Thomas:
> I agree that is the goal of `Container`, but it is not widely used in Matplotlib and the sometimes second view of the artists is....odd. It is good idea, but it did not get pushed far enough when it went in.
>
> So in summary:
> * If you have a Collection of artists that are all dirven by the same data and it is likely you would want to update them en-mass go with the grave solution (and register a custom legend handler)
> * If you have a bunch of stuff you just want "stuck together" return a tuple
> * f you want them stuck together and to have a shared legend entry duck-type `Container`, use `ax.add_container`, and register a custom legend handler

@kratsg  @lukasheinrich So it seems from Thomas's advice that we can either go for `namedtuples`, which should be enough, or we can do what `grave` did and make a `BrazilBandArtist` subclass of [`matplotlib.artist.Artist`](https://matplotlib.org/stable/api/artist_api.html#matplotlib.artist.Artist).

src/pyhf/contrib/viz/brazil.py
```diff
@@ -1,13 +1,392 @@
...
+import matplotlib.pyplot as plt
 import numpy as np
+from matplotlib.artist import Artist
+from matplotlib.container import Container
 
 
-def plot_results(ax, mutests, tests, test_size=0.05):
+class BrazilBandArtist(Artist):
+    r"""
+    TODO: Add doctring
```
matthewfeickert, line 10: Need to revise and clean docstrings.

src/pyhf/contrib/viz/brazil.py
```diff
@@ -18,33 +317,114 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+    if not no_cls:
+        brazil_band_artist = plot_brazil_band(
+            test_pois, cls_obs, cls_exp, test_size, ax, **kwargs
         )
-    ax.fill_between(mutests, cls_exp[0], cls_exp[-1], facecolor='yellow')
-    ax.fill_between(mutests, cls_exp[1], cls_exp[-2], facecolor='green')
-    ax.plot(mutests, [test_size] * len(mutests), c='red')
-    ax.set_ylim(0, 1)
+    else:
+        brazil_band_artist = BrazilBandArtist(None, None, None, None, None)
```
matthewfeickert, line 401: Not elegant, but it works.

lukasheinrich: can't those ctor args be defaulted to None?

src/pyhf/contrib/viz/brazil.py
```diff
@@ -1,13 +1,314 @@
...
+
+        Args:
+            renderer: :class:`matplotlib.backend_bases.RendererBase` subclass.
+        """
+        if not self.get_visible():
+            return
+
+        for artist in self.get_children():
+            artist.draw(renderer, *args, **kwargs)
+        self.stale = False
```
kratsg, line 114: what's the point of stale?

kratsg (approved):

src/pyhf/contrib/viz/brazil.py
matthewfeickert: Used by `matplotlib` for some state bookkeeping for artists:

> Whether the artist is 'stale' and needs to be re-drawn for the output to match the internal state of the artist.

https://matplotlib.org/stable/api/_as_gen/matplotlib.artist.Artist.stale.html

src/pyhf/contrib/viz/brazil.py
matthewfeickert: @lukasheinrich Set them to `None` by default in the `BrazilBandArtist`? I don't think we want to do that in general. The only reason I'm doing it here is to have a hacky way to get a `BrazilBandArtist` object to return later that we can then tack on the `clsb` and `clb` lines onto to get out.

This could raise the question of if we even want to have the `no_cls` kwarg ability to have a `BrailBandArtist` that doesn't have the Brazil band, and that if people just want the components then they should just call `plot_cls_components` and do the extra work seen in the example to get the plot.

lukasheinrich: isn't it bettter than to waitt witth constrtruction untiil we have all the components.. 

i.e. accumulate artists in a dict and just bbefore return do 

`return BrazilBandArtist(**artist_dict)```

matthewfeickert: Then you'd have to unpack and repack all the artists from above

```python
    if not no_cls:
        brazil_band_artist = plot_brazil_band(
            test_pois, cls_obs, cls_exp, test_size, ax, **kwargs
        )
```

src/pyhf/contrib/viz/brazil.py
```diff
@@ -1,13 +1,255 @@
...
+
+import matplotlib
+import matplotlib.pyplot as plt
 import numpy as np
 
 
-def plot_results(ax, mutests, tests, test_size=0.05):
+class BrazilBandCollection(NamedTuple):
+    r"""
+    Tuple containing the :class:`matplotlib.artist.Artist` objects of the
```
matthewfeickert, line 11: The docs are breaking when subclassing `typing.NamedTuple` and I'm not really sure why. @kratsg if you have time later to give thoughts here that could be useful.

matthewfeickert: I gave up and used `collections.namedtuple`.

src/pyhf/contrib/viz/brazil.py
matthewfeickert: This should now be resolved I think

src/pyhf/contrib/viz/brazil.py
```diff
@@ -18,33 +279,115 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+            test_pois, cls_obs, cls_exp, test_size, ax, **kwargs
         )
-    ax.fill_between(mutests, cls_exp[0], cls_exp[-1], facecolor='yellow')
-    ax.fill_between(mutests, cls_exp[1], cls_exp[-2], facecolor='green')
-    ax.plot(mutests, [test_size] * len(mutests), c='red')
-    ax.set_ylim(0, 1)
+    else:
+        # TODO: Find more elegant solution
+        brazil_band_collection = BrazilBandCollection()
+
```
lukasheinrich, line 365: perhaps the more elegant solutiion would be to make `plot_cls_components` and `plot_brazil_band` behave similarly and have `plot_brazil_band` also return a list of artists

```
brazil_band_artists = plot_brazil_band(...)
component_artists = plot_cls_components(test_pois, tail_probs, ax, **kwargs)
BrazilBandCollection(*(brazil_band_artists + component_artists))
```

src/pyhf/contrib/viz/brazil.py
```diff
@@ -18,33 +279,115 @@ def plot_results(ax, mutests, tests, test_size=0.05):
...
+          :math:`\mathrm{CL}_{s,\mathrm{exp}}` band.
+        test_size (:obj:`float`): The size, :math:`\alpha`, of the test.
+        ax (:obj:`matplotlib.axes.Axes`): The matplotlib axis object to plot on.
+
+    Returns:
+        :class:`BrazilBandArtist`: Artist containing the :obj:`matplotlib.artist`
+        objects drawn.
     """
+    if ax is None:
+        ax = plt.gca()
```
lukasheinrich, line 335: given that at this point we *know* for sure what axes all the artists will be attached on I feel we could skip the logic in the `BrazilBandCollection` and just pass the axes as an ctor argument instead of recovering it from the artists

src/pyhf/contrib/viz/brazil.py
matthewfeickert: The control flow is still there with the changes I added (can't really get away from that), but I agree that using generators makes it look more Pythonic and readable. :+1:

src/pyhf/contrib/viz/brazil.py
matthewfeickert: Hm, let me think on this a bit. 

> given that at this point we know for sure what axes all the artists will be attached on

This is true inside the plotting functions, but as the `BrazilBandCollection` is in the public API there's nothing keeping someone from trying to do something confusing and wrong like we test against now in `test_brazil_band_collection`.

> just pass the axes as an ctor argument instead of recovering it from the artists

This would get rid of the problem of assigning artists from multiple axes to the same `BrazilBandCollection`, but then you now have a namedtuple that contains both the artists as well as the `axes.Axes` object they're on.

Though at the same time, that API for all the plotting functions basically are expecting this sort of info so maybe this is pretty reasonable.

matthewfeickert: Okay, convinced. :)

lukasheinrich (approved):

lukasheinrich:
Thanks @matthewfeickert  - this looks great!

matthewfeickert:
This PR had lots of discussion on it (thanks to all involved) so for future people looking, the relevant example plot is in a comment here: https://github.com/scikit-hep/pyhf/pull/1377#issuecomment-824172523

PR #1382, opened 2021-03-26
build: Set only lower bounds on core dependencies

matthewfeickert:
# Description

Following @henryiii's and @hynek's advice (for the core dependencies now, we'll get to discussing the ones for the `setup.py` extras later) only place lower bounds on the core library dependencies, which have all reached maturity in their API.

This was more recently motivated by the fact that Henry pointed out that we'll be excluding use of `Click` `v8.0.0` (when it is released, currently [only `v8.0.0a1` on PyPI](https://pypi.org/project/click/8.0.0a1/#history)), but there's no reason to want to bound Click to only `v7.X` or `v8.X`.

Thanks also to a very good comment from @tacaswell this also reverts the effects of PR #1378:

> I think this is another library vs application difference.  Your users are adults who are responsible for keeping their own house in order :wink: It is not a random libraries job to keep their systems up-to-date and secure (that falls to the end users and the packagers).
>
> I think this is also a good case for why libraries should be as loose as possible with their version pinning as say other library was being super picky.  Say some other library was being strict to some version range of `pyyaml < 5.4` (for whatever reasons).  Bumping your minimum to `>= 5.4` with good intentions just drove someone to have unsatisfyable constraints.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Place only lower bounds (>=) on all core dependencies in setup.cfg
* Update lower bounds for core dependencies
   - Lower bounds established through empirical tests in CI with minimum supported dependencies workflow
* Effectively reverts PR #1378
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI/CD
 
 on:
-  push:
+  # push:
   pull_request:
   # Run daily at 0:01 UTC
   schedule:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI/CD
 
 on:
-  # push:
+  push:
   pull_request:
   # Run daily at 0:01 UTC
   schedule:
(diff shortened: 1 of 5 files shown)


matthewfeickert:
This is still a Draft PR as I want to go through and test if these lower bounds that we have can be safely pushed down any lower.

kratsg (approved):

PR #1388, opened 2021-03-31
fix: Stricter schema definitions for histosys and normsys

### LINKED ISSUE #1387: normsys/data and histosys/data allow additional properties, is this intended?
In the current schema definitions, the `data` properties in [`normsys`](https://github.com/scikit-hep/pyhf/blob/98bb222d9c45ad13335ab5423d866e3a1508c329/src/pyhf/schemas/1.0.0/defs.json#L141) and [`histosys`](https://github.com/scikit-hep/pyhf/blob/98bb222d9c45ad13335ab5423d866e3a1508c329/src/pyhf/schemas/1.0.0/defs.json#L104)
are of type object, but they do not specify the `additionalProperties` property. So the jsonschema default is used, meaning any and all additional properties are valid. 

Is this intended? All other defintions (outside the patchset schemata) "close" the schema by setting `"additionalProperties": false`, and that seems to me to match the expectation from HF.

kratsg:
# Pull Request Description

Resolves #1387. Histosys and Normsys should be stricter about not allowing additional properties in the `data` object in the schema.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Teach the schema to not allow additional properties for histosys and normsys
* Be more explicit about the patchset/jsonpatch metadata information (that we do allow additionalProperties)
* Add tests to check that additionalProperties for for histosys and normsys result in an invalid spec
```

### DIFF
--- a/src/pyhf/schemas/1.0.0/defs.json
+++ b/src/pyhf/schemas/1.0.0/defs.json
@@ -107,7 +107,8 @@
                             "lo_data": { "type": "array", "items": {"type": "number"}, "minItems": 1 },
                             "hi_data": { "type": "array", "items": {"type": "number"}, "minItems": 1 }
                         },
-                        "required": ["lo_data", "hi_data"]
+                        "required": ["lo_data", "hi_data"],
+                        "additionalProperties": false
                     }
                 },
                 "required": ["name", "type", "data"],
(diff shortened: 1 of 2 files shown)


matthewfeickert (approved):
LGTM. Thanks @kratsg.

Issue #1398, opened 2021-04-05
Uniform default precision for all backends?


matthewfeickert:
# Question

As seen [in the example](https://pyhf.readthedocs.io/en/v0.6.1/_generated/pyhf.set_backend.html) for `pyhf.set_backend` the default precision for the `numpy` backend is `64b` while the default precision for TensorFlow and PyTorch is `32b`. 

https://github.com/scikit-hep/pyhf/blob/98bb222d9c45ad13335ab5423d866e3a1508c329/src/pyhf/tensor/numpy_backend.py#L44

https://github.com/scikit-hep/pyhf/blob/98bb222d9c45ad13335ab5423d866e3a1508c329/src/pyhf/tensor/jax_backend.py#L58

https://github.com/scikit-hep/pyhf/blob/98bb222d9c45ad13335ab5423d866e3a1508c329/src/pyhf/tensor/tensorflow_backend.py#L16

https://github.com/scikit-hep/pyhf/blob/98bb222d9c45ad13335ab5423d866e3a1508c329/src/pyhf/tensor/pytorch_backend.py#L18

Given that there are known differences between `64b` and `32b` it is probably worth reconsidering / discussing if we should have a default precision of `64b` for all backends. My guess is that it is non-intuitive that switching from NumPy to Torch changes the precision.

This is somewhat related to Issue #981 and was put back into my mind by @dguest and @nhartman94 (c.f. Discussion #1397) (comments welcome from them both).

matthewfeickert:
If @lukasheinrich and @kratsg agree here I have a branch with the changes ready for a PR (after PR #1399 goes in).

lukasheinrich:
yeah I agree!

kratsg:
I agree. I think we had 32b at first since that was nominally the primary support when we first started with these backends. I see no reason why we shouldn't harmonize the defaults.

matthewfeickert:
> I agree. I think we had 32b at first since that was nominally the primary support when we first started with these backends.

Yeah this was [my memory too](https://github.com/scikit-hep/pyhf/discussions/1397#discussioncomment-571343).

Issue #1352, opened 2021-03-05
Revise API expectations to enforce Tensor inputs to distributions


matthewfeickert:
In PR #1351 we needed to add `self.astensor` calls back into the PyTorch backend, which we worked hard to remove in PR #582. This is the result of the public API expecting that you can pass floats and not tensors to the distribution methods and still have things work.

We should change this expectation in the API and docs to make it clear that all inputs must be tensors, and then remove the `self.astensor` calls.

_Originally posted by @matthewfeickert in https://github.com/scikit-hep/pyhf/pull/1351#r588362021_

matthewfeickert:
@kratsg rightly points out in PR #1399 that

> We need to check that the fit times for tflow don't change.

given the `self.astensor` calls that were added to the TensorFlow backend there as well.

PR #1351, opened 2021-03-05
fix: Retain continuous approximation of Poisson in torch v1.8.0

matthewfeickert:
# Description

Resolves #1348 

To retain the continuous approximation of the Poisson, set `validate_args=False` in instantiations of the `torch.distributions.Poisson` class.

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Add validate_args=False to instantiations of the torch.distributions.Poisson class in PyTorch backend to allow for the continuous approximation of the Poisson distribution in torch v1.8+
* Ensure tensor value arguments for torch.distributions.Normal
* Check that ValueErrors are raised by PyTorch v1.8+ when the validations of validate_args are violated
```

PR #1399, opened 2021-04-06
fix: Ensure tensorlib precision used throughout TensorFlow backend

matthewfeickert:
# Description

While investigating Issue #1398, it was noticed that for `64b` precision for the TensorFlow backed the precision was not properly being used throughout. `tf.float32` was used as a default instead of the result of `self.dtypemap["float"]` and if TensorFlow Probability is given floats as parameters it will default to using `32b` precision, and so _requires_ that a tensor of the desired precision be given.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Ensure that the precision requested in the set_backend API is used throughout the TensorFlow backend calculations
   - Determine dtype from dtypemap
   - Ensure TensorFlow Probability operations are given a tensor of the correct precision
```

### DIFF
--- a/src/pyhf/tensor/tensorflow_backend.py
+++ b/src/pyhf/tensor/tensorflow_backend.py
@@ -160,15 +160,12 @@ class tensorflow_backend:
             raise
 
     def outer(self, tensor_in_1, tensor_in_2):
+        dtype = self.dtypemap["float"]
         tensor_in_1 = (
-            tensor_in_1
-            if tensor_in_1.dtype != tf.bool
-            else tf.cast(tensor_in_1, tf.float32)
+            tensor_in_1 if tensor_in_1.dtype != tf.bool else tf.cast(tensor_in_1, dtype)
         )
         tensor_in_1 = (
-            tensor_in_1
-            if tensor_in_2.dtype != tf.bool
-            else tf.cast(tensor_in_2, tf.float32)
+            tensor_in_1 if tensor_in_2.dtype != tf.bool else tf.cast(tensor_in_2, dtype)
         )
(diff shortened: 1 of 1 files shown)


kratsg (approved):

kratsg (changes requested):
I don't think we should be adding so many `self.astensor` calls, should we?

matthewfeickert:
> I don't think we should be adding so many `self.astensor` calls, should we?

What do you propose instead? We had to do something similar in PR #1351, and at the moment on `master` this is a problem as setting the precision to `64b` doesn't propagate to TensorFlow Probability operations if they're given floats.

```python
>>> import pyhf
>>> pyhf.set_backend("tensorflow", precision="64b")
>>> pyhf.tensorlib.precision
'64b'
>>> pyhf.tensorlib.normal_logpdf(0.5, 0., 1.)
<tf.Tensor: shape=(), dtype=float32, numpy=-1.0439385>
>>> pyhf.tensorlib.normal(0.5, 0., 1.)
<tf.Tensor: shape=(), dtype=float32, numpy=0.35206532>
```

matthewfeickert:
We also made Issue #1352 based off of these [comments on PR #1351](https://github.com/scikit-hep/pyhf/pull/1351/files#r588118675).

kratsg (approved):
Yeah, this is the cleanest way forward for now. We need to check that the fit times for tflow don't change.

matthewfeickert:
> We need to check that the fit times for tflow don't change.

Yeah, from looking at the CI times there isn't any obvious uptick, but we should note this in the Issue as well.

PR #1400, opened 2021-04-06
refactor: Set default precision to 64b

### LINKED ISSUE #1398: Uniform default precision for all backends?

matthewfeickert:
# Description

Requires PR #1399 to go in first, as this PR relies on those changes.

Resolves #1398 

Set the default precision for all backends to `64b`.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Set the default precision for all tensorlib backends to 64b in pyhf.set_backend
* Update the TensorFlow and PyTorch docstrings
```

### DIFF
--- a/src/pyhf/tensor/tensorflow_backend.py
+++ b/src/pyhf/tensor/tensorflow_backend.py
@@ -408,8 +408,8 @@ class tensorflow_backend:
             >>> import pyhf
             >>> pyhf.set_backend("tensorflow")
             >>> t = pyhf.tensorlib.poisson_logpdf(5., 6.)
-            >>> print(t)
-            tf.Tensor(-1.8286943966417715, shape=(), dtype=float64)
+            >>> print(t) # doctest:+ELLIPSIS
+            tf.Tensor(-1.8286943966417..., shape=(), dtype=float64)
             >>> values = pyhf.tensorlib.astensor([5., 9.])
             >>> rates = pyhf.tensorlib.astensor([6., 8.])
             >>> t = pyhf.tensorlib.poisson_logpdf(values, rates)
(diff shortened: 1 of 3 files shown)


src/pyhf/tensor/tensorflow_backend.py
```diff
@@ -408,13 +408,13 @@ def poisson_logpdf(self, n, lam):
             >>> import pyhf
             >>> pyhf.set_backend("tensorflow")
             >>> t = pyhf.tensorlib.poisson_logpdf(5., 6.)
-            >>> print(t)
-            tf.Tensor(-1.8286943, shape=(), dtype=float32)
+            >>> print(t) # doctest:+ELLIPSIS
```
matthewfeickert, line 411: ELLIPSIS needed as Linux and MacOS CI system have different floating point on the TF repr print.

kratsg (approved):

PR #1409, opened 2021-04-13
fix: check POIs floating

### LINKED ISSUE #1391: workspaces with fixed poi fail in hypotest (of course)
# Description

some workspaces (like the one @elskorda provided) have a *fixed* POI in the workspace definition - if you naively put that through `pyhf cls` of course the fit doesn't really do anythinig useful.

options

* in readxml warn that POI is fixed
* in hypotest ensurer to un-fix POI
* others?

the least we could do is throw a warning  in the places where an unconstrained fit iss expected but a fixed POI is observed

lukasheinrich:
# Description

This fixes #1391 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add all_pois_floating to infer.utils to check if POIs are fixed
* Add check in hypotest that POIs are not fixed
* Add test that hypotest raises an exception if a fixed POI is used in the given model
```

### DIFF
--- a/src/pyhf/infer/__init__.py
+++ b/src/pyhf/infer/__init__.py
@@ -4,6 +4,7 @@ from . import utils
 from .. import get_backend
 from .. import exceptions
 
+
 def hypotest(
     poi_test,
     data,
@@ -131,8 +132,10 @@ def hypotest(
     par_bounds = par_bounds or pdf.config.suggested_bounds()
     fixed_params = fixed_params or pdf.config.suggested_fixed()
 
-    if not utils.all_pois_floating(pdf,fixed_params):
-        raise exceptions.InvalidModel(f'POI at index [{pdf.config.poi_index}] forced to be fixed, cannot run inference')
+    if not utils.all_pois_floating(pdf, fixed_params):
+        raise exceptions.InvalidModel(
+            f'POI at index [{pdf.config.poi_index}] forced to be fixed, cannot run inference'
+        )
 
     calc = utils.create_calculator(
         calctype,
--- a/src/pyhf/infer/__init__.py
+++ b/src/pyhf/infer/__init__.py
@@ -2,7 +2,7 @@
 
 from . import utils
 from .. import get_backend
-
+from .. import exceptions
 
 def hypotest(
     poi_test,
@@ -130,6 +130,9 @@ def hypotest(
     init_pars = init_pars or pdf.config.suggested_init()
     par_bounds = par_bounds or pdf.config.suggested_bounds()
     fixed_params = fixed_params or pdf.config.suggested_fixed()
+
+    if not utils.all_pois_floating(pdf,fixed_params):
+        raise exceptions.InvalidModel(f'POI at index [{pdf.config.poi_index}] forced to be fixed, cannot run inference')
 
     calc = utils.create_calculator(
         calctype,
--- a/src/pyhf/infer/__init__.py
+++ b/src/pyhf/infer/__init__.py
@@ -132,6 +132,11 @@ def hypotest(
     par_bounds = par_bounds or pdf.config.suggested_bounds()
     fixed_params = fixed_params or pdf.config.suggested_fixed()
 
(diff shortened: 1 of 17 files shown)


src/pyhf/infer/utils.py
```diff
@@ -9,6 +9,22 @@
 log = logging.getLogger(__name__)
 
 
+def all_pois_floating(pdf, fixed_params):
+    """
+    Checks whether all POI(s) are floatig (i.e. not within the fixed set).
+
+    Args:
+        pdf (:obj:`pyhf.Model`): The model
```
matthewfeickert, line 17:
```suggestion
        pdf (~pyhf.pdf.Model): The statistical model adhering to the schema ``model.json``.
```
Would be good to match what we have elsewhere.

src/pyhf/infer/utils.py
```diff
@@ -9,6 +9,22 @@
...
+        fixed_params: array of bools
```
matthewfeickert, line 18:
```suggestion
        fixed_params (:obj:`list` or `tensor` of :obj:`bool`): Array of :obj:`bool`s indicating if model parameters are fixed.
```

src/pyhf/infer/utils.py
```diff
@@ -9,6 +9,22 @@
...
+
+    Returns:
+        bool: The result whether all POIs are floating.
```
matthewfeickert, line 21:
```suggestion
        :obj:`bool`: The result whether all POIs are floating.
```

Giordon started defaulting to using `obj` whenever possible, and I think that's a good habit to continue.

src/pyhf/infer/__init__.py
```diff
@@ -131,6 +132,16 @@ def hypotest(
...
     fixed_params = fixed_params or pdf.config.suggested_fixed()
 
+    if pdf.config.poi_index is None:
+        raise exceptions.UnspecifiedPOI(
+            'No POI is defined. A POI is required to run a hypothesis test.'
+        )
+
+    if not utils.all_pois_floating(pdf, fixed_params):
+        raise exceptions.InvalidModel(
+            f'POI at index [{pdf.config.poi_index}] forced to be fixed, cannot run inference'
```
matthewfeickert, line 142:
```suggestion
            f'POI at index [{pdf.config.poi_index}] is set as fixed, which makes inference impossible. Please unfix the POI to continue.'
```

Maybe we should be super explicit? There's probably some cleaner/clearer way then what this suggestion has, but I'm just trying to indicate that we should probably be overly verbose in error messages like this.

matthewfeickert (changes requested):
Thanks @lukasheinrich. This looks good overall :+1:, and I don't have any code change suggestions, just docs and exception related suggestions.

src/pyhf/infer/utils.py
```diff
@@ -9,6 +9,22 @@
 log = logging.getLogger(__name__)
 
 
+def all_pois_floating(pdf, fixed_params):
+    """
+    Checks whether all POI(s) are floatig (i.e. not within the fixed set).
```
matthewfeickert, line 14:
```suggestion
    Checks whether all POI(s) are floating (i.e. not within the fixed set).
```

matthewfeickert (changes requested):
@lukasheinrich this all LGTM, but I just realized while looking at the RTD build that this isn't getting captured in the docs. Can you add `all_pois_floating` under the Utilities section in  `docs/api.rst`?

kratsg (approved):
One note: `utils.all_pois_floating` isn't written in a generic way. Something like `not all(poi_fixed)` or similar would be best. But for now, as we only consider the case of a single POI, this is fine.

lukasheinrich:
we don't have yet the infrastructure in `_ModelConfig` for multi-POI but I purposefully names the method in a way that it's genneric so we only need to change the impl once were there

matthewfeickert (approved):
LGTM. Thanks @lukasheinrich

PR #1412, opened 2021-04-20
feat: `pyhf patchset inspect`

lukasheinrich:
# Description

Convenience CLI to figurer out hwat's going on in patchsets


matthewfeickert:
Love it. :)

kratsg (approved):

kratsg:
Nice. I didn't suspect this would be overtly complicated or verbose. Probably want to dump the metadata on some level -- like the description + the ntuples of the point values, rather than the names? Unsure. Usually people look at a patchset to see what points are available.

lukasheinrich:
yeah I wanted to make a workflow possible of 

```
pyhf patchset inspect
pyhf patchset extract --name ...
```

is `p.name` different than what `--name` takes?

kratsg:
> is `p.name` different than what `--name` takes?

No. It's the same. All patchsets are looked up using `PatchSet::__getitem__`

```python
>>> patchset['patch_name_for_2100x_800y']
<pyhf.patchset.Patch object 'patch_name_for_2100x_800y(2100, 800)' at 0x...>
>>> patchset[(2100,800)]
<pyhf.patchset.Patch object 'patch_name_for_2100x_800y(2100, 800)' at 0x...>
```

For the CLI -- only the string version is supported, since we're not doing any fancy introspection to figure out how to coerce what the user specifies. And that tuple doesn't necessarily need to be a fixed number of items or only floats, it's a bit generic and analysis-dependent, but at least the API is the same.

PR #1435, opened 2021-05-07
feat: Add correlated_background API to simplemodels

### LINKED ISSUE #1434: Correlated background simplemodel
something like

```python
def simplemodel2(s,b_up,b_nom,b_dn):
    spec = {
        'channels': [
            {
                'name': 'singlechannel',
                'samples': [
                    {
                    'name': 'signal',
                    'data': s,
                    'modifiers': [{'name': 'mu', 'type': 'normfactor', 'data': None}]
                    },
                    {'name': 'background',
                    'data': b_nom,
                    'modifiers': [
                        {
                            'name': 'uncorr_bkguncrt',
                            'type': 'histosys',
                            'data': {
                                'hi_data': b_up,
                                'lo_data': b_dn
                            }
                        }
                        ]
                    }
                ]
            }
        ]
    }
    return pyhf.Model(spec)

```

with an API like `pyhf.simplemodels.correlated_bkg`

matthewfeickert:
# Description

Resolves #1434

Add a simple model with a single channel with a correlated background uncertainty.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add simplemodels.correlated_background API
   - Single channel model with a two component correlated background uncertainty
* Add tests for simplemodels.correlated_background API
```

### DIFF
--- a/src/pyhf/simplemodels.py
+++ b/src/pyhf/simplemodels.py
@@ -70,11 +70,51 @@ def hepdata_like(signal_data, bkg_data, bkg_uncerts, batch_size=None):
     return Model(spec, batch_size=batch_size)
 
 
-def correlated_background(signal, bkg_up, bkg_nominal, bkg_down, batch_size=None):
+def correlated_background(signal, bkg, bkg_up, bkg_down, batch_size=None):
+    r"""
+    Construct a simple single channel :class:`~pyhf.pdf.Model` with a
+    :class:`~pyhf.modifiers.histosys` modifier representing a correlated
+    background uncertainty.
+
+    Args:
+        signal (:obj:`list`): The data in the signal sample.
+        bkg (:obj:`list`): The data in the background sample.
+        bkg_up (:obj:`list`): The background sample under a :math:`+1\sigma` variation.
+        bkg_down (:obj:`list`): The background sample under a :math:`-1\sigma` variation.
+        batch_size (:obj:`None` or :obj:`int`): Number of simultaneous (batched) Models to compute.
+
+    Returns:
+        ~pyhf.pdf.Model: The statistical model adhering to the :obj:`model.json` schema.
+
+    Example:
+        >>> import pyhf
+        >>> pyhf.set_backend("numpy")
+        >>> model = pyhf.simplemodels.correlated_background(
+        ...     signal=[12.0, 11.0],
+        ...     bkg=[50.0, 52.0],
+        ...     bkg_up=[53.0, 59.0],
+        ...     bkg_down=[45.0, 49.0],
+        ... )
+        >>> model.schema
+        'model.json'
+        >>> model.config.channels
+        ['single_channel']
+        >>> model.config.samples
+        ['background', 'signal']
+        >>> model.config.parameters
+        ['correlated_bkg_uncertainty', 'mu']
+        >>> model.config.poi_index
+        0
+        >>> model.config.par_order
+        ['mu', 'correlated_bkg_uncertainty']
+        >>> model.expected_data(model.config.suggested_init())
+        array([62., 63.,  0.])
+
+    """
     spec = {
         "channels": [
             {
-                "name": "singlechannel",
+                "name": "single_channel",
                 "samples": [
                     {
                         "name": "signal",
@@ -85,7 +125,7 @@ def correlated_background(signal, bkg_up, bkg_nominal, bkg_down, batch_size=None
                     },
                     {
                         "name": "background",
-                        "data": bkg_nominal,
+                        "data": bkg,
                         "modifiers": [
                             {
                                 "name": "correlated_bkg_uncertainty",
--- a/src/pyhf/simplemodels.py
+++ b/src/pyhf/simplemodels.py
@@ -103,10 +103,6 @@ def correlated_background(signal, bkg, bkg_up, bkg_down, batch_size=None):
         ['background', 'signal']
         >>> model.config.parameters
         ['correlated_bkg_uncertainty', 'mu']
-        >>> model.config.poi_index
-        0
-        >>> model.config.par_order
-        ['mu', 'correlated_bkg_uncertainty']
         >>> model.expected_data(model.config.suggested_init())
         array([62., 63.,  0.])
 
--- a/src/pyhf/simplemodels.py
+++ b/src/pyhf/simplemodels.py
@@ -68,3 +68,34 @@ def hepdata_like(signal_data, bkg_data, bkg_uncerts, batch_size=None):
         ]
     }
     return Model(spec, batch_size=batch_size)
+
+
+def correlated_background(signal, bkg_up, bkg_nominal, bkg_down, batch_size=None):
+    spec = {
+        "channels": [
+            {
+                "name": "singlechannel",
+                "samples": [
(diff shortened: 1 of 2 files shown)


src/pyhf/simplemodels.py
```diff
@@ -68,3 +68,70 @@ def hepdata_like(signal_data, bkg_data, bkg_uncerts, batch_size=None):
         ]
     }
     return Model(spec, batch_size=batch_size)
+
+
+def correlated_background(signal, bkg, bkg_up, bkg_down, batch_size=None):
+    r"""
+    Construct a simple single channel :class:`~pyhf.pdf.Model` with a
+    :class:`~pyhf.modifiers.histosys` modifier representing a two component
+    correlated background uncertainty.
```
lukasheinrich, line 77: maybe `representing a background with a fully correlated bin-by-bin correlation`

matthewfeickert: "fully correlated bin-by-bin correlation" -> "fully correlated bin-by-bin uncertainty"?

lukasheinrich: yes :)

src/pyhf/simplemodels.py
```diff
@@ -68,3 +68,70 @@ def hepdata_like(signal_data, bkg_data, bkg_uncerts, batch_size=None):
...
+        ~pyhf.pdf.Model: The statistical model adhering to the :obj:`model.json` schema.
+
+    Example:
+        >>> import pyhf
+        >>> pyhf.set_backend("numpy")
+        >>> model = pyhf.simplemodels.correlated_background(
+        ...     signal=[12.0, 11.0],
+        ...     bkg=[50.0, 52.0],
+        ...     bkg_up=[53.0, 59.0],
+        ...     bkg_down=[45.0, 49.0],
```
alexander-held, line 96: In the practical applications I am aware of, `histosys` modifiers are used for pure shape variations, and the total yield of the two histograms in the modifier would be the same. The normalization effects can then be split out to a (correlated) `normsys` modifier to have an effective log-normal constraint and protect against negative yields.

Doing this kind of split automatically may be a bit too much for `simplemodels` and lead to confusion. If the aim of #1434 is the use of more standard LHC-like models, it might be worth considering.

matthewfeickert: Very good point @alexander-held. I'll fix this in the morning my time, as it appears that my brain is mush enough that I'm no longer thinking about what I'm typing. :grimacing:

lukasheinrich: yes i'd do

```
        ...     bkg=[50.0, 50.0],
        ...     bkg_up=[45.0, 55.0],
        ...     bkg_down=[55.0, 45.0],
```

or sth similar

src/pyhf/simplemodels.py
```diff
@@ -68,3 +68,70 @@ def hepdata_like(signal_data, bkg_data, bkg_uncerts, batch_size=None):
...
+    r"""
+    Construct a simple single channel :class:`~pyhf.pdf.Model` with a
+    :class:`~pyhf.modifiers.histosys` modifier representing a background
+    with a fully correlated bin-by-bin uncertainty.
+
+    Args:
+        signal (:obj:`list`): The data in the signal sample.
+        bkg (:obj:`list`): The data in the background sample.
+        bkg_up (:obj:`list`): The background sample under a :math:`+1\sigma` variation.
+        bkg_down (:obj:`list`): The background sample under a :math:`-1\sigma` variation.
```
matthewfeickert, line 83: Probably should be more explicit here.

src/pyhf/simplemodels.py
alexander-held: Here is an example for the difference this can make:
```python
model = correlated_background(
    signal=[12.0, 11.0], bkg=[50.0, 52.0], bkg_up=[73.0, 79.0], bkg_down=[25.0, 29.0]
)
print("current model:")
print(f"  nominal:  {model.expected_data([1.0, 0.0], include_auxdata=False)}")
print(f"  -1 sigma: {model.expected_data([1.0, -1.0], include_auxdata=False)}")
print(f"  -3 sigma: {model.expected_data([1.0, -3.0], include_auxdata=False)}")


print("new model:")
spec_new = {
    "channels": [
        {
            "name": "single_channel",
            "samples": [
                {
                    "name": "signal",
                    "data": [12.0, 11.0],
                    "modifiers": [{"name": "mu", "type": "normfactor", "data": None}],
                },
                {
                    "name": "background",
                    "data": [50.0, 52.0],
                    "modifiers": [
                        {
                            "name": "correlated_bkg_uncertainty",
                            "type": "histosys",
                            "data": {
                                "hi_data": [48.99, 53.01],
                                "lo_data": [47.22, 54.78],
                            },
                        },
                        {
                            "name": "correlated_bkg_uncertainty",
                            "type": "normsys",
                            "data": {"hi": 1.49, "lo": 0.53},
                        },
                    ],
                },
            ],
        }
    ]
}
model_new = pyhf.Model(spec_new)
print(f"  nominal:  {model_new.expected_data([1.0, 0.0], include_auxdata=False)}")
print(f"  -1 sigma: {model_new.expected_data([1.0, -1.0], include_auxdata=False)}")
print(f"  -3 sigma: {model_new.expected_data([1.0, -3.0], include_auxdata=False)}")
```
resulting in
```
current model:
  nominal:  [62. 63.]
  -1 sigma: [37. 40.]
  -3 sigma: [-13.  -6.]
new model:
  nominal:  [62. 63.]
  -1 sigma: [37.0266 40.0334]
  -3 sigma: [18.20221582 19.98323818]
```
where the version with the histosys/normsys split has the built-in protection against negative yields.

alexander-held: The split does not provide complete safety either, if the shapes are very large then it is possible to still have negative yield predictions in bins with sufficiently strong pulls.

kratsg: This probably should exist in a learn notebook, rather than making the simplemodel more explicit. One could certainly add a `normalized=True` kwarg that could additionally split the bkg into a shape and normalization.

src/pyhf/simplemodels.py
```diff
@@ -1,12 +1,81 @@
...
+    :class:`~pyhf.modifiers.histosys` modifier representing a background
+    with a fully correlated bin-by-bin uncertainty.
+
+    Args:
+        signal (:obj:`list`): The data in the signal sample.
+        bkg (:obj:`list`): The data in the background sample.
+        bkg_up (:obj:`list`): The background sample under an upward variation
+         corresponding to :math:`\alpha=+1`.
+        bkg_down (:obj:`list`): The background sample under a downward variation
+         corresponding to :math:`\alpha=-1`.
```
matthewfeickert, line 22: Not sure if this is clear or not, so please make comments if not.

src/pyhf/simplemodels.py
lukasheinrich: I agree we should have docs on this, but maybe we keep this trivially simple - we shouldn't necessarily be in the business of buliding production grade models for users . These simple models are mostly for quick&dirty sensitivity tests.

lukasheinrich (approved):

PR #1450, opened 2021-05-12
chore: Use newer setuptools_scm for versioning

kratsg:
# Pull Request Description

This bumps us up to a more modern version of `setuptools_scm` that supports `pyproject.toml` and will inject the right version at build time (via PEP517/8) as well as generate the corresponding `version.py` for us at build time.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Update to a modern version of setuptools_scm to support pyproject-based Python packages
   - Use local_scheme = "no-local-version" to continue uploading to TestPyPI
   - c.f. https://github.com/pypa/setuptools_scm/blob/v6.0.1/README.rst#version-number-construction
* Remove src/pyhf/version.py in favor of generating at build time
   - Remove setup.cfg and src/pyhf/version.py from bump2version control
* Set fetch depth:0 in CI to get full history and tags to generate version number
   - c.f. https://github.com/scikit-hep/pyhf/issues/1465
```

### DIFF
--- a/.github/workflows/dependencies-head.yml
+++ b/.github/workflows/dependencies-head.yml
@@ -17,6 +17,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
@@ -43,6 +45,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
@@ -69,6 +73,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
@@ -94,6 +100,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
@@ -119,6 +127,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -19,8 +19,7 @@ from pathlib import Path
 import sys
 from pkg_resources import get_distribution
 
-sys.path.insert(0, str(Path('../src').resolve()))
-sys.path.insert(1, str(Path('./exts').resolve()))
+sys.path.insert(0, str(Path('./exts').resolve()))
 
 
 def setup(app):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,10 @@
 [build-system]
 # Minimum requirements for the build system to execute.
-requires = ["wheel", "setuptools>=30.3.0", "attrs>=17.1", "setuptools_scm"]
+requires = ["wheel", "setuptools>=30.3.0", "attrs>=17.1", "setuptools_scm[toml]>=3.4"]
 build-backend = "setuptools.build_meta"
+
+[tool.setuptools_scm]
+write_to = "src/pyhf/version.py"
 
 [tool.black]
 line-length = 88
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,7 @@ build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
 write_to = "src/pyhf/version.py"
+local_scheme = "no-local-version"
 
 [tool.black]
 line-length = 88
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,5 @@
 [metadata]
 name = pyhf
-version = 0.6.1
 description = pure-Python HistFactory implementation with tensors and autodiff
 long_description = file: README.rst
 long_description_content_type = text/x-rst
@@ -28,8 +27,6 @@ classifiers =
 
 [options]
 setup_requires =
-    setuptools_scm>=1.15.0
-    setuptools_scm_git_archive>=1.0
 package_dir =
     = src
 packages = find:
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,6 @@ classifiers =
     Programming Language :: Python :: Implementation :: CPython
 
 [options]
-setup_requires =
 package_dir =
     = src
 packages = find:
--- a/src/pyhf/__init__.py
+++ b/src/pyhf/__init__.py
@@ -1,6 +1,6 @@
 from .tensor import BackendRetriever as tensor
 from .optimize import OptimizerRetriever as optimize
-from .version import __version__
+from .version import version as __version__
 from .exceptions import InvalidBackend, InvalidOptimizer, Unsupported
 from . import events
 
--- a/src/pyhf/version.py
+++ b/src/pyhf/version.py
@@ -1,5 +1,0 @@
-"""Define pyhf version information."""
-
-# Use semantic versioning (https://semver.org/)
-# The version number is controlled through bumpversion.cfg
-__version__ = '0.6.1'
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,10 +3,6 @@ current_version = 0.6.1
 commit = True
 tag = True
 
-[bumpversion:file:setup.cfg]
-
-[bumpversion:file:src/pyhf/version.py]
-
 [bumpversion:file:src/pyhf/utils.py]
 
 [bumpversion:file:README.rst]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,7 +23,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-
+        with:
+          fetch-depth: 0
       - name: Prepare
         id: prep
         run: |
--- a/.github/workflows/lower-bound-requirements.yml
+++ b/.github/workflows/lower-bound-requirements.yml
@@ -18,6 +18,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+src/pyhf/version.py
 MANIFEST
 build
 dist
--- a/src/pyhf/cli/cli.py
+++ b/src/pyhf/cli/cli.py
@@ -3,7 +3,7 @@ import logging
 
 import click
 
-from ..version import __version__
+from .. import __version__
 from . import rootio, spec, infer, patchset, complete
 from ..contrib import cli as contrib
 from .. import utils
--- a/tests/test_public_api.py
+++ b/tests/test_public_api.py
@@ -8,7 +8,7 @@ def model_setup(backend):
     np.random.seed(0)
     n_bins = 100
     # TODO: Simplify after pyhf v0.6.2 released
-    if tuple(int(part) for part in pyhf.__version__.split(".")) < (0, 6, 2):
+    if tuple(int(part) for part in pyhf.__version__.split(".")[:3]) < (0, 6, 2):
         model = pyhf.simplemodels.hepdata_like(
             [10] * n_bins, [50] * n_bins, [1] * n_bins
         )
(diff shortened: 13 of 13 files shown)


.github/workflows/dependencies-head.yml
```diff
@@ -17,6 +17,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
matthewfeickert, line 21: @kratsg I'll look at this to help debug it, but this isn't needed for this workflow. You only need this if you need to get version information.

matthewfeickert (changes requested):

src/pyhf/__init__.py
```diff
@@ -1,6 +1,6 @@
 from .tensor import BackendRetriever as tensor
 from .optimize import OptimizerRetriever as optimize
-from .version import __version__
+from .version import version as __version__
```
matthewfeickert, line 3:
```pytb
/home/runner/work/pyhf/pyhf/docs/cli.rst:4: WARNING: Failed to import "pyhf" from "pyhf.cli.cli". The following exception was raised:
  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/sphinx_click/ext.py", line 390, in _load_module
    mod = __import__(module_name, globals(), locals(), [attr_name])
  File "/home/runner/work/pyhf/pyhf/src/pyhf/__init__.py", line 3, in <module>
    from .version import version as __version__
ModuleNotFoundError: No module named 'pyhf.version'
```

matthewfeickert: > In docs, there are a few minor tweaks necessary to make sure the version is picked up correctly; just make sure you install the package and access it from there.

https://scikit-hep.org/developer/packaging

So I'll add an install bit to docs. Not sure how this will work for RTD though.

setup.cfg
```diff
@@ -1,6 +1,5 @@
 [metadata]
 name = pyhf
-version = 0.6.1
```
matthewfeickert, line 3: @kratsg @henryiii How does PyPI know the version of the release? Or is this just taken from the file name itself?

src/pyhf/__init__.py
matthewfeickert: I'm now more confused as we already install everything for the docs

https://github.com/scikit-hep/pyhf/blob/6b769fd6f5e1473deba2b4c55d49ebdb3db5b447/.github/workflows/docs.yml#L22-L32

matthewfeickert:
@kratsg this also creates a pretty ugly name for the `sdist` and `wheel`

Example:

```console
$ python -m build --sdist --wheel --outdir dist/ .
$ ls dist/
pyhf-0.6.2.dev46+g391a6793-py3-none-any.whl  pyhf-0.6.2.dev46+g391a6793.tar.gz
```

Anyway to get it to just stop at the `pyhf-0.6.2.dev46` (which is what we really care about)?

setup.cfg
henryiii: It's added to the metadata by setuptools_scm.

src/pyhf/__init__.py
henryiii: Why install setuptools and wheel? Those are only needed in the PEP 518 build environment, are they not?

henryiii: Also not sure why you have disabled the cache dir, this should be a clean build already? Ignore-installed, too?

henryiii:
> pretty ugly name for the sdist and wheel

This is an accurate name. This is _not_ version 0.6.2, it's a pre-development version. When you tag it _and_ build with exactly that tag (non-dirty directory), it will have exactly that tag's version for the version. And if it's not, it will get a unique version name as you see here.

As an added bonus, you can't upload a wheel with this name to PyPI, so you are assured tagged uploads only.

If you really need it, you can set SETUPTOOLS_SCM_PRETENT_VERSION (IIRC) and force a version, even without git info.

henryiii:
> Anyway to get it to just stop at the pyhf-0.6.2.dev46 (which is what we really care about)?

No, if I make a commit, and you make a commit, and we both build, we might get the same "N" comments past 0.6.1, but different builds. So the hash ensures this is still unique. (I think you can customize it, though, and possibly even remove the hash if you want)

setup.cfg
```diff
@@ -28,8 +27,6 @@ classifiers =
 
 [options]
 setup_requires =
```
henryiii, line 29:
```suggestion
```

src/pyhf/__init__.py
henryiii: https://github.com/scikit-hep/pyhf/blob/dd946fc5ce895636a76eb299c6dc2f7f8bcf02cc/docs/conf.py#L22

If you want the src generated inplace, you can add `-e`, then `src/pyhf/version.py` will exist. But this looks really hacky/messy? Shouldn't you use the installed version not the hacked-in version in src? Isn't the point of src layout to _avoid_ being able to import without installing?

See for example @LovelyBuggies [config in Hist](https://github.com/scikit-hep/hist/blob/main/docs/conf.py), no path manipulations.

matthewfeickert:
> > pretty ugly name for the sdist and wheel
> 
> This is an accurate name. This is _not_ version 0.6.2, it's a pre-development version. 

Yeah, that part I'm clear on given that we rely on the dev versions to be able to upload to TestPyPI on each PR merge. https://test.pypi.org/project/pyhf/#history

> When you tag it _and_ build with exactly that tag (non-dirty directory), it will have exactly that tag's version for the version. And if it's not, it will get a unique version name as you see here.

Okay, so in this terminology "dirty" just means "not tagged"? Seems from the following the answer is yes

```
$ docker run --rm -ti python:3.8 /bin/bash
# git clone https://github.com/scikit-hep/pyhf.git
# cd pyhf
# git checkout origin/chore/setuptoolsSCMpyproject -b chore/setuptoolsSCMpyproject
# python -m pip install --quiet -e .
# pyhf --version
pyhf, version 0.6.2.dev45+gdd946fc
# git config --global user.email "test@test.org"
# git config --global user.name "test"
# bump2version minor
# python -m pip install --upgrade --quiet -e .
# pyhf --version
pyhf, version 0.7.0
# python -m pip install build
# python -m build --sdist --wheel --outdir dist/ .
# ls dist/
pyhf-0.7.0-py3-none-any.whl  pyhf-0.7.0.tar.gz
```

> 
> As an added bonus, you can't upload a wheel with this name to PyPI, so you are assured tagged uploads only.
> 
> If you really need it, you can set SETUPTOOLS_SCM_PRETEND_VERSION (IIRC) and force a version, even without git info.

Okay, sounds like if we wanted to keep on uploading to TestPyPI on each PR like we have been doing that we'd need to look into using

henryiii:
No, dirty is an unclean working tree. Untagged is just number of commits + a hash. For example, let's say I made one commit and make a PR, and you made one commit and made a PR. The PR's would have identical versions if you didn't include the hash, because they are the same number of commits past a tag.

Why not just upload the wheels to GitHub Actions, and then download them from there if you need them for testing something? That's not really what Test PyPI is for. The files will go away after some number of days, but Test PyPI gets cleared eventually I assume.

matthewfeickert:
> No, dirty is an unclean working tree. Untagged is just number of commits + a hash. For example, let's say I made one commit and make a PR, and you made one commit and made a PR. The PR's would have identical versions if you didn't include the hash, because they are the same number of commits past a tag.

Yeah, I get the point of including the hash. We avoid this in general by checking the "Require branches to be up to date before merging" option in our branch restrictions for `master`, but I understand the illustration that you're making here.
 
> Why not just upload the wheels to GitHub Actions, and then download them from there if you need them for testing something? That's not really what Test PyPI is for. The files will go away after some number of days, but Test PyPI gets cleared eventually I assume.

What is TestPyPI for then? We have TestPyPI releases going all the way back to 2019 https://test.pypi.org/project/pyhf/0.0.16/. You're right that, just like PyPI, we'll eventually hit the size limit for TestPyPI and need to delete releases, but that won't be for a long long time. Do you see any downside to having every commit to `master` having a corresponding release on TestPyPI? It has been useful in the past for us to look at arbitrary points in the past.

src/pyhf/__init__.py
matthewfeickert: > Why install setuptools and wheel? Those are only needed in the PEP 518 build environment, are they not?

Boilerplate laziness of me copy and pasting things between CI bits. Doesn't `wheel` work with `pip` though if a wheel isn't available from PyPI to build one locally for installs?

> Also not sure why you have disabled the cache dir, this should be a clean build already? Ignore-installed, too?

Why not? 

> But this looks really hacky/messy? Shouldn't you use the installed version not the hacked-in version in src? Isn't the point of src layout to avoid being able to import without installing?

Yeah, tracing back the `git blame` on this makes it seem like it was originally added near the start of the project in PR #157 and then we didn't properly clean this up and change it when we switched to a `src` directory. If I'm being honest the `docs/conf.py` needs a rework, but we've just left it as nothing has broke so far.

kratsg:
Should be all working now. I made all corresponding fixes/changes with respect to this PR.

henryiii:
Are you uploading on PRs, or after PRs are merged? Anyway, TestPyPI is for testing the upload process, or making sure something works _before_ release; it is not at all archival. If you want an old commit, just do:

```console
python3 -m pip install git+https://github.com/scikit-build/pyhf@6b769fd6f5e1473deba2b4c55d49ebdb3db5b447
```

That will install any commit you want. You don't the quite the same speed as a wheel, but this is just a development/testing install, not something being installed en masse.

src/pyhf/__init__.py
henryiii: > wheel isn't available from PyPI to build one locally for installs

Only if it's a "bad" package and doesn't have a pyproject.toml. Good packages with pyproject.toml's don't need them. Soonish, the backend for pip might be switched to use pep-517 by default, which will act like a pyproject.toml is there anyway. But yes, if you _depend on_ a wheel-less classic-build package, this is useful.

> Why not?

If you are building in CI, and there's no existing cache, it's unneeded to disable caching. And if you _add_ a cache to speed up your CI, then you'd have to go remove this too.

> but we've just left it as nothing has broke so far

Ugly code should die! Kill it if you have to! :)

src/pyhf/version.py
```diff
@@ -1,5 +0,0 @@
-"""Define pyhf version information."""
```
matthewfeickert, line 1: I guess we should also add `src/pyhf/version.py` to `.gitignore` so that once it is generated we know to ignore it.

src/pyhf/__init__.py
matthewfeickert: > But yes, if you _depend on_ a wheel-less classic-build package, this is useful.

Directly as a stated core dependency, no, but `pyrsistent` shows up in our CI given that it is a requirement of `jsonschema` which is a core dependency. https://pypi.org/project/pyrsistent/0.17.3/

> If you are building in CI, and there's no existing cache, it's unneeded to disable caching

Your suggestions are applied in PR #1453. I think when I first setup a lot of this I was still not actually believing that CI did what it said it would given experiences with Travis CI having versions of NumPy installed in the environments back in the day that you'd have to clear out before you could get your required versions in.

> Ugly code should die! Kill it if you have to! :)

No disagreement here. :) Just where it falls on the priority list. Feel free to send precocious students our way to make PRs to kill said code (I know you don't have extra time to be doing this for us).

.github/workflows/dependencies-head.yml
matthewfeickert: @kratsg 

Ah, I understand why these need to have the full fetch depth now given that

https://github.com/scikit-hep/pyhf/blob/546022254a526ea62af8cbc6594ec881cddfa9f1/tests/test_public_api.py#L11

makes use of the version information.

Feel free to correct me if I'm wrong, but I think that the _only_ places that the CI needs to have the full fetch depth and tags are where the version number is going to get checked (so like `test_public_api.py`). So this means that these `fetch-depth: 0`s could get removed after `v0.6.2` is out, right? If you agree can you make this comment of mine an Issue, assign it to me, and put it on the `v0.7.0` board?

docs/conf.py
```diff
@@ -19,8 +19,7 @@
 import sys
 from pkg_resources import get_distribution
 
-sys.path.insert(0, str(Path('../src').resolve()))
-sys.path.insert(1, str(Path('./exts').resolve()))
+sys.path.insert(0, str(Path('./exts').resolve()))
```
matthewfeickert, line 22: @kratsg Is it possible for us to get rid of the path insertion entirely like @henryiii suggested? Put another way, can you remind me why we need this?

pyproject.toml
```diff
@@ -1,8 +1,12 @@
 [build-system]
 # Minimum requirements for the build system to execute.
-requires = ["wheel", "setuptools>=30.3.0", "attrs>=17.1", "setuptools_scm"]
+requires = ["wheel", "setuptools>=30.3.0", "attrs>=17.1", "setuptools_scm[toml]>=3.4"]
 build-backend = "setuptools.build_meta"
 
+[tool.setuptools_scm]
+write_to = "src/pyhf/version.py"
+local_scheme = "no-local-version"
```
matthewfeickert, line 8: @kratsg this is all fine, but can you add a link in your PR summary description that links to the relevant sections in the `setuptools_scm` docs to make the easier for people like me to understand what's happening in the future?

matthewfeickert (commented):
@kratsg Thanks for this one &mdahs; it will be nice to have! I think I agree with everything but I just have some small questions. Once those are resolved then I think we're good to go here. :+1:

docs/conf.py
kratsg: We need the xrefs: https://github.com/michaeljones/sphinx-xref -- this isn't a python package, so we are manually pulling it in for now.

pyproject.toml
kratsg: They don't have official docs. Just the readme? https://github.com/pypa/setuptools_scm#pyprojecttoml-usage -- can't do much more than that.

.github/workflows/dependencies-head.yml
matthewfeickert: This is now Issue #1465

matthewfeickert (approved):
Thanks very much for this one @kratsg!

PR #1460, opened 2021-05-17
fix: Do not configure logging in contrib.utils

### LINKED ISSUE #865: User-defined log formatting
# Description

`pyhf` uses `logging` for outputs, and calls `logging.basicConfig()` in a few places.
This has the effect of preventing the user to set their desired logging behavior after `pyhf` import.
While calling this a bug might be a bit of a stretch, I think it might be unintentional since `pyhf` does not apply any logging formatting as far as I can tell.

# Expected Behavior

I expect no calls to `logging.basicConfig()` within `pyhf` to leave the formatting fully up to the user, no matter whether they want to set it before or after importing `pyhf`.

# Actual Behavior

User-defined `logging` formatting only works before importing `pyhf`.

# Steps to Reproduce

importing `pyhf` before formatting:
```
import logging
import pyhf
print(pyhf.__version__)
logging.basicConfig(level=logging.INFO)
log.info("message")
```
output:
```
0.4.1
```
and when applying formatting before input, the expected behavior:
```
import logging
logging.basicConfig(level=logging.INFO)
import pyhf
print(pyhf.__version__)
log.info("message")
```
output:
```
0.4.1
INFO:__main__:message
``` 

# Checklist

  - no, but checked code on master to confirm that the relevant part is unchanged
- [X] Filled out the Description, Expected Behavior, Actual Behavior, and Steps to Reproduce sections above or have edited/removed them in a way that fully describes the issue

### LINKED ISSUE #1455: Logging configuration in contrib/utils
# Question

`pyhf.contrib.utils` sets up logging:
https://github.com/scikit-hep/pyhf/blob/6b769fd6f5e1473deba2b4c55d49ebdb3db5b447/src/pyhf/contrib/utils.py#L9 

This interferes with custom logging users may want to set up. To achieve this now, they would have to do so before `from pyhf.contrib.utils import download`. To avoid this issue, the logging should not be configured in this part of the code (and only for the CLI).

#865

alexander-held:
# Description

Fixes #1455. This is the equivalent of #866, which fixed #865.

`pyhf.contrib.utils` sets up logging, thereby preventing users who import that module from configuring logging themselves after the import. This removes the logging setup, which is only needed for the CLI (and done there).

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Remove the logging configuration in contrib.utils as it interferes with custom logging users may want to set up
```

### DIFF
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -6,7 +6,6 @@ from io import BytesIO
 import logging
 from .. import exceptions
 
-logging.basicConfig()
 log = logging.getLogger(__name__)
 
 __all__ = ["download"]


kratsg (approved):

lukasheinrich (approved):

lukasheinrich:
Thanks @alexander-held

matthewfeickert:
Thanks for this @alexander-held! If you don't mind, I'm going to get a fix in for Issue #1462 and then have you rebase on that so that we know workflows based on forks will work as hoped. :+1:

alexander-held:
@matthewfeickert sounds good to me!

matthewfeickert (approved):
LGTM. Many thanks for the PR @alexander-held!

PR #1461, opened 2021-05-17
fix: Check parameter shapes for pdf API calls

### LINKED ISSUE #1459: Input validation for `pyhf.pdf.Model.expected_data`
# Question

What do you think about checking the length of the input to `pyhf.pdf.Model.expected_data`? That function feels like a high-level entry point (though `model.main_model.expected_data` may also be needed as entry point for some users, see #1031). Currently an `IndexError` is raised if the argument has too few parameters, and arguments with too many parameters will cause no errors or warnings, but interpret the first `model.config.npars` parameters as those to use for the calcultion.

Example:
```python
import pyhf

)
pars = model.config.suggested_init()
print(f"model has {model.config.npars} parameters")
p = [1.1, 1.2, 1.3]
p_long = p + [1.5] * 5
p_short = [1.1, 1.2]
print(f"prediction for {p}: {model.expected_data(p)}")
print(f"prediction for {p_long}: {model.expected_data(p_long)}")
print(f"prediction for {p_short}: {model.expected_data(p_short)}")
```
Output:
```python
model has 3 parameters
prediction for [1.1, 1.2, 1.3]: [ 73.2         79.7        333.33333333  71.73877551]
prediction for [1.1, 1.2, 1.3, 1.5, 1.5, 1.5, 1.5, 1.5]: [ 73.2         79.7        333.33333333  71.73877551]
  File "test.py", line 13, in <module>
    print(f"prediction for {p_short}: {model.expected_data(p_short)}")
  File "[...]/pyhf/src/pyhf/pdf.py", line 744, in expected_data
    return self.make_pdf(pars).expected_data()
  File "[...]/pyhf/src/pyhf/pdf.py", line 788, in make_pdf
    mainpdf = self.main_model.make_pdf(pars)
  File "[...]/pyhf/src/pyhf/pdf.py", line 546, in make_pdf
    lambdas_data = self.expected_data(pars)
  File "[...]/pyhf/src/pyhf/pdf.py", line 608, in expected_data
  File "[...]/pyhf/src/pyhf/pdf.py", line 573, in _modifications
    [self.modifiers_appliers[k].apply(pars) for k in self._factor_mods],
  File "[...]/pyhf/src/pyhf/pdf.py", line 573, in <listcomp>
    [self.modifiers_appliers[k].apply(pars) for k in self._factor_mods],
  File "[...]/pyhf/src/pyhf/modifiers/shapesys.py", line 169, in apply
    shapefactors = tensorlib.gather(flat_pars, self.access_field)
  File "[...]/pyhf/src/pyhf/tensor/numpy_backend.py", line 177, in gather
    return tensor[indices]
IndexError: index 2 is out of bounds for axis 0 with size 2
```

The trace to the `IndexError` is maybe not extremely readable, but might be clear enough to find the reason for it quickly. This behavior could possibly be caught in `pyhf.pdf.Model.expected_data` with an error message?

The current handling of arguments with too many parameters could be convenient for some cases. What do you think about adding a warning? When removing a single parameter from a model, that parameter is not necessarily the last one. If the user forgets to update the input to `expected_data`, they get no feedback about the mismatch and possibly end up calculating the wrong thing.

only marginally related: #1031

kratsg:
# Pull Request Description

Resolves #1459. Check that `expected_data`, `expected_auxdata`, and `expected_actualdata` are being called with the right shape for `pars` (the parameters) before evaluating. This is usually caught by lower-level code, however the error is not as user-friendly.

A quick note: it is ok to add `if/raise` in these API calls as they're not meant to be "fast" in our code -- but this could potentially be a problem if we need to allow for autodiff capabilities. For now, we're only enforcing that `logpdf()` calls are the performant ones.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add checks on input parameter shapes for the pdf API calls to expected_auxdata,
expected_actualdata, and expected_data.
* Add tests for parameter shape checks to properly raise exceptions.InvalidPdfParameters
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -848,7 +848,7 @@ class Model:
         # Verify parameter shapes
         if pars.shape[-1] != self.config.npars:
             raise exceptions.InvalidPdfParameters(
-                f'eval failed as pars has len {pars.shape[-1]} but {self.config.npars} was expected'
+                f"Evaluation failed as parameters have length {pars.shape[-1]} but model requires {self.config.npars}."
             )
 
         return self.make_pdf(pars)[1].expected_data()
@@ -881,10 +881,10 @@ class Model:
         # Verify parameter shapes
         if pars.shape[-1] != self.config.npars:
             raise exceptions.InvalidPdfParameters(
-                f'eval failed as pars has len {pars.shape[-1]} but {self.config.npars} was expected'
+                f"Evaluation failed as parameters have length {pars.shape[-1]} but model requires {self.config.npars}."
             )
 
-        return self.make_pdf(pars)[0].expected_data()
+        return self.main_model._expected_data_unchecked(pars)
 
     def expected_data(self, pars, include_auxdata=True):
         """
@@ -902,11 +902,11 @@ class Model:
         # Verify parameter shapes
         if pars.shape[-1] != self.config.npars:
             raise exceptions.InvalidPdfParameters(
-                f'eval failed as pars has len {pars.shape[-1]} but {self.config.npars} was expected'
+                f"Evaluation failed as parameters have length {pars.shape[-1]} but model requires {self.config.npars}."
             )
 
         if not include_auxdata:
-            return self.make_pdf(pars)[0].expected_data()
+            return self.main_model._expected_data_unchecked(pars)
         return self.make_pdf(pars).expected_data()
 
     def constraint_logpdf(self, auxdata, pars):
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -625,7 +625,7 @@ class _MainModel:
         return True
 
     def make_pdf(self, pars):
-        lambdas_data = self.expected_data(pars)
+        lambdas_data = self._expected_data_unchecked(pars)
         return prob.Independent(prob.Poisson(lambdas_data))
 
     def logpdf(self, maindata, pars):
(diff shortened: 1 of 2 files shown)


lukasheinrich:
I wonder whether for these types of API we shsould have a pattern

```
def method(self,...):
    # check inputs
    ....
   self._method(...)
```

such that for performance critical or internal paths (i.e. where pyhf itself calls `method` we're able to call the "unsafe" `_method`  - relying on prior checks, while the user gets a friendly API with safety checks

kratsg:
> I wonder whether for these types of API we shsould have a pattern

I had this idea where I wanted to use decorators. It can introspect the arguments and do the checks based on something consistent, rather than copy/paste code around a lot. In this case, something like `@pyhf.checks.pars` that checks the number of parameters passed in for pdf calls, and would become a passthrough based on some `pyhf.config` configuration or similar.

src/pyhf/pdf.py
```diff
@@ -701,6 +707,12 @@ def expected_auxdata(self, pars):
...
         pars = tensorlib.astensor(pars)
+        # Verify parameter shapes
+        if pars.shape[-1] != self.config.npars:
+            raise exceptions.InvalidPdfParameters(
+                f'eval failed as pars has len {pars.shape[-1]} but {self.config.npars} was expected'
+            )
+
         return self.make_pdf(pars)[1].expected_data()
 
     def _modifications(self, pars):
```
lukasheinrich, line 718: given that we do input validation here it suggests that this has become a publicly consumable API. maybe we should add a `model.modifications` and do the input checks there and call `_modifications` if inputs are ok

kratsg: `_modifications` isn't currently a public API. This certainly could become one, although I might argue that we remove it from `Model` and keep it on `MainModel` unless there's a reason to pass-through it. e.g. `pdf.main_model.modifications` is just as clear to me. Unless the suggestion here is to remove the checks from `main_model` and `constraint_model` and keep all checks on `model` which is also possible, but feels like a mess.

alexander-held: `main_model` is needed for `return_by_sample`, so it feels similarly "public" as `Model`.

kratsg: Sure, but `main_model.expected_actualdata(..., return_by_sample)` is fine -- since that's public. However `main_model._modifications` isn't necessarily public. Although at the moment only used by `main_model.expected_data` -- so I'm fine either way.

alexander-held: > Unless the suggestion here is to remove the checks from `main_model` and `constraint_model`

My comment was meant as an example for why that may not be desirable, sorry I should have made that clear. Promoting `_modifications` to public in the longer term is a nice idea, it looks very useful for model debugging.

matthewfeickert (approved):
This all LGTM @kratsg &mdash; thanks. I'll let you and @lukasheinrich resolve the current discussion and implement any changes that you want, but I'm happy to have this merged whenever you both are.

src/pyhf/pdf.py
matthewfeickert: @lukasheinrich @kratsg If we can revisit the performance impact here soon it would be nice to have this get into `v0.7.0`.

kratsg: done

matthewfeickert (commented):
@kratsg my views have remain unchanged from https://github.com/scikit-hep/pyhf/pull/1461#pullrequestreview-661196259 it seems, so whenever we have pre-commit pass I'm happy with this.

matthewfeickert (approved):

PR #1467, opened 2021-05-18
fix: Remove 'src' from pytest test testpaths to allow for non-editable install in CI

### LINKED ISSUE #1456: pytest fails for source installs that don't use `pip install --editable .`
Strangely it seems that if we install from source with 

```
python -m pip install .[test]
```

`pytest` will fail with things like

```pytb
==================================== ERRORS ====================================
_____________________ ERROR collecting src/pyhf/compat.py ______________________
/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/_pytest/runner.py:311: in from_call
    result: Optional[TResult] = func()
/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/_pytest/runner.py:341: in <lambda>
    call = CallInfo.from_call(lambda: list(collector.collect()), "collect")
/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/_pytest/doctest.py:532: in collect
    module = import_path(self.fspath)
/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/_pytest/pathlib.py:544: in import_path
    raise ImportPathMismatchError(module_name, module_file, path)
E   _pytest.pathlib.ImportPathMismatchError: ('pyhf.compat', '/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/pyhf/compat.py', PosixPath('/home/runner/work/pyhf/pyhf/src/pyhf/compat.py'))
```

and that it will only pass if we install from source in editable mode

```
python -m pip install --editable .[test]
```

_Originally posted by @matthewfeickert in https://github.com/scikit-hep/pyhf/issues/1453#issuecomment-841541604_

It would be good to understand _why_ this is, and if possible change it. The tests _needing_ things to be installed with `pip install --editable .` seems strange and bad.

matthewfeickert:
# Description

Resolves #1456 

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,8 +39,7 @@ ignore = [
 minversion = "6.0"
 addopts = "--ignore=setup.py --ignore=validation/ --ignore=binder/ --ignore=docs/ --cov=pyhf --cov-report=term-missing --cov-config=.coveragerc --cov-report xml --doctest-modules --doctest-glob='*.rst'"
 testpaths = [
-    "src",
-    "tests",
+    "tests"
 ]
 markers = [
     "fail_jax",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
-        python -m pip install --upgrade --editable .[test]
+        python -m pip install --upgrade .[test]
     - name: List installed Python packages
       run: |
         python -m pip list
--- a/.github/workflows/dependencies-head.yml
+++ b/.github/workflows/dependencies-head.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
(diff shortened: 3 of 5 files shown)


pyproject.toml
```diff
@@ -39,8 +39,7 @@ ignore = [
 minversion = "6.0"
 addopts = "--ignore=setup.py --ignore=validation/ --ignore=binder/ --ignore=docs/ --cov=pyhf --cov-report=term-missing --cov-config=.coveragerc --cov-report xml --doctest-modules --doctest-glob='*.rst'"
 testpaths = [
-    "src",
-    "tests",
+    "tests"
```
matthewfeickert, line 42: We're not not having the docstrings get checked by `doctest` if this is removed. So we'll have to fix ths.

matthewfeickert: @kratsg You [say that for `ironman`](https://twitter.com/kratsg/status/1394483751758008323)

> Dropping "src" from "testpaths" fixes it for me and still runs doctest correctly.

but I'm not really clear on how this works then for `doctest`. Can you take a look here as well?

PR #1470, opened 2021-05-20
WIP: getting suggested_init without user config overrides

kratsg:
# Description

Just a quick idea. Normally, the `suggested_init` will often provide information about the default parameter settings, however this doesn't necessarily work when a user overrides these inits. This provides a way to get an init that does not necessarily depend on the user configs. Just an idea.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/parameters/paramsets.py
+++ b/src/pyhf/parameters/paramsets.py
@@ -26,6 +26,10 @@ class paramset:
                 f'misconfigured parameter set {self.name}. Scalar but N>1 parameters.'
             )
 
+    @property
+    def maximal_pars(self):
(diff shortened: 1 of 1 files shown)


kratsg:
perhaps allowing `suggested_init()` to have a keyword argument to handle flipping would work:
- `ignore_config`
- `ignore_user_config`
- `for_asimov`

alexander-held:
Why this is interesting: it could be used to build (pre-fit) Asimov datasets even if users provide non-default initial values for constrained parameters. It may be interesting to (optionally?) pick up non-default inits for unconstrained parameters anyway.

PR #1476, opened 2021-05-26
learn notebook for CLs asymptotics

lukasheinrich:
# Description

produces this animation

anim


cranmer:
very cool. I would still suggest a transposed version of the top right plot. Maybe in the middle right panel?
I'd also suggest increasing the size of the axis labels a little. 

Also, recommend matplotlib -> jshtml to include in static HTML
http://louistiao.me/posts/notebooks/embedding-matplotlib-animations-in-jupyter-as-interactive-javascript-widgets/

lukasheinrich:
ttherre is also an arrangement like tthis that has tthe CLs and the qµ distrtibutions in normal orientation and the gaussian µ^ distribution rotated

image

PR #1484, opened 2021-06-07
chore: Add default virtual environment on PATH for Docker image

matthewfeickert:
# Description

Create a default virtual environment and place it on `PATH` so that it is active and used to install all libraries. Then, just `COPY` the virtual environment from the builder image instead of all of `/usr/local/` reducing the amount of files copied to _only_ the necessary files.

This has the added benefit of symlinking the installed `python3` to the `virtualenv` PATH, so the equivalent of

```
ln -s $(command -v python3) /usr/local/venv/bin/python
```

This approach was recommended by Anthony Sottile in his YouTube video "[how to get pip for deadsnakes / docker pythons (intermediate) anthony explains 293](https://youtu.be/2Hg5-Hrsa6w)". Thanks Anthony!

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add a venv virtual environment to PATH
   - Adding venv also symlinks the python doing the installing to the virtual environment's python
      -  e.g. `ls -l $(which python)` shows `/usr/local/venv/bin/python -> /usr/local/bin/python`
   - Thanks to Anthony Sottile for recommending this approach
   - c.f. https://youtu.be/2Hg5-Hrsa6w
   - c.f. https://github.com/pyhf/cuda-images/pull/3
* Update hadolint GitHub Action to hadolint/hadolint-action@v1.5.0
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,7 +27,8 @@ jobs:
     - name: Lint with Black
       run: |
         black --check --diff --verbose .
+
     - name: Lint Dockerfile
-      uses: brpaz/hadolint-action@v1.1.0
+      uses: hadolint/hadolint-action@v1.5.0
       with:
-        dockerfile: "docker/Dockerfile"
+        dockerfile: docker/Dockerfile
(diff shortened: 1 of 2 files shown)


matthewfeickert:
@lukasheinrich @kratsg I used this approach first in https://github.com/pyhf/cuda-images/pull/3 and it works well there. :+1:

PR #1512, opened 2021-07-01
docs: Add release notes URL to PyPI project urls metadata

matthewfeickert:
# Description

This PR adds the release notes webpage to the PyPI project URL metadata in `setup.cfg` so that it shows up on PyPI in a similar fashion to how [`click`'s page](https://pypi.org/project/click/) has it.

The name of the `project_urls` keys is based off what [`click` has in their `setup.cfg`](https://github.com/pallets/click/blob/70358c57b20454c262bceaa901ab273b2487d7f3/setup.cfg#L5-L12) and the [template rules for the PyPA warehouse](https://github.com/pypa/warehouse/blob/3a93612489a47e37c68419821061a530064677ed/warehouse/templates/packaging/detail.html#L20-L58) that @duetosymmetry [found](https://twitter.com/duetosymmetry/status/1410687033237028876) (thanks Leo!).

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add release notes webpage to PyPI project metadata
   - Follows PyPA warehouse template rules: https://github.com/pypa/warehouse/blob/3a93612489a47e37c68419821061a530064677ed/warehouse/templates/packaging/detail.html#L20-L58
* Expand the names of the project url keys to make them more clear and explicit on the PyPI webpage
```

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,8 +3,6 @@ current_version = 0.6.2
 commit = True
 tag = True
 
-[bumpversion:file:setup.cfg]
-
 [bumpversion:file:src/pyhf/utils.py]
 
 [bumpversion:file:README.rst]
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -3,6 +3,8 @@ current_version = 0.6.2
 commit = True
 tag = True
 
+[bumpversion:file:setup.cfg]
+
 [bumpversion:file:src/pyhf/utils.py]
 
 [bumpversion:file:README.rst]
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,6 +13,7 @@ project_urls =
     Documentation = https://pyhf.readthedocs.io/
     Source Code = https://github.com/scikit-hep/pyhf
     Issue Tracker = https://github.com/scikit-hep/pyhf/issues
+    Release Notes = https://pyhf.readthedocs.io/en/v0.6.2/release-notes.html
 classifiers =
     Development Status :: 4 - Beta
     License :: OSI Approved :: Apache Software License
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,8 +11,8 @@ license_file = LICENSE
 keywords = physics fitting numpy scipy tensorflow pytorch jax
 project_urls =
     Documentation = https://pyhf.readthedocs.io/
-    Source = https://github.com/scikit-hep/pyhf
-    Tracker = https://github.com/scikit-hep/pyhf/issues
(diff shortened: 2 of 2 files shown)


setup.cfg
```diff
@@ -11,8 +11,9 @@ license_file = LICENSE
 keywords = physics fitting numpy scipy tensorflow pytorch jax
 project_urls =
     Documentation = https://pyhf.readthedocs.io/
-    Source = https://github.com/scikit-hep/pyhf
-    Tracker = https://github.com/scikit-hep/pyhf/issues
+    Source Code = https://github.com/scikit-hep/pyhf
+    Issue Tracker = https://github.com/scikit-hep/pyhf/issues
+    Release Notes = https://pyhf.readthedocs.io/en/v0.6.2/release-notes.html
```
matthewfeickert, line 16: I don't know of any way to get this URL to not be version dependent unless we use `stable`

```
https://pyhf.readthedocs.io/en/stable/release-notes.html
```

but I know that @kratsg wants to avoid using that. If we keep the version number in the URL we need to add `setup.cfg` back to `bump2version` control.

.bumpversion.cfg
```diff
@@ -3,6 +3,8 @@ current_version = 0.6.2
 commit = True
 tag = True
 
+[bumpversion:file:setup.cfg]
```
kratsg, line 6: is there a reason this got added? It should be picked up automatically from the auto-generated version.

kratsg (changes requested):

setup.cfg
kratsg: oh man, you beat me to it.

.bumpversion.cfg
matthewfeickert: c.f. below

setup.cfg
matthewfeickert: @kratsg As you requested changes in the review, do you want this to be 
```suggestion
    Release Notes = https://pyhf.readthedocs.io/en/stable/release-notes.html
```
and then remove `setup.cfg` from `.bumpversion.cfg`?

kratsg: Yep. I think fundamentally, we'd want to ensure that we're not changing these links so much. Given all previous version releases are listed on that page from `stable`, I think it's fine if we point to stable.

kratsg: Oh also, because the url for the pypi page is generally pointing to latest stable release anyway, it makes sense that we use the stable url.

matthewfeickert: Great! SGTM.

kratsg (approved):

Issue #1506, opened 2021-06-30
Exceptions related to weakref in pyhf 0.6.2


alexander-held:
# Description

With `pyhf` 0.6.2 I see a long trace of `weakref`-related exceptions at the end of a program using certain parts of `pyhf`. I have not extensively tested which parts cause this behavior, but have so far observed it with `pyhf.simplemodels.correlated_background` / `pyhf.simplemodels.uncorrelated_background`. I can also confirm that other functionality like `pyhf.set_backend("numpy")` does not trigger it.

# Expected Behavior

No stack of `weakref` exceptions after code finishes running.

# Actual Behavior

Large stack of `weakref`-related exceptions.

# Steps to Reproduce

Run a simple file using `pyhf.simplemodels.uncorrelated_background`, the example runs in a `python:3.8-slim` container with Python 3.8.11. I see the same locally in a conda environment with Python 3.8.10 on Mac OS. With `pyhf` 0.6.2 the stack of exceptions appears. When installing `pyhf==0.6.1` and using `pyhf.simplemodels.hepdata_like`, I do not see any exceptions. When running via `python -m pdb test.py` locally to debug, I do no longer see the exception. The current `pyhf` master also shows the same issue.

```python
$ cat test.py
import pyhf

model = pyhf.simplemodels.uncorrelated_background(
    signal=[5.0], bkg=[10.0], bkg_uncertainty=[2.0]
)
```
```console
$ docker run -it --rm -v$PWD:/test python:3.8-slim bash
root@8fd8a5f229c8:/# pip install pyhf
Collecting pyhf
  Downloading pyhf-0.6.2-py3-none-any.whl (140 kB)
     |████████████████████████████████| 140 kB 3.5 MB/s
Collecting pyyaml>=5.1
  Downloading PyYAML-5.4.1-cp38-cp38-manylinux1_x86_64.whl (662 kB)
     |████████████████████████████████| 662 kB 5.7 MB/s
Collecting scipy>=1.4.1
  Downloading scipy-1.7.0-cp38-cp38-manylinux_2_5_x86_64.manylinux1_x86_64.whl (28.4 MB)
     |████████████████████████████████| 28.4 MB 16.4 MB/s
Collecting jsonpatch>=1.15
  Downloading jsonpatch-1.32-py2.py3-none-any.whl (12 kB)
Collecting tqdm>=4.56.0
  Downloading tqdm-4.61.1-py2.py3-none-any.whl (75 kB)
[...]

PR #1529, opened 2021-07-20
docs: Add FAQ on reasons for need to downgrade dependencies

### LINKED ISSUE #1513: docs: Mention that downgrading dependencies might be needed at times
# Description

Given the problems brought up with `click` `v8.X` in Issue #1506, it is probably worth noting in the docs FAQ section that if a user runs into a problem with a core dependency acting weird that as `pyhf` is a library and not an application we don't overly control the range of dependencies and they might need to manually downgrade dependencies, e.g.

```
# requirements.txt
click<8.0.0
pyhf==0.6.2
```

, especially if there has been a recent release of the dependencies since the last release of `pyhf`.

matthewfeickert:
# Description

Resolves #1513

Add FAQ explaining how reasons why users might have to manually downgrade dependencies.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add FAQ explaining how reasons why users might have to manually downgrade dependencies.
   - c.f. PR #1979 for context
```

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -30,8 +30,8 @@ Use the :code:`--backend` option for :code:`pyhf cls` to specify a tensor backen
 The default backend is NumPy.
 For more information see :code:`pyhf cls --help`.
 
-I installed ``pyhf`` from PyPI, why am I getting an error from a dependency?
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+I installed an old ``pyhf`` release from PyPI, why am I getting an error from a dependency?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 You might need to manually constrain the **upper bound** on a ``pyhf`` core
 dependency.
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -50,9 +50,9 @@ If ``pyhf`` were to define upper bounds we could create situations in which
 ``requirements.txt``) could have directly conflicting dependencies that would
 result in ``pip`` failing to be able to install ``pyhf``.
 
-To give an explicit example, ``click`` ``v0.8.0`` was released *after* ``pyhf``
-``v0.6.2`` and changes in ``click``'s behavior between ``v7.X`` and ``v8.X``
-resulted in a runtime ``TypeError`` (c.f. Issue :issue:`1506`).
+To give an explicit example, changes in ``click``'s behavior between its
+``v7.X`` releases and ``v8.X`` resulted in a runtime ``TypeError``
+(c.f. Issue :issue:`1506`).
 The Issue was fixed in the next release of ``pyhf``, but the intermediate
 solution for users was to simply install an older version of ``click`` that was
 still compatible with ``pyhf``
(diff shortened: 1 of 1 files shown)


matthewfeickert:
Relevant RTD page: https://pyhf--1529.org.readthedocs.build/en/1529/faq.html#i-installed-pyhf-from-pypi-why-am-i-getting-an-error-from-a-dependency

matthewfeickert:
@alexander-held @henryiii if you also have time to let me know if this seems clear (and not too verbose) that would nice.

kratsg (approved):

docs/faq.rst
```diff
@@ -28,6 +28,39 @@ Use the :code:`--backend` option for :code:`pyhf cls` to specify a tensor backen
...
+``pyhf`` and other libraries defined in an environment file (i.e.,
+``requirements.txt``) could have directly conflicting dependencies that would
+result in ``pip`` failing to be able to install ``pyhf``.
+
+To give an explicit example, changes in ``click``'s behavior between its
+``v7.X`` releases and ``v8.X`` resulted in a runtime ``TypeError`` if ``click``
+``v8.X`` was used with ``pyhf`` ``v0.6.2`` (c.f. Issue :issue:`1506`).
+The Issue was fixed in the next release of ``pyhf``, but the intermediate
+solution for users was to simply install an older version of ``click`` that was
+still compatible with ``pyhf``
```
alexander-held, line 56: This is missing a period/colon after `pyhf`.

alexander-held:
This looks good to me and seems helpful to have. What do you think about adding the required `pip` command for the example? Something like
```bash
python -m pip install --upgrade "click<8"
```
for the `click` example. Maybe being very explicit here could help users who are getting started with Python and have not had to deal with specifying versions before. On the other hand, this is not a Python tutorial so it also seems reasonable to not go into that detail. The instructions can be found elsewhere on the internet.

henryiii:
To me this looks more like a bug in Click that happened to get added in version 8 than an intentional change. Once it's discovered I think you might want to do a bug report to Click. I'm not sure a document like this is all that useful; this is mostly not specific to PyHF, or any other library that has dependencies. If you depend on something, that something could change - or even just not work on a new platform. Even one that "caps" all dependencies is susceptible to bugs creeping into dependencies, etc. If you provide a "known working version lockfile" perhaps, but in general, as long as the library is being actively developed, this shouldn't be a problem.

matthewfeickert:
> Maybe being very explicit here could help users who are getting started with Python and have not had to deal with specifying versions before.

@alexander-held This is somewhat the reason that I put them in a `requirements.txt` as an example to indicate that there was a relationship between them. I'd probably prefer to not give too many commands here though.

> To me this looks more like a bug in Click that happened to get added in version 8 than an intentional change. Once it's discovered I think you might want to do a bug report to Click.

@henryiii if you have time can you elaborate a bit more on this (or maybe over on Issue #1506)? There was some [past discussion on the IRIS-HEP Slack](https://iris-hep.slack.com/archives/CDGH6HKMG/p1625068565357600) where it seemed like this was our fault due to using a weakref.

> If you depend on something, that something could change - or even just not work on a new platform. Even one that "caps" all dependencies is susceptible to bugs creeping into dependencies, etc.

Here you're talking about users needing to be responsible for controlling their own dependencies for an application, correct? If so, that is what I was trying to convey.

> If you provide a "known working version lockfile" perhaps, but in general, as long as the library is being actively developed, this shouldn't be a problem.

I agree that most of the time it shouldn't be an issue, but we're running into this now and while we've been busy with conferences the last few weeks `pyhf` is decently actively developed and Alex's original issue is well over two weeks old at this point (though this is also the first time we've run into something like this).

matthewfeickert:
Relevant RTD page: https://pyhf--1529.org.readthedocs.build/en/1529/faq.html#i-installed-an-old-pyhf-release-from-pypi-why-am-i-getting-an-error-from-a-dependency

PR #1536, opened 2021-07-27
feat: Propagate model parameter names to optimizers

### LINKED ISSUE #1099: Propagating model parameter names to MINUIT
# Description

It would be great to propagate model parameter names to `iminuit` when using that backend. Here is an example:

```python
import pyhf

)
print(f"model parameter names: {model.config.par_order}")
pyhf.set_backend("numpy", pyhf.optimize.minuit_optimizer())
_, result_obj = pyhf.infer.mle.fit(data, model, return_result_obj=True)
print(f"minuit parameter names: {[param.name for param in result_obj.minuit.params]}")
result_obj.minuit.minos("x0")
``` 

results in
```
model parameter names: ['mu', 'uncorr_bkguncrt']
minuit parameter names: ['x0', 'x1', 'x2']
```

Running MINOS requires translating the model parameters into `xi`. Similarly, further inspection of the `iminuit` results object would be easier with proper parameter labels.

See above, simplifies inspecting `iminuit` results / running MINOS.

Propagating of parameter names to `iminuit`. See [this](https://github.com/alexander-held/cabinetry/blob/bb21d2d39234eed3a313cbfa57cb018313bd4534/src/cabinetry/fit.py#L161-L170) for an example.

Manual translation when needed outside of `pyhf`.

For this to work, #867 needs to be resolved. At the moment there is no unique per-bin name defined within `pyhf` for staterrors/shapefactors.

# Additional context
none

kratsg:
# Pull Request Description

Resolves #1099. This passes through parameter names to the optimizers, both scipy and minuit. This is done by adding `ModelConfig.par_names()` which returns the parameter names in a format that has yet to be agreed upon (for indexing n-binned parameters). This function output is threaded through the optimizer framework via the `Optimizer._get_minimizer()` function call. However, scipy does not make use of this, so not much changes except to support the updated function signature, while Minuit is able to take advantage of this.

A format string is allowed as an optional keyword argument so we're not locked into a specific format.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Propagate model parameter names through to the optimizers (primarily Minuit)
   - Additionally allow a customizable format for the parameter names
* Add par_names method to _ModelConfig
* Add tests for _ModelConfig.par_names
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -319,6 +319,31 @@ class _ModelConfig(_ChannelSummaryMixin):
         """
         return self.par_map[name]['slice']
 
+    def par_names(self):
+        """
+        The names of the parameters in the model including binned-parameter indexing.
+
+        Returns:
+            :obj:`list`: Names of the model parameters.
+
+        Example:
+            >>> import pyhf
+            >>> model = pyhf.simplemodels.uncorrelated_background(
+            ...     signal=[12.0, 11.0], bkg=[50.0, 52.0], bkg_uncertainty=[3.0, 7.0]
+            ... )
+            >>> model.config.par_names()
+            ['mu', 'uncorr_bkguncrt[0]', 'uncorr_bkguncrt[1]']
+        """
+        _names = []
+        for name in self.par_order:
+            _npars = self.param_set(name).n_parameters
+            if _npars == 1:
+                _names.append(name)
+                continue
+
+            _names.extend([f'{name}[{i}]' for i in range(_npars)])
+        return _names
+
     def param_set(self, name):
         """
         The :class:`~pyhf.parameters.paramset` for the model parameter ``name``.
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -324,7 +324,7 @@ class _ModelConfig(_ChannelSummaryMixin):
         The names of the parameters in the model including binned-parameter indexing.
 
         Args:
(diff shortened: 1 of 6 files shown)


kratsg:
Currently, this does `[mu, uncorr_bkguncert[0], uncorr_bkguncert[1]]` -- but we could/should discuss the various naming schema allowed. One could consider allowing the user to choose how the naming schema is done by allowing them to provide a format string.

src/pyhf/pdf.py
```diff
@@ -319,6 +319,36 @@ def par_slice(self, name):
         """
         return self.par_map[name]['slice']
 
+    def par_names(self, fstring='{name}[{index}]'):
+        """
+        The names of the parameters in the model including binned-parameter indexing.
+
+        Args:
+            fstring (:obj:`str`): Format string for the parameter names using `name` and `index` variables. Default: '{name}[{index}]'.
```
matthewfeickert, line 327:
```suggestion
            fstring (:obj:`str`): Format string for the parameter names using ``name`` and ``index`` variables. Default: ``'{name}[{index}]'``.
```

For the render https://pyhf--1536.org.readthedocs.build/en/1536/_generated/pyhf.pdf._ModelConfig.html#pyhf.pdf._ModelConfig.par_names I think these should be using the double ticks to look right

matthewfeickert (commented):
Minor comments (which I'll check now on the formatting and merge if it looks okay), but this all looks good.

matthewfeickert (approved):
Yeah, as that checks out in the docs build this all LGTM. Thanks @kratsg!

PR #1539, opened 2021-07-31
refactor: Use absolute imports over explicit relative imports

matthewfeickert:
# Description

`pyhf` has used explicit relative imports for all of the development. Now that the module structure is more established, this PR changes them all to absolute imports. Use of absolute imports is easier for probably everyone to read (and probably easier for any new developer to learn the module structure).

The fact that 

```console
$ git grep "from \."
```

returns nothing now and that the CI is passing means that this all worked as expected, so this should be hopefully easy to review, unless there are concerns on this change from a developer perspective. :+1: 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Change all explicit relative imports to absolute imports
   - Provides clearer more explicit imports for new developers
```

### DIFF
--- a/src/pyhf/__init__.py
+++ b/src/pyhf/__init__.py
@@ -1,8 +1,8 @@
-from .tensor import BackendRetriever as tensor
-from .optimize import OptimizerRetriever as optimize
-from ._version import version as __version__
-from .exceptions import InvalidBackend, InvalidOptimizer, Unsupported
-from . import events
+from pyhf.tensor import BackendRetriever as tensor
+from pyhf.optimize import OptimizerRetriever as optimize
+from pyhf._version import version as __version__
+from pyhf.exceptions import InvalidBackend, InvalidOptimizer, Unsupported
+from pyhf import events
 
 tensorlib = None
 optimizer = None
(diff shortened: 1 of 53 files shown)


kratsg:
Is there a corresponding PEP? I prefer relative imports.

matthewfeickert:
> Is there a corresponding PEP?

No, none that I know of. I think this comes down just to personal preferences.

> I prefer relative imports.

Yeah, I used to as well.

I think that they are really useful when you're building a library out right from the start, but after something is somewhat established (and you're not going to be moving modules around often) it seems far easier for readability to have the imports be the same inside of the library as a user would have outside.

I'm not sure on this, but seems like it could be also easier to track down code. I remember that when working on the CLI sections of the `contrib` module additions I got confused a few times about where the `.`s were having me import from.

As I think this is a matter of preference / views on what is better readability, can you share more on your preferences for explicit relative?

alexander-held:
> Is there a corresponding PEP?

I was curious about this. The [imports section of PEP 8](https://www.python.org/dev/peps/pep-0008/#imports) says
> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured

and

> However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose

PEP 8 [used to discourage relative imports more strongly](https://stackoverflow.com/a/4209771).

The [Google style guide section](https://google.github.io/styleguide/pyguide.html#22-imports) says
> Do not use relative names in imports. Even if the module is in the same package, use the full package name. This helps prevent unintentionally importing a package twice.

I personally think that @matthewfeickert's argument
> it seems far easier for readability to have the imports be the same inside of the library as a user would have outside

is a good one, and may make a library more accessible to new users.

kratsg (approved):

PR #1546, opened 2021-08-04
feat: Sort `_ChannelSummaryMixin.channel_nbins` by keys to match `channels` order

### LINKED ISSUE #1545: Different ordering of channels between model.config.channels and mode.config.channel_nbins
# Description

We've recently observed that the ordering of channels outputted from `model.config.channels` differs from the order obtained with `model.config.channel_nbins`. This isn't really a bug, but more a quirk which we thought would be worth bringing to the attention of the developers. We ran into some issues when breaking up the `expected_data` list into individual channels by looping through the `model.config.channel_nbins` ordering, rather than the `model.config.channels` ordering (the `model.config.channels` order matches that of the model). Hopefully this issue helps save another user some time in the future, at very least. 

# Expected Behavior

We expected that the order of the channels in the two dictionaries from `model.config.channels` and `model.config.channel_nbins` would be the same. 

# Actual Behavior

The ordering of the channels is different. As an example, we are generating workspaces using two categories, and we're separating our data by year (2016, 2017, and 2018). This gives us six channels total. The outputs are: 
```
(Pdb) model.config.channels
['vbf_channel_16_high_cat', 'vbf_channel_16_low_cat', 'vbf_channel_17_high_cat', 'vbf_channel_17_low_cat', 'vbf_channel_18_high_cat', 'vbf_channel_18_low_cat']

(Pdb) model.config.channel_nbins
{'vbf_channel_16_low_cat': 12, 'vbf_channel_16_high_cat': 18, 'vbf_channel_17_low_cat': 12, 'vbf_channel_17_high_cat': 18, 'vbf_channel_18_low_cat': 12, 'vbf_channel_18_high_cat': 18}
```
I believe that `model.config.channels` is possibly re-ordering the channels so that the names are in alphabetical order. I have not confirmed this, though. The workspace .json file is filled with the ordering produced by `model.config.channel_nbins`. 

# Steps to Reproduce

I'm using pyhf version 0.6.2 along with python 3.8.8. 
I can make a dummy workspace for this issue, but I thought that since this is a pretty small issue, it might be overkill. Please let me know if this would be helpful, though. 

# Checklist

- [X] Filled out the Description, Expected Behavior, Actual Behavior, and Steps to Reproduce sections above or have edited/removed them in a way that fully describes the issue

Thanks! -Rachel

matthewfeickert:
# Description

Resolves #1545 

This PR sorts the `_ChannelSummaryMixin.channel_nbins` `dict` by its keys such that its keys are in the same order as the entries of the `_ChannelSummaryMixin.channels` `list`. This is done for aesthetics mostly and for quick checks, as users should rely on `dict` keys not `dict` ordering. It additionally adds a test.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Sort _ChannelSummaryMixin.channel_nbins dict by keys to match the order of _ChannelSummaryMixin.channels
* Add a test for the order of channel_nbins keys matching the order of channels entries
```

### DIFF
--- a/tests/test_mixins.py
+++ b/tests/test_mixins.py
@@ -53,4 +53,4 @@ def test_channel_nbins_sorted_as_channels(spec):
     spec["channels"][-1]["name"] = "a_make_first_in_sort_channel2"
     mixin = pyhf.mixins._ChannelSummaryMixin(channels=spec["channels"])
     assert mixin.channels == ["a_make_first_in_sort_channel2", "channel1"]
-    assert mixin.channel_nbins == {"a_make_first_in_sort_channel2": 2, "channel1": 2}
+    assert list(mixin.channel_nbins.keys()) == mixin.channels
--- a/tests/test_mixins.py
+++ b/tests/test_mixins.py
@@ -45,3 +45,12 @@ def test_channel_summary_mixin_empty():
     assert mixin.modifiers == []
     assert mixin.parameters == []
     assert mixin.samples == []
+
+
+def test_channel_nbins_sorted_as_channels(spec):
+    assert "channels" in spec
+    spec["channels"].append(spec["channels"][0].copy())
+    spec["channels"][-1]["name"] = "a_make_first_in_sort_channel2"
+    mixin = pyhf.mixins._ChannelSummaryMixin(channels=spec["channels"])
+    assert mixin.channels == ["a_make_first_in_sort_channel2", "channel1"]
+    assert mixin.channel_nbins == {"a_make_first_in_sort_channel2": 2, "channel1": 2}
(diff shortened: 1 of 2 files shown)


tests/test_mixins.py
```diff
@@ -45,3 +45,12 @@ def test_channel_summary_mixin_empty():
     assert mixin.modifiers == []
     assert mixin.parameters == []
     assert mixin.samples == []
+
+
+def test_channel_nbins_sorted_as_channels(spec):
+    assert "channels" in spec
+    spec["channels"].append(spec["channels"][0].copy())
+    spec["channels"][-1]["name"] = "a_make_first_in_sort_channel2"
```
matthewfeickert, line 53: @kratsg there is probably a more elegant way to do this, so please give recommendations and ask for changes here.

kratsg (approved):

Issue #1550, opened 2021-08-11
expose fitted parameter values of implicit fits in test statistic calls


lhenkelm:
# Description

The proposal is to change the API of documented functions and classes in `pyhf.infer` such that users can directly access the best-fit model parameters that implicit fits have found, toggleable by a keyword flag. Here "implicit fits" means the calls to `pyhf.infer.mle.fit` and `pyhf.infer.mle.fixed_poi_fit` in the test statistic implementations (i.e. in [`pyhf.infer.test_statistic._tmu_like`](https://github.com/scikit-hep/pyhf/blob/4b66cd2cfa0bd4bb53b14c525118d86567be8257/src/pyhf/infer/test_statistics.py#L34) and [`pyhf.infer.calculators.generate_asimov`](https://github.com/scikit-hep/pyhf/blob/4b66cd2cfa0bd4bb53b14c525118d86567be8257/src/pyhf/infer/calculators.py#L32)).
"Model parameters" means both the NPs, and (where it is being optimised) the POI.

## Related ~~problem~~ inconvenience
I had encountered some funny/unexpected p-values coming out of `pyhf.infer.hypotest`, and I wanted to inspect (and plot predicted physical distributions for), the parameter values that gave rise to the  unexpected p-values.
This is not possible in the current API, so I re-ran the fits myself to get the parameter values, which is slow on large and complex models, and has some small risk of divergence creeping in between my user-implemented calls to the fits and the internal calls.

### The solution I'd like
The documented test statistic functions in `pyhf.infer.test_statistics` (`qmu`, `tmu`, `qmu_tilde`, `tmu_tilde`, `q0`) should accept 
an optional kwarg which causes the function to return a tuple of the value of the test statistic and the tensor of fitted parameters. ([`pyhf.infer.test_statistic._tmu_like`](https://github.com/scikit-hep/pyhf/blob/4b66cd2cfa0bd4bb53b14c525118d86567be8257/src/pyhf/infer/test_statistics.py#L34) already has such a kwarg)

Similarly,  [`pyhf.infer.calculators.generate_asimov`](https://github.com/scikit-hep/pyhf/blob/4b66cd2cfa0bd4bb53b14c525118d86567be8257/src/pyhf/infer/calculators.py#L32) should have the same kwarg, 
toggling the returning of the asimov parameters alongside the asimov data if `True`.

[...]

PR #1554, opened 2021-08-19
WIP: expose fitted parameter values of implicit fits in test statistic calls

lhenkelm:
# Description

Implements issue #1550.

Functions in `pyhf.infer` that call `pyhf.infer.mle.fit` or `pyhf.infer.mle.fixed_poi_fit` (except `pyhf.infer.upperlimit`) return the best-fit parameters if they recieve `return_fitted_pars=True` as a kwarg. The `pyhf.infer.calculators.AymptoticCalculator` has an attribute holding the best-fit parameters of the most recent five fits. 

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/infer/calculators.py
+++ b/src/pyhf/infer/calculators.py
@@ -12,6 +12,7 @@ from .. import get_backend
 from . import utils
 import tqdm
 
+from dataclasses import dataclass
 import logging
 
 log = logging.getLogger(__name__)
@@ -202,6 +203,18 @@ class AsymptoticTestStatDistribution:
         )
 
 
+@dataclass(frozen=True)
+class BestFitParameters:
+    """Fitted model parameters of the fits happening in ``AsymptoticCalculator.teststatistic``"""
+
+    # ignore "F821 undefined name 'Tensor'" so as to avoid typing.Any
+    asimov_pars: 'Tensor'  # noqa: F821
+    free_fit_to_data: 'Tensor'  # noqa: F821
+    free_fit_to_asimov: 'Tensor'  # noqa: F821
+    fixed_poi_fit_to_data: 'Tensor'  # noqa: F821
+    fixed_poi_fit_to_asimov: 'Tensor'  # noqa: F821
+
+
 class AsymptoticCalculator:
     """The Asymptotic Calculator."""
 
@@ -266,6 +279,21 @@ class AsymptoticCalculator:
         self.calc_base_dist = calc_base_dist
         self.sqrtqmuA_v = None
 
+    @property
+    def fitted_pars(self):
+        """
+        Best-fit parameters of the fits that have been run in self.teststatistic.
+
+        The parameters are collected in a ~pyhf.infer.BestFitParameters instance.
+        """
+        try:
+            return self._fitted_pars
+        except AttributeError:
+            if self.sqrtqmuA_v is None:
+                raise RuntimeError("need to call .teststatistic(poi_test) first")
+            else:
+                raise
+
     def distributions(self, poi_test):
         r"""
         Probability distributions of the test statistic, as defined in
@@ -311,8 +339,12 @@ class AsymptoticCalculator:
         return sb_dist, b_dist
 
     def teststatistic(self, poi_test):
-        """
+        r"""
         Compute the test statistic for the observed data under the studied model.
+
+        The fitted parameters of the five fits that are implicitly ran at every call
+        of this method are afterwards accessible through ``self.fitted_pars``,
+        which is a ~pyhf.infer.BestFitParameters instance.
 
         Example:
 
@@ -328,6 +360,11 @@ class AsymptoticCalculator:
             >>> asymptotic_calculator.teststatistic(mu_test)
             array(0.14043184)
 
+            access the best-fit parameters
+            (here: :math:`\hat{\mu}` and :math:`\hat{\theta}` fitted to the asimov dataset):
+            >>> asymptotic_calculator.fitted_pars.free_fit_to_asimov
+            array([0.        , 1.00304893, 0.96263365])
+
         Args:
             poi_test (:obj:`float` or :obj:`tensor`): The value for the parameter of interest.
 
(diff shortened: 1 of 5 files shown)


matthewfeickert:
@lhenkelm while you're working on this can you add some commentary as well. It isn't really clear why all of what is in here so far is strictly needed.

lhenkelm:
@matthewfeickert the changes are those needed to get access to the ML estimates of the parameters in a call to  `pyhf.infer.hypotest`. The calls stack up like so: 
`hypotest` calls `AsymptoticCalculator.teststatistic` calls `generate_asimov_data` and (e.g.) `qmu_tilde`: 
 - So it is necessary that `generate_asimov_data` and all the test statistics are able to return fitted params. (https://github.com/scikit-hep/pyhf/pull/1554/commits/04b0c3babc214b1f0a87285d8bfcc0e85e952acc, https://github.com/scikit-hep/pyhf/pull/1554/commits/a2f1c86a1a5dff35728b62b7d2a76eaa68a3676a)
 - These params then need to be handed back up through the calculator, which is what the new `AsymptoticCalculator.fitted_pars` is for (https://github.com/scikit-hep/pyhf/pull/1554/commits/e804185798428931d5e618582c14fc77ebe7478b).
 -  Then finally hypotest needs the option to return the parameters as well (https://github.com/scikit-hep/pyhf/pull/1554/commits/84a8af87e734fd11644d41d313f730520a01c78c). 

So the changes to the implementation code are minimal. The bulk of the new lines are tests and docstrings,
which IMO are necessary because all the changes are in the user-visible behavior of documented parts of the API.

I tried to keep the new tests small/generic enough so that they are robust against new changes while covering all of the new behavior. But you guys have the experience of what parts change a lot and how, so let me know if the tests are overbearing! 

For the docstrings I tried to be super explicit about describing how the return values change with new option and give plenty of examples for people that find those more useful. So in that sense I think these are necessary changes, but if you think it makes the docstrings too bloated, please tell me!

lhenkelm:
regarding the tests: they all pass locally, except one unrelated test in `test_notebooks.py` that also fails when I run the test suite on master.

lukasheinrich (approved):

lukasheinrich:
I had a look at this again and this looks ok to me to go in. Thanks @lhenkelm

matthewfeickert:
Once I get release `v0.6.3` out the door I'll circle back to review this so we can get it in. Though if Lukas is already happy then it should be easy to review.

src/pyhf/infer/calculators.py
```diff
@@ -188,6 +207,21 @@ def expected_value(self, nsigma):
         )
 
 
+@dataclass(frozen=True)
+class BestFitParameters:
```
lukasheinrich, line 211: This could maybe be renamed to be something like `TestStatisticFitResults` to separate it from a single Fit

lukasheinrich:
One thing that we  should discuss is whether we can somehow unify the "data classes" with cabinetry 

i.e.  BestFitParams (or `TestStatisticFitResults` see my comment above), `OptimizerResult`,cabinetry's `FitResult` etc

alexander-held:
I think it would be great to harmonize where possible. Being able to do a `infer.mle.fit` fit and plug the result straight into a `cabinetry` visualization (instead of having to use the `cabinetry.fit` API, or having to pack up the fit result into the object `cabinetry` expects) seems very useful. I would be happy to try and find a common solution to the packaging of the return values.

src/pyhf/infer/calculators.py
lhenkelm: To my ears `TestStatisticFitResults` sounds more like the theta-hat, theta-hat-hat tuple from a test statistic call.
Maybe `ConfidenceLevelFitResults` or `LimitFitResults` or `HypoTestFitResults`?

tests/test_infer.py
```diff
@@ -157,6 +157,183 @@ def test_hypotest_return_expected_set(tmpdir, hypotest_args, test_stat):
     assert check_uniform_type(result[3])
 
 
+@pytest.mark.parametrize(
```
kratsg, line 160: a lot of tests here are asserting a lot of extra behavior that's not needed as it's covered by other tests. Just try to keep it to only one or two assertions per test if possible. The test should only be looking at the calculators, so it shouldn't care about the other parts.

src/pyhf/infer/calculators.py
```diff
@@ -252,6 +286,22 @@ def __init__(
...
+        Returns:
+            ~pyhf.infer.calculators.HypoTestFitResults:  The collection of fitted parameter tensors.
+        """
+        try:
+            return self._fitted_pars
+        except AttributeError:
+            if self.sqrtqmuA_v is None:
+                raise RuntimeError("need to call .teststatistic(poi_test) first")
+            raise
+
```
kratsg, line 304: This is not a needed function. Just rename `self._fitted_pars` to `self.fitted_pars` and it's accessible there, and harmonizes with `self.sqrtqmuA_v`.

src/pyhf/infer/calculators.py
```diff
@@ -314,6 +368,12 @@ def teststatistic(self, poi_test):
             >>> asymptotic_calculator.teststatistic(mu_test)
             array(0.14043184)
 
+            Access the best-fit parameters afterwards:
+            (here: :math:`\hat{\mu}` and :math:`\hat{\theta}` fitted to the asimov dataset):
+
+            >>> asymptotic_calculator.fitted_pars.free_fit_to_asimov
```
kratsg, line 374: add in `>>> asymptotic_calculator.fitted_pars` so people can be assured it's a `HypoTestFitResult` object.

kratsg (changes requested):

tests/test_infer.py
lhenkelm: Done. Thanks for reviewing! I took out the asserts about other return values, and collapsed the series of tests into one parameterized one. Let me know if its still asserting too much.

src/pyhf/infer/calculators.py
lhenkelm: I am a bit worried that without an error message that clearly says "need to call .teststatistic(poi_test) first" (like the other functions in the calculator that depend on it having been called) it could be very confusing for people not familiar with how the class works. But I see the point about it harmonizing, so now it is initialised to `None`, like `self.sqrtqmuA_v`

src/pyhf/infer/calculators.py
lhenkelm: Done

src/pyhf/infer/calculators.py
lukasheinrich: yeah seems like a good use for `@property` actually. maybe `sqrtqmuA_v` shold be guarded ilke tthat as well?

matthewfeickert:
Thanks for the PR @lhenkelm!

PR #1530, opened 2021-07-21
fix: Remove WeakList in favor of built-ins

matthewfeickert:
# Description

Resolves #1506

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR
```
* Rewrite the events code to handle weakrefs more properly
   - This is needed to work with Click v8.0+
* Fix up tests that were incorrectly written due to API changes in unittest.mock
```

PR #1556, opened 2021-08-21
docs: Promote comments to docstring examples for pyhf.event

### LINKED ISSUE #1555: Remove Python 2 syntax from events examples in comments
# Description

In `pyhf.events` there are a two examples of Python 2 syntax being used for 

https://github.com/scikit-hep/pyhf/blob/29c3df0e23a428004a065aed61cefb6a526a7332/src/pyhf/events.py#L46-L53

and 

https://github.com/scikit-hep/pyhf/blob/29c3df0e23a428004a065aed61cefb6a526a7332/src/pyhf/events.py#L69-L87

These examples should be updated to use Python 3 syntax.

Also the examples are wrong themselves. For example, the first example should be

```python
>>> import pyhf
>>> @pyhf.events.subscribe('myevent')
... def test(a,b):
...     print(a+b)
... 
>>> pyhf.events.trigger("myevent")(1,2)
3
```

pineappleslikei:
# Description

Fixes #1555 

Promote event examples to docstrings and use Python 3 syntax.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Promote pyhf.event.subscribe and pyhf.event.register examples to docstring examples
* Use Python 3 syntax in examples
```

### DIFF
--- a/src/pyhf/events.py
+++ b/src/pyhf/events.py
@@ -89,7 +89,9 @@ def subscribe(event):
     """
     Subscribe a function or object method as a callback to an event.
 
-    Note: this is meant to be used as a decorator.
+    .. note::
+
+     This is meant to be used as a decorator.
 
     Args:
         event (:obj:`str`): The name of the event to subscribe to.
@@ -122,7 +124,9 @@ def register(event):
     Register a function or object method to trigger an event.  This creates two
     events: ``{event_name}::before`` and ``{event_name}::after``.
 
-    Note: this is meant to be used as a decorator.
+    .. note::
+
+     This is meant to be used as a decorator.
 
     Args:
         event (:obj:`str`): The name of the event to subscribe to.
--- a/src/pyhf/events.py
+++ b/src/pyhf/events.py
@@ -47,9 +47,9 @@ def subscribe(event):
     #
     # >>> @pyhf.events.subscribe('myevent')
     # ... def test(a,b):
-    # ...   print a+b
+    # ...   print(a+b)
     # ...
-    # >>> pyhf.events.trigger_myevent(1,2)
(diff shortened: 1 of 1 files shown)


matthewfeickert (approved):
Thanks for this PR @pineappleslikei. What you had was fine :+1:, but as this is getting changed I took the opportunity to promote the examples to docstring examples as well.

I'm going to hold off on merging this in until PR #1530 is merged though as technically some of these calls could error without it.

What you have here is fine though, so you won't need to do anything else &mdash; we'll take care of shepherding it through.

kratsg:
So this is tricky, these tests will fail because of what issues that #1530 is fixing, but we also need the comments for the test coverage once the fixes are in place. Since we'd still want to acknowledge your contribution, we can either have you merge into the target branch of #1530 or add you as a contributor / acknowledgements for the #1530 PR.

matthewfeickert:
> You will need to rebase on top of #1530 once that goes in, and we can then get this PR in as well once you make the changes.

This PR was done from @pineappleslikei's `master` branch so we can't rebase (:disappointed:) else it will close the PR. I just manually resolved the merge conflict and merged in `pyhf`'s `master` branch. I'll do a quick local test to make sure that this all looks good if we ever create a "dev docs" section of the docs as we don't currently document `pyhf.events` and then this should be good to merge with @kratsg's approval.

kratsg (approved):

matthewfeickert:
This is what the docs would look like if we implemented something like Issue #1558:

if_dev_docs

matthewfeickert:
Thanks very much for your contribution, @pineappleslikei!

Out of curiosity, how did you find our project to contribute? :slightly_smiling_face:

PR #1560, opened 2021-08-25
feat: par_names handles non-scalar modifiers with 1 parameter

kratsg:
# Pull Request Description

Should fully resolve the last remaining parts of #867. There are two issues handled:

- `par_names` did allow for customizable fstring, but this would never get passed through to the underlying optimizers that use it, so it's being dropped as the benefit of configurability is outweighed by the extra work on the user to maintain and bookkeep
- `par_names` now correctly uses the fact that parameter sets which are non-scalar but only have one parameter (e.g. channel of 1 bin) will still be named with the index `{parameter}[0]` rather than just `{parameter}` to clarify that it's a non-scalar parameter explicitly (but also somewhat implicitly).

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* change par_names API to drop fstring configurability
* modify par_names to handle non-scalar parameters in 1-bin channels
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -319,12 +319,9 @@ class _ModelConfig(_ChannelSummaryMixin):
         """
         return self.par_map[name]['slice']
 
-    def par_names(self, fstring='{name}[{index}]'):
+    def par_names(self):
         """
         The names of the parameters in the model including binned-parameter indexing.
-
-        Args:
-            fstring (:obj:`str`): Format string for the parameter names using ``name`` and ``index`` variables. Default: ``'{name}[{index}]'``.
 
         Returns:
             :obj:`list`: Names of the model parameters.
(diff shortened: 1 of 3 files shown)


kratsg:
As the changes are pretty minor (especially given the original functionality was ok in the originating PR), I'm going to merge this in so @matthewfeickert can focus on getting the release out.

matthewfeickert:
@kratsg as this is API breaking do we know if this is problematic for `cabinetry` at all? cc @alexander-held.

I'm optimistic that is isn't a problem, given that you only had to _add_ a test and not remove any.

kratsg:
not API breaking as this wasn't in the v0.6.2 API. It's just changes to the API on the main branch.

matthewfeickert:
Ah that's fantastic. Thanks for clarifying that.

PR #1567, opened 2021-08-27
feat: Drop uproot3 for uproot4 for writing ROOT files

### LINKED ISSUE #1566: Drop uproot3 dependency and use uproot4 for writing
# Description

Thanks to @jpivarski's awesome and heroic work [`uproot` `v4.1.0`](https://github.com/scikit-hep/uproot4/releases/tag/4.1.0) can now write ROOT files! :rocket: 

This means that once we get everything working we can drop the `uproot3` dependency for the `xmlio` extra

https://github.com/scikit-hep/pyhf/blob/ce2ffabaaeadf2733d4d61284d750933f34cee28/setup.py#L11-L14

and in the tests

https://github.com/scikit-hep/pyhf/blob/ce2ffabaaeadf2733d4d61284d750933f34cee28/.github/workflows/dependencies-head.yml#L66-L89

and in the lower bound tests

https://github.com/scikit-hep/pyhf/blob/ce2ffabaaeadf2733d4d61284d750933f34cee28/lower-bound-requirements.txt#L8-L10

This shouldn't be too bad as there is only a few instances of `uproot3` being used at the moment

```console
$ git grep "uproot3" src/
src/pyhf/writexml.py:import uproot3 as uproot
src/pyhf/writexml.py:from uproot3_methods.classes import TH1
```

matthewfeickert:
# Description

Resolves #1566

Drop all usages of `uproot3` and only use `uproot` `v4.1.1+` for all reading and writing of ROOT files.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Remove all uproot3 dependency and use uproot4 for all reading and writing of ROOT files
* Require uproot>=4.1.1 in 'xmlio' extra
* Require uproot==4.1.1 in lower-bound-requirements.txt
* Remove uproot3 HEAD of dependencies testing workflow
* Add test to ensure pyhf.writexml writes histograms to ROOT files as expected
```

### DIFF
--- a/lower-bound-requirements.txt
+++ b/lower-bound-requirements.txt
@@ -6,8 +6,7 @@ jsonschema==3.0.0
 jsonpatch==1.15
 pyyaml==5.1
 # xmlio
-uproot3==3.14.1
-uproot==4.0.0
+uproot==4.1.0
 # minuit
 iminuit==2.4.0
 # tensorflow
--- a/lower-bound-requirements.txt
+++ b/lower-bound-requirements.txt
@@ -6,7 +6,7 @@ jsonschema==3.0.0
 jsonpatch==1.15
 pyyaml==5.1
 # xmlio
-uproot==4.1.0
+uproot==4.1.1
 # minuit
 iminuit==2.4.0
 # tensorflow
--- a/src/pyhf/writexml.py
+++ b/src/pyhf/writexml.py
@@ -45,11 +45,12 @@ def _make_hist_name(channel, sample, modifier='', prefix='hist', suffix=''):
     return f"{prefix}{middle}{suffix}"
 
 
-def _export_root_histogram(histname, data):
-    if histname in _ROOT_DATA_FILE:
-        raise KeyError(f"Duplicate key {histname} being written.")
-    hist_uproot_model = uproot.to_writable((np.asarray(data), np.arange(len(data) + 1)))
-    _ROOT_DATA_FILE[histname] = hist_uproot_model
+def _export_root_histogram(hist_name, data):
+    if hist_name in _ROOT_DATA_FILE:
+        raise KeyError(f"Duplicate key {hist_name} being written.")
+    _ROOT_DATA_FILE[hist_name] = uproot.to_writable(
+        (np.asarray(data), np.arange(len(data) + 1))
+    )
 
 
 # https://stackoverflow.com/a/4590052
--- a/src/pyhf/writexml.py
+++ b/src/pyhf/writexml.py
@@ -234,7 +234,7 @@ def build_sample(spec, samplespec, channelname):
     attrs = {
         'Name': samplespec['name'],
         'HistoName': histname,
-        'InputFile': _ROOT_DATA_FILE._path,
+        'InputFile': _ROOT_DATA_FILE.file_path,
         'NormalizeByTheory': 'False',
     }
     sample = ET.Element('Sample', **attrs)
@@ -253,7 +253,7 @@ def build_sample(spec, samplespec, channelname):
 
 def build_data(obsspec, channelname):
     histname = _make_hist_name(channelname, 'data')
-    data = ET.Element('Data', HistoName=histname, InputFile=_ROOT_DATA_FILE._path)
+    data = ET.Element('Data', HistoName=histname, InputFile=_ROOT_DATA_FILE.file_path)
 
     observation = next((obs for obs in obsspec if obs['name'] == channelname), None)
     _export_root_histogram(histname, observation['data'])
@@ -262,7 +262,7 @@ def build_data(obsspec, channelname):
 
 def build_channel(spec, channelspec, obsspec):
     channel = ET.Element(
-        'Channel', Name=channelspec['name'], InputFile=_ROOT_DATA_FILE._path
+        'Channel', Name=channelspec['name'], InputFile=_ROOT_DATA_FILE.file_path
     )
     if obsspec:
         data = build_data(obsspec, channelspec['name'])
--- a/src/pyhf/writexml.py
+++ b/src/pyhf/writexml.py
@@ -49,7 +49,6 @@ def _export_root_histogram(histname, data):
     if histname in _ROOT_DATA_FILE:
         raise KeyError(f"Duplicate key {histname} being written.")
     hist_uproot_model = uproot.to_writable((np.asarray(data), np.arange(len(data) + 1)))
-    hist_uproot_model._members["fName"] = histname
     _ROOT_DATA_FILE[histname] = hist_uproot_model
 
 
--- a/src/pyhf/writexml.py
+++ b/src/pyhf/writexml.py
@@ -6,9 +6,7 @@ import pkg_resources
 import xml.etree.ElementTree as ET
 import numpy as np
 
-# TODO: Move to uproot4 when ROOT file writing is supported
-import uproot3 as uproot
-from uproot3_methods.classes import TH1
+import uproot
 
 from pyhf.mixins import _ChannelSummaryMixin
 
@@ -48,11 +46,11 @@ def _make_hist_name(channel, sample, modifier='', prefix='hist', suffix=''):
 
 
 def _export_root_histogram(histname, data):
-    hist = TH1.from_numpy((np.asarray(data), np.arange(len(data) + 1)))
-    hist._fName = histname
     if histname in _ROOT_DATA_FILE:
         raise KeyError(f"Duplicate key {histname} being written.")
-    _ROOT_DATA_FILE[histname] = hist
+    _ROOT_DATA_FILE[histname] = np.histogram(
+        np.asarray(data), bins=np.arange(len(data) + 1)
+    )
 
 
 # https://stackoverflow.com/a/4590052
--- a/src/pyhf/writexml.py
+++ b/src/pyhf/writexml.py
@@ -48,9 +48,7 @@ def _make_hist_name(channel, sample, modifier='', prefix='hist', suffix=''):
 def _export_root_histogram(histname, data):
     if histname in _ROOT_DATA_FILE:
         raise KeyError(f"Duplicate key {histname} being written.")
-    _ROOT_DATA_FILE[histname] = np.histogram(
-        np.asarray(data), bins=np.arange(len(data) + 1)
-    )
+    _ROOT_DATA_FILE[histname] = (np.asarray(data), np.arange(len(data) + 1))
 
 
 # https://stackoverflow.com/a/4590052
--- a/src/pyhf/writexml.py
+++ b/src/pyhf/writexml.py
@@ -48,7 +48,9 @@ def _make_hist_name(channel, sample, modifier='', prefix='hist', suffix=''):
 def _export_root_histogram(histname, data):
     if histname in _ROOT_DATA_FILE:
         raise KeyError(f"Duplicate key {histname} being written.")
-    _ROOT_DATA_FILE[histname] = (np.asarray(data), np.arange(len(data) + 1))
+    hist_uproot_model = uproot.to_writable((np.asarray(data), np.arange(len(data) + 1)))
+    hist_uproot_model._members["fName"] = histname
+    _ROOT_DATA_FILE[histname] = hist_uproot_model
 
 
 # https://stackoverflow.com/a/4590052
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -394,7 +394,6 @@ def test_export_data(mocker):
 
 def test_export_duplicate_hist_name(mocker):
     mocker.patch('pyhf.writexml._ROOT_DATA_FILE', new={'duplicate_name': True})
-    mocker.patch.object(pyhf.writexml, 'TH1')
 
     with pytest.raises(KeyError):
         pyhf.writexml._export_root_histogram('duplicate_name', [0, 1, 2])
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -398,7 +398,8 @@ def test_export_data(mocker):
 def test_export_root_histogram(mocker, tmp_path):
     """
     Test that pyhf.writexml._export_root_histogram writes out a histogram
-    in the manner that uproot is expecting
+    in the manner that uproot is expecting and verifies this by reading
+    the serialized file
     """
     mocker.patch("pyhf.writexml._ROOT_DATA_FILE", {})
     pyhf.writexml._export_root_histogram("hist", [0, 1, 2, 3, 4, 5, 6, 7, 8])
@@ -406,6 +407,7 @@ def test_export_root_histogram(mocker, tmp_path):
     with uproot.recreate(tmp_path.joinpath("test_export_root_histogram.root")) as file:
         file["hist"] = pyhf.writexml._ROOT_DATA_FILE["hist"]
 
+    with uproot.open(tmp_path.joinpath("test_export_root_histogram.root")) as file:
         assert file["hist"].values().tolist() == [0, 1, 2, 3, 4, 5, 6, 7, 8]
         assert file["hist"].axis().edges().tolist() == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
         assert file["hist"].name == "hist"
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -395,7 +395,7 @@ def test_export_data(mocker):
     assert pyhf.writexml._ROOT_DATA_FILE.__setitem__.called
 
 
-def test_export_root_histogram(mocker, tmpdir):
+def test_export_root_histogram(mocker, tmp_path):
     """
     Test that pyhf.writexml._export_root_histogram writes out a histogram
     in the manner that uproot is expecting
@@ -403,7 +403,7 @@ def test_export_root_histogram(mocker, tmpdir):
     mocker.patch("pyhf.writexml._ROOT_DATA_FILE", {})
     pyhf.writexml._export_root_histogram("example", [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
 
-    with uproot.recreate(tmpdir.join("test_export_root_histogram.root")) as file:
+    with uproot.recreate(tmp_path.joinpath("test_export_root_histogram.root")) as file:
         file["hist"] = pyhf.writexml._ROOT_DATA_FILE["example"]
         counts, edges = file["hist"].to_numpy()
         values = counts * edges[:-1]
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -1,9 +1,12 @@
+import json
+import logging
+import xml.etree.ElementTree as ET
+
+import pytest
+import uproot
+
 import pyhf
 import pyhf.writexml
-import pytest
-import json
-import xml.etree.ElementTree as ET
-import logging
 
 
 def spec_staterror():
@@ -390,6 +393,25 @@ def test_export_data(mocker):
     assert data.attrib['HistoName']
     assert data.attrib['InputFile']
     assert pyhf.writexml._ROOT_DATA_FILE.__setitem__.called
+
+
+def test_export_root_histogram(mocker, tmpdir):
+    """
+    Test that pyhf.writexml._export_root_histogram writes out a histogram
+    in the manner that uproot is expecting
+    """
+    mocker.patch("pyhf.writexml._ROOT_DATA_FILE", {})
+    pyhf.writexml._export_root_histogram("example", [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+
+    with uproot.recreate(tmpdir.join("test_export_root_histogram.root")) as file:
+        file["hist"] = pyhf.writexml._ROOT_DATA_FILE["example"]
+        counts, edges = file["hist"].to_numpy()
+        values = counts * edges[:-1]
+
+        assert file["hist"].values().tolist() == [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
+        assert counts.tolist() == [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
+        assert edges.tolist() == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        assert values.tolist() == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
 
 
 def test_export_duplicate_hist_name(mocker):
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -405,13 +405,11 @@ def test_export_root_histogram(mocker, tmp_path):
 
     with uproot.recreate(tmp_path.joinpath("test_export_root_histogram.root")) as file:
         file["hist"] = pyhf.writexml._ROOT_DATA_FILE["example"]
-        counts, edges = file["hist"].to_numpy()
-        values = counts * edges[:-1]
+        values, edges = file["hist"].to_numpy()
 
-        assert file["hist"].values().tolist() == [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
-        assert counts.tolist() == [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
-        assert edges.tolist() == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        assert file["hist"].values().tolist() == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
         assert values.tolist() == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+        assert edges.tolist() == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
 
 
 def test_export_duplicate_hist_name(mocker):
--- a/tests/test_export.py
+++ b/tests/test_export.py
@@ -405,12 +405,22 @@ def test_export_root_histogram(mocker, tmp_path):
 
     with uproot.recreate(tmp_path.joinpath("test_export_root_histogram.root")) as file:
         file["hist"] = pyhf.writexml._ROOT_DATA_FILE["hist"]
-        values, edges = file["hist"].to_numpy()
 
         assert file["hist"].values().tolist() == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+        assert file["hist"].axis().edges().tolist() == [
+            0,
+            1,
+            2,
+            3,
+            4,
+            5,
+            6,
+            7,
+            8,
+            9,
+            10,
+        ]
         assert file["hist"].name == "hist"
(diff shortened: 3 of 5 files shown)


matthewfeickert:
I'm a bit stuck as I'm having some trouble understanding why the fllowing traceback isn't finding the directories

```pytb
______________________ test_import_roundtrip[example-one] ______________________

self = <[AttributeError("'MemmapSource' object has no attribute '_fallback'") raised in repr()] MemmapSource object at 0x7fe0b4a33590>

    def _open(self):
        try:
>           self._file = numpy.memmap(self._file_path, dtype=self._dtype, mode="r")

/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/uproot/source/file.py:113: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

subtype = <class 'numpy.memmap'>
filename = '/tmp/pytest-of-runner/pytest-0/test_import_roundtrip_example_0/()'
dtype = dtype('uint8'), mode = 'r', offset = 0, shape = None, order = 'C'

    def __new__(subtype, filename, dtype=uint8, mode='r+', offset=0,
                shape=None, order='C'):
        # Import here to minimize 'import numpy' overhead
        import mmap
        import os.path
        try:
            mode = mode_equivalents[mode]
        except KeyError:
            if mode not in valid_filemodes:
                raise ValueError("mode must be one of %s" %
                                 (valid_filemodes + list(mode_equivalents.keys())))
    
        if mode == 'w+' and shape is None:
            raise ValueError("shape must be given")
    
        if hasattr(filename, 'read'):
            f_ctx = contextlib_nullcontext(filename)
        else:
>           f_ctx = open(os_fspath(filename), ('r' if mode == 'c' else mode)+'b')
E           FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pytest-of-runner/pytest-0/test_import_roundtrip_example_0/()'

/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/numpy/core/memmap.py:225: FileNotFoundError

self = <uproot.source.file.FileResource object at 0x7fe0b4a33650>
file_path = '/tmp/pytest-of-runner/pytest-0/test_import_roundtrip_example_0/()'

    def __init__(self, file_path):
        self._file_path = file_path
        try:
>           self._file = open(self._file_path, "rb")
E           FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pytest-of-runner/pytest-0/test_import_roundtrip_example_0/()'

/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/uproot/source/file.py:37: FileNotFoundError

tmpdir = local('/tmp/pytest-of-runner/pytest-0/test_import_roundtrip_example_0')
toplvl = 'validation/xmlimport_input/config/example.xml'
basedir = 'validation/xmlimport_input/'

    @pytest.mark.parametrize(
        'toplvl, basedir',
        [
            (
                'validation/xmlimport_input/config/example.xml',
                'validation/xmlimport_input/',
            ),
            (
                'validation/xmlimport_input2/config/example.xml',
                'validation/xmlimport_input2',
            ),
            (
                'validation/xmlimport_input3/config/examples/example_ShapeSys.xml',
                'validation/xmlimport_input3',
            ),
        ],
        ids=['example-one', 'example-two', 'example-three'],
    )
    def test_import_roundtrip(tmpdir, toplvl, basedir):
        parsed_xml_before = pyhf.readxml.parse(toplvl, basedir)
        spec = {
            'channels': parsed_xml_before['channels'],
            'parameters': parsed_xml_before['measurements'][0]['config']['parameters'],
        }
        pdf_before = pyhf.Model(spec, poi_name='SigXsecOverSM')
    
        tmpconfig = tmpdir.mkdir('config')
        tmpdata = tmpdir.mkdir('data')
        tmpxml = tmpdir.join('FitConfig.xml')
        tmpxml.write(
            pyhf.writexml.writexml(
                parsed_xml_before,
                tmpconfig.strpath,
                tmpdata.strpath,
                Path(tmpdir.strpath).joinpath('FitConfig'),
            ).decode('utf-8')
        )
>       parsed_xml_after = pyhf.readxml.parse(tmpxml.strpath, tmpdir.strpath)

tests/test_validation.py:846: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/pyhf/readxml.py:347: in parse
    ET.parse(str(Path(rootdir).joinpath(inp))), rootdir, track_progress
src/pyhf/readxml.py:228: in process_channel
    parsed_data = process_data(data[0], rootdir, inputfile, histopath)
src/pyhf/readxml.py:212: in process_data
    data, _ = import_root_histogram(rootdir, inputfile, histopath, histoname)
src/pyhf/readxml.py:61: in import_root_histogram
    f = uproot.open(fullpath)
/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/uproot/reading.py:156: in open
    **options  # NOTE: a comma after **options breaks Python 2
/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/uproot/reading.py:589: in __init__
    file_path, **self._options  # NOTE: a comma after **options breaks Python 2
/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/uproot/source/file.py:109: in __init__
    self._open()
/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/uproot/source/file.py:120: in _open
    self._file_path, **opts  # NOTE: a comma after **opts breaks Python 2
/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/uproot/source/file.py:257: in __init__
    self._open()
/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/uproot/source/file.py:263: in _open
    for x in uproot._util.range(self._num_workers)
/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/uproot/source/file.py:263: in <listcomp>
    for x in uproot._util.range(self._num_workers)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <uproot.source.file.FileResource object at 0x7fe0b4a33650>
file_path = '/tmp/pytest-of-runner/pytest-0/test_import_roundtrip_example_0/()'

    def __init__(self, file_path):
        self._file_path = file_path
        try:
            self._file = open(self._file_path, "rb")
        except uproot._util._FileNotFoundError:
>           raise uproot._util._file_not_found(file_path)
E           FileNotFoundError: file not found
E           
E               '/tmp/pytest-of-runner/pytest-0/test_import_roundtrip_example_0/()'
E           
E           Files may be specified as:
E              * str/bytes: relative or absolute filesystem path or URL, without any colons
E                    other than Windows drive letter or URL schema.
E                    Examples: "rel/file.root", "C:\abs\file.root", "http://where/what.root"
E              * str/bytes: same with an object-within-ROOT path, separated by a colon.
E                    Example: "rel/file.root:tdirectory/ttree"
E              * pathlib.Path: always interpreted as a filesystem path or URL only (no
E                    object-within-ROOT path), regardless of whether there are any colons.
E                    Examples: Path("rel:/file.root"), Path("/abs/path:stuff.root")
E           
E           Functions that accept many files (uproot.iterate, etc.) also allow:
E              * glob syntax in str/bytes and pathlib.Path.
E                    Examples: Path("rel/*.root"), "/abs/*.root:tdirectory/ttree"
E              * dict: keys are filesystem paths, values are objects-within-ROOT paths.
E                    Example: {"/data_v1/*.root": "ttree_v1", "/data_v2/*.root": "ttree_v2"}
E              * already-open TTree objects.
E              * iterables of the above.

/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/site-packages/uproot/source/file.py:39: FileNotFoundError
```

If I run this locally

```console
$ python -m pytest -sx tests/test_validation.py -k test_import_roundtrip[example-one]
...
self = <uproot.source.file.FileResource object at 0x7f7057103820>, file_path = '/tmp/pytest-of-feickert/pytest-42/test_import_roundtrip_example_0/()'

    def __init__(self, file_path):
        self._file_path = file_path
        try:
            self._file = open(self._file_path, "rb")
        except uproot._util._FileNotFoundError:
>           raise uproot._util._file_not_found(file_path)
E           FileNotFoundError: file not found
E           
E               '/tmp/pytest-of-feickert/pytest-42/test_import_roundtrip_example_0/()'
E           
E           Files may be specified as:
E              * str/bytes: relative or absolute filesystem path or URL, without any colons
E                    other than Windows drive letter or URL schema.
E                    Examples: "rel/file.root", "C:\abs\file.root", "http://where/what.root"
E              * str/bytes: same with an object-within-ROOT path, separated by a colon.
E                    Example: "rel/file.root:tdirectory/ttree"
E              * pathlib.Path: always interpreted as a filesystem path or URL only (no
E                    object-within-ROOT path), regardless of whether there are any colons.
E                    Examples: Path("rel:/file.root"), Path("/abs/path:stuff.root")
E           
E           Functions that accept many files (uproot.iterate, etc.) also allow:
E              * glob syntax in str/bytes and pathlib.Path.
E                    Examples: Path("rel/*.root"), "/abs/*.root:tdirectory/ttree"
E              * dict: keys are filesystem paths, values are objects-within-ROOT paths.
E                    Example: {"/data_v1/*.root": "ttree_v1", "/data_v2/*.root": "ttree_v2"}
E              * already-open TTree objects.
E              * iterables of the above.

../../../.pyenv/versions/pyhf-CPU/lib/python3.8/site-packages/uproot/source/file.py:39: FileNotFoundError
```

The directory exists
```
$ file /tmp/pytest-of-feickert/pytest-42/test_import_roundtrip_example_0
/tmp/pytest-of-feickert/pytest-42/test_import_roundtrip_example_0: directory
$ ls /tmp/pytest-of-feickert/pytest-42/test_import_roundtrip_example_0
config  data  FitConfig_channel1.xml  FitConfig.xml  HistFactorySchema.dtd
```

Nothing should have changed in [`src/pyhf/readxml.py`](https://github.com/scikit-hep/pyhf/blob/ce2ffabaaeadf2733d4d61284d750933f34cee28/src/pyhf/readxml.py) so I'm not really sure what's going on from a first pass.

I'll tag @jpivarski in case he can think of something that is obvious that I'm missing, but I'm not expecting you to work on this PR Jim as my guess is that something subtle has changed that I need to fix in `pyhf`.

jpivarski:
The first thing I'm noticing is that the filename has parentheses in it. That might be some string formatting (maybe in Uproot) gone awry. This is the error message for "can't find a file at that file path" and I'd be surprised if your filename really did have parentheses in it.

```
filename = '/tmp/pytest-of-runner/pytest-0/test_import_roundtrip_example_0/()'
```

matthewfeickert:
> The first thing I'm noticing is that the filename has parentheses in it. ...
> This is the error message for "can't find a file at that file path" and I'd be surprised if your filename really did have parentheses in it.
> 
> ```
> filename = '/tmp/pytest-of-runner/pytest-0/test_import_roundtrip_example_0/()'
> ```

@jpivarski Yeah, all the filenames are just `pathlib.Path()` wrapped in `str` (though the `str` bit probably isn't needed), so they should all be properly formed paths.

>  That might be some string formatting (maybe in Uproot) gone awry. 

It seems maybe so(?). If I revert the following diff

```diff
diff --git a/src/pyhf/writexml.py b/src/pyhf/writexml.py
index f1595c57..650c313a 100644
--- a/src/pyhf/writexml.py
+++ b/src/pyhf/writexml.py
@@ -6,9 +6,7 @@
 import xml.etree.ElementTree as ET
 import numpy as np
 
-# TODO: Move to uproot4 when ROOT file writing is supported
-import uproot3 as uproot
-from uproot3_methods.classes import TH1
+import uproot
 
 from pyhf.mixins import _ChannelSummaryMixin
 
@@ -48,11 +46,11 @@ def _make_hist_name(channel, sample, modifier='', prefix='hist', suffix=''):
 
 
 def _export_root_histogram(histname, data):
-    hist = TH1.from_numpy((np.asarray(data), np.arange(len(data) + 1)))
-    hist._fName = histname
     if histname in _ROOT_DATA_FILE:
         raise KeyError(f"Duplicate key {histname} being written.")
-    _ROOT_DATA_FILE[histname] = hist
+    _ROOT_DATA_FILE[histname] = np.histogram(
+        np.asarray(data), bins=np.arange(len(data) + 1)
+    )
 
 
 # https://stackoverflow.com/a/4590052
```

so that I'm using `uproot3` again, then 

````
python -m pytest -sx tests/test_validation.py -k test_import_roundtrip
````

passes.

I don't think(?) the assignment to `_ROOT_DATA_FILE[histname]` should matter a huge amount here, so it seems that comes down to the `uproot.recreate` block in `writexml`

https://github.com/scikit-hep/pyhf/blob/ce2ffabaaeadf2733d4d61284d750933f34cee28/src/pyhf/writexml.py#L288-L303

**Comment by matthewfeickert** [comment]
I wonder if `InputFile=_ROOT_DATA_FILE._path` in

https://github.com/scikit-hep/pyhf/blob/ce2ffabaaeadf2733d4d61284d750933f34cee28/src/pyhf/writexml.py#L265-L268

is causing problems.

**Comment by matthewfeickert** [comment]
Ah seems so. If I add in the following addition

```diff
$ git diff
diff --git a/src/pyhf/writexml.py b/src/pyhf/writexml.py
index 650c313a..e9391ae8 100644
--- a/src/pyhf/writexml.py
+++ b/src/pyhf/writexml.py
@@ -261,6 +261,7 @@ def build_data(obsspec, channelname):
 
 
 def build_channel(spec, channelspec, obsspec):
+    print(f"\n{_ROOT_DATA_FILE._path=}\n")
     channel = ET.Element(
         'Channel', Name=channelspec['name'], InputFile=_ROOT_DATA_FILE._path
     )
```

then with the `uproot4` in this PR

```console
$ python -m pytest -sx tests/test_validation.py -k test_import_roundtrip[example-one]
...
_ROOT_DATA_FILE._path=()
...
```

If I revert back to `uproot3`

```console
$ python -m pytest -sx tests/test_validation.py -k test_import_roundtrip[example-one]
... 
_ROOT_DATA_FILE._path='/tmp/pytest-of-feickert/pytest-51/test_import_roundtrip_example_0/data/data.root'
...
```

where for both `_ROOT_DATA_FILE` is set by

https://github.com/scikit-hep/pyhf/blob/ce2ffabaaeadf2733d4d61284d750933f34cee28/src/pyhf/writexml.py#L288-L290

matthewfeickert:
Debugging note to self: If I print out the `spec` from 

https://github.com/scikit-hep/pyhf/blob/ce2ffabaaeadf2733d4d61284d750933f34cee28/tests/test_validation.py#L846-L850

and then compare them for `uproot3` and `uproot4`

```console
$ diff -y /tmp/uproot3_spec.py /tmp/uproot4_spec.py 
{								{
    "channels": [						    "channels": [
        {							        {
            "name": "channel1",					            "name": "channel1",
            "samples": [					            "samples": [
                {						                {
                    "name": "signal",				                    "name": "signal",
                    "data": [20.0, 10.0],		      |	                    "data": [0.0, 0.0],
                    "modifiers": [				                    "modifiers": [
                        {					                        {
                            "name": "syst1",			                            "name": "syst1",
                            "type": "normsys",			                            "type": "normsys",
                            "data": {"lo": 0.95, "hi": 1.05},	                            "data": {"lo": 0.95, "hi": 1.05},
                        },					                        },
                        {"name": "SigXsecOverSM", "type": "no	                        {"name": "SigXsecOverSM", "type": "no
                    ],						                    ],
                },						                },
                {						                {
                    "name": "background1",			                    "name": "background1",
                    "data": [100.0, 0.0],		      |	                    "data": [1.0, 0.0],
                    "modifiers": [				                    "modifiers": [
                        {"name": "lumi", "type": "lumi", "dat	                        {"name": "lumi", "type": "lumi", "dat
                        {					                        {
                            "name": "staterror_channel1",	                            "name": "staterror_channel1",
                            "type": "staterror",		                            "type": "staterror",
                            "data": [5.000000074505806, 0.0], |	                            "data": [2.0, 0.0],
                        },					                        },
                        {					                        {
                            "name": "syst2",			                            "name": "syst2",
                            "type": "normsys",			                            "type": "normsys",
                            "data": {"lo": 0.95, "hi": 1.05},	                            "data": {"lo": 0.95, "hi": 1.05},
                        },					                        },
                    ],						                    ],
                },						                },
                {						                {
                    "name": "background2",			                    "name": "background2",
                    "data": [0.0, 100.0],		      |	                    "data": [1.0, 0.0],
                    "modifiers": [				                    "modifiers": [
                        {"name": "lumi", "type": "lumi", "dat	                        {"name": "lumi", "type": "lumi", "dat
                        {					                        {
                            "name": "staterror_channel1",	                            "name": "staterror_channel1",
                            "type": "staterror",		                            "type": "staterror",
                            "data": [0.0, 10.0],	      |	                            "data": [2.0, 0.0],
                        },					                        },
                        {					                        {
                            "name": "syst3",			                            "name": "syst3",
                            "type": "normsys",			                            "type": "normsys",
                            "data": {"lo": 0.95, "hi": 1.05},	                            "data": {"lo": 0.95, "hi": 1.05},
                        },					                        },
                    ],						                    ],
                },						                },
            ],							            ],
        }							        }
    ],								    ],
    "parameters": [						    "parameters": [
        {							        {
            "name": "lumi",					            "name": "lumi",
            "auxdata": [1.0],					            "auxdata": [1.0],
            "bounds": [[0.5, 1.5]],				            "bounds": [[0.5, 1.5]],
            "inits": [1.0],					            "inits": [1.0],
            "sigmas": [0.1],					            "sigmas": [0.1],
            "fixed": True,					            "fixed": True,
        },							        },
        {"name": "SigXsecOverSM", "bounds": [[0.0, 3.0]], "in	        {"name": "SigXsecOverSM", "bounds": [[0.0, 3.0]], "in
        {"name": "syst1", "fixed": True},			        {"name": "syst1", "fixed": True},
    ],								    ],
}								}
```

the only parts that differ are the `data` fields

```console
$ diff /tmp/uproot3_spec.py /tmp/uproot4_spec.py 
8c8
<                     "data": [20.0, 10.0],
---
>                     "data": [0.0, 0.0],
20c20
<                     "data": [100.0, 0.0],
---
>                     "data": [1.0, 0.0],
26c26
<                             "data": [5.000000074505806, 0.0],
---
>                             "data": [2.0, 0.0],
37c37
<                     "data": [0.0, 100.0],
---
>                     "data": [1.0, 0.0],
43c43
<                             "data": [0.0, 10.0],
---
>                             "data": [2.0, 0.0],
```

so I need to look at how those are getting written.

matthewfeickert:
> it is clear to me then that how I've written `_export_root_histogram` in this PR is wrong.

Yeah, I wrote it 100% wrong. Don't write code when very tired is the obvious lesson here.

lower-bound-requirements.txt
```diff
@@ -6,8 +6,7 @@ jsonschema==3.0.0
 jsonpatch==1.15
 pyyaml==5.1
 # xmlio
-uproot3==3.14.1
-uproot==4.0.0
+uproot==4.1.0
```
jpivarski, line 9: You might want to loosen that to anything in 4.1.*. They're will be a 4.1.1 to fix a performance bug soon.

tests/test_export.py
```diff
@@ -392,9 +395,25 @@ def test_export_data(mocker):
...
+    """
+    Test that pyhf.writexml._export_root_histogram writes out a histogram
+    in the manner that uproot is expecting
+    """
+    mocker.patch("pyhf.writexml._ROOT_DATA_FILE", {})
+    pyhf.writexml._export_root_histogram("example", [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
+
+    with uproot.recreate(tmp_path.joinpath("test_export_root_histogram.root")) as file:
+        file["hist"] = pyhf.writexml._ROOT_DATA_FILE["example"]
+        values, edges = file["hist"].to_numpy()
```
jpivarski, line 408: This literally writes and reads the data back. If you want to keep an Uproot Model of the data for later processing, you could use `uproot.to_writable` to turn it into a Model, which is implicitly the first step in writing anyway.

jpivarski (approved):
Looks great! I just have a few suggestions that could streamline things, nothing critical.

lower-bound-requirements.txt
matthewfeickert: This is intentional. The `lower-bound-requirements.txt` is only used in the minimum supported dependencies workflow that we run to empirically test the lower bounds we state in `setup.cfg` and `setup.py`.

https://github.com/scikit-hep/pyhf/blob/ce2ffabaaeadf2733d4d61284d750933f34cee28/.github/workflows/lower-bound-requirements.yml#L30 

If there were bug fixes that we became dependent on this would help us catch them so we could update the minimum required version of `uproot` in the `xmlio` extra in `setup.py`. :+1:

tests/test_export.py
matthewfeickert: This was more just meant as a very simple sanity check that `pyhf.writexml._export_root_histogram` is writing what we expect it to into the ROOT file. I see though that

```python
>>> import uproot
>>> import numpy as np
>>> data = np.arange(10)
>>> edges = np.arange(len(data) + 1)
>>> uproot.to_writable((data, edges))
<TH1D (version 3) at 0x7f421bf86df0>
```

verifies the structure. So in your mind should I add the suggestion that I've placed in `pyhf.writexml._export_root_histogram`?

src/pyhf/writexml.py
```diff
@@ -48,11 +46,9 @@ def _make_hist_name(channel, sample, modifier='', prefix='hist', suffix=''):
 
 
 def _export_root_histogram(histname, data):
-    hist = TH1.from_numpy((np.asarray(data), np.arange(len(data) + 1)))
-    hist._fName = histname
     if histname in _ROOT_DATA_FILE:
         raise KeyError(f"Duplicate key {histname} being written.")
-    _ROOT_DATA_FILE[histname] = hist
+    _ROOT_DATA_FILE[histname] = (np.asarray(data), np.arange(len(data) + 1))
```
matthewfeickert, line 51:
```suggestion
    _ROOT_DATA_FILE[histname] = uproot.to_writable(
        (np.asarray(data), np.arange(len(data) + 1))
    )
```

@jpivarski given your other comment, this seems better then directly writing, correct? As if this was ever messed up then we'd get a `TypeError` raised.

matthewfeickert (commented):
Thanks for the super fast response, @jpivarski! Just one follow up question for you though.

src/pyhf/writexml.py
```diff
@@ -48,11 +46,9 @@ def _make_hist_name(channel, sample, modifier='', prefix='hist', suffix=''):
 
 
 def _export_root_histogram(histname, data):
-    hist = TH1.from_numpy((np.asarray(data), np.arange(len(data) + 1)))
-    hist._fName = histname
```
matthewfeickert, line 52: @jpivarski I lied and have _another_ follow up question. Is it possible to set a `name` attribute if creating a `TH1D` object from an `uproot.Model` with some later call?

```python
>>> import uproot
>>> import numpy as np
>>> data = np.arange(10)
>>> edges = np.arange(len(data) + 1)
>>> hist_model = uproot.to_writable((np.asarray(data), np.arange(len(data) + 1)))
>>> hist_model.name  # not set
>>> hist_model.name = "test"
AttributeError: can't set attribute
```

Looking at the `help(hist_model)` show that 

```
class Model_TH1D_v3(uproot.behaviors.TH1.TH1, uproot.model.VersionedModel)
...
 |  ----------------------------------------------------------------------
 |  Readonly properties inherited from uproot.behaviors.TH1.Histogram:
 |  
 |  name
 |      The name of the histogram.
 |  
 |  title
 |      The title of the histogram.

```

Is there anyway to set this using something in the non-public API lke we were here with the `_fName` attribute with `uproot3_methods.classes.TH1`?

matthewfeickert: Ah. Yes:

```python
>>> import uproot
>>> import numpy as np
>>> data = np.arange(10)
>>> edges = np.arange(len(data) + 1)
>>> hist_model = uproot.to_writable((np.asarray(data), np.arange(len(data) + 1)))
>>> hist_model.name  # empty here
>>> hist_model._members["fName"] = "test"
>>> hist_model.name
'test'
```

matthewfeickert: ...and I now see that none of this really mattered as `_ROOT_DATA_FILE[histname] = hist_uproot_model` will set the `_ROOT_DATA_FILE[histname].name` to `histname`.

tests/test_export.py
jpivarski: If it's a sanity check, then writing and reading back is a good idea. It would be even more of a sanity check if you did so with different file objects:

```python
    with uproot.recreate(tmp_path.joinpath("test_export_root_histogram.root")) as file:
        file["hist"] = pyhf.writexml._ROOT_DATA_FILE["example"]

    with uproot.open(tmp_path.joinpath("test_export_root_histogram.root")) as file:
        values, edges = file["hist"].to_numpy()
```

That would test more—the whole file structure, and the certainty that these really are bytes on disk. Using the same `file` does invoke the system operation to read from the file it's just written to, but with the same file handle in `"r+b"` mode and without going back to the beginning of the file and seeking through all the pointers to get to the object again. `to_writable` does still less, just turning the 2-tuple of arrays Python object into an Uproot Model Python object, which is what you'd want to do if performance were your goal, rather than sanity-checking.

So it's up to you, depending on what your goal is here.

src/pyhf/writexml.py
jpivarski: Just

```python
writable = uproot.to_writable((np.asarray(data), np.arange(len(data) + 1)))
```

by itself ensures that the data passed to `to_writable` will be writable. It will raise a TypeError if it is not recognized as a writable Python type.

The next step of assigning to the file,

```python
_ROOT_DATA_FILE[histname] = writable
```

actually writes it to the file. The disk I/O operations are only invoked when this assignment happens, not when the `to_writable` function is called.

matthewfeickert: > The disk I/O operations are only invoked when this assignment happens, not when the `to_writable` function is called.


Right. Okay, then it seems what I have here (now) is doing as expected. :+1:

tests/test_export.py
matthewfeickert: Good point. I've added this in the latest commit.

matthewfeickert:
I bumped the lower bound to `uproot` `v4.1.1` as there's no real downside to requiring it given that @jpivarski just applied some bug fixes and `v4.1.0` is way too new for anyone to have developed any hard dependency on it in the last 4 day.s

lukasheinrich (approved):

PR #1568, opened 2021-08-28
refactor: Remove unnecessary str wraps of pathlib.Path objects

matthewfeickert:
# Description

Remove `str` wraps around `pathlib.Path` objects that are not necessary at runtime. Additionally use [`tmp_path`](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture) fixture in `tests/test_validation.py` to remove need for `pathlib` import and `strpath` calls.

Requires PR #1567 to go in first.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Remove str wraps fo pathlib.Path objects in readxml and writexml if the str wrap is not needed
* Use pytest tmp_path fixture to have `test_import_roundtrip` use pathlib ojbects
   - Removes the need for converting to str and for import of pathlib.Path
* Apply isort to tests/test_validation.py
```


matthewfeickert:
@lukasheinrich @kratsg I'm just going to merge this to move forward with `v0.6.3` as this should be an uncontroversial change, but I will happily revisit or revert any part of this.

PR #1576, opened 2021-09-01
docs: Use GitHub Issue forms as Issue templates

### LINKED ISSUE #1452: Use pip's Issue Template design to guide users towards the right forum
# Description

TIL that `pip` has a super nice Issue template page

[pip_Issue_templates](https://github.com/pypa/pip/issues/new/choose)

that uses a [`.github/ISSUE_TEMPLATE/config.yml`](https://github.com/pypa/pip/blob/e6414d6db6db37951988f6f2b11ec530ed0b191d/.github/ISSUE_TEMPLATE/config.yml) file to add the ability to have additional buttons (here Discourse and IRC) to guide user's reporting.

This seems like a _great_ idea that we should encorperate as well. It would seem to also be useful for basically all of the Scikit-HEP projects, especially those with Gitter channels. cc @eduardo-rodrigues @jpivarski @henryiii

matthewfeickert:
# Description

Resolves #1452

Add [GitHub Issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) to provide better context and guide users. As the forms can also require information to be entered before the Issue can be created, they can also streamline input from users.

These Issue forms are heavily influenced / copied from [Ansible](https://github.com/scikit-hep/pyhf/issues/1452#issuecomment-840746775)'s (thanks @webknjaz!).

options

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use GitHub Issue form Issue templates to provide more consistent context of requested information from users
   - c.f. 
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms
   - Remove all previous GitHub Issue templates
   - These Issue forms are heavily influenced / copied from Ansible's https://github.com/ansible/ansible/
   - Intentionally add a second space after emojis to have them render better on macOS
* Add config.yml to also link users to GitHub Discussions, the pyhf tutorial, and the Code of Conduct

Co-authored-by: Sviatoslav Sydorenko <wk.cvs.github@sydorenko.org.ua>
```


matthewfeickert:
As there is no good way to preview these in a PR, I did all the writing on my fork https://github.com/matthewfeickert/pyhf. I'd invite everyone to go over to [my fork's Issues](https://github.com/matthewfeickert/pyhf/issues/new/choose) to review the PR.

@alexander-held as you open some of the best Issues on `pyhf` if we can get your review on what parts of this are good and what parts might slow you down / be obnoxious that would be really helpful!

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,107 @@
...
+      pyhf, version 0.6.2
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: OS / Environment
+    description: >-
+      Provide the name of your operating system and the output of `uname --all`.
+    placeholder: |
```
webknjaz, line 50: Maybe render as console? Also, I'd add `cat /etc/os-release` and `cat /etc/lsb-release` (IIRC)

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,77 @@
...
+  attributes:
+    value: >
+      **Thank you for wanting to suggest a feature for pyhf!**
+
+
+      ⚠
+      Verify first that your issue is not [already reported on
+      GitHub][issue search].
+      Make sure to check the closed issues as well as it may
+      already be implimented in a development release.
```
webknjaz, line 17:
```suggestion
      already be implemented in a development release.
```

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,15 @@
+# Ref: https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser
+blank_issues_enabled: true
+contact_links:
+- name: 🙋 Use Questions
```
alexander-held, line 4: Maybe "Usage" instead of "Use"? It took me a while to understand that "use" was not a verb here.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,107 @@
...
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: pyhf Version
+    description: >-
+      Paste verbatim output from `pyhf --version` below, under
+      the prompt line. Please don't wrap it with tripple backticks &mdash; your
+      whole input will be turned into a code snippet automatically.
```
alexander-held, line 37: I am assuming that users with editable installs who checkout different versions of `pyhf` will generally understand the system enough so that this is not an issue, but in principle the output of `pyhf --version` may not match the actual version in use (this is for example the case for me right now).

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,107 @@
...
+    placeholder: >-
+      When I try to do X with pyhf and the following workspace, Y breaks or
+      Z happens in an unexpected manner.
+      Here are all the details I know about this problem.
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: pyhf Version
```
alexander-held, line 33: Would suggest moving this towards the end of the form, in my experience with existing issues the version seems to rarely matter a lot.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,107 @@
...
+      $ pyhf --version
+      pyhf, version 0.6.2
+  validations:
+    required: true
+
+- type: textarea
+  attributes:
+    label: OS / Environment
+    description: >-
+      Provide the name of your operating system and the output of `uname --all`.
```
alexander-held, line 49: This part seems like it would not matter for the large majority of existing issues. I think it does not hurt to have for completeness, but would suggest moving to the end of the fields. `uname --all` will not always work, for example on Mac OS I need `uname -a`.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,107 @@
...
+    label: Steps to Reproduce
+    description: >-
+      Describe exactly how to reproduce the problem, using a minimal test-case.
+      It would *really* help us understand your problem if you paste in the Python code
+      that you're running.
+
+
+      **HINT:** You can paste [GitHub Gist](https://gist.github.com) links for larger files.
+    value: |
+      <!--- Paste your minimal failing Python example code between the quotes below -->
```
alexander-held, line 67: Maybe easier to use if the comment is above the cell, and the whole cell is `python` formatted? Not clear to me why this different style is used here.

alexander-held (commented):
The templates look very good to me. I have a few rather minor comments. Is there still the possibility to upload files (e.g. a workspace) in the new bug report form? I spontaneously would not know how to do it.

webknjaz:
> Is there still the possibility to upload files (e.g. a workspace) in the new bug report form?

Every `textarea` field is a separate WYSIWYG, just like the ones where you type in comments. So people can attach files/pictures there as usual.

.github/ISSUE_TEMPLATE/bug-report.yml
webknjaz: Probably to give people a chance to also add text, not just code.

alexander-held: Good point. I assumed extra text could go into summary / expected results / actual results, but for more complicated setups the opportunity for more text seems useful.

webknjaz: You could have two separate fields for code and text, of course. But then, you can never guess how many code blocks the reporter would need.

matthewfeickert: Good idea. I should probably add additional optional entries for reporting CLI API use bugs and for uploading workspaces.

.github/ISSUE_TEMPLATE/bug-report.yml
matthewfeickert: @alexander-held You just saying that if you are locally installing `pyhf` and not using a supported release from PyPI then you might get something like

```console
$ pyhf --version
pyhf, version 0.6.3.dev17
```

? Or do you mean something in addition to this?

.github/ISSUE_TEMPLATE/bug-report.yml
matthewfeickert: > You could have two separate fields for code and text, of course. But then, you can never guess how many code blocks the reporter would need.

Yeah. I think I may stick with doing the hint and then just adding an optional place for people to upload a workspace file. This way I don't have to have both an entry for CLI and for Python APIs.

matthewfeickert:
Thanks for these comments! I think I've applied all of them now. :+1:

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,148 @@
+---
+name: 🐛 Bug Report
```
kratsg, line 2:
```suggestion
name: 🐛  Bug Report
```

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,15 @@
+# Ref: https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser
+blank_issues_enabled: true
+contact_links:
+- name: 🙋 Useage Questions
```
kratsg, line 4:
```suggestion
- name: 🙋  Usage Questions
```

.github/ISSUE_TEMPLATE/documentation-report.yml
```diff
@@ -0,0 +1,52 @@
+---
+name: 📝 Documentation Report
```
kratsg, line 2:
```suggestion
name: 📝  Documentation Report
```

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,80 @@
+---
+name: ✨ Feature Request
```
kratsg, line 2:
```suggestion
name: ✨  Feature Request
```

.github/ISSUE_TEMPLATE/~release-checklist.md
```diff
@@ -1,5 +1,5 @@
 ---
-name: (Maintainers Only) Release Checklist
+name: ✅ Release Checklist (Maintainers Only)
```
kratsg, line 2:
```suggestion
name: ✅  Release Checklist (Maintainers Only)
```

.github/ISSUE_TEMPLATE/bug-report.yml
kratsg: During this portion, I wonder if we should provide a `pyhf debug` or similar CLI so users can copy/paste that output entirely.

matthewfeickert: @kratsg want to Issue that for follow up?

.github/ISSUE_TEMPLATE/bug-report.yml
matthewfeickert: @kratsg Why do you want two spaces for all of these?

.github/ISSUE_TEMPLATE/bug-report.yml
kratsg: #1580

.github/ISSUE_TEMPLATE/bug-report.yml
matthewfeickert: Ah okay, that must be some bug with mac OS then as they render fine elsewhere. 

spaces

matthewfeickert: Does this also happen with Firefox?

.github/ISSUE_TEMPLATE/bug-report.yml
alexander-held: I mean something closely related, yes. I won't get the dev tags if I clone the repo, check out tag `v0.6.2`, `pip install -e .`, and then check out the master branch. `pyhf --version` will show `pyhf, version 0.6.2` nevertheless, even though I am using a newer version. This is so minor though that it is probably not worth any changes.

matthewfeickert: Ah gotcha yeah. :+1: I think this is probably okay, as you're like our ultimate power user and you know what you're doing here. I'm more assuming that people who are reporting bugs are installing directly from PyPI.

Though I'll mention here that when you report bugs in the future feel free to just give whatever information is the most relevant here (e.g. git commit hash) or something.

kratsg (approved):

PR #1589, opened 2021-09-04
ci: Skip doctest for 'Minimum supported dependencies' workflow

matthewfeickert:
# Description

Run pytest explicitly over the tests/ directory only to avoid in the 'Minimum supported dependencies' workflow to avoid running doctest on the docstring examples. The latest version of the JAX docsting examples have some print failures that are incompatible with the older versions of JAX, even thought the API tests pass.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Run pytest explicitly over the tests/ directory only to avoid in the 'Minimum supported dependencies' workflow to avoid running doctest on the docstring examples
   - The docstring examples use the latest dependencies which might have different text output than older releases. So focus only on the API tests
```

### DIFF
--- a/.github/workflows/lower-bound-requirements.yml
+++ b/.github/workflows/lower-bound-requirements.yml
@@ -18,10 +18,12 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
+
     - name: Install dependencies and force lowest bound
       run: |
         python -m pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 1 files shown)


.github/workflows/lower-bound-requirements.yml
```diff
@@ -18,17 +18,21 @@ jobs:
...
         python -m pip --no-cache-dir --quiet install --requirement lower-bound-requirements.txt
         python -m pip --no-cache-dir --quiet install --editable .[test]
         python -m pip install --requirement lower-bound-requirements.txt
         python -m pip list
+
     - name: Test with pytest
       run: |
-        python -m pytest -r sx --ignore tests/benchmarks/ --ignore tests/contrib --ignore tests/test_notebooks.py
+        # Run on tests/ to skip doctests of src given examples are for latest APIs
+        python -m pytest -r sx --ignore tests/benchmarks/ --ignore tests/contrib --ignore tests/test_notebooks.py tests/
```
matthewfeickert, line 38: @kratsg I'm not sure if there is an easier way to do this that running explicitly over `tests/`.

Is there a way at the command line to negate the `--doctest-modules` from

https://github.com/scikit-hep/pyhf/blob/173c3c711c8037928a966d381c922fd535be956f/pyproject.toml#L44

?

kratsg: Nope

matthewfeickert: Okay, cool. This is ready for review then. :)

matthewfeickert:
I'm going to just merge this as it should be uncontroversial as it is extra tests, but let me know if anyone wants changes made in another PR. :+1:

PR #1625, opened 2021-10-12
feat: Add setup for custom modifiers

lukasheinrich:
# Description

Closes #1188

a cleaner version of PR #1188 on the way to addressing Issue #850

# Before Merging

For the PR Assignees:

```
* This API breaking change allows for the start of addition of custom
modifiers.
* `pyhf.workspace.Workspace.parameters` are fully removed from the
API through removal from `pyhf.mixins`. Instead, parameters is added
as an attribute to the model config.
* Remove the modifier registry and in place add {modifier}_builder
classes for the modifiers. Additionally add
pyhf.modifiers.histfactory_set which is the map of all the
predefined modifiers that HistFactory specifies.
* All workspace parameters related tests are removed and tests are
added for the change to model.config
* The change from workspace.parameters to model.config.parameters
is made in all affected example notebooks.
* Apply a tiny relative tolerance change in the regression tests to
increase to rel=1.5e-5. The resulting tolerance is on the order of 1e-7
and so the change is inconsequential.
```

### DIFF
--- a/src/pyhf/parameters/paramview.py
+++ b/src/pyhf/parameters/paramview.py
@@ -57,7 +57,6 @@ class ParamViewer:
         self._all_indices = default_backend.reshape(flat_indices, shape)
 
         # a tensor viewer that can split and stitch parameters
-
         self.allpar_viewer = _tensorviewer_from_parmap(par_map, batch_size)
 
         # a tensor viewer that can split and stitch the selected parameters


alexander-held:
This looks very promising, thanks for all the work! I noticed one thing when testing the downstream effects: the order of model parameters can change. While this may not be a breaking change for many users (the order is something that should probably not be relied on anyway), it may cause some unexpected effects. I think it would be useful to highlight this change (assuming it is an intended side-effect of this PR).

PR #1636, opened 2021-10-13
feat: Allow POI-less models via Workspace.model

kratsg:
# Pull Request Description

No associated issues. Primarily drop the `poi_name` functionality in `pyhf.Workspace.get_measurement` as it is not really used and is kind of a bit weird anyway. Instead, allow for users to specify `poi_name` in `pyhf.Workspace.model()` in order to override the suggested POI that comes from the measurement they would like to use. This allows a user to use an existing measurement, but override the `poi_name`.

For example, `pyhf.Workspace.model(poi_name=None)` will allow for background-only models to be created from `pyhf.Workspace` which was not something that was allowed in the past.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Remove `poi_name` functionality in `pyhf.Workspace.get_measurement`
* Allow for `poi_name` to be optionally set in `pyhf.Workspace.model` to override suggested POI
* Add tests for setting poi_name through `pyhf.Workspace.model`
* Provide intersphinx for jsonpatch docs
```

### DIFF
--- a/docs/examples/notebooks/binderexample/StatisticalAnalysis.ipynb
+++ b/docs/examples/notebooks/binderexample/StatisticalAnalysis.ipynb
@@ -148,8 +148,8 @@
     }
    ],
    "source": [
-    "print(f'Samples:\\n {workspace.samples}')\n",
-    "print(f'Parameters:\\n {workspace.parameters}')"
+    "print(f'Samples:\\n {pdf.config.samples}')\n",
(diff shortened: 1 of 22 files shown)


matthewfeickert:
Relevant RTD build: https://pyhf--1636.org.readthedocs.build/en/1636/_generated/pyhf.workspace.Workspace.html

alexander-held:
related: https://github.com/scikit-hep/pyhf/discussions/1602

PR #1644, opened 2021-10-15
fix: custom modifier / new parameter support and test

lukasheinrich:
# Description

this fixes a bug in #1625 and a test that would catch it

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Fix a regression introduced in PR #1625
* Check that the poi set is an existing parameter in the model
* Add tests for custom modifiers to cover the use case and make sure it's caught
```


lukasheinrich:
@kratsg added

Issue #1486, opened 2021-06-09
Make download tests robust to HEPData outages by mocking of tar.gz files


matthewfeickert:
@kratsg is pointing out that we should probably have these mocked so that the test suite doesn't rely on HEPData behavior.

Example of something similar he's done recently

```python
def test_user_expires_reauthenticate(user_temp, requests_mock, mocker):
    user, _ = user_temp
    assert user.is_authenticated()
    assert user.is_expired() == False
    assert user.expires_in > 0
    user._id_token['exp'] = time.time() - 1
    assert user.is_authenticated()
    assert user.is_expired()
    assert user.expires_in == 0

    mock = mocker.patch.object(user, '_parse_id_token')
    requests_mock.post(
        requests.compat.urljoin(itkdb.settings.ITKDB_AUTH_URL, 'grantToken'),
        text=json.dumps(
            {
                'id_token': {
                    'exp': time.time() + 3600,
                    'name': _name,
                    'uuidentity': _identity,
                },
                'access_token': _access_token,
            }
        ),
    )
    user.authenticate()
    user._id_token = {'exp': time.time() + 3600, 'name': _name, 'uuidentity': _identity}
    assert user.is_authenticated()
    assert user.is_expired() == False
```

_Originally posted by @matthewfeickert in https://github.com/scikit-hep/pyhf/issues/1485#issuecomment-857370600_

Issue #1451, opened 2021-05-13
fail.org exists and is for sale


kratsg:
Previously had an SSL error, now exists which is somewhat problematic. We should use https://httpstat.us/ or similar.

kratsg:
The other option is to mock out the underlying requests to return the expected error, and only check that the rest of the code behaves as expected.

matthewfeickert:
@kratsg This error isn't from what you say it is, given that if I rerun the CI on `master` right now everything passes. So as it is passing with 

https://github.com/scikit-hep/pyhf/blob/671ed56483bc602876f611cf10ff9add7f4a2137/tests/test_scripts.py#L563-L568

then I don't think it is related to https://www.fail.org/ existing or not.

kratsg:
See this failure: https://github.com/scikit-hep/pyhf/runs/2571497115?check_suite_focus=true

matthewfeickert:
> See this failure: https://github.com/scikit-hep/pyhf/runs/2571497115?check_suite_focus=true

I'm confused though. This is a `ConnectionResetError` that doesn't seem to have anything to do with the code and doesn't show up again anywhere else in the runs on PR #1450

```pytb
>       assert (
            "SSLCertificateError" and "hostname 'www.fail.org' doesn't match" in ret.stderr
        )
E       assert ('SSLCertificateError' and "hostname 'www.fail.org' doesn't match" in 'Traceback (most recent call last):\n  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/urllib...uests.exceptions.ConnectionError: (\'Connection aborted.\', ConnectionResetError(104, \'Connection reset by peer\'))\n')
E        +  where 'Traceback (most recent call last):\n  File "/opt/hostedtoolcache/Python/3.8.10/x64/lib/python3.8/site-packages/urllib...uests.exceptions.ConnectionError: (\'Connection aborted.\', ConnectionResetError(104, \'Connection reset by peer\'))\n' = <pytest_console_scripts.RunResult object at 0x7fae1ed21040>.stderr
```

Can you explain in like very basic terms why this needs to get changed? I'm missing how we're doing anything wrong.

kratsg:
This happened because the site is not stable. You're relying on `fail.org` which is not owned and therefore you can't guarantee proper SSL connections or that it will work. You'll want to rely on something more stable, or otherwise the CI will fail randomly like it did in this case.

matthewfeickert:
[...]

PR #1645, opened 2021-10-15
docs: Add and apply codespell as a pre-commit hook

alexander-held:
# Description

This adds the spell-checker [`codespell`](https://github.com/codespell-project/codespell/) to run in `pre-commit`, as suggested [on the Scikit-HEP website](https://scikit-hep.org/developer/style#spelling-extra). Most spelling mistakes are also fixed, with the exception of mistakes coming from ROOT (the HistFactory `.dtd` schema, and stdout from ROOT). The fixes are applied in separate commits to be able to remove parts as needed.

`codespell` is configured as follows:
```bash
codespell -w -L "hist,gaus" -S "*.ipynb,*.dtd,*.eps,*.bib" -x "validation/xmlimport_input/log"
```
- `-w`: apply changes in place (consistent with other `pre-commit` hooks used)
- `-L`: ignore the provided list of words
- `-S`: ignore the provided list of file types
  - `.ipynb` since many short strings appearing in images are flagged and I do not know how to prevent this (but two typos in notebooks were also fixed by running over notebooks and parsing output manually) 
  - `.dtd` since the spelling mistakes come from ROOT and consistency with ROOT may be more desirable than fixing the mistakes
  - `.eps` since it seems to flag EPS details instead of actual words
  - `.bib` since a Journal name is (presumably incorrectly) flagged, could also add "ser" to list of ignored words otherwise
- `-x`: ignore specific file (ROOT stdout with spelling mistakes coming from ROOT)

Instead of trying to run over everything and excluding files that are problematic, the setup could alternatively also be to run only over a subset of folders (like `src/` and `tests/`). This is done for example in `cabinetry` (see [PR](https://github.com/scikit-hep/cabinetry/pull/296) / [pre-commit config](https://github.com/scikit-hep/cabinetry/blob/b89bb09b97608e984864cde6d45f9b8901612a06/.pre-commit-config.yaml#L41-L45).

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add and apply codespell pre-commit hook
   - c.f. https://github.com/codespell-project/codespell/
* Configure codespell to run only over *.py, *.md, *.rst files
* Ignore the HEP specific terms "hist" and "gaus"
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -629,6 +629,7 @@ class Model:
             modifier_set, self.config, self.spec, self.batch_size
         )
 
+        poi_name = None if poi_name == "" else poi_name
         if poi_name is not None:
             self.config.set_poi(poi_name)
 
--- a/tests/test_pdf.py
+++ b/tests/test_pdf.py
@@ -278,6 +278,39 @@ def test_poiless_model(backend):
         pyhf.infer.hypotest(1.0, data, model)
 
 
+def test_poiless_model_empty_string(backend):
+    spec = {
+        'channels': [
+            {
+                'name': 'channel',
+                'samples': [
+                    {
+                        'name': 'goodsample',
+                        'data': [10.0],
+                        'modifiers': [
+                            {
+                                'type': 'normsys',
+                                'name': 'shape',
+                                'data': {"hi": 0.5, "lo": 1.5},
+                            }
+                        ],
+                    },
+                ],
+            }
+        ]
+    }
+    model = pyhf.Model(spec, poi_name="")
+
+    data = [12] + model.config.auxdata
+    pyhf.infer.mle.fit(data, model)
+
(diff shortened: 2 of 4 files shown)


alexander-held:
I found another approach to make the original suggestion work: `pre-commit` can match files via regex. This means that
```yaml
files: ^.*\.(py|md|rst)$
```
selects all matching files (`*.py`, `*.md`, `*.rst`) and then sends them to `codespell`. The only downside of this approach is that it is harder to run `codespell` standalone to reproduce the same result, but that might not be needed.

Not sure whether piping can be made to work to handle notebooks. As far as I can see, they would have to first be written to a file for `codespell` to handle them.

matthewfeickert:
> The CI failures look unrelated to the changes here, I am not sure what is causing them.

They can fail at times due to Issue #1451 and Issue #1486.

PR #1646, opened 2021-10-15
feat: Configurable default backend

### LINKED ISSUE #1004: Explicit setting of the default backend

kratsg:
# Pull Request Description

Supersedes #1121. Should allow for default backend to be changeable.

Resolves #1004.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add conditional logic in `set_backend` and `get_backend` based on
a new kwarg `default=False` that will set or get the default backend
instead of the tensorlib backend. Particularly useful for differentiable
model construction, which needs to be propagated through tensors
instead of numpy arrays.
* Migrate relevant code in src/pyhf/__init__.py into a submodule
instead and clean things up more
* Add tests to ensure autodifferentiability and jit with the
changing of default backends
```

### DIFF
--- a/src/pyhf/__init__.py
+++ b/src/pyhf/__init__.py
@@ -1,4 +1,5 @@
 from pyhf.tensor import BackendRetriever as tensor
+from pyhf.optimize import OptimizerRetriever as optimize  # noqa
 from pyhf.tensor.manager import get_backend
 from pyhf.tensor.manager import set_backend
 from pyhf._version import version as __version__


kratsg:
> @kratsg I think we don't need all the `import pyhf` everywhere. I know that you said you're doing this so avoid circular imports, but I think this can be avoided.

we need this to access `pyhf.default_backend`. I cannot `from pyhf import default_backend` anymore fyi.

kratsg:
> Can you add a test that shows that a function that uses both default backend and tensorlib with both set to jax becomes diffable?

Yes, I can definitely add a test. I'm hoping @phinate can provide some suggestions on a way to ensure diffability.

phinate:
> I would try something like
> 
> def func(x): y = default_backend.power(x, 2) z = tensorlib.power(y, 2) return z
> 
> jax.jacrev(func)(default_backend.astensor([1.,2.]))

Yeah, this is a nice way to check there’s no backend foul play on the scale of individual ops. Probably not a planned use case, but I imagine this is also possible to test for the other backends.

To ensure the most general case in terms of model construction, one could also do an example similar to the one posted in #742 that differentiates with respect to a latent param that controls the yields, but with a more general set of models that explores the different types of systematics, though probably out of scope for this PR!

phinate:
> thanks @phinate ccan you verify that this works too?
> 
> ```
>     def example_op(x):
>         y = pyhf.default_backend.power(x,2)
>         z = pyhf.tensorlib.sum(x)
>         return z
> ```

sure -- guessing you meant `z = pyhf.tensorlib.sum(y)`?

phinate:
> Great tests.. Would be nice to also test that you can jit through this if the two backwnds are jax

nice idea! still works :)

```python3
def test_diffable_backend():
    pyhf.set_backend("jax", default=True)

    @jax.jit
    def example_op(x):
        y = pyhf.default_backend.astensor(x)
        return 2 * y

    assert jax.jacrev(example_op)([1.0]) == [2.0]

    @jax.jit
    def example_op2(x):
        y = pyhf.default_backend.power(x, 2)
        z = pyhf.tensorlib.sum(y)
        return z

    assert jax.jacrev(example_op2)(pyhf.tensorlib.astensor([2.0, 3.0])).tolist() == [
        4.0,
        6.0,
    ]


def test_diffable_backend_failure():
    pyhf.set_backend("numpy", default=True)
    pyhf.set_backend("jax")

    def example_op(x):
        y = pyhf.default_backend.astensor(x)
        return 2 * y

    with pytest.raises(Exception):
        jax.value_and_grad(example_op)(1.0)

    def example_op2(x):
        y = pyhf.default_backend.power(x, 2)
        z = pyhf.tensorlib.sum(y)
        return z

    with pytest.raises(Exception):
        jax.jacrev(example_op2)(pyhf.tensorlib.astensor([2.0, 3.0]))
```

phinate:
> > @kratsg it seems like the first error encountered in the failure test has type `jax.errors.ConcretizationTypeError` instead of the one you have listed -- just what it hits first I guess from the numpy asarray call, doesn't even get to the tracer error. The other one works :)
> 
> That's confusing. It hits the other one first on my mac?

agh, that is confusing. could be os-dependent (yikes), since the CI and me are running on Ubuntu.

if you want, I don't think we lose anything by dropping the first failure mode of the test -- it's just to catch if we're mixing backends when diffing, which should always fail with any `default_backend` call if it's set to numpy, regardless of the op.

phinate:
just for clarity, [here's the relevant page in the jax docs](https://jax.readthedocs.io/en/latest/notebooks/Common_Gotchas_in_JAX.html#non-array-inputs-numpy-vs-jax) that goes over this

kratsg:
Huh, we do have `tests/test_simplemodels.py` but those don't cover the other backends.

Issue #882, opened 2020-05-29
Making pyhf differentiable


phinate:
## Description

It's become clear to me through the discussions that have taken place in IRIS-HEP and [gradhep](https://github.com/gradhep/) that it would be awesome to have pyhf as part of the modelling step in a differentiable analysis workflow.

Having backends that support autograd is most of the work, but me and @lukasheinrich have found through experiments in the development of [neos](https://github.com/pyhf/neos) that there are still operations that don't play nice with the differentiability of `pyhf.Model` construction, for instance. This is also just in the situation of using `pyhf.tensor.jax_backend()` -- I haven't tried this with other backends.

## Pull requests/issues
There's already been an attempt to make part of this differentiable in #742, which I made small modifications to in [my fork of this branch](https://github.com/phinate/pyhf/tree/diffable_json). (I make an explicit call to override the backend at one point, which isn't pretty...)

## Scope
 So far, this idea has only been used for the model construction and likelihood evaluation steps, but I wonder if this could also extend to inference in `pyhf.infer`? This could influence design decisions made for [a library to provide differentiable HEP operations](https://github.com/gradhep/smooth), which could be imported into `pyhf.infer` down the line if the actual underlying functionality doesn't change, e.g. [wrapping a `scipy` optimizer with the two-phase method in the `fax` library.](https://github.com/pyhf/neos/issues/4#issuecomment-597812969)

All this seems big enough a task that I felt like it warranted an issue as a documented way forward :)

kratsg:
Rather than overriding `defeault_backend` to the current tensorlib -- we should provide a way for users to specify an explicit `**kwargs = {backend=default_backend}` and pass things through as needed. This should maintain backwards-compatibility on some level -- to first attempt.

We generally want to support users switching backends transparently without a huge amount of re-computation.

phinate:
cc'ing @kratsg about this again (and the relevant PR in #742)

kratsg:
@phinate I just went ahead and did the rebase. Try it now. To be honest, there's not a ton of changes now...

lukasheinrich:
I'm reactivating this based on the recent #1625 PR. The goal is to make this fucnttion diffable (and removiing the `.tolist()` calls.

```
def makeit(
    pars,
    nominal,
    uncorr_data,
    corrup_data,
    corrdn_data,
    stater_data,
    normsys_up,
    normsys_dn,
    lumi_sigma,
  ):
  spec = {
      "channels": [
[...]

PR #1647, opened 2021-10-15
feat: Add support for arrayful JSON

lukasheinrich:
# Description

this is a step towards #882 and allows tensor-like values in the spec

we relied a lot on `+=` being concatenatoin for python lists.. but now we first collect a list of lists/arrays and then do `tb.concatenate`

 @phinate 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Allow for tensor-like values in model spec by building the
modifier builder_data from a list of lists/arrays and then
using tensorlib.concatenate to built up the sample properties.
* Add test for building a model with tensor inputs.
```

### DIFF
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -553,19 +553,18 @@ def test_patchset_download(tmpdir, script_runner, archive):
     ret = script_runner.run(*shlex.split(command))
     assert ret.success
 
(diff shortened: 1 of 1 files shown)


lukasheinrich:
added!

matthewfeickert:
Oh, I think it is actually Issue #1451. Testing and if so then fixing it now.

ponyisi:
Thanks for taking care of this!

Issue #1654, opened 2021-10-19
Support vanilla lists when the default backend is non-numpy


kratsg:
(Make builders diffable)

#1646 will pull out a bug that we somewhat have in our code about assumptions of the default backend... we'll need to fix this.

This is just revealing an underlying feature we never actually supported

```pytb
>>> import json
>>> ws = pyhf.Workspace(json.load(open('mysigfit_brZ_100_brH_0_brW_0_bre_33_brm_33_brt_34_mass_100.json')))
>>> pdf = ws.model()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/pyhf/workspace.py", line 425, in model
    return Model(modelspec, **config_kwargs)
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/pyhf/pdf.py", line 632, in __init__
    modifiers, _nominal_rates = _nominal_and_modifiers_from_spec(
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/pyhf/pdf.py", line 129, in _nominal_and_modifiers_from_spec
    nominal_rates = nominal.finalize()
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/pyhf/pdf.py", line 69, in finalize
    [
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/pyhf/pdf.py", line 70, in <listcomp>
    pyhf.default_backend.concatenate(self.mega_samples[sample]['nom'])
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/pyhf/tensor/jax_backend.py", line 298, in concatenate
    return jnp.concatenate(sequence, axis=axis)
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 3382, in concatenate
    _check_arraylike("concatenate", *arrays)
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 560, in _check_arraylike
    raise TypeError(msg.format(fun_name, type(arg), pos))
[...]

PR #1655, opened 2021-10-19
fix: Resolve differences between`jax_backend.concatenate` and other backends

phinate:
Resolves part of the `jax` component for #1654.

This should make the behaviour consistent with numpy, e.g. when using a list as input. The perf slowdown in the trivial case (input == `jax` array) is on the order of µs on my machine  (8 core mbp) with an example 1D array of size 10.

I made this PR purely for syntactical consistency -- of course, the slightly neater way may be to just always cast sequence-type inputs to tensors explicitly, so this isn't a hidden detail.
I couldn't think of a way to best test this consistency in the scope of this PR; if there was a suite that tested the output of common ops, then it fits nicely in there, but AFAIK you only test the core `pyhf` functionality instead.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/lower-bound-requirements.txt
+++ b/lower-bound-requirements.txt
@@ -1,5 +1,5 @@
 # core
-scipy==1.4.1  # c.f. PR #1001
+scipy==1.1.0
 click==7.0
 tqdm==4.56.0
 jsonschema==3.0.0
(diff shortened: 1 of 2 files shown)


phinate:
@kratsg thoughts on ths type of approach?

PR #1001, opened 2020-07-28
fix: Ensure TensorFlow backend Poisson compatibility with other backends

matthewfeickert:
# Description

Resolves #997 

~~As TensorFlow Probability `v0.11.0+` is needed for PR #817, and as Issue #997 makes it clear that TF/TFP minor releases can be as breaking as normal major releases, require TensorFlow `v2.3.0+` in the `v2.3.X` range and TensorFlow Probability `v0.11.0+` in the `v0.11.X` range.~~

Ensure compatibility of TensorFlow Poisson's behavior with PyTorch and JAX while Issue #293 is being resolved by using the differentiable TensorFlow operations `tf.logical_or` and `tf.where` to allow for checking if the situation in which `Poisson(n=0 | lam=0)` is being encountered. This provides a working solution for both the TF `v2.2.X` and TFP `V0.10.X` API behavior as well as for the TF `v2.3.+` and TFP `v0.11.0+` API behavior. Also disallow TF `v2.3.0` given https://github.com/tensorflow/tensorflow/pull/40789.

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Use differentiable TensorFlow operations to determine if Poisson(n=0 | lam=0) is being encountered and ensure a return value compatible with PyTorch and JAX
* Use TensorFlow releases compatible with v2.X with a lower bound of v2.2.1
   - Disallow TensorFlow v2.3.0 for accidental pinning of SciPy
* Use TensorFlow Probability releases compatible with v0.X with a lower bound of v0.10.1
* Use tf.errors.InvalidArgumentError for compatibility across TensorFlow releases
```

PR #1657, opened 2021-10-21
feat: Allow zero rate Poisson 

### LINKED ISSUE #293: Should Poisson[0,0] be handled especially?
# Question

Filing on behalf of @wiso for now since we need an issue to move and focus the discussion better. The initial PR #277 added a feature to use `scipy.special.xlogy` which was actually reverted in #280 because we started seeing divergence in calculations among the backends around the special lambda=0 result.

That said, the poisson is not well-defined for lambda=0, so it's not clear if we should define it for Poisson[0,0]. @wiso's discussion here https://github.com/diana-hep/pyhf/pull/277#issuecomment-423261532 is copied/pasted here verbatim. From this point on.

-------------------

By the way if I well understood the implementation for the poisson in the other backends is a gaussian, e.g.

https://github.com/diana-hep/pyhf/blob/cbb899fc134b7d156550dfdc9130154f264a5188/pyhf/tensor/tensorflow_backend.py#L198-L200

So I don't think you want to implement such a check:

```
In [2]: pyhf.tensorlib.poisson(3, 2)
Out[2]: 0.18044704431548356

In [3]: pyhf.set_backend(pyhf.tensor.mxnet_backend())

In [4]: pyhf.tensorlib.poisson(3, 2)
Out[4]: 

[0.21969564]
<NDArray 1 @cpu(0)>

In [5]: pyhf.set_backend(pyhf.tensor.pytorch_backend())

In [6]: pyhf.tensorlib.poisson(3, 2)
Out[6]: tensor([0.2197])
```

I don't know how the nan vs 1 problem can affect the optimization (probably you should reason in that way and not from the mathematical point of view (e.g. Poisson not defined for lam=0), in the same way as you are using a form of continuous Poissonian). By the way if we observe n=0 then the formula become `exp(-lam)` and its limit is 1 for `lam->0`, which is also the maximum. This is true also for the (discrete) Poisson distribution.

Here a stupid case where having a number for Poisson[0, 0] may be useful.

```
from pyhf import get_backend
from pyhf.utils import loglambdav
import pyhf.simplemodels

_, optimizer = get_backend()

pdf = pyhf.simplemodels.hepdata_like(signal_data=[1], bkg_data=[0.],
                                     bkg_uncerts=[10])  # last parameter should not count in this case

init_pars = pdf.config.suggested_init()
par_bounds = pdf.config.suggested_bounds()

NOBS = 10
print (optimizer.unconstrained_bestfit(loglambdav, [NOBS] + pdf.config.auxdata,
                                       pdf, init_pars, par_bounds)[0])
```

with this branch, result is 10 as expected. Also, if I put Let put `NOBS=0` as a result I get 0, without any RuntimeWarning.

With master branch (just pulled) with any numbers (I tried 10 or 0):

```
/home/turra/pyhf/pyhf/tensor/numpy_backend.py:186: RuntimeWarning: divide by zero encountered in log
  return np.exp(n * np.log(lam) - lam - gammaln(n + 1.))
/home/turra/pyhf/pyhf/tensor/numpy_backend.py:186: RuntimeWarning: invalid value encountered in multiply
  return np.exp(n * np.log(lam) - lam - gammaln(n + 1.))
     fun: nan
     jac: array([nan, nan])
 message: 'Iteration limit exceeded'
    nfev: 1404
     nit: 101
    njev: 101
  status: 9
 success: False
       x: array([nan, nan])
  File "mytest.py", line 15, in <module>
    pdf, init_pars, par_bounds)[0])
  File "/home/turra/pyhf/pyhf/optimize/opt_scipy.py", line 14, in unconstrained_bestfit
    assert result.success
AssertionError
```

It works only if I put a non-zero background.

matthewfeickert:
# Description

* Resolves #293
* Closes https://github.com/tensorflow/probability/issues/1025

Allow zero rate Poisson under the case that the limit `\lambda -> 0` is well defined as

```
    P(n | \lambda = 0) = { 1 if n = 0
                         { 0 otherwise
```

This PR enforces the [default behavior across all backends](https://github.com/scikit-hep/pyhf/issues/293#issuecomment-949044513).

This PR effectively reverts most of PR #1001 and PR #280, and reapplies most of @wiso's PR #277.

To allow for zero rate Poisson across all backends, the mimimum required versions are adjusted:
- Zero rate Poisson support was added to TensorFlow Probability in release `v0.11.0`, which requires `tensorflow>=2.3` (c.f. https://github.com/tensorflow/probability/issues/1025). As a result, the minimum compatible releases are bumped to TensorFlow `v2.3.1` and TensorFlow Probability `v0.11.0`.
- Zero rate Poisson support was added to PyTorch in release `v1.10.0`, so the minimum compatible release is set to that.
- This is already supported in SciPy stats, though @andrewfowlie asked that this be clarified in the docs in https://github.com/scipy/scipy/issues/12113

Note, that Issue #293 introduces the idea of a cutoff. That isn't off the table, but should be discussed in a separate Issue.

(Self note: As it will probably come up again at some point, the first citation of this being a thing that I can find is from this [Cross Validated post](https://stats.stackexchange.com/questions/190458/why-is-the-poisson-distribution-only-defined-for-a-positive-rate).)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Effectively reverts most of PR #1001 and PR #280, reapplies most
of PR #277
* Use scipy.special.xlogy in Poisson computation for numpy backend
and use jax.scipy.special.xlogy for jax backend
* Set minimum required PyTorch to v1.10 for API stability
   - c.f. https://github.com/pytorch/pytorch/pull/61511 in torch v1.10.0
* Set minimum required TensorFlow to v2.3.1 and TensorFlow Probability
to v0.11.0
   - tfp v0.11.0 supports zero rate Poisson and requires tensorflow>=2.3.0
* Add note to docs that limit Poisson(n = 0 | lambda -> 0) = 1 is being used
* Update tests to use limit Poisson(n = 0 | lambda -> 0) = 1 result
* Run doctest on only the latest Python release

Co-authored-by: Ruggero Turra <giurrero@gmail.com>
```

### DIFF
--- a/tests/test_tensor.py
+++ b/tests/test_tensor.py
@@ -287,7 +287,7 @@ def test_pdf_calculations(backend):
         ],
         nan_ok=True,
     )
-    # Allow poisson(lambda=0) under limit Poisson(lambda -> 0) = 1
+    # Allow poisson(lambda=0) under limit Poisson(n = 0 | lambda -> 0) = 1
     assert tb.tolist(
         tb.poisson(tb.astensor([0, 0, 1, 1]), tb.astensor([0, 1, 0, 1]))
     ) == pytest.approx([1.0, 0.3678794503211975, 0.0, 0.3678794503211975])
@@ -330,7 +330,7 @@ def test_pdf_calculations_pytorch(backend):
         ],
     )
 
-    # Allow poisson(lambda=0) under limit Poisson(lambda -> 0) = 1
+    # Allow poisson(lambda=0) under limit Poisson(n = 0 | lambda -> 0) = 1
     assert tb.tolist(
         tb.poisson(tb.astensor([0, 0, 1, 1]), tb.astensor([0, 1, 0, 1]))
     ) == pytest.approx([1.0, 0.3678794503211975, 0.0, 0.3678794503211975])


matthewfeickert:
Relevant RTD build page: https://pyhf--1657.org.readthedocs.build/en/1657/_generated/pyhf.tensor.numpy_backend.numpy_backend.html#pyhf.tensor.numpy_backend.numpy_backend.poisson

PR #1665, opened 2021-10-22
feat: Allow schema validation with tensor types

### LINKED ISSUE #1662: Extend schema validation to support tensor types
### Summary

Schema validation for lists looks just at regular python lists. It should be possible to extend it to support the tensor types that `pyhf` supports.

We can get this working if each tensor backend gives the tensor type: `(list, numpy_backend.array_type, jax_backend.array_type, ...)`. 

I'm not sure if we can or can't rely on the backend here. We've got the actual tensor object that we need to validate against... and the validator won't necessarily have access to the current backend -- but we can teach it that if needed.

```python
isinstance(instance, list) or isinstance(instance, pyhf.tensorlib.array_type)
```

and this should fix some common issues where we need to pass in `validate=False` explicitly to build the models...

### Additional Information

```python
>>> from jsonschema import Draft6Validator
>>> from jsonschema.validators import extend
>>> import numpy as np
>>> def is_array_or_tensor(checker, instance):
...   return isinstance(instance, (list, np.ndarray))
... 
>>> type_checker = Draft6Validator.TYPE_CHECKER
>>> type_checker = Draft6Validator.TYPE_CHECKER.redefine('array', is_array_or_tensor)
>>> Validator = extend(Draft6Validator, type_checker=type_checker)
>>> validator_foo_required = Validator(schema={"type": "array"})
>>> example = [1, 2, 3]
>>> example_np = np.array([1,2,3])
>>> isinstance(example_np, np.ndarray)
True
>>> validator_foo_required.validate(example)
>>> validator_foo_required.validate(example_np)
```

Inspired by Julian/jsonschema#727

### Code of Conduct

kratsg:
# Pull Request Description

Teach pyhf.schema.validate a new trick!

Resolves #1662. This allows for schema validation to work with objects (specifications) that contain tensors in them. This is done by extending the base type checker to teach it about the tensor types from the backends used in pyhf. Should also help with validation for differentiability (#882).

Notes:
- numpy_backend documented but with array_type first and array_subtype second
- torch/jax/tensorflow documented with array_subtype first and array_type second, but missing the values of the attributes
- https://github.com/sphinx-doc/sphinx/issues/10293

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Teach pyhf.schema.validate about the tensors from the backends.
* Define backend.array_type and backend.array_subtype to be used by the
  type checking in schema validation.
* Define pyhf.tensor.array_types and pyhf.tensor.array_subtypes to list
  all types currently loaded in the session.
* Add docstring example for pyhf.schema.load_schema.
* Add array information to docs class summary template.
```


kratsg:
This bug is tricky. Need a way for sphinx to document a class attribute as an attribute, and not the underlying value. I wonder how I can hide this away better, because I need access to `class.attr` and I don't want it to be a `property` object. Maybe @henryiii or @jpivarski have ideas?

the short of it is

```python
class backend:
  # my documentation for array_type
  array_type = np.ndarray
```

is getting expanded in sphinx with documentation/docstrings for `np.ndarray` and treated like a class/function, instead of treating `array_type` as an attribute, e.g. via something like

```python
class backend:

  @property
  def array_type(self):
  """my documentation for array_type"""
    return np.ndarray
```

which is fine, but feels pretty annoying to wrap around to get access to it at the class level (`backend.fget(None)`) . Is there anything better? Maybe I need to mimic a `@classproperty` like django has.

henryiii:
Can you add it to `:exclude-members:`?

kratsg:
> Can you add it to `:exclude-members:`?

Not a bad idea. I would prefer to document it as a property.

kratsg:
To be clear, by assigning a class variable to a function, sphinx is now treating it as a function, rather than a member variable.

kratsg:
> To be clear, by assigning a class variable to a function, sphinx is now treating it as a function, rather than a member variable. At least it shows up under the object's members, rather than under attributes. The only way I guess it to override sphinx's default?

Filed an issue for now: sphinx-doc/sphinx#10293 with a hacky-ish workaround.

matthewfeickert:
Relevant RTD build: https://pyhf--1665.org.readthedocs.build/en/1665/_generated/pyhf.schema.load_schema.html

matthewfeickert (approved):
Just stepped through this and LGTM. :+1: Thanks a lot for this @kratsg!

PR #1691, opened 2021-11-10
fix: Speed-up readxml by caching key lookup instead of using try/except

### LINKED ISSUE #1690: DeserializationError too frequent when `HistoPath` is specified in HiFa XML
### Summary

When a `<Sample>` specifies the `HistoPath` and it's not empty, we will run into very frequent deserialization errors from `uproot` which is very slow to raise an exception (see scikit-hep/uproot4#504).

`pyhf` needs to fix this to be a bit smarter in how to check valid keys, and in particular, fix up its logic to not hit a/rely on `DeserializationError`.

### OS / Environment

```console
$ system_profiler -detailLevel mini SPSoftwareDataType | head -n 6
Software:

    System Software Overview:

      System Version: macOS 10.14.6 (18G9323)
      Kernel Version: Darwin 18.7.0
```

### Steps to Reproduce

See #1687 for the fundamental issue. It's reproducible using private workspaces for now, but can be confirmed reproducible.

```
$ time pyhf xml2json monotop_twmetComb0L1LBoosted_allCRs_normDStoDR_unblind_sigTheo_envelope__pmoder_sig_a250_DM10_H900_tb1_st0p7/NormalMeasurement.xml --basedir monotop_twmetComb0L1LBoosted_allCRs_normDStoDR_unblind_sigTheo_envelope__pmoder_sig_a250_DM10_H900_tb1_st0p7
Processing ./NormalMeasurement_CR_0LBoosted_ttbar_cuts.xml:   0%|                           | 0/1 [00:00<?, ?channel/smonotop_twmetComb0L1LBoosted_allCRs_normDStoDR_unblind_sigTheo_envelope__pmoder_sig_a250_DM10_H900_tb1_st0p7/results/monotop_twmetComb0L1LBoosted_allCRs_normDStoDR_unblind_sigTheo_envelope__pmoder_sig_a250_DM10_H900_tb1_st0p7/Exclusion_combined_NormalMeasurement_model.root not in filecache
path=CR_0LBoosted_ttbar_cuts_hists/data, name=hData_CR_0LBoosted_ttbar_obs_cuts
deserialization error, trying fullname=CR_0LBoosted_ttbar_cuts_hists/data/hData_CR_0LBoosted_ttbar_obs_cuts instead
                                                                                                                      monotop_twmetComb0L1LBoosted_allCRs_normDStoDR_unblind_sigTheo_envelope__pmoder_sig_a250_DM10_H900_tb1_st0p7/results/monotop_twmetComb0L1LBoosted_allCRs_normDStoDR_unblind_sigTheo_envelope__pmoder_sig_a250_DM10_H900_tb1_st0p7/Exclusion_combined_NormalMeasurement_model.root in filecache
path=CR_0LBoosted_ttbar_cuts_hists/Top0LBoosted, name=hTop0LBoostedNom_CR_0LBoosted_ttbar_obs_cuts
deserialization error, trying fullname=CR_0LBoosted_ttbar_cuts_hists/Top0LBoosted/hTop0LBoostedNom_CR_0LBoosted_ttbar_obs_cuts instead
                                                                                                                      monotop_twmetComb0L1LBoosted_allCRs_normDStoDR_unblind_sigTheo_envelope__pmoder_sig_a250_DM10_H900_tb1_st0p7/results/monotop_twmetComb0L1LBoosted_allCRs_normDStoDR_unblind_sigTheo_envelope__pmoder_sig_a250_DM10_H900_tb1_st0p7/Exclusion_combined_NormalMeasurement_model.root in filecache
path=CR_0LBoosted_ttbar_cuts_hists/Top0LBoosted, name=hTop0LBoostedEG_EffLow_CR_0LBoosted_ttbar_obs_cutsNorm
deserialization error, trying fullname=CR_0LBoosted_ttbar_cuts_hists/Top0LBoosted/hTop0LBoostedEG_EffLow_CR_0LBoosted_ttbar_obs_cutsNorm instead
```

### File Upload (optional)

_No response_

### Expected Results

`pyhf xml2json` should be fast.

### Actual Results

```console
`pyhf xml2json` is slow.
```

### pyhf Version

```console
This impacts all pyhf versions up to 0.6.4.
```

### Code of Conduct

kratsg:
# Pull Request Description

This resolves #1690 and was brought up by the discussion in #1687.

Fundamentally, when accessing a path that does not exist in `uproot` via `f.__getitem__`, it would then trigger a user-friendly levenshtein distance calculation to recommend similarly-named keys. However, in the case of `pyhf`, it was assumed that `f[name]` would be fast to return or at least fast to raise an exception, but this is a bad assumption.

Instead, the set of keys in the file (without cycle numbers) is cached along with the file itself, and we will rely on a key lookup in the set instead.

This only occurs in HistFactory when the XML has `<Sample/>` with `HistoPath` attributes and a suboptimal computation path in `pyhf` is then used, because we need to build `fullname` in order to figure out the right path for the histogram.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Refactor the readxml logic
* Cache the set of keys in a file and use that to look up if a histogram
exists or not
   - Avoids issue of accessing a path that does not exist in uproot via
     f.__getitem__, triggering a user-friendly levenshtein distance
     calculation to recommend similarly-named keys.
* Add tests and test files for key name lookups
```


kratsg:
@gollumben - this has the fix.

matthewfeickert (approved):
LGTM. :+1: Thanks @kratsg!

src/pyhf/readxml.py
```diff
@@ -59,19 +59,21 @@ def import_root_histogram(rootdir, filename, path, name, filecache=None):
...
-    except (KeyError, uproot.deserialization.DeserializationError):
-        fullname = "/".join([path, name])
-        try:
-            hist = f[fullname]
-        except KeyError:
-            raise KeyError(
-                f'Both {name} and {fullname} were tried and not found in {fullpath}'
-            )
+    elif fullname in keys:
+        hist = f[fullname]
```
matthewfeickert, line 72: @kratsg Can we get coverage here though?

kratsg: done

matthewfeickert (approved):
Thanks for the coverage boost. LGTM now.

matthewfeickert:
@gollumben If you are willing to [install a development version of `pyhf` from GitHub](https://pyhf.readthedocs.io/en/v0.6.3/installation.html#install-latest-development-version-from-github) your Issue described in Discussion #1687 should be fixed now. Please let us know if you encounter any other problems. :+1:

gollumben:
Hi @kratsg, @matthewfeickert and @lukasheinrich, thank you very much for your help and the very fast turn around time! It takes 3 minutes to convert the workspace now instead of the previous 30 h (~600x speed increase), which is amazing!

PR #1697, opened 2021-11-12
refactor: Make contrib.utils.download robust to archive file types

### LINKED ISSUE #1111: Extend pyhf contrib download to allow for uncompressed targets
# Description

This feature request was a suggestion of @lukasheinrich's:

> I wonder whether we should have something like 
>
> ```
> $ pyhf contrib download http://pyhf.scikit-hep.org/examples/shapesysexample
> ```
> 
> or similar, as a way to get starter workspaces.

This should be easy to implement and I think mostly will just require making

https://github.com/scikit-hep/pyhf/blob/c1727cec1ea29d2ffe834ebfcdb0d3b9add47396/src/pyhf/contrib/utils.py#L55-L58

be able to deal with non-compressed targets.

### LINKED ISSUE #1486: Make download tests robust to HEPData outages by mocking of tar.gz files

### LINKED ISSUE #1519: Make `pyhf contrib download` be able to handle multiple compression types
# Description

@alexander-held has pointed out that at the moment (`pyhf` `v0.6.2`) `pyhf contrib download` can't download the probability models for:

- https://www.hepdata.net/record/ins1827025 (has [models as `.tar` archive](https://www.hepdata.net/record/resource/1737595?view=true))
- https://www.hepdata.net/record/ins1831504 (has [models as `.zip` archive](https://www.hepdata.net/record/resource/1981552?view=true))

as they both don't have the models compressed as `tar.gz`. At the moment, `pyhf contrib download` can only handle `.tar.gz`

https://github.com/scikit-hep/pyhf/blob/cc7cf0f50ff891abce3c3f8a922f57b3d9e25a65/src/pyhf/contrib/utils.py#L56-L64

but that should be expanded.

It is reasonable to have `pyhf contirb download` be more robust, but ti would also be helpful if HEPData archive uploads had a standardized archive compression format.

cc @kratsg

### LINKED ISSUE #1672: Refactor `contrib.utils.download` internals
Some other suggestions for improvements to this code:
* Check the `response.status_code` and return an error message if not OK.
* Use [`tarfile.is_tarfile`](https://docs.python.org/3/library/tarfile.html#tarfile.is_tarfile) to check that `response.content` is actually a tarball and return an error message if not.
* Remove `mode="r|gz"` or replace it with `mode="r"` or `mode="r:*"` for reading with transparent compression, so that the code works also with uncompressed tarballs (see #1111 and #1519), where the media type is still `application/x-tar`.
* Maybe add an option to download a zipfile instead of a tarball (see #1519), then you'd need `headers={'Accept': 'application/zip'}` in the request and [`zipfile.is_zipfile`](https://docs.python.org/3/library/zipfile.html#zipfile.is_zipfile) to check the response content.  You could use the Python [`zipfile`](https://docs.python.org/3/library/zipfile.html) module to unpack, but maybe easier to use [`shutil.unpack_archive`](https://docs.python.org/3/library/shutil.html#shutil.unpack_archive) for both tarballs and zipfiles.

Making these changes should not break the functionality with the current situation (where https://doi.org/10.17182/hepdata.89408.v1/r2 returns the tarball directly).  I'd therefore recommend you make them ASAP before the next `pyhf` release.  After we redirect the DOI to the landing page, probably in the next few weeks, the DOI will return the HTML landing page instead of the tarball unless the request contains the `Accept: application/x-tar` header.

_Originally posted by @GraemeWatt in https://github.com/scikit-hep/pyhf/issues/1491#issuecomment-951810848_

matthewfeickert:
# Description

* Resolves #1672
* Resolves #1111
* Resolves #1519
* Resolves #1486

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Refactor the internals of contrib.utils.download to be more robust and
handle uncompressed tarfile targets as well as zipfile targets.
* Mock HEPData responses to avoid load on HEPData API and avoid testing
problems due to HEPData outages. This generally makes the test suite more
robust.
* Add requests-mock as a dependency to the 'test' extra. requests-mock makes
mocking requests calls easy and allows for a pytest fixture.
* Add tests for contrib.utils.download Python API to test outside of script runner
```

### DIFF
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -66,10 +66,8 @@ try:
         # The HEPData landing page for the resource file can check if the Accept
         # request HTTP header matches the content type of the resource file and
         # return the content directly if so.
-        # TODO: Figure out how to accept both x-tar and zip headers. :?
-        # with requests.get(
-        #     archive_url, headers={"Accept": "application/zip"}
-        # ) as response:
+        # TODO: Figure out how to accept headers of both application/x-tar and
+        # application/zip.
         with requests.get(
             archive_url, headers={"Accept": "application/x-tar"}
         ) as response:
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -5,6 +5,7 @@ import tarfile
 import zipfile
 from io import BytesIO
 from pathlib import Path
+from shutil import rmtree
 from urllib.parse import urlparse
 
 from pyhf import exceptions
@@ -100,6 +101,8 @@ try:
                         archive.extractall(output_directory)
                 else:
                     output_directory = Path(output_directory)
+                    if output_directory.exists():
+                        rmtree(output_directory)
                     with zipfile.ZipFile(BytesIO(response.content)) as archive:
                         archive.extractall(output_directory)
 
@@ -109,7 +112,9 @@ try:
                     # temporary path and then replace the output directory target with
                     # the contents at the temporary path.
                     child_path = [child for child in output_directory.iterdir()][0]
-                    _tmp_path = Path(output_directory.name + "_tmp_")
+                    _tmp_path = output_directory.parent.joinpath(
+                        Path(output_directory.name + "_tmp_")
+                    )
                     child_path.replace(_tmp_path).replace(output_directory)
 
 
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -71,6 +71,11 @@ try:
                     + "There is either something temporarily wrong with the archive host"
                     + f" or {archive_url} is an invalid URL."
                 )
+            if not tarfile.is_tarfile(BytesIO(response.content)):
+                raise exceptions.InvalidArchive(
+                    f"The archive downloaded from {archive_url} is not a tarfile"
+                    + " and so can not be opened as one."
+                )
             if compress:
                 with open(output_directory, "wb") as archive:
                     archive.write(response.content)
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -122,7 +122,8 @@ try:
                             Path(output_directory.name + "__tmp__")
                         )
                         child_path.replace(_tmp_path)
-                        # the zipfile could contain remnant __MACOSX directories from creation time
+                        # the zipfile could contain remnant __MACOSX directories
+                        # from creation time
                         rmtree(output_directory)
                         _tmp_path.replace(output_directory)
 
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -81,7 +81,7 @@ try:
                     archive.write(response.content)
             else:
                 with tarfile.open(
-                    mode="r|gz", fileobj=BytesIO(response.content)
+                    mode="r:*", fileobj=BytesIO(response.content)
                 ) as archive:
                     archive.extractall(output_directory)
 
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -90,6 +90,7 @@ try:
                     archive.write(response.content)
             else:
                 if _format == "tar":
+                    # Use transparent compression to allow for .tar or .tar.gz
                     with tarfile.open(
                         mode="r:*", fileobj=BytesIO(response.content)
                     ) as archive:
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -121,11 +121,9 @@ try:
                         _tmp_path = output_directory.parent.joinpath(
                             Path(output_directory.name + "__tmp__")
                         )
-                        # TODO: Once Python 3.7 support is dropped these two lines can be
-                        # combined by chaining the replace calls. This won't work in
-                        # Python 3.7 as pathlib.Path.replace was:
-                        # > Changed in version 3.8: Added return value, return the new Path instance.
                         child_path.replace(_tmp_path)
+                        # the zipfile could contain remnant __MACOSX directories from creation time
+                        rmtree(output_directory)
                         _tmp_path.replace(output_directory)
 
 
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -85,9 +85,10 @@ try:
                     archive.write(response.content)
             else:
                 # Support for file-like objects for tarfile.is_tarfile was added
-                # in Python 3.9, so as can't do tarfile.is_tarfile(BytesIO(response.content))
-                # just use a 'try except' block to determine if archive is a
-                # valid tarfile.
+                # in Python 3.9, so as pyhf is currently Python 3.7+ then can't
+                # do tarfile.is_tarfile(BytesIO(response.content)).
+                # Instead, just use a 'try except' block to determine if the
+                # archive is a valid tarfile.
                 # TODO: Simplify after pyhf is Python 3.9+ only
                 try:
                     # Use transparent compression to allow for .tar or .tar.gz
--- a/src/pyhf/contrib/utils.py
+++ b/src/pyhf/contrib/utils.py
@@ -4,6 +4,7 @@ import logging
 import tarfile
 import zipfile
 from io import BytesIO
+from pathlib import Path
 from urllib.parse import urlparse
 
 from pyhf import exceptions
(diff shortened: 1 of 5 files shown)


src/pyhf/contrib/utils.py
```diff
@@ -62,17 +65,52 @@ def download(archive_url, output_directory, force=False, compress=False):
         # The HEPData landing page for the resource file can check if the Accept
         # request HTTP header matches the content type of the resource file and
         # return the content directly if so.
+        # TODO: Figure out how to accept both x-tar and zip headers. :?
+        # with requests.get(
+        #     archive_url, headers={"Accept": "application/zip"}
+        # ) as response:
```
matthewfeickert, line 71: @kratsg You have any experience with this?

matthewfeickert:
I've put some temporary files up at https://feickert.web.cern.ch/feickert/pyhf/tests/ to test this out and at this point the following work:

```console
$ pyhf contrib download --force --verbose https://feickert.web.cern.ch/feickert/pyhf/tests/1Lbb-likelihoods.tar.gz /tmp/1Lbb-likelihoods
/tmp/1Lbb-likelihoods/patchset.json
/tmp/1Lbb-likelihoods/BkgOnly.json
/tmp/1Lbb-likelihoods/README.md
$ pyhf contrib download --force --verbose https://feickert.web.cern.ch/feickert/pyhf/tests/1Lbb-likelihoods.tar /tmp/1Lbb-likelihoods
/tmp/1Lbb-likelihoods/patchset.json
/tmp/1Lbb-likelihoods/BkgOnly.json
/tmp/1Lbb-likelihoods/README.md
# The zipfile should work soon
```

src/pyhf/contrib/utils.py
matthewfeickert: Actually `.zip` seems to work with `application/x-tar`(?) so maybe this isn't needed?

src/pyhf/contrib/utils.py
```diff
@@ -62,17 +66,64 @@ def download(archive_url, output_directory, force=False, compress=False):
...
+                    with tarfile.open(
+                        mode="r:*", fileobj=BytesIO(response.content)
+                    ) as archive:
+                        archive.extractall(output_directory)
+                except tarfile.ReadError:
+                    if not zipfile.is_zipfile(BytesIO(response.content)):
+                        raise exceptions.InvalidArchive(
+                            f"The archive downloaded from {archive_url} is not a tarfile"
+                            + " or a zipfile and so can not be opened as one."
+                        )
```
matthewfeickert, line 104: @kratsg as you have a lot more experience mocking out things in the codebase then I do, do you have any thoughts on how to best approach mocking the requests `response.content` as binary representations of a tarfile and zipfile?

alexander-held: You could mock all the way through the calls I think: mock `tarfile.open` to succeed / fail, `BytesIO` to return anything you'd like, and mock `zipfile.is_zipfile` to return `False` given whatever return value you pick from `BytesIO`. That would assume though that those APIs do not break, which is probably safe enough for the standard library. That way it would not matter what `response.content` is either, that could be a mock itself too.

matthewfeickert:
There's a Python 3.7 problem for `pahtlib` that I'm not seeing in Python 3.9

```pytb
                        child_path = [child for child in output_directory.iterdir()][0]
                        _tmp_path = output_directory.parent.joinpath(
                            Path(output_directory.name + "__tmp__")
                        )
>                       child_path.replace(_tmp_path).replace(output_directory)
E                       AttributeError: 'NoneType' object has no attribute 'replace'
```

I'll need to resolve this later, but I'm pretty sure this is because [`pathlib.Path.replace`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.replace) 
> Changed in version 3.8: Added return value, return the new Path instance.

so I think I can't chain these until we drop Python 3.7 support.

matthewfeickert:
@GraemeWatt your review is welcome here as well. :)

kratsg (approved):

matthewfeickert:
Can confirm with regards to Issue #1519 that the following now works :+1: 

```console
$ curl -sL https://www.hepdata.net/record/resource/1981552?view=true -o SUSY-2018-14.zip
$ python -m zipfile --list SUSY-2018-14.zip 
File Name                                             Modified             Size
likelihood/                                    2021-02-15 08:21:20            0
__MACOSX/._likelihood                          2021-02-15 08:21:20          312
likelihood/SRmm_left_bkgonly.json              2020-11-02 16:29:16         2023
__MACOSX/likelihood/._SRmm_left_bkgonly.json   2020-11-02 16:29:16          312
likelihood/Stauleft_patchset.json              2021-02-15 08:21:14       569930
likelihood/SRmm_patchset.json                  2021-02-15 08:20:34       476160
likelihood/SRmm_bkgonly.json                   2020-11-02 16:29:14         2023
__MACOSX/likelihood/._SRmm_bkgonly.json        2020-11-02 16:29:14          312
likelihood/SRee_bkgonly.json                   2020-11-02 16:29:08         2023
__MACOSX/likelihood/._SRee_bkgonly.json        2020-11-02 16:29:08          368
likelihood/SRmm_right_patchset.json            2021-02-15 08:20:54       469313
likelihood/Stau_patchset.json                  2021-02-15 08:21:04       568408
likelihood/README.md                           2020-12-17 11:02:54         1911
__MACOSX/likelihood/._README.md                2020-12-17 11:02:54          368
likelihood/SRee_patchset.json                  2021-02-15 08:20:24       392287
likelihood/SRee_left_bkgonly.json              2020-11-02 16:29:10         2023
__MACOSX/likelihood/._SRee_left_bkgonly.json   2020-11-02 16:29:10          312
likelihood/SRee_right_patchset.json            2021-02-15 08:20:14       395526
likelihood/Comb_patchset.json                  2021-02-15 08:19:46       563860
__MACOSX/likelihood/._Comb_patchset.json       2021-02-15 08:19:46          276
likelihood/Stauleft_bkgonly.json               2020-11-02 16:29:22         3516
__MACOSX/likelihood/._Stauleft_bkgonly.json    2020-11-02 16:29:22          312
likelihood/SRee_right_bkgonly.json             2020-11-02 16:29:12         2023
__MACOSX/likelihood/._SRee_right_bkgonly.json  2020-11-02 16:29:12          312
likelihood/Stau_bkgonly.json                   2020-11-02 16:29:20         3516
__MACOSX/likelihood/._Stau_bkgonly.json        2020-11-02 16:29:20          312
likelihood/SRmm_left_patchset.json             2021-02-15 08:20:44       478671
likelihood/SRmm_right_bkgonly.json             2020-11-02 16:29:18         2023
__MACOSX/likelihood/._SRmm_right_bkgonly.json  2020-11-02 16:29:18          312
likelihood/SRee_left_patchset.json             2021-02-15 08:20:04       396176
likelihood/Comb_bkgonly.json                   2020-11-02 16:29:06         3516
__MACOSX/likelihood/._Comb_bkgonly.json        2020-11-02 16:29:06          368
$ pyhf contrib download --verbose https://www.hepdata.net/record/resource/1981552?view=true /tmp/ins1831504
/tmp/ins1831504/SRmm_right_patchset.json
/tmp/ins1831504/Stauleft_patchset.json
/tmp/ins1831504/Comb_bkgonly.json
/tmp/ins1831504/SRmm_left_bkgonly.json
/tmp/ins1831504/SRee_right_patchset.json
/tmp/ins1831504/SRmm_bkgonly.json
/tmp/ins1831504/SRmm_left_patchset.json
/tmp/ins1831504/Stau_patchset.json
/tmp/ins1831504/Stauleft_bkgonly.json
/tmp/ins1831504/SRee_left_bkgonly.json
/tmp/ins1831504/SRee_patchset.json
/tmp/ins1831504/SRmm_right_bkgonly.json
/tmp/ins1831504/SRee_bkgonly.json
/tmp/ins1831504/SRee_right_bkgonly.json
/tmp/ins1831504/README.md
/tmp/ins1831504/Comb_patchset.json
/tmp/ins1831504/SRee_left_patchset.json
/tmp/ins1831504/Stau_bkgonly.json
/tmp/ins1831504/SRmm_patchset.json

```

matthewfeickert:
I'm going to merge this now, but if people have additional comments please either add them here after the fact or make a new Issue and I'll address them.

src/pyhf/contrib/utils.py
GraemeWatt: Sorry I didn't comment before you merged.  The `Accept` headers will be ignored if `archive_url` is a HEPData URL (ending in `?view=true`) that returns the content directly.  The `Accept` headers will only become relevant when `archive_url` is a HEPData DOI for a resource file after we redirect to a landing page.  We're still working on the relevant PR (HEPData/hepdata#412) after it expanded in scope.  The default `Accept` header is `*/*` for Python `requests` (and `curl`), different from the [default `Accept` values of most web browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values) where `*/*` has a lower weighting ([quality value](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values)) than `text/html`.  With our current implementation, if `*/*` has weight 1 in a request to the landing page, the content will be returned directly instead of the HTML landing page.  This means you don't strictly need to specify the `Accept` header with Python `requests`.  But I'd recommend you keep the `Accept` header to avoid accepting content types that your download code can't process.  To accept both tar and zip files, you just need to use `headers={"Accept": "application/x-tar, application/zip"}`.  We will return a [406 Not Acceptable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406) status code if a request is made to the landing page where the `Accept` header does not match the media type of the content.

matthewfeickert: Thanks @GraemeWatt! This is all super helpful and I truly appreciate the level of detail &mdash; helps a lot for someone like me where this isn't my forte.

PR #1639, opened 2021-10-14
refactor: Simplified parameters

lukasheinrich:
# Description

- Resolves #1634 
- Resolves #856 
- Resolves #540

this is just a commit on top of #1625 that cleans up parameter handling and will a lot of nice things (user-configurable constraint terms, dropping constraints, etc)

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Remove parameter pruning due to non-defined constraints (will
move all parameter allocation to modifier_builders
* Remove dead code in src/pyhf/constraints.py
* Refactor pyhf.pdf._nominal_builder
* Add tests for modifier holes and fixed sets
```

PR #1706, opened 2021-11-18
fix: Ensure `_ModelConfig.suggested_fixed` list contains only booleans for all modifiers

### LINKED ISSUE #1701: Revert slight change in return types of `suggested_fixed`
### Summary

https://github.com/scikit-hep/pyhf/pull/1639 very slightly changed the return type of `model.config.suggested_fixed()`. It used to be `List[bool]`, but now some elements can be `numpy.bool_`, which behaves differently (see https://github.com/numpy/numpy/issues/9646):
```python
>>> import numpy as np
>>> isinstance(np.bool_(True), bool)
False
```
This makes typing / type checking in libraries using `pyhf` more complicated (spotted via https://github.com/scikit-hep/cabinetry/pull/301). Would it be possible to consistently return `List[bool]` again instead?

### Additional Information

The example below shows how `numpy.bool_` entries appear for some, but not all models.

```python (paste below)
import pyhf

spec = {
    "channels": [
        {
            "name": "SR",
            "samples": [
                {
                    "data": [25, 5],
                    "modifiers": [
                        {
                            "data": [5, 2],
                            "name": "staterror_SR",
                            "type": "staterror",
                        },
                    ],
                    "name": "Signal",
                }
            ],
        },
    ],
    "measurements": [{"config": {"parameters": [], "poi": ""}, "name": "fit"}],
    "observations": [
        {"data": [35, 8], "name": "SR"},
    ],
    "version": "1.0.0",
}


def check_types(model):
    for i, par in enumerate(model.config.suggested_fixed()):
        if not isinstance(par, bool):
            print(f"{model.config.par_names()[i]} has type {type(par)}")


model = pyhf.simplemodels.correlated_background([5], [10], [11], [9])
print("checking correlated_background model")
check_types(model)

model = pyhf.Workspace(spec).model()
print("checking custom model")
check_types(model)
```
Output for 9fbbbf9740175fcdfe73fedc4351953be18d2664 and later (including current `master`):
```
checking correlated_background model
checking custom model
staterror_SR[0] has type <class 'numpy.bool_'>
staterror_SR[1] has type <class 'numpy.bool_'>
```
Output for 5ea4e0a7751aa446c286b9697e01a9439622ebb5:
```
checking correlated_background model
checking custom model
```

### Code of Conduct

alexander-held:
# Description

The `staterror` refactoring in #1639 caused `model.config.suggested_fixed()` to return `numpy.bool_` elements for `staterror` modifiers. Those behave differently from `bool` objects, and for consistency I believe it is useful to have a consistent return type, in line with the documented return type of `List[bool]`:

https://github.com/scikit-hep/pyhf/blob/f25b8ddc6ab34d089954eb4898d54220f7f2defa/src/pyhf/pdf.py#L349-L355

#1701 includes an example to showcase the current status.

There are many places at which a change could be introduced to restore the old behavior. The current implementation applies the change at the first possible point, where the list is initially created. Alternatively, a conversion could probably happen in `paramset` or at the last point in `_ModelConfig.suggested_fixed`. Please let me know if you would prefer another implementation. I picked this approach for consistency with other modifiers, such that `required_parset()["fixed"]` consistently returns `bool` / `List[bool]`.

This also adds tests for the return type behavior for `staterror` modifiers, I'm happy to remove them or add more for other modifiers if desired.

Since this is about types, I thought about adding type hints for the affected code and am happy to do so if you'd like.

resolves #1701 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Have the fixed parameters selection in staterror_builder be a list
of Bools over numpy.bool_
* Add type hints for the relevant sections
* Add test to test_modifiers.py to require the fixed parameters
selection list to be Bool
```

### DIFF
--- a/src/pyhf/modifiers/staterror.py
+++ b/src/pyhf/modifiers/staterror.py
@@ -106,7 +106,7 @@ class staterror_builder:
                 modifier_data['data']['mask'] = masks[modname]
             sigmas = relerrs[masks[modname]]
             # list of bools, consistent with other modifiers (no numpy.bool_)
-            fixed = (sigmas == 0).tolist()
+            fixed = default_backend.tolist(sigmas == 0)
             # ensures non-Nan constraint term, but in a future PR we need to remove constraints for these
             sigmas[fixed] = 1.0
             self.required_parsets.setdefault(parname, [required_parset(sigmas, fixed)])
--- a/src/pyhf/modifiers/staterror.py
+++ b/src/pyhf/modifiers/staterror.py
@@ -1,4 +1,5 @@
 import logging
+from typing import List
 
 import pyhf
 from pyhf import events
@@ -8,7 +9,7 @@ from pyhf.parameters import ParamViewer
 log = logging.getLogger(__name__)
 
 
-def required_parset(sigmas, fixed):
+def required_parset(sigmas, fixed: List[bool]):
     n_parameters = len(sigmas)
     return {
         'paramset_type': 'constrained_by_normal',
--- a/src/pyhf/modifiers/staterror.py
+++ b/src/pyhf/modifiers/staterror.py
@@ -104,7 +104,8 @@ class staterror_builder:
             for modifier_data in self.builder_data[modname].values():
                 modifier_data['data']['mask'] = masks[modname]
             sigmas = relerrs[masks[modname]]
-            fixed = [s == 0 for s in sigmas]
+            # list of bools, consistent with other modifiers (no numpy.bool_)
+            fixed = (sigmas == 0).tolist()
             # ensures non-Nan constraint term, but in a future PR we need to remove constraints for these
             sigmas[fixed] = 1.0
             self.required_parsets.setdefault(parname, [required_parset(sigmas, fixed)])
--- a/src/pyhf/parameters/paramsets.py
+++ b/src/pyhf/parameters/paramsets.py
@@ -1,3 +1,5 @@
+from typing import List
+
 import pyhf
 
 __all__ = [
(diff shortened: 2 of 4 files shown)


matthewfeickert (approved):
Thanks for this PR @alexander-held. :+1: 

> Since this is about types, I thought about adding type hints for the affected code and am happy to do so if you'd like.

If you'd like to add them we certainly won't complain. :) Though I'm happy to approve this as is.

alexander-held:
I added type hints for the relevant part of the code, only relating to `suggested_fixed`. I verified them all at runtime with `typeguard` as well and they are correct for the examples I tried (#1701 and slight modifications).

matthewfeickert:
(I'm not sure why the CodeQL check is hanging, but this PR is obviously fine with or without it).

src/pyhf/modifiers/staterror.py
```diff
@@ -104,7 +105,8 @@ def finalize(self):
             for modifier_data in self.builder_data[modname].values():
                 modifier_data['data']['mask'] = masks[modname]
             sigmas = relerrs[masks[modname]]
-            fixed = [s == 0 for s in sigmas]
+            # list of bools, consistent with other modifiers (no numpy.bool_)
+            fixed = (sigmas == 0).tolist()
```
kratsg, line 109:
```suggestion
            fixed = default_backend.tolist(sigmas == 0)
```

we can't guarantee the default_backend is numpy or has `.tolist`.

kratsg (changes requested):

src/pyhf/modifiers/staterror.py
matthewfeickert: Good catch. I'm surprised this isn't caught already in the tests.

kratsg (approved):

PR #1707, opened 2021-11-18
ci: Quote GitHub Action python-version number as YAML strings

matthewfeickert:
# Description

As preventative actions, quote the `python-version` numbers in CI as without quotes [YAML will treat `3.10` as a float and interpret it as `3.1`](https://twitter.com/HEPfeickert/status/1445860336565948416) which is certainly not desired. :) This won't affect us until all backends are on Python 3.10, but it will keep us from having the CI fail weirdly in the future.

Also simplify the macos-latest inclusion with `include` keyword.

Please describe the purpose of this pull request in some detail. Reference and link to any relevant issues or pull requests.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Quote all python-version numbers in CI. This is done as a preventive
action to set a YAML string precedence for when Python 3.10 is added
to testing, as without quotes YAML will treat `3.10` as a float and 
interpret it as `3.1`.
* Simplify the inclusion of the latest Python version on macos-latest
by using the `include` keyword in the CI strategy matrix to add it
instead of excluding all Python versions to not test.
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,13 +21,11 @@ jobs:
       || (github.event_name == 'push' && github.ref != 'refs/heads/master' && contains(github.event.head_commit.message, '[ci all]'))
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest]
         python-version: ['3.7', '3.8', '3.9']
-        exclude:
+        include:
           - os: macos-latest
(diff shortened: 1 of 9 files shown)


matthewfeickert:
I'm going to merge this without asking for additional review only because it is quoting the version number and the CI still passes. I assume that people don't mind me not annoying them here with a trivial :+1:, but let me know if people would prefer that in the future.

Issue #1699, opened 2021-11-15
Option to skip workspace validation


alexander-held:
### Summary

Workspace validation is currently not skippable:

https://github.com/scikit-hep/pyhf/blob/767ed59d06274f810e47944f1fe4c95d94968543/src/pyhf/workspace.py#L289-L298

In some scenarios the validation takes a significant amount of time, and it would be convenient if it was possible to skip it.

### Additional Information

This is using [NormalMeasurement_combined.txt](https://github.com/scikit-hep/pyhf/files/7522251/NormalMeasurement_combined.txt) from https://github.com/scikit-hep/pyhf/discussions/1695. While working with [this gist](https://gist.github.com/alexander-held/2f6727df4cc8f65d75368e293ca83b5a) used in https://github.com/scikit-hep/pyhf/discussions/1695#discussioncomment-1635955, I noticed that workspace validation can take a long time for this example. Below is a minimal reproducer:

```python (paste below)
import pyhf
import json

with open("NormalMeasurement_combined.txt") as f:
    ws = pyhf.Workspace(json.load(f))
```

This runs in around 5.4 seconds for me locally.

When commenting out 

https://github.com/scikit-hep/pyhf/blob/767ed59d06274f810e47944f1fe4c95d94968543/src/pyhf/workspace.py#L298

this time goes down to 1.6 seconds.

The workspace validation takes about as long as the model construction (via `ws.model()`), which for me takes about 4 seconds by itself (again dominated by validation time in this example). It is possible to skip model validation (via `validate=False`), and I believe it would be useful to allow the same for the workspace.

I also noticed that arbitrary kwargs can be passed to workspace construction, e.g. in the above example
```python
ws = pyhf.Workspace(json.load(f), abcdefg=False)
```
also runs fine. It may be convenient to catch these kwargs to help users identify typos.

### Code of Conduct

- [X] I agree to follow the Code of Conduct

matthewfeickert:
This seems reasonable to add with it being turned off by default so that only expert users will try to use it.

PR #1709, opened 2021-11-22
feat: Catch unexpected keyword arguments in workspace construction

alexander-held:
# Description

When building a model, unexpected kwargs are caught in `_ModelConfig` and an error is raised:
https://github.com/scikit-hep/pyhf/blob/34dd45c84426c5e14a280b7da8742d0d4fd7c83d/src/pyhf/pdf.py#L218-L221

This PR ports the same behavior over to `Workspace` construction as well. I believe this is useful to catch typos. Slightly related to #1699, where I came across this originally.

A test has been added for the new behavior. Since the workspace tests used the same `simplemodels` setup a few times for different purposes, I factored that out into a fixture. Happy to implement any other solution you may prefer for this instead.

- [x] Tests are passing
- [X] "WIP" removed from the title of the pull request

# Before Merging

For the PR Assignees:

```
* Raise an Unsupported exception if unexpected kwargs are passed during
pyhf.Workspace creation.
* Add test for this behavior.
* Refactor workspace test model and data into a pytest fixture.
```

### DIFF
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -866,8 +866,7 @@ def test_closure_over_workspace_build(simplemodels_model_data):
 def test_wspace_immutable(simplemodels_model_data):
     model, data = simplemodels_model_data
     workspace = pyhf.Workspace.build(model, data)
-
-    spec = json.loads(json.dumps(workspace))
+    spec = dict(workspace)
 
     ws = pyhf.Workspace(spec)
     model = ws.model()
@@ -895,8 +894,7 @@ def test_workspace_poiless(datadir):
 def test_wspace_unexpected_keyword_argument(simplemodels_model_data):
     model, data = simplemodels_model_data
     workspace = pyhf.Workspace.build(model, data)
-
-    spec = json.loads(json.dumps(workspace))
+    spec = dict(workspace)
 
     with pytest.raises(pyhf.exceptions.Unsupported):
         pyhf.Workspace(spec, abc=True)
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -36,6 +36,15 @@ def workspace_factory(workspace_xml):
     return lambda: pyhf.Workspace(workspace_xml)
 
 
+@pytest.fixture
+def simplemodels_model_data():
+    model = pyhf.simplemodels.uncorrelated_background(
+        signal=[12.0, 11.0], bkg=[50.0, 52.0], bkg_uncertainty=[3.0, 7.0]
+    )
+    data = [51, 48]
+    return model, data
+
+
 def test_build_workspace(workspace_factory):
     w = workspace_factory()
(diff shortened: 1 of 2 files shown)


tests/test_workspace.py
```diff
@@ -887,3 +890,13 @@ def test_workspace_poiless(datadir):
 
     assert model.config.poi_name is None
     assert model.config.poi_index is None
+
+
+def test_wspace_unexpected_keyword_argument(simplemodels_model_data):
+    model, data = simplemodels_model_data
+    workspace = pyhf.Workspace.build(model, data)
+
+    spec = json.loads(json.dumps(workspace))
```
alexander-held, line 899: This approach was taken from `test_wspace_immutable`, but
```python
spec = dict(workspace)
```
works fine as far as I can tell. Is there a reason to not use it? Otherwise I can refactor this.

tests/test_workspace.py
```diff
@@ -36,6 +36,15 @@ def workspace_factory(workspace_xml):
     return lambda: pyhf.Workspace(workspace_xml)
 
 
+@pytest.fixture
```
matthewfeickert, line 39:
```suggestion
@pytest.fixture(scope="module")
```

I _think_ this can safely be scoped to module.

tests/test_workspace.py
matthewfeickert: If 

```python
spec = dict(workspace)
```

works then I'm more than happy to go with that.

matthewfeickert (changes requested):
Thanks for this PR @alexander-held! This all LGTM, and I'm only marking it as "request changes" to suggest that you implement the changes you suggested (and then try the 1 comment I made).

tests/test_workspace.py
alexander-held: Switched to the conversion via `dict`. I tried to look for possible caveats, and the only thing I came across was [in the documentation of `json.dumps`](https://docs.python.org/3/library/json.html#json.dumps): 

> Note Keys in key/value pairs of JSON are always of the type `str`. When a dictionary is converted into JSON, all the keys of the dictionary are coerced to strings. As a result of this, if a dictionary is converted into JSON and then back into a dictionary, the dictionary may not equal the original one. That is, `loads(dumps(x)) != x` if x has non-string keys.

I don't think that this is a problem here, since the spec is defined in JSON to begin with, and all keys are strings.

matthewfeickert (approved):
LGTM. :+1:

As a workspace is a JSON-serializable object you can actually pass the output of `Workspace.build` to `Workspace`

```python
workspace = pyhf.Workspace.build(model, data)
pyhf.Workspace(workspace)
```

but in these tests I wouldn't change it as I think the way that you have it is clearer now. :+1: Thanks!

kratsg (approved):

PR #1710, opened 2021-11-22
feat: Allow skipping validation when constructing workspaces

### LINKED ISSUE #1699: Option to skip workspace validation

alexander-held:
# Description

This adds a `validate` kwarg to the `Workspace` constructor and to `Workspace.build`, replicating the existing functionality in `Model` to skip validation against the JSON schema. For complex workspaces / models, where validation can take an appreciable amount of time (see #1699), this can be a useful option if users are sure that their inputs follow the required format.

There are two things in here that are optional, I tried to split them across commits to easily take out anything you may not want:
- added typing for the `validate` kwargs,
- moved the log info about reference schema to only trigger if validation is actually performed.

There was previously no documentation for `validate` in the `Model` constructor, so this has also been added.

A test for workspace construction without validation is also included, for both construction from a specification and from model + data. This uses the same `simplemodels` setup also used elsewhere in the workspace tests, ~~so this could be refactored if #1709 is merged first (or otherwise could be refactored in #1709). Since #1709 also adds a test at the end, merging one probably results in a conflict with the other PR as well.~~ -> #1709 was merged before, and the implementation here was changed to re-use the fixture added in that PR.

resolves #1699 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add validate kwarg to pyhf.workspace.Workspace to allow skipping validation
if the user desires.
   - The motivation for this comes from some workspace validations taking
     multiple seconds. Power users might want the ability to turn off
     validation in these cases to save time. This behavior is discouraged
     in general, but should be allowed.
* Add docstring for pyhf.workspace.Workspace and add validate kwarg to
pyhf.pdf.Model docstring.
* Add tests for workspace validate kwarg.
```


src/pyhf/workspace.py
```diff
@@ -286,16 +286,17 @@ class Workspace(_ChannelSummaryMixin, dict):
 
     valid_joins = ['none', 'outer', 'left outer', 'right outer']
 
-    def __init__(self, spec, **config_kwargs):
+    def __init__(self, spec, validate: bool = True, **config_kwargs):
         """Workspaces hold the model, data and measurements."""
```
alexander-held, line 290: How does the documentation here work? I am not sure where to document the `validate` kwarg. Should I just expand this existing docstring, and also add `spec`?

tests/test_workspace.py
```diff
@@ -887,3 +887,18 @@ def test_workspace_poiless(datadir):
 
     assert model.config.poi_name is None
     assert model.config.poi_index is None
+
+
+def test_workspace_without_validation(mocker):
+    # TODO: refactor if #1709 is accepted / merged
```
alexander-held, line 893: #1709 refactors the `simplemodels` setup to use a fixture instead, so this could be adopted here too for consistency.

matthewfeickert:
Relevant RTD builds:
* [`pyhf.pdf.Model`](https://pyhf--1710.org.readthedocs.build/en/1710/_generated/pyhf.pdf.Model.html#model)
* [`pyhf.workspace.Workspace`](https://pyhf--1710.org.readthedocs.build/en/1710/_generated/pyhf.workspace.Workspace.html#workspace)

src/pyhf/pdf.py
```diff
@@ -635,14 +635,21 @@ class Model:
...
+        **config_kwargs,
     ):
         """
         Construct a HistFactory Model.
 
         Args:
             spec (:obj:`jsonable`): The HistFactory JSON specification
-            batch_size (:obj:`None` or :obj:`int`): Number of simultaneous (batched) Models to compute.
+            batch_size (:obj:`None` or :obj:`int`): Number of simultaneous (batched)
+            Models to compute.
```
matthewfeickert, line 651:
```suggestion
             Models to compute.
```

space is just for alignment issues in docs

src/pyhf/workspace.py
matthewfeickert: Yeah, the docstrings get a bit fincky for the classes. What you suggest sounds great, and you can take a look at https://scikit-hep.org/pyhf/_generated/pyhf.patchset.PatchSet.html for an example of how we've got things kinda working now.

tests/test_workspace.py
matthewfeickert: Yup, this all sounds good. We can merge PR #1709 first and then rebase this one on top.

matthewfeickert (commented):
Thanks for the PR @alexander-held. This all is looking good to me and your plan for docstrings sounds great. :+1:

src/pyhf/workspace.py
alexander-held: I've updated the docstring, it's also quite similar to the one for the `Model` constructor.

matthewfeickert:
Okay this one can get rebased and refactored now. :+1:

tests/test_workspace.py
alexander-held: Done now in e737e4d.

matthewfeickert (approved):

kratsg (approved):

kratsg:
Congrats on the 1k'th commit @alexander-held :)

PR #1711, opened 2021-11-24
test: Use scikit-hep-testdata to provide probability models for regression tests

matthewfeickert:
# Description

With https://github.com/scikit-hep/scikit-hep-testdata/pull/67, the sbottom probability models that are used for regression tests are added to `scikit-hep-testdata`. This provides stability in testing that is independent of HEPData's availability and doesn't abuse HEPData either as we currently ping them _a lot_ with our test suite.

This also simplifies the tests a bit by removing the need for a fixture to provide the tarfile archive.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add scikit-hep-testdata to the 'test' extra
* Use scikit-hep-testdata to provide the file path to the tar.gz that
contains the sbottom probability models used for regression tests
(pyhf-ins1748602-probability-models.tar.gz)
   - File added in release scikit-hep-testdata v0.4.11
* Refactor get_json_from_tarfile pytest fixture to open the tarfile archive
* Remove sbottom_likelihoods_download pytest fixture as no longer needed
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ extras_require['test'] = sorted(
         + extras_require['contrib']
         + extras_require['shellcomplete']
         + [
+            'scikit-hep-testdata>=0.4.10',
             'pytest>=6.0',
             'pytest-cov>=2.5.1',
             'pytest-mock',
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ extras_require['test'] = sorted(
         + extras_require['contrib']
         + extras_require['shellcomplete']
         + [
-            'scikit-hep-testdata>=0.4.10',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -29,6 +29,7 @@
         + extras_require['contrib']
         + extras_require['shellcomplete']
         + [
+            'scikit-hep-testdata>=0.4.10',
```
matthewfeickert, line 32: I opened this PR before `scikit-hep-testdata` `v0.4.11` was released so everything will fail. Once it is though, everything should pass.

kratsg (approved):

Issue #1685, opened 2021-11-08
chore: some possible cleanup / extra testing


henryiii:
The old black Jupyter code still needs to be removed:
https://github.com/scikit-hep/pyhf/blob/c493c7c5ad3e5aeda7d022cdf5fe77e9fdab43a2/pyproject.toml#L78-L79
(and a line below)

https://github.com/scikit-hep/pyhf/blob/c493c7c5ad3e5aeda7d022cdf5fe77e9fdab43a2/pyproject.toml#L81-L82

The pytest adpdotps probably could do without the ignores? What are they for?

https://github.com/scikit-hep/pyhf/blob/c493c7c5ad3e5aeda7d022cdf5fe77e9fdab43a2/pyproject.toml#L44-L55

Some of the options here for pytest might be useful:
https://scikit-hep.org/developer/pytest#configuring-pytest

I’d probably at least put the lower bound requirements into tests, that’s not a public thing really. Keeping the main directory from getting too big reduces user scrolling.

matthewfeickert:
Thanks for following up here like I requested @henryiii. :) I'll do some cleaning on this soon.

PR #1713, opened 2021-11-29
chore: Use constraints.txt for lower bound testing

matthewfeickert:
# Description

As suggested by @henryiii [on Twitter](https://twitter.com/HenrySchreiner3/status/1464603725562105858?s=20), rename the `lower-bound-requirements.txt` to `constraints.txt` and use it as a `pip` constraints file instead of as a `requirements.txt`.

```console
$ pip install --help | grep constraints
  -c, --constraint <file>     Constrain versions using the given constraints
```

and from the [`pip` docs on constraints files](https://pip.pypa.io/en/stable/user_guide/#constraints-files):

> Constraints files are requirements files that only control which version of a requirement is installed, not whether it is installed or not. Their syntax and contents is a subset of Requirements Files, with several kinds of syntax not allowed: constraints must have a name, they cannot be editable, and they cannot specify extras. In terms of semantics, there is one key difference: Including a package in a constraints file does not trigger installation of the package.

As also suggested by @henryiii in Issue #1685, move the `constraints.txt` under the `tests` directory to reduce clutter in the top level directory.

As constraints files work like requirements files they are still able to follow `--find-links` URLs and so can download old `jaxlib` releases from `https://storage.googleapis.com/jax-releases/jax_releases.html`. :+1: 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use pip's options for constraints files to test lower bounds of install
requirements.
   - Rename 'lower-bound-requirements.txt' to 'constraints.txt'.
   - Use pip's --constraint option while installing the 'test' extra that
     includes all backends and user extras.
   - c.f. https://pip.pypa.io/en/stable/user_guide/#constraints-files
   - Note that transitioning from a requirements file to a constraints file
     works only because the 'test' extra already specifies all of the
     libraries found in the constraints file. Constraints files do NOT
     ensure that a library is installed. They only ensure that if it IS
     installed that it meets the constraints specified in the constraints
     files.
     > Constraints files are requirements files that only control which
     > version of a requirement is installed, not whether it is installed
     > or not.
* Move the constraints.txt under the tests/ directory to reduce clutter
in the top level of the repository.
```


kratsg (approved):

matthewfeickert:
If I diff the output of `pip list` for the `lower-bound-requirements.txt` vs. `constraints.txt` I get

```diff
$ diff --color -u /tmp/lower-bound-requirements-output.txt /tmp/constraints-output.txt 
--- /tmp/lower-bound-requirements-output.txt	2021-11-29 20:44:15.692252460 -0600
+++ /tmp/constraints-output.txt	2021-11-29 20:44:41.536190380 -0600
@@ -1,11 +1,12 @@
 absl-py                 1.0.0
 ansiwrap                0.8.4
+appdirs                 1.4.4
 argcomplete             1.12.3
 argon2-cffi             21.1.0
 astunparse              1.6.3
 attrs                   21.2.0
 backcall                0.2.0
-black                   21.11b1
+black                   20.8b0
 bleach                  4.1.0
 cachetools              4.2.4
 certifi                 2021.10.8
@@ -80,7 +81,6 @@
 pickleshare             0.7.5
 Pillow                  8.4.0
 pip                     21.3.1
-platformdirs            2.4.0
 pluggy                  1.0.0
 prometheus-client       0.12.0
 prompt-toolkit          3.0.23
```

which seems acceptable as none of those are specified in either files. So I'll merge this now that there is 1 approval.

PR #1727, opened 2021-12-08
ci: Publish to TestPyPI on tag or by workflow dispatch trigger

matthewfeickert:
# Description

There is a build up of releases on TestPyPI that will eventually need to get deleted into the future. This is a manual process (which will be done by me) and is very slow. To avoid this, only publish to TestPyPI on tags (that start with `v`) or if a workflow dispatch event is run with the user selecting the `true` option from the drop down menu in response to the question 'Publish to TestPyPI?'

dispatch_option

Using workflow dispatch will still allow for using TestPyPI to check images before release.

https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/

I've tested this setup of the CI config on my fork.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Publish to TestPyPI on push events where a tag is pushed that starts with
a 'v' (release tags of form vX.Y.Z) or on workflow dispatch events where the
maintainer running the workflow has selected the 'true' option from the
drop-down menu for whether to publish.
   - The workflow dispatch option defaults to 'false' to avoid accidental
     publishing.
* Update development docs to reflect that not every commit is published to
TestPyPI
```

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -83,8 +83,9 @@ jobs:
       run: python -m zipfile --list dist/pyhf-*.whl
 
     - name: Publish distribution 📦 to Test PyPI
-      # every PR will trigger a push event on master, so check the push event is actually coming from master
-      if: github.event_name == 'push' && github.ref == 'refs/heads/master' && github.repository == 'scikit-hep/pyhf'
+      if: >-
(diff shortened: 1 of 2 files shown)


kratsg (approved):

matthewfeickert:
And the publishing workflow to TestPyPI via workflow dispatch works too:

all_good
[works](https://test.pypi.org/project/pyhf/0.6.4.dev68/)
:+1:

PR #1753, opened 2022-01-24
feat: Alternative Schema Locations

### LINKED ISSUE #1609: Expose `base_uri` in `utils.validate`
### Summary

Would it be possible to expose [the `base_uri` passed to the schema reference resolver](https://github.com/scikit-hep/pyhf/blob/4b66cd2cfa0bd4bb53b14c525118d86567be8257/src/pyhf/utils.py#L55) as a parameter of `pyhf.utils.validate`? 
(and similarly [the `path` used in `pyhf.utils.load_schema`](https://github.com/scikit-hep/pyhf/blob/4b66cd2cfa0bd4bb53b14c525118d86567be8257/src/pyhf/utils.py#L38), too?)
 
After the proposed change,  `validate` would take an optional argument, defaulting to the currently used path to the package install location, and also forward that path to `load_schema`, which similarly would take it as an argument (also with a default).
This would make it much easier to validate specs against a custom schema (e.g. one extracted from a reference to the `defs.json`).

#### Related problem
When trying to validate components of a to-be-assembled `Workspace` or `Model` spec, the `pyhf.utils` helper functions (`validate` and `load_schema`) search for schemas only in the package installation path. 
I do not really want to add files to the package install directories manually, 
so I cannot use the `pyhf.utils` helpers to validate partial specs against component schemas (e.g. `lumi`, `sample`, or `observation` specs).

#### My current work-around
I am currently working around the issue by having copy-pasted implementations of [`pyhf.utils.validate`](https://gitlab.cern.ch/lhenkelm/dynamicpyhf/-/blob/master/python/dynpyhf/core/schematic.py#L448) and [`pyhf.utils.load_schema`](https://gitlab.cern.ch/lhenkelm/dynamicpyhf/-/blob/master/python/dynpyhf/core/schematic.py#L470) in my code, 
which I fall back on if the schema is not found by `pyhf.utils.load_schema` (meaning for everything except `Workspace` or `Model`):
```python
try:
  pyhf.utils.validate(...)
except FileNotFoundError:
  homebrew_validate(...) # has a different base path hardcoded in ...
```
But I feel this is a bit silly (and risks diverging from the pyhf implementation over time)
since the base path to look up schemas in is the only real difference of my copies to the originals. 

#### Example usage after the suggestion is implemented

Validate a custom schema (that is extracted from `defs.json` or custom-defined) using the pyhf validation helper function:

```python
import pyhf
my_lumi = dict(type='lumi', name='lumi', data=None) # for example 
pyhf.utils.validate(my_lumi, 'lumi.json', '1.0.0', schema_base = 'path/to/local/user/defined/schemas/')
```
#### Relevant Issues and Pull Requests
Both of these are more tangential, but maybe there are examples of where this could be a useful feature:
- I guess this change would make the answer to a previous question I confused myself with (issue #1360) easier to implement for newcomers like me.
- It might also make it easier to keep schema-validation a requirement when using custom modifiers a la #850, #1188, if the solution there uses user-defined schema extensions (somehow).

### Code of Conduct

kratsg:
# Description

Resolves #1609.

Refactor:
- migrate from `pkg_resources` to `importlib`. See [migration guide](https://importlib-resources.readthedocs.io/en/latest/migration.html) and details about [importlib](https://docs.python.org/3/library/importlib.html).
- `load_schema` will handle functionality for finding the schema locations locally in multiple places
- `load_schema` will not handle `version` identification (that is done through `validate` only)

Feature:
- support overriding the paths for finding schemas, using the `pyhf` installed location as a base via `pyhf.utils.schemas`
- support for offline access to currently supported version of schemas (via extra `load_schema` commands)
- `SchemaNotFound` exception for when a schema cannot be found locally (outside of `jsonschema.RefResolver` calls)

Example Usage:

```python
import pyhf, pathlib, json
pyhf.utils.schemas = pathlib.Path('/Users/kratsg/pyhf/myschemas')
ws = pyhf.Workspace(json.load(open('myschemas/custom.json')))
```

and then additionally, a nice benefit (maybe?) we get for free is that code which generates against a different schema won't work and fails more loudly (where it might have been more silent in the past).

```python 
>>> pyhf.simplemodels.uncorrelated_background([10], [5], [1]).spec
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/pyhf/simplemodels.py", line 141, in uncorrelated_background
    return Model(spec, batch_size=batch_size)
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/pyhf/pdf.py", line 669, in __init__
    utils.validate(self.spec, self.schema, version=self.version)
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/pyhf/utils.py", line 74, in validate
    schema = load_schema(f'{version}/{schema_name}')
  File "/Users/kratsg/.pyenv/versions/pyhf-dev/lib/python3.8/site-packages/pyhf/utils.py", line 53, in load_schema
    raise pyhf.exceptions.SchemaNotFound(
pyhf.exceptions.SchemaNotFound: The schema 1.0.0/model.json was not found. Do you have the right version or the right path? /Users/kratsg/pyhf/myschemas/1.0.0/model.json
```

Documentation: https://pyhf--1753.org.readthedocs.build/en/1753/api.html#schema

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
- Refactors
  * Migrate from `pkg_resources` to `importlib`. See importlib's migration guide.
    https://importlib-resources.readthedocs.io/en/latest/migration.html
  * Add importlib_resources as core dependency for Pythons older than 3.9.
  * `load_schema` will handle functionality for finding the schema locations
    locally in multiple places.
  * `load_schema` will not handle `version` identification (that is done through
    `validate` only).

- Features
  * Support overriding the paths for finding schemas, using the `pyhf` installed
    location as a base via `pyhf.schema.variables.path`.
  * Add support for offline access to currently supported version of schemas
    (via extra `load_schema` commands).
  * Add `SchemaNotFound` exception for when a schema cannot be found locally
    (outside of `jsonschema.RefResolver` calls)

- Python API
  * `pyhf.schema` introduced as a new API for all things schema-related.
  * `pyhf.schema.version`, `pyhf.schema.path`, `pyhf.schema.load_schema`, and
    `pyhf.schema.validate` are migrated over from `pyhf.utils`.
```


src/pyhf/utils.py
```diff
@@ -1,11 +1,25 @@
...
+else:
+    import importlib_resources as resources
+schemas = resources.files('pyhf') / "schemas"
+
+import pyhf.exceptions
+
+if sys.version_info >= (3, 9):
+    from importlib import resources
+else:
+    import importlib_resources as resources
```
matthewfeickert, line 21: Is this double block of 

```python
if sys.version_info >= (3, 9):
    from importlib import resources
else:
    import importlib_resources as resources
```

intentional? If so, why do we need it?

Can we also make it clear that the reason the check for Python 3.9 is needed is because [`importlib.resources.as_file`](https://docs.python.org/3.10/library/importlib.html#importlib.resources.as_file) wasn't added until Python 3.9?

matthewfeickert (commented):
@kratsg Just a question, the general idea of

```python
import pyhf, pathlib, json
pyhf.utils.schemas = pathlib.Path('/Users/kratsg/pyhf/myschemas')
ws = pyhf.Workspace(json.load(open('myschemas/custom.json')))
```

of being able to update the schema path seems fine, but instead of having users overwriting the value by assignment, do you think there is value in offering a `pyhf.utils.update_schema_path` or something that performs this operation through the API?

Other than that things seem to be looking good here I think. Though maybe worth also asking @alexander-held and maybe @lhenkelm as they asked for this in Issue #1609, and so might have some thoughts.

src/pyhf/utils.py
lhenkelm: Thanks @kratsg for taking up my suggestion!

I second @matthewfeickert's suggestion to change schemas via function call rather than assignment. It is much simpler to add functionality to a function if it ever becomes necessary, than to replace an entire existing module with a class instance with a property.

I'd even go so far to say that a context manager doubling as an update function would be ideal IMO:
```python
# in pyhf.utils
_SCHEMAS = Path(...)
class use_schema_path: # snake_case to remind of function-like usage
    def  __init__(self, path):
        global _SCHEMAS
        self._old_schemas = _SCHEMAS
        _SCHEMAS = pathlib.Path(path)
    def __enter__(self):
        pass
    def __exit__(self, *args, **kwargs):
        global _SCHEMAS
        _SCHEMAS = self._old_schemas
```
which can still be called as a function (only executing `__init__`), so short scripts etc. are not forced to use `with` blocks.
But it can also be used like so:
```python
def make_my_workspace(spec):
    with pyhf.utils.use_schema_path('/my/very/special/schemas'):
        return pyhf.Workspace(spec)
```
So as a user writing code on top of pyhf, I don't have to worry about resesetting the global variable, the CM does it for me, and there are fewer mistakes to make.

lhenkelm:
> instead of having users overwriting the value by assignment, do you think there is value in offering a `pyhf.utils.update_schema_path` or something that performs this operation through the API?

Thanks @kratsg for taking up my suggestion!

I second @matthewfeickert's suggestion to change schemas via function call rather than assignment. It is much simpler to add functionality to a function if it ever becomes necessary, than to replace an entire existing module with a class instance with a property.

I'd even go so far to say that a context manager doubling as an update function would be ideal IMO:
```python
# in pyhf.utils
_SCHEMAS = Path(...)
class use_schema_path: # snake_case to remind of function-like usage
    def  __init__(self, path):
        global _SCHEMAS
        self._old_schemas = _SCHEMAS
        _SCHEMAS = pathlib.Path(path)
    def __enter__(self):
        pass
    def __exit__(self, *args, **kwargs):
        global _SCHEMAS
        _SCHEMAS = self._old_schemas
```
which can still be called as a function (only executing `__init__`), so short scripts etc. are not forced to use `with` blocks.
But it can also be used like so:
```python
def make_my_workspace(spec):
    with pyhf.utils.use_schema_path('/my/very/special/schemas'):
        return pyhf.Workspace(spec)
```
So as a user writing code on top of pyhf, I don't have to worry about resesetting the global variable, the CM does it for me, and there are fewer mistakes to make.

src/pyhf/utils.py
kratsg: resolving this, made the changes.

kratsg:
> So as a user writing code on top of pyhf, I don't have to worry about resesetting the global variable, the CM does it for me, and there are fewer mistakes to make.

This is filed as #1815 so this PR isn't overtly complicated. The API for this PR will be `pyhf.schema(new_path: pathlib.Path)`

src/pyhf/exceptions/__init__.py
```diff
@@ -44,6 +44,12 @@ class InvalidNameReuse(Exception):
     pass
 
 
+class SchemaNotFound(FileNotFoundError):
```
kratsg, line 47: new exception

src/pyhf/schema/__init__.py
```diff
@@ -0,0 +1,32 @@
+import pathlib
```
kratsg, line 1: all brand new, to support the new api for `pyhf.schema(...)` and will support contextlib wrapping :)

src/pyhf/schema/loader.py
```diff
@@ -0,0 +1,31 @@
+from pathlib import Path
```
kratsg, line 1: copied over from `pyhf.utils`

src/pyhf/schema/validator.py
```diff
@@ -0,0 +1,25 @@
+import jsonschema
```
kratsg, line 1: copied over from `pyhf.utils`

src/pyhf/schema/variables.py
```diff
@@ -0,0 +1,12 @@
+import sys
```
kratsg, line 1: copied over from `pyhf.utils`

src/pyhf/utils.py
```diff
@@ -140,14 +99,9 @@ def citation(oneline=False):
...
     """
-    path = Path(
-        pkg_resources.resource_filename(
-            __name__, str(Path('data').joinpath('citation.bib'))
-        )
-    )
-    with path.open() as fp:
-        # remove end-of-file newline if there is one
-        data = fp.read().strip()
+    ref = resources.files('pyhf') / 'data' / 'citation.bib'
```
kratsg, line 102: use `importlib`

setup.cfg
```diff
@@ -41,6 +41,7 @@ install_requires =
     jsonschema>=3.0.0  # for utils
     jsonpatch>=1.15
     pyyaml>=5.1  # for parsing CLI equal-delimited options
+    importlib_resources; python_version < "3.9"
```
matthewfeickert, line 44: It is always good to add a lower bound for dependencies to make it explicit and easy for `pip` to do a correct solve. As we need `importlib_resources.as_file` and `importlib_resources.files` (and I think that's it) here then we should add the lowest bound for that functioning API. From looking at the [`importlib_resources` docs](https://importlib-resources.readthedocs.io/en/latest/history.html#v1-3-0) and testing this out myself the first time the `importlib_resources.as_file` exists is `v1.3.0`.

```suggestion
    importlib_resources>=1.3.0; python_version < "3.9"  # for resources in schema
```

We should then additionally add

```
importlib_resources==1.3.0
```

to `tests/constraints.txt` just after `pyyaml`

https://github.com/scikit-hep/pyhf/blob/1884c6c08bc5b629724dd9ef3c60ec4386353b94/tests/constraints.txt#L1-L7

src/pyhf/schema/loader.py
```diff
@@ -0,0 +1,40 @@
...
+import sys
+import json
+import pyhf.exceptions
+from pyhf.schema import variables
+
+# importlib.resources.as_file wasn't added until Python 3.9?
```
matthewfeickert, line 7: Correct. So let's add a link to remind us as well?

```suggestion
# importlib.resources.as_file wasn't added until Python 3.9
# c.f. https://docs.python.org/3.9/library/importlib.html#importlib.resources.as_file
```

src/pyhf/schema/variables.py
```diff
@@ -0,0 +1,12 @@
...
+
+# importlib.resources.as_file wasn't added until Python 3.9?
```
matthewfeickert, line 3:
```suggestion
# importlib.resources.as_file wasn't added until Python 3.9
# c.f. https://docs.python.org/3.9/library/importlib.html#importlib.resources.as_file
```

src/pyhf/schema/loader.py
```diff
@@ -0,0 +1,40 @@
...
+if sys.version_info >= (3, 9):
+    from importlib import resources
+else:
+    import importlib_resources as resources
```
matthewfeickert, line 11: In a follow up PR we should see if it would be worth making this

https://github.com/scikit-hep/pyhf/blob/1884c6c08bc5b629724dd9ef3c60ec4386353b94/.github/workflows/ci.yml#L70

be for both the oldest and newest Python supported (so Python 3.7 and Python 3.10) so that we cover these sorts of things in our reports.

src/pyhf/schema/validator.py
```diff
@@ -0,0 +1,41 @@
...
+import pyhf.exceptions
+from pyhf.schema.loader import load_schema
+from pyhf.schema import variables
+from typing import Union
+
+
+def validate(spec: dict, schema_name: str, version: Union[str, None] = None):
```
matthewfeickert, line 8: As we don't have `mypy` in the tests yet have these been verified yet?

src/pyhf/utils.py
```diff
@@ -1,69 +1,28 @@
...
 import click
 import hashlib
 
-from pyhf.exceptions import InvalidSpecification
+import sys
 
-SCHEMA_CACHE = {}
-SCHEMA_BASE = "https://scikit-hep.org/pyhf/schemas/"
-SCHEMA_VERSION = '1.0.0'
+# importlib.resources.as_file wasn't added until Python 3.9?
```
matthewfeickert, line 8:
```suggestion
# importlib.resources.as_file wasn't added until Python 3.9
# c.f. https://docs.python.org/3.9/library/importlib.html#importlib.resources.as_file
```

tests/test_schema/customschema/1.1.0/defs.json
```diff
@@ -0,0 +1,315 @@
+{
+    "$schema": "http://json-schema.org/draft-06/schema#",
+    "$id": "1.1.0/defs.json",
```
matthewfeickert, line 3: I haven't looked at this too deeply, but is this what the `v1.1.0` schema def will actually be? Or are these just examples for the sake of demoing an alternative schema?

If these are actually a new schema can we add these in a separate PR that would go in first? Otherwise it seems like we have two important changes going into 1 squashed PR.

matthewfeickert (changes requested):
Thanks for this nice work @kratsg. Overall it looks really solid :+1: and stepping through it with you annotations (thanks! :pray:) I see that it isn't really that bad at all (just looks like a lot).

I have a few super minor suggestions that are more just around testing and adding notes for ourselves in the future and then just one question (though it might not even be necessary if my question comes from me being confused).

src/pyhf/schema/validator.py
kratsg: yes. i ran locally. `Union[str, None]` can become `Optional[str]` but I was going to pick that up in a different set of sweeping changes.

tests/test_schema/customschema/1.1.0/defs.json
kratsg: this is just for testing purposes. In this case, the easiest change to the schema is to just change the version and watch it break.

kratsg: e.g. the point is to check the behavior of `pyhf.schema()` is working as expected, by the fact that we're able to get a previously working spec to not validate when we change the schema.

src/pyhf/schema/loader.py
matthewfeickert: This is now Issue #1816.

tests/constraints.txt
```diff
@@ -5,6 +5,7 @@ tqdm==4.56.0
 jsonschema==3.0.0
 jsonpatch==1.15
 pyyaml==5.1
+importlib_resources==1.3.0
```
matthewfeickert, line 8: I've verified that this passes the Minimum supported dependencies workflow, so LGTM. :+1:

matthewfeickert (approved):
Thanks @kratsg. LGTM now!

lukasheinrich:
LGTM!

lukasheinrich (approved):

PR #1806, opened 2022-03-07
refactor: Clarify exception message applies only to profile likelihood ratio

matthewfeickert:
# Description

Clarify that the warning RE: 'inference [being] impossible' the user gets if `pyhf.infer.hypotest` is called with fixed POIs is limited to profile likelihood ratio based methods, and not to methods like MLE fits. Thanks to a discussion with @alexander-held for helping me realize this warning wasn't very clear in general.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Clarify that the message RE: 'inference [being] impossible' the user
gets if pyhf.infer.hypotest is called with fixed POIs is limited to
profile likelihood ratio based methods, and not to methods like MLE fits.
```

### DIFF
--- a/src/pyhf/infer/__init__.py
+++ b/src/pyhf/infer/__init__.py
@@ -13,7 +13,7 @@ def _check_hypotest_prerequisites(pdf, data, init_pars, par_bounds, fixed_params
 
     if not utils.all_pois_floating(pdf, fixed_params):
         raise exceptions.InvalidModel(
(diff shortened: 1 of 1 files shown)


alexander-held (approved):
Thanks for this addition, I think this is quite useful to avoid potential confusion. Looks good to me!

kratsg (approved):

PR #1818, opened 2022-03-23
feat: Add contextlib support to pyhf.schema API

### LINKED ISSUE #1815: Add contextlib support to the pyhf.schema API
> instead of having users overwriting the value by assignment, do you think there is value in offering a `pyhf.utils.update_schema_path` or something that performs this operation through the API?
>
> Thanks @kratsg for taking up my suggestion!
> 
> I second @matthewfeickert's suggestion to change schemas via function call rather than assignment. It is much simpler to add functionality to a function if it ever becomes necessary, than to replace an entire existing module with a class instance with a property.
> 
> I'd even go so far to say that a context manager doubling as an update function would be ideal IMO:
> ```python
> # in pyhf.utils
> _SCHEMAS = Path(...)
> class use_schema_path: # snake_case to remind of function-like usage
>     def  __init__(self, path):
>         global _SCHEMAS
>         self._old_schemas = _SCHEMAS
>         _SCHEMAS = pathlib.Path(path)
>     def __enter__(self):
>         pass
>     def __exit__(self, *args, **kwargs):
>         global _SCHEMAS
>         _SCHEMAS = self._old_schemas
> ```
> which can still be called as a function (only executing `__init__`), so short scripts etc. are not forced to use `with` blocks.
> But it can also be used like so:
> ```python
> def make_my_workspace(spec):
>     with pyhf.utils.use_schema_path('/my/very/special/schemas'):
>         return pyhf.Workspace(spec)
> ```
> So as a user writing code on top of pyhf, I don't have to worry about resesetting the global variable, the CM does it for me, and there are fewer mistakes to make.
> 
_Originally posted by @lhenkelm in https://github.com/scikit-hep/pyhf/issues/1753#issuecomment-1026678066_

kratsg:
# Pull Request Description

Resolves #1815, originally asked on #1753.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add contextlib support to the pyhf.schema API.
   - Follow up to PR #1753.
* Add tests for use of the schema contextlib to tests/test_schema.py.
* Add tests for use of the schema contextlib to the public API tests.
```

### DIFF
--- a/src/pyhf/schema/__init__.py
+++ b/src/pyhf/schema/__init__.py
@@ -23,17 +23,40 @@ class Schema(sys.modules[__name__].__class__):
     """
     A module-level wrapper around :mod:`pyhf.schema` which will provide additional functionality for interacting with schemas.
 
-    Example:
+    .. rubric:: Example (callable)
+
+    .. doctest::
+
         >>> import pyhf.schema
         >>> import pathlib
+        >>> new_path = pathlib.Path("/home/root/my/new/path")
         >>> curr_path = pyhf.schema.path
-        >>> curr_path # doctest: +ELLIPSIS
+        >>> curr_path  # doctest: +ELLIPSIS
         PosixPath('.../pyhf/schemas')
-        >>> pyhf.schema(pathlib.Path('/home/root/my/new/path'))
+        >>> pyhf.schema(new_path)  # doctest: +ELLIPSIS
+        <module 'pyhf.schema' from ...>
         >>> pyhf.schema.path
         PosixPath('/home/root/my/new/path')
-        >>> pyhf.schema(curr_path)
-        >>> pyhf.schema.path # doctest: +ELLIPSIS
+        >>> pyhf.schema(curr_path)  # doctest: +ELLIPSIS
+        <module 'pyhf.schema' from ...>
+        >>> pyhf.schema.path  # doctest: +ELLIPSIS
+        PosixPath('.../pyhf/schemas')
+
+    .. rubric:: Example (context-manager)
+
+    .. doctest::
+
+        >>> import pyhf.schema
+        >>> import pathlib
+        >>> new_path = pathlib.Path("/home/root/my/new/path")
+        >>> curr_path = pyhf.schema.path
+        >>> curr_path  # doctest: +ELLIPSIS
+        PosixPath('.../pyhf/schemas')
+        >>> with pyhf.schema(new_path):
+        ...     print(repr(pyhf.schema.path))
+        ...
+        PosixPath('/home/root/my/new/path')
+        >>> pyhf.schema.path  # doctest: +ELLIPSIS
         PosixPath('.../pyhf/schemas')
 
     """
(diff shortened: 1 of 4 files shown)


matthewfeickert:
Relevant RTD build: https://pyhf--1818.org.readthedocs.build/en/1818/_generated/pyhf.schema.Schema.html

src/pyhf/schema/__init__.py
```diff
@@ -23,17 +23,40 @@ class Schema(sys.modules[__name__].__class__):
...
         >>> import pyhf.schema
         >>> import pathlib
+        >>> new_path = pathlib.Path("/home/root/my/new/path")
         >>> curr_path = pyhf.schema.path
-        >>> curr_path # doctest: +ELLIPSIS
+        >>> curr_path  # doctest: +ELLIPSIS
         PosixPath('.../pyhf/schemas')
-        >>> pyhf.schema(pathlib.Path('/home/root/my/new/path'))
+        >>> pyhf.schema(new_path)  # doctest: +ELLIPSIS
+        <module 'pyhf.schema' from ...>
```
matthewfeickert, line 37: This is perhaps personal preference, but when reading the example I would assume the reader would want to see all the information as:
* Stuff with current path
* Introduction of a new path
* Change to the new path and do stuff with new path
* Restore to the 'current' path to show that we're back where we started

This is basically done, but with the nitpick exception of moving `new_path` to just before use, like the following (I can't make a recommendation in the PR across all these lines else I would have just done that sorry):

```python
        >>> curr_path = pyhf.schema.path
        >>> curr_path  # doctest: +ELLIPSIS
        PosixPath('.../pyhf/schemas')
        >>> new_path = pathlib.Path("/home/root/my/new/path")
        >>> pyhf.schema(new_path)  # doctest: +ELLIPSIS
        <module 'pyhf.schema' from ...>
```

src/pyhf/schema/__init__.py
```diff
@@ -23,17 +23,40 @@ class Schema(sys.modules[__name__].__class__):
...
+    .. rubric:: Example (context-manager)
+
+    .. code-block:: pycon
+
+        >>> import pyhf.schema
+        >>> import pathlib
+        >>> new_path = pathlib.Path("/home/root/my/new/path")
+        >>> curr_path = pyhf.schema.path
+        >>> curr_path  # doctest: +ELLIPSIS
+        PosixPath('.../pyhf/schemas')
```
matthewfeickert, line 54: Same comment as above, except now I can actually make a code suggestion in the PR.

```suggestion
        >>> curr_path = pyhf.schema.path
        >>> curr_path  # doctest: +ELLIPSIS
        PosixPath('.../pyhf/schemas')
        >>> new_path = pathlib.Path("/home/root/my/new/path")
```

matthewfeickert (approved):
Thanks for this PR following up on your nice PR #1753 @kratsg. :slightly_smiling_face: You made this easy to review, so thanks for splitting it out. :+1: 

I have one super tiny suggestion for the examples, but this might be a nit, and so feel free to ignore. Other than that this LGTM, so I'll approve.

kratsg:
All resolved.

matthewfeickert (approved):

PR #1820, opened 2022-03-23
docs: Add JupyterLite REPL for interactive pyhf in docs

### LINKED ISSUE #1736: Explore pyodide for pyhf
### Summary

Pyodide allows pure-python wheels to be "imported" into the browser and run as-is.. this means we could ship an interactive console as part of our docs that just.. works.. 

https://pyodide.org/en/stable/index.html

However, one of our main dependencies is on scipy which is currently being worked on (pyodide/pyodide#1293) but it seems like all of our other deps should be ok.

### Additional Information

_No response_

### Code of Conduct

### LINKED ISSUE #1775: Try to get pyhf to run in Pyolite / JupyterLite
### Summary

As [`pyodide` is now running with `scipy` `v1.7.3`](https://github.com/pyodide/pyodide/blob/950c1dcbd271f90faea0d205a7c95fcd40710a69/packages/scipy/meta.yaml#L1-L3) ([demoed on the NumPy docs](https://twitter.com/jtpio/status/1491698955859767298)) all the components to be able to use `pyhf` in WebAssembly should be there. It would be interesting to see how much work would be needed to get a minimal working example going and then to run some examples in the docs with [JupyterLite](https://github.com/jupyterlite/jupyterlite).

cc @kratsg as he was thinking the same thing as me when he saw the NumPy tweet.

### Additional Information

_No response_

### Code of Conduct

matthewfeickert:
# Description

* Resolves #1775
* Resolves #1736

PR summary:

* Add jupyterlite file to docs that is spliced into the README content to insert the Pyolite kernel example as the first element after the introduction on the docs landing page.
   - Add `generate_jupyterlite_iframe.py` script to build the URL and reStructuredText for the iframe in jupyterlite.rst.
   - Use an iframe to insert a Pyolite JupyterLite shell that uses [piplite](https://jupyterlite.readthedocs.io/en/latest/configuring.html#adding-pyolite-wheels) to install the latest stable release of pyhf and requests from PyPI and imports pyhf.
* Add [`sphinx-togglebutton`](https://github.com/executablebooks/sphinx-togglebutton) to the 'docs' extra and Sphinx conf.py so that it can be used for the instructional admonition to [make it a toggleable dropdown](https://sphinx-togglebutton.readthedocs.io/en/latest/reference/index.html#admonition-toggles).
* Add docs/generate_jupyterlite_iframe.py and docs/jupyterlite.rst to tbump config to ensure version of pyhf installed by piplite is locked to the latest PyPI release.

Relevant RTD build: https://pyhf--1820.org.readthedocs.build/en/1820/

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add jupyterlite file to docs that is spliced into the README content to insert
the Pyolite kernel example as the first element after the introduction on the
docs landing page.
   - Add generate_jupyterlite_iframe.py script to build the URL and
     reStructuredText for the iframe in jupyterlite.rst.
   - Use an iframe to insert a Pyolite JupyterLite shell that uses piplite
     to install the latest stable release of pyhf and requests from PyPI and
     imports pyhf.
* Add sphinx-togglebutton to the 'docs' extra and Sphinx conf.py so that it
can be used for the instructional admonition to make it a toggleable dropdown.
* Add docs/generate_jupyterlite_iframe.py and docs/jupyterlite.rst to tbump config
to ensure version of pyhf installed by piplite is locked to the latest PyPI release.
```


README.rst
```diff
@@ -32,6 +32,17 @@ to support modern computational graph libraries such as PyTorch and
 TensorFlow in order to make use of features such as autodifferentiation
 and GPU acceleration.
 
+Try out now with Pyolite
+------------------------
+
```
matthewfeickert, line 37: This will currently render with nothing below it on GitHub's preview and on PyPI as mentioned in the first concern in https://github.com/scikit-hep/pyhf/issues/1775#issuecomment-1076791563.

README.rst
```diff
@@ -32,6 +32,17 @@ to support modern computational graph libraries such as PyTorch and
...
+.. raw:: html
```
matthewfeickert, line 38: Ah the use of a `raw` directive is a no go in the `README` as `twine` forbids it

```
Checking dist/pyhf-0.6.4.dev124-py3-none-any.whl: FAILED
  `long_description` has syntax errors in markup and would not be rendered on PyPI.
    line 3[8](https://github.com/scikit-hep/pyhf/runs/5666625641?check_suite_focus=true#step:10:8): Warning: "raw" directive disabled.
Checking dist/pyhf-0.6.4.dev[12](https://github.com/scikit-hep/pyhf/runs/5666625641?check_suite_focus=true#step:10:12)4.tar.gz: FAILED
  `long_description` has syntax errors in markup and would not be rendered on PyPI.
    line 38: Warning: "raw" directive disabled.
```

matthewfeickert:
Relevant RTD build: https://pyhf--1820.org.readthedocs.build/en/1820/

README.rst
```diff
@@ -32,6 +32,9 @@ to support modern computational graph libraries such as PyTorch and
 TensorFlow in order to make use of features such as autodifferentiation
 and GPU acceleration.
 
+..
+  Comment: JupyterLite segment goes here in docs
```
matthewfeickert, line 36: This comment is used by Sphinx to find where it should insert the contents of docs/jupyterlite.rst.

docs/index.rst
```diff
@@ -31,7 +31,16 @@
 
    <p id="dev-version"><strong>Warning:</strong> This is a development version. The latest stable version is at <a href="https://pyhf.readthedocs.io/">ReadTheDocs</a>.</p>
 
+..
+  Comment: Splice the JupyterLite example into the README by looking for a particular comment
+
+.. include:: ../README.rst
+    :end-before: Comment: JupyterLite segment goes here in docs
```
matthewfeickert, line 38: Searching for the comment above to do injection.

docs/jupyterlite.rst
```diff
@@ -0,0 +1,25 @@
+Try out now with JupyterLite_
+-----------------------------
+
+.. admonition:: To try pyhf code in the browser with the Pyolite JupyterLite kernel:
```
matthewfeickert, line 4: What do people think? If what is currently there is too long and technical we can shorten it to something like

```suggestion
.. admonition:: To try pyhf code in the browser:
```

setup.py
```diff
@@ -58,6 +58,7 @@
             'ipywidgets',
             'sphinx-issues',
             'sphinx-copybutton>=0.3.2',
+            'sphinx-togglebutton>=0.3.0',
```
matthewfeickert, line 61: Setting to v0.3.0 as lower bound as that's what is available on PyPI today: https://pypi.org/project/sphinx-togglebutton/0.3.0/

tbump.toml
```diff
@@ -54,6 +54,9 @@ src = "codemeta.json"
 [[file]]
 src = "CITATION.cff"
 
+[[file]]
+src = "docs/jupyterlite.rst"
```
matthewfeickert, line 58: Bump the `pyhf%3D%3D0.6.3` to the next version.

docs/jupyterlite.rst
```diff
@@ -0,0 +1,25 @@
...
+   <iframe
+      src="https://jupyterlite.github.io/demo/repl/index.html?kernel=python&toolbar=1&code=import%20piplite%0Aawait%20piplite.install%28%5B%22pyhf%3D%3D0.6.3%22%2C%20%22requests%22%5D%29%0Aimport%20pyhf"
+      width="100%"
+      height="500px"
+   ></iframe>
+
+..
+  Comment: Add an extra blank line as a spacer
+
+|
```
matthewfeickert, line 23: Spacing to "User Guide" section looks strange without it.

matthewfeickert:
@lukasheinrich @kratsg I think I have it now. :) Let me know what you think.

@henryiii @jpivarski @alexander-held I would welcome style feedback and any and all comments to have this looking good!

matthewfeickert:
@kratsg is reporting that for some reason the RTD build won't display in Chrome on his Mac, but it works fine for me using Chrome on Ubuntu. :? 

For the record, this is what I see:

example

@kratsg if you can check out this PR and build the docs locally it would be interesting to see if you get the same result or not.

kratsg:
https://user-images.githubusercontent.com/761483/159817887-45fc2118-1ce6-4251-8109-414366b5ff15.mp4

alexander-held:
This is a very cool feature! Two comments from a brief look:
- The iframe does not load for me in an incognito Chrome tab on macOS.
- This automatically downloads the relevant packages upon loading the docs page, right? How large are they, as this is perhaps a concern for people with limited data who would not expect the docs to consume much bandwith.

matthewfeickert:
> * The iframe does not load for me in an incognito Chrome tab on macOS. edit: turns out to be the same cookie issue as shown in @kratsg's video

Do you think this is a big enough problem that we should make a note in the admonition about it?

> * This automatically downloads the relevant packages upon loading the docs page, right? How large are they, as this is perhaps a concern for people with limited data who would not expect the docs to consume much bandwith.

Henry might know more about the specifics, but I think the largest part of all of this is the download of Pyodide, as `pyhf` and `requests` are a both about `100 kB` in size. The rest of the page can load and be used before any Pyodide code is run, so I don't think that this will slow down anyone trying to view the docs. While it might be true that someone visiting the docs on mobile with limited data could be surprised that they just got a bigger download than they were expecting, my guess is that the amount of data the docs would consume compared to their normal browsing is very small. I'm not an expert on web stuff in general though, so my naive assumptions about how caching and things work are probably wrong so maybe someone with much more experience can correct me here.

alexander-held:
> Do you think this is a big enough problem that we should make a note in the admonition about it?

I'd say yes this is worth a note, yes.

>  I think the largest part of all of this is the download of Pyodide, as pyhf and requests are a both about 100 kB in size

I was wondering about the dependencies of `pyhf` as well, or are they already included in this? I don't know anything about how that setup works, but expected that `numpy` and other dependencies may add a bit more to the size.

I don't have any strong opinion here, but wanted to bring up the idea of delaying the loading until the user takes some action (clicks a button or so), ensuring they would want things to be loaded. Not sure whether this is even feasible and how much it actually matters though.

docs/jupyterlite.rst
```diff
@@ -0,0 +1,25 @@
...
+-----------------------------
+
+.. admonition:: To try pyhf code in the browser with the Pyolite JupyterLite kernel:
+    :class: dropdown
+
+     #. Type (or copy and paste) code in the input cell.
+     #. To execute the code, press ``Shift + Enter`` or click on the (Run) ▶ button in the toolbar.
+
+     To get going try copying and pasting the "Hello World" example below!
+
```
matthewfeickert, line 11: @kratsg @alexander-held Do you think that this info would be helpful? Or is this overestimating the problem and severity?

```suggestion

     .. admonition:: Kernel not loading?
      :class: dropdown

       If the kernel isn't loading and you just see a blank iframe, you will need your
       browser to accept cookies from jupyterlite.github.io and then reload the page.

```

matthewfeickert: Ah I see Alex already said this would be helpful. I'll add.

alexander-held: Definitely feel free to not include it, I just figured it does not hurt to have and may help a few people.

matthewfeickert:
> I was wondering about the dependencies of `pyhf` as well, or are they already included in this? I don't know anything about how that setup works, but expected that `numpy` and other dependencies may add a bit more to the size.

[NumPy and SciPy are built into Pyodide](https://pyodide.org/en/stable/usage/packages-in-pyodide.html), so we only have to install `pyhf` itself here. With [`micropip`](https://pyodide.org/en/stable/usage/loading-packages.html)/`piplite` you can only install pure-Python wheels, so packages with compiled dependencies need to already be built in, for the time being.

> I don't have any strong opinion here, but wanted to bring up the idea of delaying the loading until the user takes some action (clicks a button or so), ensuring they would want things to be loaded. Not sure whether this is even feasible and how much it actually matters though.

I don't think this is possible as implemented.

matthewfeickert:
Ah [`Jinja2` `v3.1.0`](https://pypi.org/project/Jinja2/3.1.0/#history) was released 2 hours ago and is breaking Sphinx. :(

jpivarski (commented):
I think this is a great idea and it's awesome if you got it to work. Is there a way to see it? GitHub's rendering of README.md at this point in history doesn't show it.

matthewfeickert:
> I think this is a great idea and it's awesome if you got it to work. Is there a way to see it? GitHub's rendering of README.md at this point in history doesn't show it.

It isn't in the README (we're injecting it into the docs at build time with some RsT directives so that we don't hit twine/PyPI errors), but check out the RTD build: https://pyhf--1820.org.readthedocs.build/en/1820/ :slightly_smiling_face:

henryiii:
> but I think the largest part of all of this is the download of Pyodide

FYI, the "download" is ~10 MB for Pyodide, and it does not "include" numpy or SciPy (SciPy, especially, is quite large). Those get downloaded when you run the micropip command. They get download from the CDN instead of PyPI, but it's still a download. Once a user visits the page, they should be cached. I'm guessing this will be painfully slow for someone on a slow connection, but then again, they don't really need to sit here and wait for it to load, they can just read the docs as normal while it's loading.

matthewfeickert:
> As a suggestion, maybe you want
> 
> ```
> %matplotlib inline
> ```
> 
> to be included in the cell of preloaded statements, so that any examples someone tries that generate a plot will show up without the extra step of `plt.show()`.

Very good point. I'll add that. Thanks @jpivarski! 

> For the future, maybe a small [gallery](https://matplotlib.org/stable/gallery/index.html) of typical analyses you might do with pyhf as a toolbar: clicking on an image of a Brazil band plot fills and evaluates a code cell that makes the Brazil band plot, which the user can then interact with in subsequent evaluations? pyhf needs data to be useful: there should be exploratory-clickable ways of inserting data to play with.

This is now Issue #1825. :+1: 

> FYI, the "download" is ~10 MB for Pyodide, and it does not "include" numpy or SciPy (SciPy, especially, is quite large). Those get downloaded when you run the micropip command. They get download from the CDN instead of PyPI, but it's still a download. Once a user visits the page, they should be cached. I'm guessing this will be painfully slow for someone on a slow connection, but then again, they don't really need to sit here and wait for it to load, they can just read the docs as normal while it's loading.

Ah thanks very much for this clarification @henryiii! This is good to know in general. Yeah, I'm hoping that for people who are on a slow connection this won't be too annoying, but hopefully they will also not be coming to the docs with the express intention of running things in JupyterLite in that situation.

jpivarski:
> FYI, the "download" is ~10 MB for Pyodide, and it does not "include" numpy or SciPy (SciPy, especially, is quite large).

But fortunately, it's the last thing loaded and not strictly necessary to understand the page. Visitors can be reading other things while it loads. From where I am on the network, it evaluated the first command "20-Mississippi" seconds after the initial page load, but the page didn't look broken in the meantime.

From a psychological/usability perspective, this is much better than a Binder link or even Thebe, which make you wait _after_ you've decided that you want the thing.

That load time does include NumPy and SciPy. When I tried `import numpy, scipy`, they loaded right away and there were no pip errors. (I don't know how pyhf would work without NumPy, anyway.)

matthewfeickert:
@kratsg I have Jim's suggestions added locally, but as PR #1824 still needs to go in so the docs can build and its jobs are still queued up (big hugs to all the tired GitHub engineers out there today) I'm not going to push any more updates to this branch until things are moving again.

kratsg:
> (I don't know how pyhf would work without NumPy, anyway.)

actuality, we don't need that with the `set_backend('torch', default=True)`. Primary point is that we do try to load numpy as the start as the default upon import, but we could definitely find a better way to allow for the default to change before `pyhf` is loaded which would avoid `numpy` entirely. (This was a requirement for making it differentiable).

jpivarski (approved):
I see there's `%matplotlib inline` in there. I think it's great!

henryiii:
Loading numpy doesn't matter, "installing" it (via micropip) is what takes some time, that's when it's being copied to your local browser cache. But that's fine. SciPy is worse, but also fine.  Importing it just does the normal Python thing (importing) from that local file.

I think there's a way to "hide" this, that is, cause it to load automatically. Some places (like numpy.org) load it as part of loading the season. It just saves a little visible code (the micropip part), but nice because it would look more like "native" code and not confuse users with a strange pyodide construct. I haven't worked that out, though (it might just be calling the load manually from JS).

alexander-held:
The concern I had regarding the download size is that people might not be aware that this is a thing, and would want to not download this on connections with limited data (browsing docs under the assumption that they are lightweight). But perhaps that is a minor concern. I see the argument outlined above regarding the convenience of everything being "ready to go" without the user having to do extra things as well.

kratsg (approved):

Issue #1840, opened 2022-04-06
nbclient.exceptions.DeadKernelError from papermill on GHA runners


matthewfeickert:
### Summary

When trying to get the Notebook tests to work again, it seems that any notebook that happens to 

### OS / Environment

```console
GHA runner for `ubuntu-latest`
```

### Steps to Reproduce

On a GHA runner execute

```console
$ pytest --override-ini filterwarnings= -sx tests/test_notebooks.py -k test_xml_importexport
```

### File Upload (optional)

[log.txt](https://github.com/scikit-hep/pyhf/files/8424800/log.txt)

### Expected Results

For 

```console
$ pytest --override-ini filterwarnings= -sx tests/test_notebooks.py -k test_xml_importexport
```

to pass on the runner like it does locally

### Actual Results

```console
It seems to fail at the 


!ls -lavh ../../../validation/xmlimport_input
```

cell
```


### pyhf Version

```console
dev
```

### Code of Conduct

- [X] I agree to follow the Code of Conduct

matthewfeickert:
@kratsg @lukasheinrich I'm unable to replicate the error that the GHA runners in PR #1841 are seeing when I (on the same branch) run

```console
$ pytest --override-ini filterwarnings= tests/test_notebooks.py -k test_xml_importexport
================================================================= test session starts ==================================================================
platform linux -- Python 3.9.6, pytest-7.1.1, pluggy-1.0.0
Matplotlib: 3.5.1
Freetype: 2.6.1
benchmark: 3.4.1 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/feickert/Code/GitHub/pyhf, configfile: pyproject.toml
plugins: console-scripts-1.3.1, mock-3.7.0, mpl-0.14.0, requests-mock-1.9.3, benchmark-3.4.1, cov-3.0.0, anyio-3.3.3
collected 12 items / 11 deselected / 1 selected                                                                                                        

tests/test_notebooks.py .                                                                                                                        [100%]
```

Can one of you try to replicate the GHA runner behavior before I try to bug `papermill` about this?

matthewfeickert:
Oh. It fails if you add `PYTHONWARNINGS='default'` into the environment

```console
[...]

PR #1841, opened 2022-04-06
fix: Override error on filterwarnings to pass notebook tests

matthewfeickert:
# Description

To get the notebook tests to pass given the many warnings that pop up from the Jupyter infrastructure and `papermill`, override the ini option for `filterwarnings` (added in PR #1773) with an empty list to disable error on filterwarnings as the notebook tests are testing for notebooks to run with the latest API, not if Jupyter infrastructure is warning free.

The use of `PYTHONWARNINGS='default'` should be possible, but is causing errors and is now Issue #1840.

Use `scrapbook` as it is the new name of `nteract-scrapbook` and `nteract-scrapbook` is no longer maintained or developed.
Restrict `papermill` and `scrapbook` to compatible releases at the patch level for API stability.

Use the `papermill.execute_notebook` `cwd` optional argument to change the execution directory safely to avoid having the state of operations change in the test.

Add concurrency group to avoid wasting run time.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use 'scrapbook' as it is the new name of 'nteract-scrapbook', which is no longer
maintained or developed.
* Restrict 'papermill' and 'scrapbook' to compatible releases at the patch level
for API stability.
* Override error on filterwarnings for the 'notebooks' GHA workflow as it is testing
for the notebooks to run with the latest pyhf API, not if the Jupyter infrastructure
is warning free.
* Use the papermill.execute_notebook 'cwd' optional argument to change the execution
directory safely.
* Use a concurrency group to limit the notebook testing workflow to only one run
at a time.
   - c.f. https://docs.github.com/en/actions/using-jobs/using-concurrency
```

### DIFF
--- a/.github/workflows/notebooks.yml
+++ b/.github/workflows/notebooks.yml
@@ -39,6 +39,4 @@ jobs:
         # Override the ini option for filterwarnings with an empty list to disable error
         # on filterwarnings as testing for notebooks to run with the latest API, not if
         # Jupyter infrastructure is warning free.
-        # Though still show warnings by setting warning control to 'default'.
-        # export PYTHONWARNINGS='default'
         pytest --override-ini filterwarnings= tests/test_notebooks.py
--- a/.github/workflows/notebooks.yml
+++ b/.github/workflows/notebooks.yml
@@ -37,11 +37,13 @@ jobs:
       run: python -m pip list
 
     - name: REMOVE test a failing test by self
+      shell: bash
       run: |
         export PYTHONWARNINGS='default'
         pytest --override-ini filterwarnings= tests/test_notebooks.py -k test_xml_importexport
 
     - name: Test example notebooks
+      shell: bash
       run: |
         # Override the ini option for filterwarnings with an empty list to disable error
         # on filterwarnings as testing for notebooks to run with the latest API, not if
--- a/.github/workflows/notebooks.yml
+++ b/.github/workflows/notebooks.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
-        python -m pip --quiet install --upgrade .[test]
+        python -m pip install --upgrade .[test]
 
     - name: List installed Python packages
       run: python -m pip list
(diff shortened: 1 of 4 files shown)


.github/workflows/notebooks.yml
```diff
@@ -16,15 +22,32 @@ jobs:
...
+        python -m pip install --upgrade .[test]
+
+    - name: List installed Python packages
+      run: python -m pip list
+
+    - name: REMOVE test a failing test by self
+      shell: bash
+      run: |
+        export PYTHONWARNINGS='default'
+        pytest --override-ini filterwarnings= tests/test_notebooks.py -k test_xml_importexport
```
matthewfeickert, line 43: This is just here to understand what is happening in Issue #1840 at the single test level. This should be removed after that.

tests/test_notebooks.py
```diff
@@ -1,16 +1,19 @@
...
 
 @pytest.fixture()
 def common_kwargs(tmpdir):
     outputnb = tmpdir.join('output.ipynb')
     return {
-        'output_path': str(outputnb),
+        'output_path': Path(outputnb),
         'kernel_name': f'python{sys.version_info.major}',
+        'progress_bar': False,
+        'start_timeout': 120,
```
matthewfeickert, line 16: Added to try to avoid issues, but maybe not needed.

.github/workflows/notebooks.yml
```diff
@@ -16,15 +20,29 @@ jobs:
...
+
     - name: Test example notebooks
       run: |
-        pytest tests/test_notebooks.py
+        # Override the ini option for filterwarnings with an empty list to disable error
+        # on filterwarnings as testing for notebooks to run with the latest API, not if
+        # Jupyter infrastructure is warning free.
+        # Though still show warnings by setting warning control to 'default'.
+        # export PYTHONWARNINGS='default'
```
matthewfeickert, line 47: I don't understand why having `PYTHONWARNINGS='default'` doesn't work, so maybe this will become an Issue unto itself and then we can just move on here.

matthewfeickert: I've just summarized things in Issue #1840.

PR #1845, opened 2022-04-08
feat: Allow for clipping negative yields in the main model

kratsg:
# Description

Provides functionality for clipping `expected_data` in the `_MainModel` via `pyhf.Model(clip_sample_data=0.0, clip_bin_data=0.0)` keyword arguments. Added to the docstrings.

Backwards functionality is preserved, which is that clipping is not done. There does not seem to be a performance hit in checking for clipping or not. Clipping can be done at two levels:

- the level at which we have the individual samples (perhaps useful if we need the granularity and a specific sample is providing large negative yields)
- the level at which we have the bin-by-bin data (this is primarily the expected use case that makes the most sense? Maybe...)

Note that it seems ROOT primarily goes for the level of the individual samples, so adding both options and leaving it up to the user to decide if they want to use this. If you have all but one sample which is ok, and one sample providing large negative yields, you can either zero out the sample, or zero out the bin.

See documentation changes here: https://pyhf--1845.org.readthedocs.build/en/1845/_generated/pyhf.pdf.Model.html#pyhf.pdf.Model

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* API change to allow for the _MainModel to clip expected data at the sample-wise
data, or bin-wise data levels.
   - Should help make Minuit a little more robust in particular cases where the
     workspace provided has regions of phase-space where the model is not
     well-behaved (negative yields).
   - Add warning to initialization of _MainModel if clipping is used to ensure that users are
     aware of what they are doing.
* Add tests for clipping by-sample and clipping by-bin.
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -556,7 +556,6 @@ class _MainModel:
         return True
 
     def make_pdf(self, pars):
-        tensorlib, _ = get_backend()
         lambdas_data = self.expected_data(pars)
         return prob.Independent(prob.Poisson(lambdas_data))
 
@@ -820,8 +819,6 @@ class Model:
             pdf: A distribution object implementing the main measurement pdf of HistFactory
 
         """
-        tensorlib, _ = get_backend()
-
         pdfobjs = []
         mainpdf = self.main_model.make_pdf(pars)
         if mainpdf:
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -631,7 +631,7 @@ class _MainModel:
         allfac = tensorlib.concatenate(factors + [nom_plus_delta])
 
         newbysample = tensorlib.product(allfac, axis=0)
-        if self.clip_sample_data:
+        if self.clip_sample_data is not None:
             newbysample = tensorlib.clip(
                 newbysample, self.clip_sample_data, max_value=None
             )
@@ -643,7 +643,7 @@ class _MainModel:
             return batch_first
 
         newresults = tensorlib.sum(newbysample, axis=0)
-        if self.clip_bin_data:
+        if self.clip_bin_data is not None:
             newresults = tensorlib.clip(newresults, self.clip_bin_data, max_value=None)
 
         if self.batch_size is None:
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -526,6 +526,14 @@ class _MainModel:
         self.clip_sample_data = clip_sample_data
         self.clip_bin_data = clip_bin_data
 
+        if self.clip_sample_data is not None:
+            log.warning(
+                f"Clipping expected data per-bin for each sample below {self.clip_sample_data}"
+            )
+
+        if self.clip_bin_data is not None:
+            log.warning(f"Clipping expected data per-bin below {self.clip_bin_data}")
+
         self._nominal_rates = default_backend.tile(
             nominal_rates, (1, 1, self.batch_size or 1, 1)
         )
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -545,7 +545,9 @@ class _MainModel:
         return True
 
     def make_pdf(self, pars):
+        tensorlib, _ = get_backend()
         lambdas_data = self.expected_data(pars)
+        lambdas_data = tensorlib.clip(lambdas_data, 1e-6, max_value=None)
         return prob.Independent(prob.Poisson(lambdas_data))
 
     def logpdf(self, maindata, pars):
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -2,7 +2,7 @@
 
 import copy
 import logging
-from typing import List
+from typing import List, Union
 
 import pyhf.parameters
 import pyhf
@@ -506,7 +506,15 @@ class _ConstraintModel:
 class _MainModel:
     """Factory class to create pdfs for the main measurement."""
 
-    def __init__(self, config, modifiers, nominal_rates, batch_size=None):
+    def __init__(
+        self,
+        config,
+        modifiers,
+        nominal_rates,
+        batch_size=None,
+        clip_sample_data: Union[float, None] = None,
+        clip_bin_data: Union[float, None] = None,
+    ):
         default_backend = pyhf.default_backend
 
         self.config = config
@@ -514,6 +522,9 @@ class _MainModel:
         self._factor_mods = []
         self._delta_mods = []
         self.batch_size = batch_size
+
+        self.clip_sample_data = clip_sample_data
+        self.clip_bin_data = clip_bin_data
 
         self._nominal_rates = default_backend.tile(
             nominal_rates, (1, 1, self.batch_size or 1, 1)
@@ -547,7 +558,6 @@ class _MainModel:
     def make_pdf(self, pars):
         tensorlib, _ = get_backend()
         lambdas_data = self.expected_data(pars)
-        lambdas_data = tensorlib.clip(lambdas_data, 1e-6, max_value=None)
         return prob.Independent(prob.Poisson(lambdas_data))
 
     def logpdf(self, maindata, pars):
@@ -621,6 +631,11 @@ class _MainModel:
         allfac = tensorlib.concatenate(factors + [nom_plus_delta])
 
         newbysample = tensorlib.product(allfac, axis=0)
+        if self.clip_sample_data:
+            newbysample = tensorlib.clip(
+                newbysample, self.clip_sample_data, max_value=None
+            )
+
         if return_by_sample:
             batch_first = tensorlib.einsum('ij...->ji...', newbysample)
             if self.batch_size is None:
@@ -628,6 +643,9 @@ class _MainModel:
             return batch_first
 
         newresults = tensorlib.sum(newbysample, axis=0)
+        if self.clip_bin_data:
+            newresults = tensorlib.clip(newresults, self.clip_bin_data, max_value=None)
+
         if self.batch_size is None:
             return newresults[0]
         return newresults
@@ -642,6 +660,8 @@ class Model:
         modifier_set=None,
         batch_size=None,
         validate: bool = True,
+        clip_sample_data: Union[float, None] = None,
+        clip_bin_data: Union[float, None] = None,
         **config_kwargs,
     ):
         """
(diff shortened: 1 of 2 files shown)


kratsg:
./cc @mdhank

src/pyhf/pdf.py
```diff
@@ -545,6 +556,7 @@ def has_pdf(self):
         return True
 
     def make_pdf(self, pars):
+        tensorlib, _ = get_backend()
```
kratsg, line 559:
```suggestion
```

kratsg:
@HDembinski - this is a possible interesting edge-case for Minuit. Minuit has issues with convergence on a (admittedly hacked-together) workspace that throws negative yields. If we explicitly add clipping to make sure the data isn't negative, there's no issue. (SciPy seems to be "robust" regardless).

tests/test_pdf.py
```diff
@@ -1070,3 +1070,115 @@ def make_model(
...
+    assert tensorlib.tolist(
+        model_clip_sample.expected_actualdata(pars)
+    ) == pytest.approx([1.830496e02, 0.0, 0.0], abs=1e-3)
+
+    # Check with clipping by-bin
+    model_clip_bin = ws.model(clip_bin_data=0.0)
+    assert all(value >= 0 for value in model_clip_bin.expected_actualdata(pars))
+    assert tensorlib.tolist(model_clip_bin.expected_actualdata(pars)) == pytest.approx(
+        [1.830496e02, 0.0, 0.0], abs=1e-3
+    )
```
matthewfeickert, line 1173: Are these tolerances of `abs=1e-3` selected as they are reasonable or were these as tight as they could go? Is there any reason to not tighten these a bit?

matthewfeickert (approved):
@kratsg Overall LGTM. :+1: Thanks! 

I just have a question on the tolerances, but beyond that this seems good. Also is there any associated GitHub Issue or Discussion for this? Or was this only brought to our attention through email?

matthewfeickert:
(We can ignore that Semantic Pull Request isn't reporting at the moment given https://github.com/zeke/semantic-pull-requests/issues/183 and just visually check. This PR is fine. :+1:)

HDembinski:
This is not a Minuit issue. You should not have negative predictions for bin entries in a maximum likelihood fit, ever. Negative expectations are mathematically forbidden. Construct and limit the model parameters so that this never happens. If scipy handles this better, it is by chance.

tests/test_pdf.py
kratsg: Tensorflow has larger deviations. But there's no reason they need to be tight.. the point is just to see zeros, and the problem is that the first bin varies by 7e-3 in tensorflow (and 1e-6 in other backends).

matthewfeickert:
> There is a related study in the iminuit docs about using scipy as an optimisier and a constrained fit. I can't link it here because I am on the phone.

(Linking here before I sign off for the night) I think the study is this one: [SciPy minimizers and constraints](https://iminuit.readthedocs.io/en/stable/notebooks/scipy_and_constraints.html) (which is the [`scipy_and_constraints.ipynb` notebook in the repo](https://github.com/scikit-hep/iminuit/blob/9884b3ffc019730fdee673e4eec1ed6489dfb31d/doc/notebooks/scipy_and_constraints.ipynb)).

HDembinski:
To add to my previous comment: this workaround is not going to work well, because by clipping the predictions you introduce discontinuities in the the cost function, but Minuit and other gradient-based minimisers expect that the cost function is analytical, an infinitesimal changes in the parameters introduces an infinitesimal change in the value of the cost function.

kratsg:
> To add to my previous comment: this workaround is not going to work well, because by clipping the predictions you introduce discontinuities in the cost function, but Minuit and other gradient-based minimisers expect that the cost function is analytical, an infinitesimal changes in the parameters produces an infinitesimal change in the value of the cost function. When you clip, this is no longer the case.

Yeah, this is actually the reason why I and others like @alexander-held are surprised to see this *turned on* by default in ROOT, so we're not entirely sure why. We can add specific documentation in the docstring to say that this option is "dangerous" (or that they should be careful trusting any results that needs this option).

HDembinski:
> 


> @HDembinski , In general, I would agree that well-designed likelihood should not spit out negative yields. I think it happens when you have a lot of shape systematics that are uncorrelated but pulled in the same direction. I think limiting the model parameters manually would make the most sense if your model is simple enough (some of the issues from within ATLAS correspond with models with > 100 parameters, so it becomes slightly more intractable to hunt down all problematic parameters).

I think it is not so difficult, because most of the parameters are shape parameters. You can change those as you like, it will not generate negative bin expectations. You may get zero expectation in some bin, which is also bad, but easier to work around. You cannot work around negative expectations. The main problem is the use Chebychev polynomials for background and allowing amplitudes of pdfs to be negative. Both is a bad tradition that we need to break away from.
 
> In ROOT, I will point out that this is enabled by default, which means ROOT is using Minuit with the forced positive-definite.

Can you please link to that code in ROOT, because I don't know what you mean? You talk about positive-definiteness, but that is a different thing from what I am talking about. Positive-definiteness means that the curvature of the cost function is positive in all directions. The matrix of second derivatives has to be positive-definite for Minuit/Migrad both in ROOT and in iminuit.

> I think it would be useful to understand why SciPy seems to be able to converge in these cases while Minuit [which is more strict and careful] will rightfully complain.

Minuit's Migrad algorithm is not more strict and careful. On the contrary, Migrad as implemented in Minuit2, is less robust than some of the algorithms in SciPy in my experience. Migrad cannot recover well from situations where the cost function returns NaN. I think there is a potential for improvement there, but it is not an easy fix.
 
> Does this indicate, even with the comparison studies you've done, that SciPy is wrong and should be more strict? Or is the argument you're making that for poorly-defined workspaces, you throw everything out of the window when it comes to comparing Minuit and SciPy?

My comparisons indicate that sometimes Migrad works better than other minimisers and sometimes the other minimisers work better. It depends on the specific problem. My argument is that people need to define their models properly and there is little that you can do on the level of pyhf to fix bad models constructed by users. Another (unrelated) common problem is that users generate models with strongly correlated parameters. There is also nothing you can do in pyhf to prevent that. You should educate your users about this as I am doing.

alexander-held:
> I think it is not so difficult, because most of the parameters are shape parameters. You can change those as you like, it will not generate negative bin expectations.

In the context of `pyhf`, "shape" variations are encoded by `histosys` modifiers which leave the total yield across all bins in a channel unchanged. If the shape is sufficiently strong, it can lead to scenarios where a change of the related parameter can drive the predicted yield in some of the bins to be negative (compensating for increases in yield in other bins). An example for that is [in this gist](https://gist.github.com/alexander-held/2807ab29ea4246c47bd17eb73ec35920) for the case of two `histosys` modifiers, the same also can happen for a single `histosys` modifier if the parameter is pulled beyond one sigma. A way to address this in practice is the use of `normsys` modifiers that provide exponential extrapolation, but any remaining shape effects can still cause this behavior in principle.

alexander-held:
https://github.com/scikit-hep/pyhf/discussions/1884 might also be of interest in this context: it's not specifically about protecting against negative yields, but protecting against NaNs due to ill-defined templates. I don't know where/how exactly ROOT protects against this.

src/pyhf/pdf.py
```diff
@@ -619,13 +630,21 @@ def expected_data(self, pars, return_by_sample=False):
         allfac = tensorlib.concatenate(factors + [nom_plus_delta])
 
         newbysample = tensorlib.product(allfac, axis=0)
+        if self.clip_sample_data is not None:
```
matthewfeickert, line 633: @kratsg this is probably something that should go into a new PR rather than this one to avoid delaying this, but if we allow for this should we add logging that explicitly warns the user about what they are doing? We're mirroring ROOT's behavior here, but potentially allowing for users who might be a bit naive to wander into unexpected territory. Thoughts from @lukasheinrich and @alexander-held here?

src/pyhf/pdf.py
```diff
@@ -619,13 +630,21 @@ def expected_data(self, pars, return_by_sample=False):
...
+            )
+
         if return_by_sample:
             batch_first = tensorlib.einsum('ij...->ji...', newbysample)
             if self.batch_size is None:
                 return batch_first[0]
             return batch_first
 
         newresults = tensorlib.sum(newbysample, axis=0)
+        if self.clip_bin_data is not None:
```
matthewfeickert, line 645: ditto the above

matthewfeickert (approved):
> I still would like to provide this functionality in `pyhf` to at least match what exists in ROOT, but keep it as non-default. I do want to add more verbose description in the documentation to clarify the expectations of the minimizers and the expectations of the users, to make sure people understand that it is not up to `pyhf` to enforce that workspaces are well-defined, but we'll do our best.

@kratsg (thanks for your patience and I hope that I never do such a poor job of responding to you again) I think this is a reasonable path forward for now. Though to avoid more delays on this I think making a dedicated docs Issue for the descriptions/discussions would be a good idea.

src/pyhf/pdf.py
alexander-held: A warning sounds like a good idea to me. I think there's a chance that using this becomes the standard and people copy&paste things around without paying much attention to this kwarg. Having a warning could help ensure that users are aware of what is happening. What would perhaps be most useful is a message that is only emitted when the clipping does something, but that presumably requires an additional runtime check that might be expensive.

kratsg: A problem with a warning is that this function gets called a lot of times... so we don't want to spam the user. We could add a warning during model initialization...

mdhank:
I have also tested this some on my workspaces, I've found it does seem to help. Interestingly, it seems clipping bins might improve stability more than clipping samples; I've run into some cases where the former works but not the latter.

src/pyhf/pdf.py
kratsg: Added warning to initialization of model.

matthewfeickert: Thanks! (just saw that you did this)

matthewfeickert:
> I have also tested this some on my workspaces, I've found it does seem to work as intended in my cases. Interestingly, it seems clipping bins might improve stability more than clipping samples; I've run into some cases where the former works but not the latter.

@mdhank Thanks for testing this out. I'm quite behind on things at the moment (this PR is testimony to that) but if you have examples of where this is helping can you share them?

PR #1851, opened 2022-04-27
docs: Deprecate use of git.io URL shortener

matthewfeickert:
# Description

[GitHub is killing off the `git.io` URL shortening service this Friday (2022-04-29)](https://github.blog/changelog/2022-04-25-git-io-deprecation/) so all URLs in the docs that use it need to be updated.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Replace all instances of git.io shortened URLs with full URLs as git.io is
being killed off by GitHub on 2022-04-29.
   - c.f. https://github.blog/changelog/2022-04-25-git-io-deprecation/
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,7 +70,8 @@ Alternatively the statistical model and observational data can be read from its 
    >>> import pyhf
    >>> import requests
    >>> pyhf.set_backend("numpy")
-   >>> wspace = pyhf.Workspace(requests.get("https://git.io/JJYDE").json())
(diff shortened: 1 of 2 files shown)


matthewfeickert:
@kratsg if you have a favorite URL shortener that we can rely on to be stable for a long time to come it would be good to switch to that instead of using the full URLs.

kratsg (approved):

PR #1856, opened 2022-04-29
docs: Clarify staterror modifier specification

Moelf:
clarify and fix typo for `Staterror` modifier.

# Description
The `\delta` parameter is the absolute error in each bin due to MC Stat at +/- 1 sigma. Thus, the `\sigma` of the Gaussian constraint term should have `\delta / nominal` value not the other way around.

And by play around with the `logpdf()` output, I'm convinced that the input in JSON is `\delta` rather than `\delta^2` (https://scikit-hep.org/pyhf/intro.html#id25)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Correct the description of the calculation of the scale of the constraint
to indicate it is calculated from the relative statistical uncertainty from the
channel's event rate.
* Add additional channel subscripts to make the notation be more explicit that
staterror is per-channel.
* Remove typo of additional 'are'.
* Add Jerry Ling to contributor list.
```

### DIFF
--- a/docs/likelihood.rst
+++ b/docs/likelihood.rst
@@ -199,11 +199,13 @@ each sample would yield a very large number of nuisance parameters with limited
 utility. Therefore a set of bin-wise scale factors :math:`\gamma_b` is
 introduced to model the overall uncertainty in the bin due to MC statistics.
 The constraint term is constructed as a set of Gaussian constraints with a
-central value equal to unity for each bin in the channel. The scales
-:math:`\sigma_b` of the constraint are computed from the individual
+central value equal to unity (one Gaussian for each bin). The scales
+:math:`\sigma_b` of the Gaussian are computed from the individual
 uncertainties of samples defined within the channel relative to the total event
-rate of all samples: :math:`\delta_{csb} = \sigma_{csb}/\sum_s \nu^0_{scb}`. As
-not all samples are within a channel are estimated from MC simulations, only
+rate of all samples: :math:`\sigma_{csb} = \delta_{csb}/\sum_s \nu^0_{scb}`.
+Where :math:`\delta_{csb}` is the absolute yield uncertainty in each bin.
+
+As not all samples are within a channel are estimated from MC simulations, only
 the samples with a declared statistical uncertainty modifier enter the sum.
 An example of a statistical uncertainty modifier for a single bin channel is
 shown below:
--- a/docs/likelihood.rst
+++ b/docs/likelihood.rst
@@ -202,8 +202,8 @@ The constraint term is constructed as a set of constraints with a
 central value equal to unity (e.g. `Gauss(:math:`\mu` = 1, :math:`\sigma_b`)` for each bin). The scales
 :math:`\sigma_b` of the constraints are computed from the individual
 uncertainties of samples defined within the channel relative to the total event
-rate of all samples: :math:`\sigma_{csb} = \delta_{csb}/\sum_s \nu^0_{scb}`.
-Where :math:`\delta_{csb}` is the absolute yield uncertainty in each bin.
+rate of all samples: :math:`\sigma_{b} = \sqrt{\sum_s\delta_{sb}}/\sum_s \nu^0_{sb}`.
+Where :math:`\delta_{sb}` is the absolute yield uncertainty in each bin, each sample.
 
 As not all samples are within a channel are estimated from MC simulations, only
 the samples with a declared statistical uncertainty modifier enter the sum.
--- a/docs/likelihood.rst
+++ b/docs/likelihood.rst
@@ -196,14 +196,15 @@ As the sample counts are often derived from Monte Carlo (MC) datasets, they
 necessarily carry an uncertainty due to the finite sample size of the datasets.
 As explained in detail in :cite:`likelihood-Cranmer:1456844`, adding uncertainties for
 each sample would yield a very large number of nuisance parameters with limited
-utility. Therefore a set of bin-wise scale factors :math:`\gamma_{cb}` is
+utility.
+Therefore a set of bin-wise scale factors :math:`\gamma_{cb}` is
 introduced to model the overall uncertainty in the bin due to MC statistics.
 The constraint term is constructed as a set of constraints with a
 central value equal to unity (e.g. Gauss(:math:`\mu` = 1, :math:`\sigma_{cb}`) for
-each bin in the channel). The scales :math:`\sigma_{cb}` of the
-constraints are computed from the individual uncertainties of samples
-defined within the channel relative to the total event rate of all samples:
-:math:`\sigma_{cb} = \sqrt{\sum_s\delta_{csb}}/\sum_s \nu^0_{csb}`.
+each bin in the channel).
+The scales :math:`\sigma_{cb}` of the constraints are computed from the individual
+uncertainties of samples defined within the channel relative to the total event rate
+of all samples: :math:`\sigma_{cb} = \sqrt{\sum_s\delta_{csb}}/\sum_s \nu^0_{csb}`.
 Where :math:`\delta_{csb}` is the absolute yield uncertainty in each bin.
 
 As not all samples are within a channel are estimated from MC simulations, only
--- a/docs/likelihood.rst
+++ b/docs/likelihood.rst
@@ -200,8 +200,8 @@ utility.
 Therefore a set of bin-wise scale factors :math:`\gamma_{cb}` is
 introduced to model the overall uncertainty in the bin due to MC statistics.
 The constraint term is constructed as a set of constraints with a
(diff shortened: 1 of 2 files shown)


matthewfeickert:
Thanks for the PR @Moelf. We'll try to get to it before the end of the weekend.

As you didn't make a GitHub Issue for this can you give a full summary in the PR body and then also finish off the rest of the [CONTRIBUTING guidelines](https://github.com/scikit-hep/pyhf/blob/master/CONTRIBUTING.md#pull-request-procedure) (hint: adding yourself to contributors. :wink:)?

Moelf:
I think the `\delta^2` in the Table can be understood as:
1. we take the quadrature sum of `\delta_b` across all the samples with the same MC error
2. we pass the `sqrt(sum)` into the `staterror` constructor.

So I think the table is not wrong.

docs/likelihood.rst
```diff
@@ -199,11 +199,13 @@ each sample would yield a very large number of nuisance parameters with limited
 utility. Therefore a set of bin-wise scale factors :math:`\gamma_b` is
 introduced to model the overall uncertainty in the bin due to MC statistics.
 The constraint term is constructed as a set of Gaussian constraints with a
-central value equal to unity for each bin in the channel. The scales
-:math:`\sigma_b` of the constraint are computed from the individual
+central value equal to unity (one Gaussian for each bin). The scales
```
kratsg, line 202: Well, one can _change_ the constraint for the staterror typically... so I'm slightly concerned about specifically mentioning the distribution here.

docs/likelihood.rst
```diff
@@ -199,11 +199,13 @@ each sample would yield a very large number of nuisance parameters with limited
...
+:math:`\sigma_b` of the Gaussian are computed from the individual
```
kratsg, line 203: same comment here

docs/likelihood.rst
```diff
@@ -199,11 +199,13 @@ each sample would yield a very large number of nuisance parameters with limited
...
 uncertainties of samples defined within the channel relative to the total event
-rate of all samples: :math:`\delta_{csb} = \sigma_{csb}/\sum_s \nu^0_{scb}`. As
-not all samples are within a channel are estimated from MC simulations, only
+rate of all samples: :math:`\sigma_{csb} = \delta_{csb}/\sum_s \nu^0_{scb}`.
```
kratsg, line 205: this deviates from the notation in the table...

docs/likelihood.rst
Moelf: what's a possible alternative? I can add `e.g.` maybe?

docs/likelihood.rst
Moelf: true, fixed

docs/likelihood.rst
kratsg: Well, I think we mention Gaussian already originally so maybe... leave this for now.

docs/likelihood.rst
```diff
@@ -198,12 +198,14 @@ As explained in detail in :cite:`likelihood-Cranmer:1456844`, adding uncertaint
...
-The constraint term is constructed as a set of Gaussian constraints with a
-central value equal to unity for each bin in the channel. The scales
-:math:`\sigma_b` of the constraint are computed from the individual
+The constraint term is constructed as a set of constraints with a
+central value equal to unity (e.g. `Gauss(:math:`\mu` = 1, :math:`\sigma_b`)` for each bin). The scales
+:math:`\sigma_b` of the constraints are computed from the individual
 uncertainties of samples defined within the channel relative to the total event
-rate of all samples: :math:`\delta_{csb} = \sigma_{csb}/\sum_s \nu^0_{scb}`. As
-not all samples are within a channel are estimated from MC simulations, only
+rate of all samples: :math:`\sigma_{b} = \sqrt{\sum_s\delta_{sb}}/\sum_s \nu^0_{sb}`.
```
kratsg, line 205: this is missing the channel, as this is done _per_ channel separately.

matthewfeickert:
Note to myself for when I review this tomorrow: Relevant RTD build section is here: https://pyhf--1856.org.readthedocs.build/en/1856/likelihood.html#mc-statistical-uncertainty-staterror

docs/likelihood.rst
```diff
@@ -198,12 +198,14 @@ As explained in detail in :cite:`likelihood-Cranmer:1456844`, adding uncertaint
 each sample would yield a very large number of nuisance parameters with limited
 utility. Therefore a set of bin-wise scale factors :math:`\gamma_b` is
 introduced to model the overall uncertainty in the bin due to MC statistics.
-The constraint term is constructed as a set of Gaussian constraints with a
-central value equal to unity for each bin in the channel. The scales
-:math:`\sigma_b` of the constraint are computed from the individual
+The constraint term is constructed as a set of constraints with a
+central value equal to unity (e.g. `Gauss(:math:`\mu` = 1, :math:`\sigma_b`)` for each bin). The scales
```
matthewfeickert, line 202: I will unhelpfully note before I sign off tonight that there is a Sphinx render issue here with `:math:`

sphinx-issue

I don't mind fixing it tomorrow though myself while reviewing. So don't take this as criticism or a demand &mdash; just a note.

matthewfeickert (changes requested):

docs/likelihood.rst
```diff
@@ -194,16 +194,20 @@ MC Statistical Uncertainty (staterror)
 
 As the sample counts are often derived from Monte Carlo (MC) datasets, they
 necessarily carry an uncertainty due to the finite sample size of the datasets.
-As explained in detail in :cite:`likelihood-Cranmer:1456844`, adding uncertainties for
+As explained in detail in :cite:`likelihood-Cranmer:1456844`, adding uncertainties for
 each sample would yield a very large number of nuisance parameters with limited
-utility. Therefore a set of bin-wise scale factors :math:`\gamma_b` is
+utility.
+Therefore a set of bin-wise scale factors :math:`\gamma_{cb}` is
```
matthewfeickert, line 200: @kratsg @lukasheinrich I need to refresh myself on this, but if we clarify here that staterror is a per-channel modifier, then why don't we need to additionally denote this in the [Modifiers and Constraints table](https://scikit-hep.org/pyhf/intro.html#tab-modifiers-and-constraints)?

modifiers

Did we have a good reason for this?

docs/likelihood.rst
```diff
@@ -194,16 +194,20 @@ MC Statistical Uncertainty (staterror)
...
-not all samples are within a channel are estimated from MC simulations, only
+The constraint term is constructed as a set of constraints with a
+central value equal to unity, e.g. :math:`\mathrm{Gauss} (\mu = 1, \sigma_{cb})`, for
+each bin in the channel.
+The scales :math:`\sigma_{cb}` of the constraints are computed from the individual
+uncertainties of samples defined within the channel relative to the total event rate
+of all samples: :math:`\sigma_{cb} = \sqrt{\sum_s\delta_{csb}}/\sum_s \nu^0_{csb}`,
+where :math:`\delta_{csb}` is the absolute yield uncertainty in each bin.
+
+As not all samples are within a channel are estimated from MC simulations, only
```
matthewfeickert, line 210:
```suggestion
As not all samples within a channel are estimated from MC simulations, only
```

I think we had a typo here in the original docs as well. oops.

matthewfeickert (changes requested):

matthewfeickert (approved):
@lukasheinrich @kratsg while I think we should maybe discuss more the notational consistency across the docs I don't think that this needs to hold up Jerry's PR. So if we get approvals from both of you then I this LGTM for now.

docs/likelihood.rst
Moelf: I think in the end per-bin implies per-channel, since different channels may not have the same binning.

matthewfeickert: > I think in the end per-bin implies per-channel, since different channels may not have the same binning.

Yeah, this is somewhat my point. Though while per-bin notation should imply per-channel, one could also say that being overly explicit probably wouldn't hurt or new users and I don't think would confuse people experienced with the spec.

I could go either way here, so would be good to get @lukasheinrich to weight in given that he is the only one who hasn't yet.

kratsg: > Yeah, this is somewhat my point. Though while per-bin notation should imply per-channel, one could also say that being overly explicit probably wouldn't hurt or new users and I don't think would confuse people experienced with the spec.

This probably depends on how we define the indexing right? Arguably, one could say that `(c,b) = (2, 0)` is equivalent to the flattened coordinate `(b,) = (10,)`. So the notation of `sigma_{cb}` should be equivalent to `sigma_{b}` but the `b` have different interpretations.

matthewfeickert:
@kratsg @lukasheinrich if this can get reviewed today by one (ideally both) of you then I think we can merge this.

kratsg (approved):

matthewfeickert:
Thanks for your PR @Moelf!

PR #1865, opened 2022-05-23
feat: Allow readxml to demote validation exception to warning

### LINKED ISSUE #1864: Skip schema validation for xml2json
### Summary

Sometimes upstream tools like HistFitter can produce non-valid XML workspaces, but we need to be able to convert them to JSON. Schema validation could be configured off in some of these cases (force the error to a warning instead).

### Additional Information

_No response_

### Code of Conduct

kratsg:
# Pull Request Description

This resolves #1864. Added functionality `pyhf.readxml` to push the exception down to a warning if needed, so conversion still works for invalid XML workspaces, but will still warn that it is invalid. This also adds it to `pyhf xml2json` as well.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add ability to passing validation_as_error=False to readxml.parse to warn when the workspace
is invalid instead of just raising an exception (default is to raise exception).
* Add --validation-as-error/validation-as-warning flag to xml2json command line interface to
switch between the two (default is to raise exception).
```

### DIFF
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -1,5 +1,6 @@
 from pyhf import schema
 from pyhf import compat
+from pyhf import exceptions
 
 import logging
 
@@ -333,7 +334,7 @@ def dedupe_parameters(parameters):
     return list({v['name']: v for v in parameters}.values())
 
 
-def parse(configfile, rootdir, track_progress=False, skip_validation=False):
+def parse(configfile, rootdir, track_progress=False, validation_exception=True):
     toplvl = ET.parse(configfile)
     inputs = tqdm.tqdm(
         [x.text for x in toplvl.findall('Input')],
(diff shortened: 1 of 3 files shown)


src/pyhf/readxml.py
```diff
@@ -366,8 +367,13 @@ def parse(configfile, rootdir, track_progress=False):
         ],
         'version': schema.version,
     }
-    schema.validate(result, 'workspace.json')
-
+    try:
+        schema.validate(result, 'workspace.json')
+    except exceptions.InvalidSpecification as exc:
+        if validation_as_error:
+            raise exc
```
matthewfeickert, line 374: Hm. @kratsg do you know why this isn't getting covered?

kratsg: Oh, we don't have anything that covers reading an XML that's invalid.. I can add that pretty easily as another validation.

kratsg: Or just as a mock.

matthewfeickert (approved):
LGTM. :+1: Thanks for fixing the coverage!

PR #1895, opened 2022-06-23
build: Update lower bound on iminuit to v2.7.0 

matthewfeickert:
# Description

This was caught in the [Minimum supported dependencies workflow](https://github.com/scikit-hep/pyhf/blob/2411ed32e4077259a6676bd95f3e381e1b5b3507/.github/workflows/lower-bound-requirements.yml) after PR #1845 was merged.

Update lower bound on `iminuit` to `v2.7.0` as it is the 'oldest' SemVer release that has the behavior (as described in the [`iminiuit` `v2.7.0` changelog](https://iminuit.readthedocs.io/en/stable/changelog.html#july-4-2021)):
> `Minuit.hesse` no longer sets the status of the `FunctionMinimum` unconditionally to valid if it was invalid before.

Without this fix, the situation can arise where `Minuit.hesse()` can cause `Minuit.valid` to be `True` when it should be `False`. Specifically in 

https://github.com/scikit-hep/pyhf/blob/2411ed32e4077259a6676bd95f3e381e1b5b3507/src/pyhf/optimize/opt_minuit.py#L135-L140

during

```
pytest tests/test_pdf.py -k test_pdf_clipping[numpy_minuit]
```

https://github.com/scikit-hep/pyhf/blob/2411ed32e4077259a6676bd95f3e381e1b5b3507/tests/test_pdf.py#L1075

Also update `tests/constraints.txt` to use `iminuit==2.7.0`.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Update lower bound on iminuit to v2.7.0 as it is the 'oldest' SemVer release
that has the behavior:
> Minuit.hesse no longer sets the status of the FunctionMinimum unconditionally
> to valid if it was invalid before.

Without this fix, the situation can arise where `Minuit.hesse()` can cause
`Minuit.valid` to be True when it should be False.
* Update tests/constraints.txt to use iminuit==2.7.0.
* Amends PR #1845
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ extras_require = {
     'torch': ['torch>=1.10.0'],  # c.f. PR #1657
     'jax': ['jax>=0.2.10', 'jaxlib>=0.1.60,!=0.1.68'],  # c.f. Issue 1501
     'xmlio': ['uproot>=4.1.1'],  # c.f. PR #1567
-    'minuit': ['iminuit>=2.7.0'],  # c.f. PR #1893
+    'minuit': ['iminuit>=2.7.0'],  # c.f. PR #1895
 }
 extras_require['backends'] = sorted(
     set(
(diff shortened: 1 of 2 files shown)


matthewfeickert:
Minimum supported dependencies workflow is passing now. :+1:

kratsg (approved):

Issue #1882, opened 2022-06-10
xml2json extremely slow


cburgard:
### Summary

When trying to convert some XMLs to pyhf JSON, I found that the runtime was unacceptably slow (order of hours).

### OS / Environment

```console
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux-logo
```

### Steps to Reproduce

The XMLs are ATLAS internal, but I'm happy to share them with ATLAS members of the dev team on request.

### Expected Results

I was hoping it would finish relatively fast.

### Actual Results

```console
It took hours.
```

### pyhf Version

```console
pyhf, version 0.6.3
uproot 4.2.2
```

### Code of Conduct

- [X] I agree to follow the Code of Conduct

matthewfeickert:
Thanks for the report @cburgard. Can you please share the files as a private GitHub Gist with the dev team?

matthewfeickert:
@cburgard Please first verify that you have read through the [translations docs](https://pyhf.readthedocs.io/en/v0.6.3/babel.html#translations) and tell us how the workspaces were created. Then please tell us the exact commands that you are using.

cburgard:
The workspaces were created with SFramwork, specifically with this function:
https://gitlab.cern.ch/atlas-caf/CAFCore/-/blob/master/SFramework/Root/TSModelFactory.cxx#L176

The exact command I was using is
`pyhf xml2json hww-xml/HWWRun2GGF.xml`

matthewfeickert:
> The workspaces were created with SFramwork, specifically with this function: https://gitlab.cern.ch/atlas-caf/CAFCore/-/blob/b1d40c29b0b4635c28d563db1483317d4a8b25d5/SFramework/Root/TSModelFactory.cxx#L176

[...]

PR #1909, opened 2022-07-01
feat: Handle absolute paths in XML config files (xml2json / readxml)

kratsg:
# Pull Request Description

- support XML configs with absolute paths by being able to prune the paths out/replace them to point elsewhere
- typehint the entire thing and fix some minor holes

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add support for XML configurations with absolute paths by pruning the
paths out and renaming paths on-the-fly in readxml.
* Add -v/--mounts command line arguments for xml2json with similar behavior
to the Docker equivalent to support this feature.
* Add tests and test XML files with absolute paths.
```


matthewfeickert:
From @kratsg's testing so far it looks like this will be able to address Issue #1882. :+1:

src/pyhf/typing.py
```diff
@@ -0,0 +1,132 @@
+import os
```
matthewfeickert, line 1: I still need to actually review this, but if this is only providing the types can this get moved out to a separate PR that we merge in first so that things stay as tight as possible for scope? Some of our PRs have gotten big enough that it becomes difficult to understand later what the atomic scope of things was (e.g. the staterror problem we have now).

kratsg: tricky because there are bugs caught in this PR that are fixed related to resolving that would not have otherwise been caught without the typehints added. This is all being imported in.

kratsg: #1931, #1933 for the particular bugs

kratsg:
@matthewfeickert this is ready, but codecov is broken for this one (some glitch on their side of things) probably related to my rebasing to clean things up.

matthewfeickert:
Relevant RTD build: https://pyhf--1909.org.readthedocs.build/en/1909/_generated/pyhf.readxml.html

src/pyhf/readxml.py
```diff
@@ -343,20 +367,41 @@ def dedupe_parameters(parameters):
...
+    rootdir: PathOrStr,
+    mounts: T.Optional[MountPathType] = None,
+    track_progress: bool = False,
+    validation_as_error: bool = True,
+):
+    """
+    Parse the configfile with respect to the rootdir.
+
+    Args:
+        mounts (:obj:`None` or :obj:`list` of 2-:obj:`tuple` of :class:`pathlib.Path` objects): The first field is the local path to where files are located, the second field is the path where the file or directory are saved in the XML configuration. This is similar in spirit to docker. Default is ``None``.
```
matthewfeickert, line 381:
```suggestion
        mounts (:obj:`None` or :obj:`list` of 2-:obj:`tuple` of :class:`pathlib.Path` objects): The first field is the local path to where files are located, the second field is the path where the file or directory are saved in the XML configuration. This is similar in spirit to Docker volume mounts. Default is ``None``.
```

I'd explicitly state that we're talking about volume mounts for the uninitiated.

src/pyhf/readxml.py
```diff
@@ -1,18 +1,28 @@
-from pyhf import schema
-from pyhf import compat
-from pyhf import exceptions
+from __future__ import annotations
 
 import logging
-
-from pathlib import Path
+import os
+import typing as T
```
matthewfeickert, line 5: Any objection to just following what seems to be more standard and just importing everything we'll need from `typing`?

https://github.com/scikit-hep/hist/blob/208244597b1313a5623b5d41331d7056860bc830/src/hist/typing.py#L4

```suggestion
from typing import TYPE_CHECKING, Callable, Iterable, Tuple, Union
```

src/pyhf/readxml.py
```diff
@@ -1,18 +1,28 @@
...
+from pyhf import compat
+from pyhf import exceptions
+from pyhf import schema
+
 log = logging.getLogger(__name__)
 
+if T.TYPE_CHECKING:
+    PathOrStr = T.Union[str, os.PathLike[str]]
+else:
+    PathOrStr = T.Union[str, "os.PathLike[str]"]
```
matthewfeickert, line 22: Uninformed question: As `typing.TYPE_CHECKING` is never run but only analyzed by `mypy` then there's no way to not take a hit here in coverage, correct?

coverage_codecov

src/pyhf/readxml.py
```diff
@@ -343,20 +367,41 @@ def dedupe_parameters(parameters):
...
-def parse(configfile, rootdir, track_progress=False, validation_as_error=True):
+def parse(
+    configfile: PathOrStr | T.IO[bytes] | T.IO[str],
+    rootdir: PathOrStr,
+    mounts: T.Optional[MountPathType] = None,
+    track_progress: bool = False,
+    validation_as_error: bool = True,
+):
+    """
+    Parse the configfile with respect to the rootdir.
```
matthewfeickert, line 378:
```suggestion
    Parse the ``configfile`` with respect to the ``rootdir``.
```

formatting suggestion for the docs

src/pyhf/readxml.py
matthewfeickert, line 381: Uniformed question: Why are we only adding docstrings for the last 3 args? Why do we not also add them for `configfile` and `rootdir`?

src/pyhf/utils.py
```diff
@@ -41,6 +42,24 @@ def convert(self, value, param, ctx):
             self.fail(f'{value:s} is not a valid equal-delimited string', param, ctx)
 
 
+class VolumeMountPath(click.Path):
```
matthewfeickert, line 45: As this is a public class should we add a docstring and at least add it to `__all__`?

matthewfeickert (commented):
Thanks, this is looking super good.

I have a few questions and suggestions, but nothing major, just things that we could decide if we want to fixup now as they're getting added for the first time.

src/pyhf/readxml.py
kratsg: limitation on codecov's part. can't do much.

src/pyhf/utils.py
kratsg: can't be used outside of click (and not sure why we want to document it right now for users). I hope nobody needs this outside of our code.

src/pyhf/readxml.py
kratsg: added

src/pyhf/utils.py
matthewfeickert: That's a good point (RE: needing to be used witch Click). Should we just make this "Python private" then and name it `_VolumeMountPath`?

kratsg: We didn't do the same with `EqDelimStringParamType`.

matthewfeickert: Yeah. We can revise this later if needed.

matthewfeickert (approved):
LGTM. :+1: Thanks @kratsg!

src/pyhf/readxml.py
matthewfeickert: @henryiii I assume that this is just something that we live with, correct?

matthewfeickert: Never mind. PR #1937 fixes this with `.coveragerc`.

henryiii: What the if here? The string form is fine until you drop Pythons that don’t support it.

Issue #1894, opened 2022-06-22
How to manually override the yields inputs to the model spec


phinate:
### Summary

As discussed briefly on other channels, the number one thing preventing [neos](github.com/gradhep/neos) from being used on a realistic use-case is the technical hurdle that comes with differentiating through `pyhf` model construction (#882).

A recently talked about way to circumvent this issue would be to skip the model construction entirely, and just modify the spec information in-place for a given model. I've tested that this works on a technical level, and it's super fast. However, there are problems with validating the likelihood shape for the modified model, i.e. not all information is propagating in the correct way.

Here's an example of hand-altering the nominal rates of a model to match those of a different model, but their likelihoods don't match:

```python3
import pyhf

pyhf.set_backend("jax")

def make_model(s, b, bup, bdown):
    return pyhf.simplemodels.correlated_background(s, b, bup, bdown)

s = [1,2,3]
b = [21,20,22]

bup = [22,22,24]
bdown = [20, 18, 20]

# some target model
model1 = make_model(s, b, bup, bdown)

# a skeleton model with different yields, but the same up/down variations
model2 = make_model([0,0,0], [0,0,0], bup, bdown)

# simply set the yields of the old model to the new one
model2.main_model.nominal_rates = model1.main_model.nominal_rates

# Plotting:
import numpy as np
import jax
import jax.numpy as jnp
import matplotlib.pyplot as plt

plt.rc("figure", figsize=(6, 2), dpi=200, facecolor="w")

data = [23, 23, 23] + model1.config.auxdata

def plot_model(model, ax, model_name=None):
    grid = (500, 500)
    x_range = np.linspace(0,10,grid[0])
    y_range = np.linspace(0,10,grid[1]) 
    xx, yy = np.meshgrid(x_range, y_range)
    xy = np.concatenate([xx.reshape(-1, 1), yy.reshape(-1, 1)], axis=1)
    @jax.vmap
    def logpdf(pars):
        return model.logpdf(pars, data)[0]
    z = logpdf(jnp.array(xy)).reshape(*grid)
    ax.contourf(xx, yy, z)
    if model_name is not None:
        ax.set_title(model_name)

def plot_difference(model1, model2, ax):
    grid = (500, 500)
    x_range = np.linspace(0,10,grid[0])
    y_range = np.linspace(0,10,grid[1]) 
    xx, yy = np.meshgrid(x_range, y_range)
[...]

PR #1912, opened 2022-07-04
Feat: model building (and gradients thereof) with jax as the default backend

phinate:
# Description

Building on work by @lukasheinrich in #1676 to address #882, this PR attempts to keep going down the debugging rabbit hole to make pyhf model construction properly differentiable for all modifier types.

Right now, there is some jax-only syntax that I plan to refactor for all tensor backends if we get this working, e.g. `array = array.at[index].set(value)`, which is the jax equivalent of `array[index] = value` since jax arrays are immutable.

A notable other addition is the changing of the jax concatenate implementation as referenced in #1655 to something many factors slower, but it's necessary to match numpy's behaviour. It also doesn't produce a user-facing slowdown that I experience, since the arrays involved are small and the operation is infrequent.

One roadblock right now as discussed offline with @kratsg concerns the following logic in the parameter "finalize" step: https://github.com/scikit-hep/pyhf/blob/acde7f4ff8d0db2351f5d6e31ff5584e34da0cf0/src/pyhf/parameters/utils.py#L37-L40

When building models with arrays in them, the `v` here can contain jax arrays, which are not hashable. I tried patching in a hashable construct, but this also failed -- when taking gradients, these values will become `JVPTracer` objects that I don't know a good way to hash (I was using `jax.numpy.array_str` then).

To add some optimism though: when I bypassed this logic by incorrectly using a list here, I seemed to be able to run the new test that this PR adds, which really would make model construction differentiable for all modifiers. That would mitigate the need from my standpoint to work on #1894, which was my most recent attempt in bypassing the model building logic for the sake of gradient calculations, though it may have other utility.

So: if we can get that logic re-written with collections that don't require hashable elements, then we could be good to go here! :) 

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/constraints.py
+++ b/src/pyhf/constraints.py
@@ -10,6 +10,7 @@ __all__ = ["gaussian_constraint_combined", "poisson_constraint_combined"]
 def __dir__():
     return __all__
 
+
 class gaussian_constraint_combined:
     def __init__(self, pdfconfig, batch_size=None):
         default_backend = pyhf.default_backend
@@ -44,9 +45,7 @@ class gaussian_constraint_combined:
             if not parset.pdf_type == 'normal':
                 continue
 
-            normal_constraint_data.append(
-                default_backend.astensor(thisauxdata)
-            )
+            normal_constraint_data.append(default_backend.astensor(thisauxdata))
 
             # many constraints are defined on a unit gaussian
             # but we reserved the possibility that a paramset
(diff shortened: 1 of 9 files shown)


phinate:
@kratsg should I wait on your suggested refactor of the above? or do you think we could hack in a short-term solution?

kratsg:
@phinate do you mind rebasing and fixing the conflicts in staterror? We had to fix a bug/regression there.

phinate:
I looked at the changes, but the only actual code change looked to be the removal of `self.__staterror_uncrt` -- i've left it in, but can you confirm if that was part of the bug?

kratsg:
> @kratsg I looked at the changes, but the only actual code change looked to be the removal of `self.__staterror_uncrt` -- i've left it in, but can you confirm if that was part of the bug?

it is. remove it.

phinate:
> > @kratsg I looked at the changes, but the only actual code change looked to be the removal of `self.__staterror_uncrt` -- i've left it in, but can you confirm if that was part of the bug?
> 
> it is. remove it.

Done! This bit of code was written by Lukas so I just wanted to double check.

PR #1917, opened 2022-07-06
fix: Populate the schema cache with local `defs.json` again

### LINKED ISSUE #1916: Schema validation crashes when running in an environment without internet access
### Summary

In master and the 0.7.0 release candidate, pyhf operations involving model validation will crash in offline environments with a RefResolutionError. This is a common situation e.g. with worker nodes on HTC clusters.
The bug was introduced after 0.6.3, I think in #1753 where the [pre-loading was dropped](https://github.com/scikit-hep/pyhf/pull/1753/files#diff-01a944844c3739d996c27da33c727473ec48ebcac65f16b4001384bc3ae4e725L48).

### OS / Environment

```console
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="http://cern.ch/linux/"
BUG_REPORT_URL="http://cern.ch/linux/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

### Steps to Reproduce

I don't know a good way to prepare the environment to demonstrate this. 
But the below test exposes the attempt by the RefResolver to resolve the schema id through the https URL, and fails against the release candidate/master, but passes in 0.6.3

```python (paste below)
from functools import partial

import pytest
import jsonschema
import pyhf

def make_asserting_handler(origin):
    def asserting_handler(*args, **kwargs):
        raise AssertionError(
            f'called URL request handler from {origin} with args={args!r}, kwargs={kwargs!r} '
            'when no call should have been needed'
        )

    return asserting_handler


@pytest.fixture
def no_http_jsonschema_ref_resolving(monkeypatch):
    asserting_handler = make_asserting_handler('handlers')
    handlers = {
        'https': asserting_handler,
        'http': asserting_handler,
    }
    WrappedResolver = partial(jsonschema.RefResolver, handlers=handlers)
    monkeypatch.setattr('jsonschema.RefResolver', WrappedResolver, raising=True)

def test_preloaded_cache(
    no_http_jsonschema_ref_resolving,
):
    spec = {
        'channels': [
            {
                'name': 'singlechannel',
                'samples': [
                    {
                        'name': 'signal',
                        'data': [10],
                        'modifiers': [
                            {'name': 'mu', 'type': 'normfactor', 'data': None}
                        ],
                    },
                    {
                        'name': 'background',
                        'data': [20],
                        'modifiers': [
                            {
                                'name': 'uncorr_bkguncrt',
                                'type': 'shapesys',
                                'data': [30],
                            }
                        ],
                    },
                ],
            }
        ]
    }
    try:
        pyhf.schema.validate(spec, 'model.json')
    except AttributeError:
        pyhf.utils.validate(spec, 'model.json')
                                               
```

### File Upload (optional)

_No response_

### Expected Results

I expect schema validation to succeed without crashing even when there is no network access that allows resolving the https schema-ids.

### Actual Results

```console
jsonschema.exceptions.RefResolutionError: HTTPSConnectionPool(host='scikit-hep.org', port=443): Max retries exceeded with url: /pyhf/schemas/1.0.0/defs.json (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x2b2bb8457c40>: Failed to establish a new connection: [Errno 101] Network is unreachable'))
```

### pyhf Version

```console
pyhf, version 0.7.0rc2
```

### Code of Conduct

lhenkelm:
# Description

Resolves #1916 by re-introducing the pre-population of the schema cache that had been done before the re-write of the schema utilities in #1753. Since  `defs.json` is in the cache under its id, `jsonschema.RefResolver` will not attempt to connect to `"https://scikit-hep.org"` to find it, which can crash code using pyhf in environments where the resource is not reachable.

I also tried to include a test to avoid future regression, but I am not happy with it. 
A better/more robust test would be a dedicated run in a sandboxed (i.e. no internet access / relevant ports blocked) environment, e.g. as part of the CI. But I have no clue how to set that up. 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Load `defs.json` into `schema.variables.SCHEMA_CACHE` from the local file.
  - in line with what was done before in PR #1753
* Add pytest-socket as a dependency of the 'test' extra. `pytest_socket.socket_disabled`
  was added in pytest-socket v0.2.0.
   - c.f. https://github.com/miketheman/pytest-socket/releases/tag/0.2.0
* Add a unit test to catch this regression in the future.
* `pyhf.schema` clears the cache to avoid stale schemas (and restores the cache if used as a CM).
* Amend unit tests for `pyhf.schema` to test the correct treatment of the cache.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,6 +40,7 @@ extras_require['test'] = sorted(
             'scrapbook~=0.5.0',
             'jupyter',
             'graphviz',
+            'pytest-socket>=0.2',
         ]
     )
 )
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ extras_require['test'] = sorted(
             'scrapbook~=0.5.0',
             'jupyter',
             'graphviz',
-            'pytest-socket>=0.2',
+            'pytest-socket>=0.2',  # 0.2 adds the socket_disabled fixture, cf. https://github.com/scikit-hep/pyhf/pull/1917#discussion_r951943084
         ]
     )
 )
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ extras_require['test'] = sorted(
             'scrapbook~=0.5.0',
             'jupyter',
             'graphviz',
-            'pytest-socket>=0.2',  # 0.2 adds the socket_disabled fixture, cf. https://github.com/scikit-hep/pyhf/pull/1917#discussion_r951943084
+            'pytest-socket>=0.2.0',  # c.f. PR #1917
         ]
     )
 )
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -620,10 +620,12 @@ def no_sockets(monkeypatch):
 
 @pytest.fixture
 def refresh_pyhf(monkeypatch):
+    global pyhf
     modules_to_clear = [name for name in sys.modules if name.split('.')[0] == 'pyhf']
     for module_name in modules_to_clear:
         monkeypatch.delitem(sys.modules, module_name)
-    importlib.import_module(pyhf.__name__)
+    pyhf = importlib.import_module(pyhf.__name__)
+    return pyhf
 
 
 def test_defs_always_cached(
@@ -639,6 +641,7 @@ def test_defs_always_cached(
     Otherwise using pyhf in contexts where the jsonschema.RefResolver cannot lookup the definition by the schema-id,
     it will crash (e.g. a cluster node without network access).
     """
+    pyhf = refresh_pyhf
     spec = {
         'channels': [
             {
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -609,8 +609,8 @@ def test_defs_always_cached(
     """
     Schema definitions should always be loaded from the local files and cached at first import.
 
-    Otherwise using pyhf in contexts where the jsonschema.RefResolver cannot lookup the definition by the schema-id,
-    it will crash (e.g. a cluster node without network access).
+    Otherwise pyhf will crash in contexts where the jsonschema.RefResolver cannot lookup the definition by the schema-id
+    (e.g. a cluster node without network access).
     """
     pyhf = refresh_pyhf
     spec = {
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1,10 +1,9 @@
 import pyhf
 import pytest
 import json
-import jsonschema
-from functools import partial
 import importlib
 import sys
+from pytest_socket import socket_disabled  # noqa: F401
 
 
 @pytest.mark.parametrize('version', ['1.0.0'])
@@ -593,51 +592,6 @@ def test_patchset_fail(datadir, patchset_file):
         pyhf.schema.validate(patchset, 'patchset.json')
 
 
-def make_asserting_handler(origin):
-    def asserting_handler(*args, **kwargs):
-        raise AssertionError(
-            f'called URL request handler from {origin} with args={args!r}, kwargs={kwargs!r} '
-            'when no call should have been needed'
-        )
-
-    return asserting_handler
-
-
-@pytest.fixture
-def no_http_jsonschema_ref_resolving(monkeypatch):
-    asserting_handler = make_asserting_handler('handlers')
-    handlers = {
-        'https': asserting_handler,
-        'http': asserting_handler,
-    }
-    WrappedResolver = partial(jsonschema.RefResolver, handlers=handlers)
-    monkeypatch.setattr('jsonschema.RefResolver', WrappedResolver, raising=True)
-
-
-@pytest.fixture
-def no_requests(monkeypatch):
-    monkeypatch.delattr('requests.sessions.Session.request', raising=True)
-    monkeypatch.setattr(
-        'requests.get', make_asserting_handler('requests.get'), raising=True
-    )
-
-
-@pytest.fixture
-def no_urllib(monkeypatch):
-    monkeypatch.setattr(
-        'urllib.request.urlopen',
-        make_asserting_handler('urllib.request.urlopen'),
-        raising=True,
-    )
-
-
-@pytest.fixture
-def no_sockets(monkeypatch):
-    monkeypatch.setattr(
-        'socket.socket', make_asserting_handler('socket.socket'), raising=True
-    )
-
-
 @pytest.fixture
 def refresh_pyhf(monkeypatch):
     global pyhf
@@ -649,10 +603,7 @@ def refresh_pyhf(monkeypatch):
 
 
 def test_defs_always_cached(
-    no_http_jsonschema_ref_resolving,  # this should catch the request and raise the error if it happens
-    no_requests,  # future jsonschema code may try to fall back to to explicit/default handlers
-    no_urllib,
-    no_sockets,
+    socket_disabled,  # noqa: F811
     refresh_pyhf,  # ensure there is not a pre-existing cache hiding the issue
 ):
     """
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1,9 +1,11 @@
-import pyhf
-import pytest
-import json
 import importlib
+import json
 import sys
+
+import pytest
 from pytest_socket import socket_disabled  # noqa: F401
+
+import pyhf
 
 
 @pytest.mark.parametrize('version', ['1.0.0'])
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -1,6 +1,10 @@
 import pyhf
 import pytest
 import json
+import jsonschema
+from functools import partial
+import importlib
+import sys
 
 
 @pytest.mark.parametrize('version', ['1.0.0'])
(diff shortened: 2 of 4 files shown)


lhenkelm:
> I'm trying to figure out or remind myself if this was removed to allow for a custom `v1.0.0` to be loaded by the user and override the default `defs.json` that would have been shipped otherwise in this package. I guess I'm wondering if one should even allow this situation or not (probably not).

I don't remember an explicit discussion in that set of issues/PRs. I might have missed it though.
You are ofc. correct that this creates a bit of a trap if one provides a custom `defs.json`, with version `'1.0.0'`, and the same schema-id, then your cache and your set schema will disagree, quite confusingly. 

I guess strictly speaking this breaks some assumptions of what the version and schema-id mean, if someine changes the schema they should change at least one of those too.
But the simplest way of approaching a custom schema (copy the builtins and modify step by step) does lead straight into this trap :(. 

Maybe we could add some logic to the schema-path setter that checks and warns/raises if some of the schemas found at the new path would cause this error? Because as long as the schema-id+version is unique to a definition, the cache should work fine.

lhenkelm:
> For your tests, the easiest way is to monkeypatch any part of the necessary call to `requests` or `urllib` under the hood and force it to return an exception instead -- and you can also use `mocker.spy` to check if it's been called or not. Ideally, a local ref resolver shouldn't have to grab anything via internet connection.

Yeah this is why the tests the PR is currently proposing monkey-patch `requests.get` (the default w/o handlers)  and `urlopen` (the fallback), according to the [jsonschema implementation](https://github.com/python-jsonschema/jsonschema/blob/8819f463d7e01b2f7e3f8ca3a010d38301aa7e69/jsonschema/validators.py#L953).
Also deleting `requests.sessions.Session.request` and patching out `socket.socket` is just an attempt to anticipate any other workarounds. The sockets might be overkill though.

matthewfeickert:
@lhenkelm just to check, is this ready for review? I assume yes, but I just wanted to make sure that you didn't have anything that you wanted to come back and add before it got reviewed and went in.

lhenkelm:
@matthewfeickert having  thought some more about @kratsg's concern about users overriding the default schema using the default version and id, I would like to add a bit of code to at least clear the cache if the schema path/version are changed, to preempt these confusions. 

A second step might be to also merge the `jsonschema.RefResolver`'s cache into the `pyhf` cache at the end of validation, to allow custom schema users to benefit from the caching implicitly. But that can be second PR too (it goes beyond just a fix).

I am a bit busy ATM but I'll get on this this week and ping you once I am done.

kratsg:
I wonder if we can simplify the PR via https://pypi.org/project/pytest-socket/ using the (`socket_disabled`) fixture.

tests/test_schema.py
```diff
@@ -567,3 +590,54 @@ def test_patchset_fail(datadir, patchset_file):
...
+    modules_to_clear = [name for name in sys.modules if name.split('.')[0] == 'pyhf']
+    for module_name in modules_to_clear:
+        monkeypatch.delitem(sys.modules, module_name)
+    pyhf = importlib.import_module(pyhf.__name__)
+    return pyhf
+
+
+def test_defs_always_cached(
+    socket_disabled,  # noqa: F811
+    refresh_pyhf,  # ensure there is not a pre-existing cache hiding the issue
```
kratsg, line 607: I think this could be fixed by monkeypatching the cache directly instead of messing with the importlib.

kratsg (changes requested):

tests/test_schema.py
```diff
@@ -567,3 +590,54 @@ def test_patchset_fail(datadir, patchset_file):
...
+
+
+@pytest.fixture
+def refresh_pyhf(monkeypatch):
+    global pyhf
+    modules_to_clear = [name for name in sys.modules if name.split('.')[0] == 'pyhf']
+    for module_name in modules_to_clear:
+        monkeypatch.delitem(sys.modules, module_name)
+    pyhf = importlib.import_module(pyhf.__name__)
+    return pyhf
```
matthewfeickert, line 602: I like @kratsg's suggestion about not having to try to deal with `importlib` if it can be ignored, but could the equivalent effect be done with just `reload` like https://github.com/scikit-hep/pyhf/blob/97664bcfcef3b4a459a7f700b0703b40aec9b842/tests/test_scripts.py#L620-L625?

setup.py
```diff
@@ -40,6 +40,7 @@
             'scrapbook~=0.5.0',
             'jupyter',
             'graphviz',
+            'pytest-socket>=0.2',
```
matthewfeickert, line 43: Is this lower bound tested, or just a guess?

matthewfeickert (commented):
Thanks for the PR @lhenkelm. I haven't stepped through this deeply, but it looks like you and @kratsg have iterated a bit, so I think if you can work out a solution with @kratsg's suggestion then this looks good!

setup.py
lhenkelm: The fixture used in this PR (`socket_disabled`) was [only added to the API in 0.2](https://github.com/miketheman/pytest-socket/compare/0.1.0...0.2.0). I added a comment to clarify this.

tests/test_schema.py
lhenkelm: I think if we monkeypatch the cache directly, it makes it impossible to catch the regression in the test.
The behavior the test is meant to assert on is the _initial state of the cache_, i.e. the module-level code. This is only executed exactly once at first import. So we need to re-run the import to be able to catch this behavior.

However, framing it like this makes it clear that the module-level code is really the unit under test here. 
So this should be part of the test body, not a fixture. (it is the act in prepare-act-assert)

tests/test_schema.py
lhenkelm: Thanks for suggesting `mock.patch.dict`, it does look a little cleaner now.

However it seems leaving `pyhf` in `sys.modules` and `reload`ing it instead somehow causes the old cache to stick around after the import, so the test fails to catch the regression. This is a little mysterious to me. Maybe it is related to this line in the [docs](https://docs.python.org/3.7/library/importlib.html#importlib.reload):

> When a module is reloaded, its dictionary (containing the module’s global variables) is retained.

I would have expected the imports in `pyhf/__init__.py` to be re-interpreted during `reload`, since those are still missing from `sys.modules` at this point. But it seems not (perhaps bc. their local symbols are part of the retained dictionary for the  `pyhf` module?)

for future reference, this is the patch trying `reload`  that fails to catch the regression:
```diff
--- a/tests/test_schema.py
+++ b/tests/test_schema.py
@@ -611,10 +611,12 @@ def test_defs_always_cached(
     """

     global pyhf
-    modules_to_clear = [name for name in sys.modules if name.split('.')[0] == 'pyhf']
+    modules_to_clear = [
+        nom for nom in sys.modules if (nom.split('.')[0] == 'pyhf') and (nom != 'pyhf')
+    ]
     for module_name in modules_to_clear:
         del monkeypatched_sys_modules[module_name]
-    pyhf = importlib.import_module(pyhf.__name__)
+    pyhf = importlib.reload(pyhf)

     spec = {
         'channels': [
```

tests/test_schema.py
```diff
@@ -593,26 +594,28 @@ def test_patchset_fail(datadir, patchset_file):
...
     Schema definitions should always be loaded from the local files and cached at first import.
 
-    Otherwise using pyhf in contexts where the jsonschema.RefResolver cannot lookup the definition by the schema-id,
-    it will crash (e.g. a cluster node without network access).
+    Otherwise pyhf will crash in contexts where the jsonschema.RefResolver cannot lookup the definition by the schema-id
+    (e.g. a cluster node without network access).
     """
-    pyhf = refresh_pyhf
+
+    global pyhf
```
kratsg, line 613: don't use global

tests/test_schema.py
```diff
@@ -593,26 +594,28 @@ def test_patchset_fail(datadir, patchset_file):
...
+    modules_to_clear = [name for name in sys.modules if name.split('.')[0] == 'pyhf']
+    for module_name in modules_to_clear:
+        del monkeypatched_sys_modules[module_name]
+    pyhf = importlib.import_module(pyhf.__name__)
```
kratsg, line 617: just `importlib.import_module('pyhf')` as we won't be renaming it anytime soon...

kratsg (changes requested):

tests/test_schema.py
kratsg: Fair enough. The test should be the one doing the reimporting in an isolated modules, as you have it. So that looks ok to me (to force the reload).

setup.py
```diff
@@ -40,6 +40,7 @@
             'scrapbook~=0.5.0',
             'jupyter',
             'graphviz',
+            'pytest-socket>=0.2',  # 0.2 adds the socket_disabled fixture, cf. https://github.com/scikit-hep/pyhf/pull/1917#discussion_r951943084
```
matthewfeickert, line 43:
```suggestion
            'pytest-socket>=0.2',  # c.f. PR #1917
```

We can just add this information RE: the API in the commit message

kratsg (approved):

matthewfeickert (approved):
LGTM. :+1: Thanks very much for your contribution @lhenkelm!

PR #1919, opened 2022-07-17
fix: Manually set uncertainties for fixed parameters to 0

### LINKED ISSUE #1918: CI failing for `iminuit` `v2.12.2`
### Summary

On 2022-07-15 [`iminuit` `v2.12.2`](https://github.com/scikit-hep/iminuit/releases/tag/v2.12.2) was released to "fix bug in error heuristic and prevent assigning non-positive values". With this release the CI is failing for `tests/test_infer.py`, `tests/test_otim.py`, `tests/test_jit.py`, `tests/test_public_api.py`, `tests/test_scripts.py`, and `tests/test_validation.py`

### OS / Environment

```console
All
```

### Steps to Reproduce

Run the test suite in CI

### File Upload (optional)

_No response_

### Expected Results

CI to pass

### Actual Results

```console
All the above tests fail.
```

### pyhf Version

```console
`HEAD` which is currently 8e242c9c829963c714f9b10acd28494f004a2452
```

### Code of Conduct

matthewfeickert:
# Description

Resolves #1918

* Manually set the uncertainties for fixed parameters to zero after minimization, and remove setting the uncertainty/step_sizes during minimization, to avoid warnings from `iminuit`.
For `iminuit` `v2.12.2+` "assigning zero to errors is invalid, but in the past this was
not caught." This approach harmonizes with `cabinetry` for fixed parameters for `iminuit`
`v2.12.2+`.
   - c.f. https://github.com/scikit-hep/iminuit/issues/762
   - c.f. https://github.com/scikit-hep/cabinetry/pull/346
   - c.f. https://iminuit.readthedocs.io/en/stable/changelog.html#july-15-2022
     > fix a bug in error heuristic when parameters have negative values and prevent
     > assigning negative values to errors
* Remove `tests/test_optim.py` `test_step_sizes_fixed_parameters_minuit` as the uncertainties/step sizes values are no longer set during minimization and so are no longer in pyhf's control.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Manually set the uncertainties for fixed parameters to zero after minimization,
and remove setting the uncertainty/step_sizes during minimization, to avoid warnings
from iminuit. For iminuit v2.12.2+ "assigning zero to errors is invalid, but in the
past this was not caught." This approach harmonizes with cabinetry for fixed
parameters for iminuit v2.12.2+.
   - c.f. https://github.com/scikit-hep/iminuit/issues/762
   - c.f. https://github.com/scikit-hep/cabinetry/pull/346
   - c.f. https://iminuit.readthedocs.io/en/stable/changelog.html#july-15-2022
     > fix a bug in error heuristic when parameters have negative values and prevent
     > assigning negative values to errors
* Remove tests/test_optim.py test_step_sizes_fixed_parameters_minuit as the
uncertainties/step sizes values are no longer set during minimization and so are no
longer in pyhf's control.
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,6 +89,7 @@ filterwarnings = [
     'ignore:Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with:UserWarning',  #FIXME: tests/test_optim.py::test_minimize[no_grad-scipy-pytorch-no_stitch]
     'ignore:divide by zero encountered in (true_)?divide:RuntimeWarning',  #FIXME: pytest tests/test_tensor.py::test_pdf_calculations[numpy]
     'ignore:[A-Z]+ is deprecated and will be removed in Pillow 10:DeprecationWarning',  # keras
+    'ignore:Assigned errors must be positive. Non-positive values are replaced by a heuristic.:UserWarning',  # iminuit
 ]
 
 [tool.nbqa.mutate]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,7 +89,6 @@ filterwarnings = [
     'ignore:Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with:UserWarning',  #FIXME: tests/test_optim.py::test_minimize[no_grad-scipy-pytorch-no_stitch]
     'ignore:divide by zero encountered in (true_)?divide:RuntimeWarning',  #FIXME: pytest tests/test_tensor.py::test_pdf_calculations[numpy]
     'ignore:[A-Z]+ is deprecated and will be removed in Pillow 10:DeprecationWarning',  # keras
-    'ignore:Assigned errors must be positive. Non-positive values are replaced by a heuristic.:UserWarning',  # iminuit
 ]
 
 [tool.nbqa.mutate]
--- a/src/pyhf/optimize/mixins.py
+++ b/src/pyhf/optimize/mixins.py
@@ -1,9 +1,11 @@
 """Helper Classes for use of automatic differentiation."""
-from pyhf.tensor.manager import get_backend
+import logging
+
+import numpy as np
+
 from pyhf import exceptions
 from pyhf.optimize.common import shim
-
-import logging
+from pyhf.tensor.manager import get_backend
 
 log = logging.getLogger(__name__)
 
@@ -82,6 +84,13 @@ class OptimizerMixin:
         if uncertainties is not None:
             # extract number of fixed parameters
             num_fixed_pars = len(fitted_pars) - len(fitresult.x)
+
+            # FIXME: Set uncertainties for fixed parameters to 0 manually
+            # https://github.com/scikit-hep/iminuit/issues/762
+            # https://github.com/scikit-hep/pyhf/issues/1918
+            # https://github.com/scikit-hep/cabinetry/pull/346
+            uncertainties = np.where(fitresult.minuit.fixed, 0.0, uncertainties)
+
             # stitch in zero-uncertainty for fixed values
             uncertainties = stitch_pars(
                 tensorlib.astensor(uncertainties),
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -376,7 +376,7 @@ def test_optim_uncerts(backend, source, spec, mu):
         return_uncertainties=True,
     )
     assert result.shape == (2, 2)
-    assert pytest.approx([0.26418431, 0.1]) == pyhf.tensorlib.tolist(result[:, 1])
+    assert pytest.approx([0.26418431, 0.01]) == pyhf.tensorlib.tolist(result[:, 1])
 
 
 @pytest.mark.parametrize('mu', [1.0], ids=['mu=1'])
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -376,7 +376,7 @@ def test_optim_uncerts(backend, source, spec, mu):
         return_uncertainties=True,
     )
     assert result.shape == (2, 2)
-    assert pytest.approx([0.26418431, 0.0]) == pyhf.tensorlib.tolist(result[:, 1])
+    assert pytest.approx([0.26418431, 0.1]) == pyhf.tensorlib.tolist(result[:, 1])
 
 
 @pytest.mark.parametrize('mu', [1.0], ids=['mu=1'])
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -376,7 +376,7 @@ def test_optim_uncerts(backend, source, spec, mu):
         return_uncertainties=True,
     )
     assert result.shape == (2, 2)
-    assert pytest.approx([0.26418431, 0.01]) == pyhf.tensorlib.tolist(result[:, 1])
+    assert pytest.approx([0.26418431, 0.0]) == pyhf.tensorlib.tolist(result[:, 1])
 
 
 @pytest.mark.parametrize('mu', [1.0], ids=['mu=1'])
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -520,22 +520,6 @@ def test_init_pars_sync_fixed_values_minuit(mocker):
     assert minimizer.fixed == [True, False, False]
 
 
-def test_step_sizes_fixed_parameters_minuit(mocker):
-    opt = pyhf.optimize.minuit_optimizer()
-
-    # patch all we need
-    from pyhf.optimize import opt_minuit
-
-    minuit = mocker.patch.object(getattr(opt_minuit, 'iminuit'), 'Minuit')
-    minimizer = opt._get_minimizer(None, [9, 9, 9], [(0, 10)] * 3, fixed_vals=[(0, 1)])
-
-    assert minuit.called
-    assert minimizer.fixed == [True, False, False]
-    # iminuit v2.12.2+ sets non-zero step sizes for all parameters including fixed parameters
-    # c.f. https://github.com/scikit-hep/iminuit/issues/762
-    assert minimizer.errors == [0.01, 0.01, 0.01]
-
-
 def test_solver_options_behavior_scipy(mocker):
     opt = pyhf.optimize.scipy_optimizer(solver_options={'arbitrary_option': 'foobar'})
 
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -531,7 +531,7 @@ def test_step_sizes_fixed_parameters_minuit(mocker):
 
     assert minuit.called
     assert minimizer.fixed == [True, False, False]
-    # iminuit v2.12.1+ sets non-zero step sizes for all parameters including fixed parameters
+    # iminuit v2.12.2+ sets non-zero step sizes for all parameters including fixed parameters
     # c.f. https://github.com/scikit-hep/iminuit/issues/762
     assert minimizer.errors == [0.01, 0.01, 0.01]
 
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -531,7 +531,9 @@ def test_step_sizes_fixed_parameters_minuit(mocker):
 
     assert minuit.called
     assert minimizer.fixed == [True, False, False]
-    assert minimizer.errors == [0.0, 0.01, 0.01]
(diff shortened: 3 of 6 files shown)


matthewfeickert:
This just needs to get fixed in `pytest`'s comparison. We can survive with the iminuit difference.

matthewfeickert:
@kratsg @lukasheinrich I'm going to leave this in draft until I have enough time to look at this and https://github.com/scikit-hep/iminuit/issues/762 and determine if the change in uncertainties is desired and create some minimal examples that demonstrate the differences.

lukasheinrich (approved):

pyproject.toml
```diff
@@ -89,6 +89,7 @@ filterwarnings = [
     'ignore:Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with:UserWarning',  #FIXME: tests/test_optim.py::test_minimize[no_grad-scipy-pytorch-no_stitch]
     'ignore:divide by zero encountered in (true_)?divide:RuntimeWarning',  #FIXME: pytest tests/test_tensor.py::test_pdf_calculations[numpy]
     'ignore:[A-Z]+ is deprecated and will be removed in Pillow 10:DeprecationWarning',  # keras
+    'ignore:Assigned errors must be positive. Non-positive values are replaced by a heuristic.:UserWarning',  # iminuit
```
alexander-held, line 92: Removing this line https://github.com/scikit-hep/pyhf/blob/2cd8c0be20b6004a3fdb5305afd13d970589f547/src/pyhf/optimize/opt_minuit.py#L60 should fix this warning, and the line no longer does anything useful in new `iminuit` versions. If you keep the line, users will see different errors depending on what version of `iminuit` they have installed so that might not be desirable.

alexander-held (commented):
Based on https://github.com/scikit-hep/iminuit/issues/762#issuecomment-1193413707, https://github.com/scikit-hep/pyhf/blob/2cd8c0be20b6004a3fdb5305afd13d970589f547/src/pyhf/optimize/opt_minuit.py#L53 could also in principle be removed I think (does not need to happen here, but came up in the same context, and presumably the effect is negligible).

I am not yet sure how to handle this in `cabinetry` but I am leaning towards going through the fit results and manually setting errors for fixed parameters to `0.0` again. I will put my motivation for that into https://github.com/scikit-hep/iminuit/issues/762. It would be best to harmonize the behavior of `pyhf` and `cabinetry` though so I'm also interested to hear your thoughts on this so we can ideally agree on the approach.

pyproject.toml
matthewfeickert: Yeah @alexander-held we should for sure reach a common agreement here as a way to move forward. I think the iminuit Issue that you opened is a good place to have some discussion and come to an agreement in general but we can also follow up on Slack. 👍 (I'm typing this on my phone while on holiday, so everything in my side will be a bit delayed.)

matthewfeickert: @alexander-held I haven't had time to step through everything in the Issue (as I'm on holiday) but it looks like regardless of the discussion results will probably be needing to update the iminuit lower bound (as any revised behavior will be in future releases). Do you have any concerns about merging this PR now and then having an issue to follow this discussion and plan the pyhf and cabinetry changes?

@lukasheinrich @kratsg if you have time to follow up this week (I know it is busy for both of you) that would be helpful! If not I will follow up on 2022-08-08.

alexander-held: I don't have anything against merging this to get the CI fixed, but I think it would be useful to agree on an approach before the next release. If you do want to merge this, I think it makes sense to remove the line I pointed out above (which will do nothing anymore as the setting will get overridden inside `iminuit`). That allows you to also skip the ignore for this new warning (which will no longer appear).

What I did in `cabinetry` (https://github.com/scikit-hep/cabinetry/pull/346) is to set the post-fit errors manually to zero again for fixed parameters. That recovers the old behavior no matter what the `iminuit` version is (no need to change the pin in that case). The advantage is that the behavior is unchanged for users, the disadvantage is that it diverges from what pure `iminuit` would do, and if this is merged as-is, it will also disagree with what `pyhf` will do by extension.

matthewfeickert: > I don't have anything against merging this to get the CI fixed, but I think it would be useful to agree on an approach before the next release. If you do want to merge this, I think it makes sense to remove the line I pointed out above (which will do nothing anymore as the setting will get overridden inside `iminuit`). That allows you to also skip the ignore for this new warning (which will no longer appear).

Okay that is now removed, and the changed test value updated as well.

> What I did in `cabinetry` ([scikit-hep/cabinetry#346](https://github.com/scikit-hep/cabinetry/pull/346)) is to set the post-fit errors manually to zero again for fixed parameters. That recovers the old behavior no matter what the `iminuit` version is (no need to change the pin in that case). The advantage is that the behavior is unchanged for users, the disadvantage is that it diverges from what pure `iminuit` would do, and if this is merged as-is, it will also disagree with what `pyhf` will do by extension.

As can now be seen in the failing `test_step_sizes_fixed_parameters_minuit` tests now, we'll have to rectify this in general. There may be a more obvious solution that I'm missing, but I think we'll need to do a similar approach of setting the post-fit errors manually to zero in

https://github.com/scikit-hep/pyhf/blob/2cd8c0be20b6004a3fdb5305afd13d970589f547/src/pyhf/optimize/mixins.py#L68

I haven't thought much about this as this is my first day back though.

matthewfeickert: @kratsg @lukasheinrich If we follow what `cabinetry` is doing here then the test

https://github.com/scikit-hep/pyhf/blob/2cd8c0be20b6004a3fdb5305afd13d970589f547/tests/test_optim.py#L523-L534

breaks and indeed it doesn't make much sense to test this specifically (or at least the same way) as we're now overriding the `iminuit` behavior for step sizes for fixed parameters (`minimizer.errors`). @kratsg you wrote this test, so do you have any input on how you would like this to be rewritten?

matthewfeickert:
Note to self in case needed in the future: Previous commit message summary was

> * Update lower bound on `iminuit` to `v2.12.2` as it is the 'oldest' SemVer release to ensure uncertainties are positive.
>    - c.f. https://iminuit.readthedocs.io/en/stable/changelog.html#july-15-2022
>      > fix a bug in error heuristic when parameters have negative values and prevent assigning negative values to errors
> * Update the tests/test_optim.py test_optim_uncerts uncertainties values to be nonzero as "assigning zero to errors is invalid, but in the past this was not caught."
>    - c.f. https://github.com/scikit-hep/iminuit/issues/762
> * Update [`tests/constraints.txt`](https://github.com/scikit-hep/pyhf/blob/2cd8c0be20b6004a3fdb5305afd13d970589f547/tests/constraints.txt) to use `iminuit==2.12.2`.
> * Add a ignore to `filterwarnings` to avoid iminuit UserWarning:
> 
> > UserWarning: Assigned errors must be positive. Non-positive values are replaced by a heuristic.
> 
> This was introduced in `iminuit` `v2.12.2` as the result of a step size guess heuristic being applied to fixed parameters. This results in the error of fixed parameters changing during MIGRAD.
>    - c.f. https://github.com/scikit-hep/iminuit/issues/762

tests/test_optim.py
```diff
@@ -531,7 +531,9 @@ def test_step_sizes_fixed_parameters_minuit(mocker):
 
     assert minuit.called
     assert minimizer.fixed == [True, False, False]
-    assert minimizer.errors == [0.0, 0.01, 0.01]
+    # iminuit v2.12.2+ sets non-zero step sizes for all parameters including fixed parameters
+    # c.f. https://github.com/scikit-hep/iminuit/issues/762
+    assert minimizer.errors == [0.01, 0.01, 0.01]
```
matthewfeickert, line 536: @lukasheinrich @kratsg I'm for now sidestepping this by just accepting that `iminuit` has its step size behavior that we need to deal with and that the minimizer step size and the fit result uncertainties might not match. I'm also setting the `v2.12.2` lower bound to make it so this isn't confusing between versions. Let me know if you find this unacceptable.

matthewfeickert: Though maybe given @alexander-held's comment https://github.com/scikit-hep/pyhf/pull/1919#pullrequestreview-1053689055 it would be better to just say that `iminuit` will control the step size and not have this test?

alexander-held: I think this test can be removed, it previously checked that the zero values were set properly for fixed parameters, but now this test no longer checks any `pyhf` behavior (only `iminuit` defaults).

alexander-held (commented):
I think the current implementation allows for leaving the `iminuit` version requirement unchanged. The change in 2.12.2 affects parameters with an assigned step size of zero. Since `pyhf` no longer sets a step size of zero to any parameters (and instead corrects the post-fit parameter uncertainties for fixed parameters now), `iminuit` versions before and after 2.12.2 should behave the same in combination with these changes to `pyhf`.

The one thing that might break, but for which I do not see a good solution, is user code that directly uses the Minuit instance, reads out `.errors` on that and gets a different behavior starting with 2.12.2. I don't think `pyhf` can (or should) do anything about that though.

matthewfeickert:
> I think the current implementation allows for leaving the `iminuit` version requirement unchanged. The change in 2.12.2 affects parameters with an assigned step size of zero. Since `pyhf` no longer sets a step size of zero to any parameters (and instead corrects the post-fit parameter uncertainties for fixed parameters now), `iminuit` versions before and after 2.12.2 should behave the same in combination with these changes to `pyhf`.
> 
> The one thing that might break, but for which I do not see a good solution, is user code that directly uses the Minuit instance, reads out `.errors` on that and gets a different behavior starting with 2.12.2. I don't think `pyhf` can (or should) do anything about that though.

Yeah the only reason I included the version bump was to ensure that there could only be one behavior if the user tried to access `Minuit.errors`, but I think I agree that I'm trying to be too protective of the user in this instance and if they are running into this that we should just direct them to Issue https://github.com/scikit-hep/iminuit/issues/762.

alexander-held:
> Yeah the only reason I included the version bump was to ensure that there could only be one behavior if the user tried to access `Minuit.errors`

The behavior should still be the same with old and new versions, i.e. non-zero `.errors` for fixed parameters (since you're no longer setting the zero value before the fit). A change appears though for someone installing e.g. `pyhf` 0.6.3 with old and new versions of `iminuit` (I think that change cannot be prevented from the side of `pyhf` though).

alexander-held (approved):

matthewfeickert:
>  A change appears though for someone installing e.g. `pyhf` 0.6.3 with old and new versions of `iminuit` (I think that change cannot be prevented from the side of `pyhf` though).

Exactly. Though this [not being able to force users to install newer versions software] is what you, Henry, Hans, and Thomas have all reminded me about many times and that I have to let users make their own decisions.

src/pyhf/optimize/mixins.py
```diff
@@ -82,6 +84,13 @@ def _internal_postprocess(self, fitresult, stitch_pars, return_uncertainties=Fal
         if uncertainties is not None:
             # extract number of fixed parameters
             num_fixed_pars = len(fitted_pars) - len(fitresult.x)
+
+            # FIXME: Set uncertainties for fixed parameters to 0 manually
+            # https://github.com/scikit-hep/iminuit/issues/762
+            # https://github.com/scikit-hep/pyhf/issues/1918
+            # https://github.com/scikit-hep/cabinetry/pull/346
+            uncertainties = np.where(fitresult.minuit.fixed, 0.0, uncertainties)
```
kratsg, line 92: check this behavior against all the backends.

matthewfeickert: @kratsg It is already. It is fine to use NumPy operations here (and the reason I'm doing so it so avoid expensive tensor casts for small arrays when it isn't needed) as `iminuit` returns NumPy-like containers for `fitresult.minuit.fixed` and `uncertainties`. `uncertainties` is then cast into a tensor when needed a few lines further down when the sticthing is done.

kratsg: Ok, that was not obvious. I thought minuit returned tensors that you fed in rather than always numpy-like.

kratsg (approved):

src/pyhf/optimize/mixins.py
matthewfeickert: Yeah, sorry. I agree it isn't obvious as I first did tensor casts myself and then I looked at what @alexander-held was doing in https://github.com/scikit-hep/cabinetry/pull/346 and then realized I could do the same here.

PR #1933, opened 2022-08-10
fix: Handle importing XML measurements with no POI attribute

kratsg:
# Description

See #1909.

Add runtime check for missing POI in XML measurements. XML measurements are not necessarily required to have a POI, but we require it, so need to fix coverage.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add runtime check for missing POI in XML measurements. XML measurements are not
necessarily required to have a POI, but we require it, so need to fix coverage.
* Add test and test XML files for missing POI in XML configs.
```


matthewfeickert (approved):
LGTM. :+1: Thanks @kratsg!

PR #1934, opened 2022-08-11
feat: Enable typehints in pre-commit and add typehints to readxml

kratsg:
# Description

This will need to be rebased again on top of `master` once #1909 is merged in.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add mypy to pre-commit.
* Add typehints to readxml.
* Enable typing for the pyhf package.
* Add nearly all other files to mypy.overrides.
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -59,6 +59,13 @@ repos:
     - id: flake8
       args: ["--count", "--statistics"]
 
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.961
+    hooks:
+      - id: mypy
+        files: src
+        args: []
+
 -   repo: https://github.com/nbQA-dev/nbQA
     rev: 1.4.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,7 +60,7 @@ repos:
       args: ["--count", "--statistics"]
 
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.961
+    rev: v0.971
     hooks:
       - id: mypy
         files: src
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ install_requires =
     jsonpatch>=1.15
     pyyaml>=5.1  # for parsing CLI equal-delimited options
     importlib_resources>=1.3.0; python_version < "3.9"  # for resources in schema
-    typing_extensions; python_version == "3.7"  # for typing.TypedDict
+    typing_extensions; python_version == "3.7"  # for TypedDict, Literal
 
 [options.packages.find]
 where = src
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,6 +42,7 @@ install_requires =
     jsonpatch>=1.15
     pyyaml>=5.1  # for parsing CLI equal-delimited options
     importlib_resources>=1.3.0; python_version < "3.9"  # for resources in schema
+    typing_extensions; python_version == "3.7"  # for typing.TypedDict
 
 [options.packages.find]
 where = src
--- a/src/pyhf/readxml.py
+++ b/src/pyhf/readxml.py
@@ -2,7 +2,17 @@ from __future__ import annotations
 
 import logging
 import os
-from typing import TYPE_CHECKING, Callable, Iterable, Tuple, Union, IO, Literal
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    Iterable,
+    Tuple,
+    Union,
+    IO,
+    Literal,
+    Dict,
+    Set,
+)
 from typing_extensions import TypedDict  # for python 3.7 only (3.8+ has T.TypedDict)
 
 import xml.etree.ElementTree as ET
@@ -11,6 +21,7 @@ from pathlib import Path
 import numpy as np
 import tqdm
 import uproot
+import sys
 
 from pyhf import compat
 from pyhf import exceptions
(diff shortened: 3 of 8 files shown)


.pre-commit-config.yaml
```diff
@@ -59,6 +59,13 @@ repos:
     - id: flake8
       args: ["--count", "--statistics"]
 
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.961
```
matthewfeickert, line 63:
```suggestion
    rev: v0.971
```

As [`v0.971`](https://github.com/pre-commit/mirrors-mypy/releases/tag/v0.971) is out now is there a reason to not use it? If so, then we'll need to manually ignore it when pre-commit tries to autoupdate.

.pre-commit-config.yaml
```diff
@@ -59,6 +59,13 @@ repos:
...
+    hooks:
+      - id: mypy
+        files: src
+        args: []
```
matthewfeickert, line 67: Uniformed question: Why is the `[]` needed for `args`? If it isn't can we drop it to avoid confusion in the future?

.pre-commit-config.yaml
```diff
@@ -59,6 +59,13 @@ repos:
     - id: flake8
       args: ["--count", "--statistics"]
 
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v0.961
+    hooks:
+      - id: mypy
+        files: src
```
matthewfeickert, line 66: In `pyproject.toml` you're adding

```
[tool.mypy]
files = "src"
```

so why do we need to specify the files here in `pre-commit`. It would be good to avoid duplicating this information to avoid potential future conflicts.

setup.cfg
```diff
@@ -42,6 +42,7 @@ install_requires =
     jsonpatch>=1.15
     pyyaml>=5.1  # for parsing CLI equal-delimited options
     importlib_resources>=1.3.0; python_version < "3.9"  # for resources in schema
+    typing_extensions; python_version == "3.7"  # for TypedDict, Literal
```
matthewfeickert, line 45: These are strictly just for type checking right, and not runtime dependencies, so they don't _need_ to be added to [`tests/constraints.txt`](https://github.com/scikit-hep/pyhf/blob/97664bcfcef3b4a459a7f700b0703b40aec9b842/tests/constraints.txt), correct?

src/pyhf/readxml.py
```diff
@@ -1,8 +1,20 @@
...
+    Tuple,
+    IO,
+    List,
+    Set,
+    Union,
+    Sequence,
+    MutableMapping,
+    MutableSequence,
+    cast,
+)
```
matthewfeickert, line 16: This is just a suggestion to avoid more import movement noise if we end up eventually turning on `isort` in `pre-commit`. Feel free to ignore though.

```suggestion
from typing import (
    IO,
    Callable,
    Iterable,
    List,
    MutableMapping,
    MutableSequence,
    Sequence,
    Set,
    Tuple,
    Union,
    cast,
)
```

src/pyhf/readxml.py
```diff
@@ -13,16 +25,32 @@
...
+    StatError,
+    ShapeSys,
+    ShapeFactor,
+    Measurement,
+    ParameterBase,
+    Channel,
+    Observation,
+    Workspace,
+    PathOrStr,
+)
```
matthewfeickert, line 45: This is just a suggestion to avoid more import movement noise if we end up eventually turning on `isort` in `pre-commit`. Feel free to ignore though.

```suggestion
from pyhf.typing import (
    Channel,
    HistoSys,
    LumiSys,
    Measurement,
    Modifier,
    NormFactor,
    NormSys,
    Observation,
    Parameter,
    ParameterBase,
    PathOrStr,
    Sample,
    ShapeFactor,
    ShapeSys,
    StatError,
    Workspace,
)
```

src/pyhf/typing.py
```diff
@@ -0,0 +1,137 @@
+import os
+import sys
+
+from typing import TYPE_CHECKING, Union, Sequence, MutableSequence
+
+if sys.version_info >= (3, 8):
+    from typing import TypedDict, Literal
+else:
+    from typing_extensions import TypedDict, Literal
```
matthewfeickert, line 9: This is just a suggestion to avoid more import movement noise if we end up eventually turning on `isort` in `pre-commit`. Feel free to ignore though.

```suggestion
from typing import TYPE_CHECKING, MutableSequence, Sequence, Union

if sys.version_info >= (3, 8):
    from typing import Literal, TypedDict
else:
    from typing_extensions import Literal, TypedDict
```

matthewfeickert (commented):
Thanks for this nice work moving us towards typehints @kratsg! This looks good overall, I just have a few questions on `pre-commit` and then some suggestions (that can be ignored it you don't like them) that are simply reordering of imports to mitigate the amount of noise if we eventually turn on `isort` in `pre-commit`.

.pre-commit-config.yaml
kratsg: next PR is going to flesh it out to fix up missing typehints for imports. It'll be needed.

.pre-commit-config.yaml
kratsg: This is not duplicating information (confusingly) and there's some awkward interplay between how pre-commit finds things and how mypy finds things. pre-commit is using `files` to determine when to call `mypy` and pass it in, but `mypy` ignores whatever `pre-commit` gives it (it seems) and just uses that. You're going to have to live with it.

setup.cfg
kratsg: It's needed for runtime checks (as the import is evaluated at runtime and `typing` is missing some stuff in 3.7)

.pre-commit-config.yaml
matthewfeickert: Lol. Amazing. Thanks for the summary though!

setup.cfg
matthewfeickert: Okay then I'll later go in and figure out what is the minimum version required (some super late at night PR work for the future).

matthewfeickert (approved):

PR #1937, opened 2022-08-12
test: Ignore TYPE_CHECKING for coverage

kratsg:
# Pull Request Description

As introduced in #1909, we had a drop in coverage due to the `TYPE_CHECKING` blocks. We'll ignore it to fix coverage as those blocks aren't meant to be seen during runtime.

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR
```
* Add .coveragerc file with configuration to ignore TYPE_CHECKING.
  - ref: https://coverage.readthedocs.io/en/6.4.3/excluding.html#advanced-exclusion
```

PR #1942, opened 2022-08-12
refactor: Use string literal types of generic builtins for type checking

### LINKED ISSUE #1941: Use string form for type checking until Python 3.7 dropped
What the if here? The string form is fine until you drop Pythons that don’t support it.

_Originally posted by @henryiii in https://github.com/scikit-hep/pyhf/pull/1909#discussion_r944456765_

This was in reference to 

```python
if T.TYPE_CHECKING:
    PathOrStr = T.Union[str, os.PathLike[str]]
else:
    PathOrStr = T.Union[str, "os.PathLike[str]"]
```

in PR #190 now in 

https://github.com/scikit-hep/pyhf/blob/ad1dd86f1d7c1bcbf737805b6821e07c4ef75fca/src/pyhf/typing.py#L30-L33

So until Python 3.7 is dropped (I think this is the version I am not very up to date with my type checking knowledge) we could instead just drop the `if TYPE_CHECKING` and use

```python
#TODO: Switch to os.PathLike[str] once Python 3.7 dropped
PathOrStr = Union[str, "os.PathLike[str]"]
```

This would also allow for reverting PR #1937.

matthewfeickert:
# Description

Resolves #1941 

* Use [string literal types](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#string-literal-types) of generic builtins for type checking until Python 3.8 support is dropped (so pyhf is Python 3.9+ only) to avoid having to check for `if typing.TYPE_CHECKING`.
   - c.f. https://mypy.readthedocs.io/en/stable/runtime_troubles.html#string-literal-types
   - c.f. https://mypy.readthedocs.io/en/stable/builtin_types.html#generic-types
   - c.f. https://mypy.readthedocs.io/en/stable/runtime_troubles.html#generic-builtins
   - c.f. https://peps.python.org/pep-0585/
* Remove `.coveragerc` as no longer needed.
   - Reverts PR #1937

Thanks @henryiii for the nudge on this.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use string literal types of generic builtins for type checking until Python
  3.8 support is dropped (so pyhf is Python 3.9+ only) to avoid having to
  check for `if typing.TYPE_CHECKING`.
   - c.f. https://mypy.readthedocs.io/en/stable/runtime_troubles.html#string-literal-types
   - c.f. https://mypy.readthedocs.io/en/stable/builtin_types.html#generic-types
   - c.f. https://mypy.readthedocs.io/en/stable/runtime_troubles.html#generic-builtins
   - c.f. https://peps.python.org/pep-0585/
* Remove .coveragerc as no longer needed.
   - Reverts PR #1937
```

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,0 @@
-[report]
-exclude_lines =
-    if TYPE_CHECKING:
--- a/src/pyhf/typing.py
+++ b/src/pyhf/typing.py
@@ -27,7 +27,7 @@ __all__ = (
     "Workspace",
 )
 
-# TODO: Switch to os.PathLike[str] once Python 3.7 dropped
+# TODO: Switch to os.PathLike[str] once Python 3.8 support dropped
 PathOrStr = Union[str, "os.PathLike[str]"]
 
 
--- a/src/pyhf/typing.py
+++ b/src/pyhf/typing.py
@@ -1,6 +1,6 @@
 import os
 import sys
-from typing import TYPE_CHECKING, MutableSequence, Sequence, Union
+from typing import MutableSequence, Sequence, Union
 
 if sys.version_info >= (3, 8):
     from typing import Literal, TypedDict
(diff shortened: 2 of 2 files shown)


matthewfeickert:
I didn't do a good job of describing the details in the PR body, so we should make this more clear before merging.

kratsg (approved):

Issue #1907, opened 2022-06-29
Add pre-release support to publish workflow


matthewfeickert:
Creating the GitHub pre-release [`v0.7.0rc1`](https://github.com/scikit-hep/pyhf/releases/tag/v0.7.0rc1) should have triggered the publish to PyPI workflow, but it didn't. I don't think that there is a `pre_release` `github.event_name`

https://github.com/scikit-hep/pyhf/blob/acde7f4ff8d0db2351f5d6e31ff5584e34da0cf0/.github/workflows/publish-package.yml#L104-L105

so we'll need to figure out how release candidate pre-releases can get integrated in.

matthewfeickert:
So seems that I need to learn the `github.event` API better as there is apparently a [`github.event.release.prerelease`](https://github.community/t/release-prerelease-action-triggers/17275). So I now don't know how `github.event_name` and `github.event.release`'s attributes differ.

matthewfeickert:
I think https://github.com/orgs/community/discussions/26281#discussioncomment-3251177 is how to do this.

PR #1705, opened 2021-11-18
docs: Add v0.7.0 release notes

matthewfeickert:
# Description

Add release notes for pyhf v0.7.0.

Relevant RTD build: https://pyhf--1705.org.readthedocs.build/en/1705/release-notes.html#release-v0-7-0

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Add release notes for pyhf v0.7.0.
* Add Graeme Watt to contributors list.
```

PR #1965, opened 2022-08-27
fix: Keep staterror from affecting multiple samples

### LINKED ISSUE #1720: Large cls value difference between v0.6.3 and master
### Summary

When I try to run

```console
$ pyhf cls --test-poi 0.16 signal.json
```
(file provided to developers) with pyhf, the result depends on the version. For versions before https://github.com/scikit-hep/pyhf/pull/1639, the CLS_exp is about 50% lower than for versions on or after that commit. Testing revealed that this is likely due to the handling of staterrors, as pruning these removed the discrepancy.

### OS / Environment

```console
cat /etc/os-release
NAME="Scientific Linux"
VERSION="7.9 (Nitrogen)"
ID="scientific"
ID_LIKE="rhel centos fedora"
VERSION_ID="7.9"
PRETTY_NAME="Scientific Linux 7.9 (Nitrogen)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:scientificlinux:scientificlinux:7.9:GA"
HOME_URL="http://www.scientificlinux.org//"
BUG_REPORT_URL="mailto:scientific-linux-devel@listserv.fnal.gov"

REDHAT_BUGZILLA_PRODUCT="Scientific Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.9
REDHAT_SUPPORT_PRODUCT="Scientific Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.9"
```

### Steps to Reproduce

```console (paste below)
pip install --ignore-installed -U "git+https://github.com/scikit-hep/pyhf.git@9fbbbf9740175fcdfe73fedc4351953be18d2664"
pyhf cls --test-poi 0.16 signal.json
```

differs from

```console
pip install --ignore-installed -U "git+https://github.com/scikit-hep/pyhf.git@5ea4e0a7751aa446c286b9697e01a9439622ebb5
pyhf cls --test-poi 0.16 signal.json
```

### File Upload (optional)

pyhf devs sent internal file.

### Expected Results

I expect the behavior to only change slightly between pyhf versions.

### Actual Results

```console
The behavior changes up to 50%.
```

### pyhf Version

```console
pyhf, version 0.6.4.dev33
pyhf, version 0.6.4.dev34
```

### Code of Conduct

### LINKED ISSUE #1945: staterror regressions blocking v0.7.0
Before `v0.7.0rc2` can go out and then `v0.7.0` we have to fix the regressions in `staterror`. This Issue is just to keep track of them and help us squash them all.

* [ ] Issue #1720
* [x] Issue #1944

lukasheinrich:
# Pull Request Description

* Resolves #1720 
* Resolves #1945

This PR is supposed to fix the issue w/ staterrors which erroneously affected multiple samples. 

- [x] Would be great to get add. test coverage and cross-checks w/ real world models

@alexander-held @matthewfeickert @kratsg 

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Fix staterror bug introduced after v0.6.3 where staterror erroneously affected
  multiple samples.
   - Amends PR #1639 which introduced regression.
   - c.f. Issue #1720
* Remove staterror_combined.__staterror_uncrt as it is unused by the codebase.
* Add test to verify that staterror does not affect more samples than shapesys equivalently.

```

### DIFF
--- a/src/pyhf/modifiers/staterror.py
+++ b/src/pyhf/modifiers/staterror.py
@@ -100,8 +100,12 @@ class staterror_builder:
                 ],
                 axis=0,
             )
+            # here relerrs still has all the bins, while the staterror are usually per-channel
+            # so we need to pick out the masks for this modifier to extract the
+            # modifier configuration (sigmas, etc..)
+            # so loop over samples and extract the first mask
+            # while making sure any subsequent mask is consistent
             relerrs = default_backend.sqrt(relerrs)
-
             masks = {}
             for modifier_data in self.builder_data[modname].values():
                 mask_this_sample = default_backend.astensor(
@@ -113,12 +117,13 @@ class staterror_builder:
                     else:
                         assert (mask_this_sample == masks[modname]).all()
 
-            for modifier_data in self.builder_data[modname].values():
-                modifier_data['data']['mask'] = masks[modname]
+            #extract sigmas using this modifiers mask
             sigmas = relerrs[masks[modname]]
+
+            # sigmas that are zero will be fixed to 1.0 arbitrarily
             # list of bools, consistent with other modifiers (no numpy.bool_)
+            # to ensure non-Nan constraint term, but in a future PR we need to remove constraints for these
             fixed = default_backend.tolist(sigmas == 0)
-            # ensures non-Nan constraint term, but in a future PR we need to remove constraints for these
             sigmas[fixed] = 1.0
             self.required_parsets.setdefault(parname, [required_parset(sigmas, fixed)])
         return self.builder_data
@@ -145,18 +150,6 @@ class staterror_combined:
             [[builder_data[m][s]['data']['mask']] for s in pdfconfig.samples]
             for m in keys
         ]
-        self.__staterror_uncrt = default_backend.astensor(
-            [
-                [
-                    [
-                        builder_data[m][s]['data']['uncrt'],
-                        builder_data[m][s]['data']['nom_data'],
-                    ]
-                    for s in pdfconfig.samples
-                ]
-                for m in keys
-            ]
-        )
         global_concatenated_bin_indices = [
             [[j for c in pdfconfig.channels for j in range(pdfconfig.channel_nbins[c])]]
         ]
--- a/src/pyhf/modifiers/staterror.py
+++ b/src/pyhf/modifiers/staterror.py
@@ -120,10 +120,11 @@ class staterror_builder:
             # extract sigmas using this modifiers mask
             sigmas = relerrs[masks[modname]]
 
-            # sigmas that are zero will be fixed to 1.0 arbitrarily
             # list of bools, consistent with other modifiers (no numpy.bool_)
-            # to ensure non-Nan constraint term, but in a future PR we need to remove constraints for these
             fixed = default_backend.tolist(sigmas == 0)
+            # FIXME: sigmas that are zero will be fixed to 1.0 arbitrarily to ensure
+            # non-Nan constraint term, but in a future PR need to remove constraints
+            # for these
             sigmas[fixed] = 1.0
             self.required_parsets.setdefault(parname, [required_parset(sigmas, fixed)])
         return self.builder_data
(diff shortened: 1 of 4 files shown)


matthewfeickert:
> Would be great to get add. test coverage and cross-checks w/ real world models

:+1: From the list in Issue #1945, we should be able to check against Issue #1720.

alexander-held (commented):
Thanks for this fix! Works as intended in the setups I have tested and I did not spot any problems with it.

src/pyhf/modifiers/staterror.py
```diff
@@ -145,18 +151,6 @@ def __init__(self, modifiers, pdfconfig, builder_data, batch_size=None):
             [[builder_data[m][s]['data']['mask']] for s in pdfconfig.samples]
             for m in keys
         ]
-        self.__staterror_uncrt = default_backend.astensor(
```
matthewfeickert, line 148: I'm sure this will be obvious after I've slept, but why is this chunk removed?

tests/test_scripts.py
```diff
@@ -297,7 +297,7 @@ def test_testpoi(tmpdir, script_runner):
     command = f'pyhf xml2json validation/xmlimport_input/config/example.xml --basedir validation/xmlimport_input/ --output-file {temp.strpath:s}'
     ret = script_runner.run(*shlex.split(command))
 
-    pois = [1.0, 0.5, 0.0]
+    pois = [1.0, 0.5, 0.001]
```
matthewfeickert, line 300: While I'm not fundamentally against this very tiny change, is it understood why a POI of `0` results in a `RuntimeWarning: invalid value encountered in double_scalars`

```console
$ pytest tests/test_scripts.py -k test_testpoi
```
```pytb
...
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/infer/calculators.py", line 418, in _false_case
    teststat = (qmu - qmu_A) / (2 * self.sqrtqmuA_v)
RuntimeWarning: invalid value encountered in double_scalars

```


? I haven't stepped through enough things yet to understand why this is getting hit, but given this example's RuntimeWarning indicating that things are blowing up (given a super tiny denominator), does this particular example just happen to have an pathology that is causing this?

matthewfeickert (commented):
Thanks @lukasheinrich for this PR and to @kratsg for adding the very nice test! I just have some questions that hopefully have very easy answers that I'm just missing as it is late for me.

(@kratsg As I was the one who originally moved this PR to draft I moved it to ready for review. So we were waiting for me and not Lukas I think.)

tests/test_scripts.py
kratsg: Extremely small or extremely large values are occurring. In Linux, this crashes - on Mac, this passes and returns NaN for the obs and 1.0 for all expected CLs which means we're hitting infinities somewhere in the code.

See the failing test here: https://github.com/scikit-hep/pyhf/runs/8051304900?check_suite_focus=true . Not sure why this crashes the way it does, but as we don't care about what POI we're testing on this code, just that we test different ones and see different values, I changed it.

src/pyhf/modifiers/staterror.py
kratsg: it's not used by anything.

matthewfeickert: Okay so just cleaning of vestigial code.

matthewfeickert (approved):
LGTM. :+1: Thanks @lukasheinrich and @kratsg!

Late tonight after I'm done packing I'll try to get Issue #1907 resolved so that we can do a `v0.7.0rc2` correctly and then I'll send out a request email for testing and being drafting up the `v0.7.0` release notes in PR #1705 so that we can release `v0.7.0` early this week.

tests/test_scripts.py
matthewfeickert: The reason for this was Issue #1992, which was fixed in PR #1993.

PR #1972, opened 2022-08-30
docs: Document Channel Summary Mixin

### LINKED ISSUE #1747: Documentation for parameter-related properties of `_ModelConfig`
### Summary

There are a few different things available in a `model.config` for accessing parameters:

- `model.config.parameters` for an alphabetically sorted, non-expanded (staterror/shaperror) list,
- `model.config.par_order` for a non-expanded list in the order that `pyhf` uses internally when handling and returning parameters,
- `model.config.par_names()` for an expanded version of the list in the order used internally.

Only `par_names()` shows up in the [`_ModelConfig` documentation](https://pyhf.readthedocs.io/en/v0.6.3/_generated/pyhf.pdf._ModelConfig.html) at the moment. `parameters` presumably comes from the `_ChannelSummaryMixin` mixin, but I do not see this class in the documentation (if that was there, clicking through to it from the `_ModelConfig` page could be a convenient way to find additional properties coming from the base class).

`model.config.parameters` may not be that commonly needed, and could be obtained from `model.config.par_order`. The latter is very useful when using `model.config.param_set()`.

See also #1024 for a few more points related to `parameters` vs `par_order`.

### Documentation Page Link

https://pyhf.readthedocs.io/en/v0.6.3/_generated/pyhf.pdf._ModelConfig.html

### Code of Conduct

### LINKED ISSUE #1982: Include `model.config` property in the `pyhf.pdf.Model` API documentation
I would also suggest somehow pointing to `_ModelConfig` from the `pyhf.pdf.Model` page. The `.config` API is not listed there, and it is not particularly intuitive for a user to look at `_ModelConfig` to find out about that `.config` attribute, where it says

> `_ModelConfig` should not be called directly. It should instead by accessed through the `config` attribute of `Model`

Given that `_ModelConfig` implies internal API, I wonder whether all of this page should just live within the `Model` API documentation page? This is where I would expect a user to look for it.

_Originally posted by @alexander-held in https://github.com/scikit-hep/pyhf/pull/1972#pullrequestreview-1095476721_

kratsg:
# Pull Request Description

* Resolves #1747
* Resolves #1982.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Make _ChannelSummaryMixin parameter related attributes properties.
* Add documentation for parameter related properties of _ModelConfig by documenting
  the _ChannelSummaryMixin.
* Add typehints to mixins.
* Update tests to mock the properties.
```

### DIFF
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -170,11 +170,16 @@ def test_get_workspace_data(workspace_factory, include_auxdata):
     assert w.data(m, include_auxdata=include_auxdata)
 
 
-def test_get_workspace_data_bad_model(workspace_factory, caplog):
+def test_get_workspace_data_bad_model(workspace_factory, caplog, mocker):
     w = workspace_factory()
     m = w.model()
     # the iconic fragrance of an expected failure
-    m.config.channels = [c.replace('channel', 'chanel') for c in m.config.channels]
+
+    mocker.patch(
+        'pyhf.mixins._ChannelSummaryMixin.channels',
+        new_callable=mocker.PropertyMock,
+        return_value=['chanel'],
+    )
     with caplog.at_level(logging.INFO, 'pyhf.pdf'):
         with pytest.raises(KeyError):
             assert w.data(m)
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -183,7 +183,7 @@ def test_get_workspace_data_bad_model(workspace_factory, caplog, mocker):
     with caplog.at_level(logging.INFO, 'pyhf.pdf'):
         with pytest.raises(KeyError):
             assert w.data(m)
-        assert 'Invalid channel' in caplog.text
+        assert "Invalid channel" in caplog.text
 
 
 def test_json_serializable(workspace_factory):
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -178,7 +178,7 @@ def test_get_workspace_data_bad_model(workspace_factory, caplog, mocker):
     mocker.patch(
         "pyhf.mixins._ChannelSummaryMixin.channels",
         new_callable=mocker.PropertyMock,
-        return_value=["channel"],
+        return_value=["fakechannel"],
     )
     with caplog.at_level(logging.INFO, 'pyhf.pdf'):
         with pytest.raises(KeyError):
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -183,7 +183,7 @@ def test_get_workspace_data_bad_model(workspace_factory, caplog, mocker):
     with caplog.at_level(logging.INFO, 'pyhf.pdf'):
         with pytest.raises(KeyError):
             assert w.data(m)
-            assert 'Invalid channel' in caplog.text
+        assert 'Invalid channel' in caplog.text
 
 
 def test_json_serializable(workspace_factory):
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -176,9 +176,9 @@ def test_get_workspace_data_bad_model(workspace_factory, caplog, mocker):
     # the iconic fragrance of an expected failure
 
     mocker.patch(
-        'pyhf.mixins._ChannelSummaryMixin.channels',
+        "pyhf.mixins._ChannelSummaryMixin.channels",
         new_callable=mocker.PropertyMock,
-        return_value=['chanel'],
+        return_value=["chanel"],
     )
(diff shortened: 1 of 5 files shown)


kratsg:
Screen Shot 2022-08-30 at 12 09 49 PM

matthewfeickert:
Relevant RTD build: https://pyhf--1972.org.readthedocs.build/en/1972/_generated/pyhf.pdf._ModelConfig.html

tests/test_workspace.py
```diff
@@ -170,11 +170,16 @@ def test_get_workspace_data(workspace_factory, include_auxdata):
...
+def test_get_workspace_data_bad_model(workspace_factory, caplog, mocker):
     w = workspace_factory()
     m = w.model()
     # the iconic fragrance of an expected failure
-    m.config.channels = [c.replace('channel', 'chanel') for c in m.config.channels]
+
+    mocker.patch(
+        "pyhf.mixins._ChannelSummaryMixin.channels",
+        new_callable=mocker.PropertyMock,
+        return_value=["channel"],
```
matthewfeickert, line 181:
```suggestion
        return_value=["chanel"],
```

Hm. pre-commit is pushing back a "fix" here for what it thinks is a `codespell` error (`chanel` -> `channel`), but this is intentional to cause a `KeyError`. So this will need to get ignored somehow, or a different erroneous `return_value` will need to be used.

matthewfeickert (changes requested):

tests/test_workspace.py
matthewfeickert: Also the

```python
    with caplog.at_level(logging.INFO, 'pyhf.pdf'):
        with pytest.raises(KeyError):
            assert w.data(m)
            assert 'Invalid channel' in caplog.text
```

test block isn't executing any code after

```python
            assert w.data(m)
```

so the

```python
            assert 'Invalid channel' in caplog.text
```

never gets run. I checked this by putting a `breakpoint` in after

```python
        with pytest.raises(KeyError):
            assert w.data(m)
            breakpoint()
            assert 'Invalid channel' in caplog.text
```

which is never reached.

alexander-held (commented):
I might have misunderstood the purpose of the changes, but I do not see `model.config.parameters` / `.par_order` mentioned in #1747 in https://pyhf--1972.org.readthedocs.build/en/1972/_generated/pyhf.pdf._ModelConfig.html. `.par_names()` had already been included previously, and that is very useful, but I think `.par_order` would also be needed. I think `.parameters` is not particularly relevant and can be obtained by sorting `.par_order` instead.

I would also suggest somehow pointing to `_ModelConfig` from the `pyhf.pdf.Model` page. The `.config` API is not listed there, and it is not particularly intuitive for a user to look at `_ModelConfig` to find out about that `.config` attribute, where it says

> `_ModelConfig` should not be called directly. It should instead by accessed through the `config` attribute of `Model`

Given that `_ModelConfig` implies internal API, I wonder whether all of this page should just live within the `Model` API documentation page? This is where I would expect a user to look for it.

kratsg:
> I might have misunderstood the purpose of the changes, but I do not see `model.config.parameters` / `.par_order` mentioned in #1747 in https://pyhf--1972.org.readthedocs.build/en/1972/_generated/pyhf.pdf._ModelConfig.html. `.par_names()` had already been included previously, and that is very useful, but I think `.par_order` would also be needed. I think `.parameters` is not particularly relevant and can be obtained by sorting `.par_order` instead.

`par_order` is on `_ModelConfig` and not the mixin. This PR is just documenting the mixin.

> I would also suggest somehow pointing to `_ModelConfig` from the `pyhf.pdf.Model` page. The `.config` API is not listed there, and it is not particularly intuitive for a user to look at `_ModelConfig` to find out about that `.config` attribute, where it says

File an issue specifically for this one, as that's different from the other issue.

The fundamental reason that `par_order` and others didn't show up is because they were object attributes and not class attributes (so sphinx won't see it by introspection).

alexander-held:
Looks like I was a bit too slow and created #1982 before the edit of the previous comment, so that would be another issue this PR closes.

I see that `par_order` is also there now, plus the other `_ModelConfig` pieces so this looks great. Thank you!

alexander-held:
I confused myself and was looking at `_ModelConfig` instead of `Model` where `par_order` etc. appear, but given that this is now cross-linked through the `.config` property of `Model` this is discoverable and I think that works well.

matthewfeickert (approved):
LGTM. :+1: Thanks @kratsg &mdash; this will help!

PR #1974, opened 2022-08-30
fix: Form data through loop over channels for multichannel coupled histo notebook

### LINKED ISSUE #1313: multichannel-coupled-histo notebook giving NAN with v0.6.0 API
# Description

The notebook `multichannel-coupled-histo.ipynb` on the branch `fix/api-for-multichannel-coupled-histo` produces `nan` in the final cell of the notebook. This was found in PR #1312, and should be fixed up.

kratsg:
# Pull Request Description

Resolves #1313.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use pdf.config.channels API instead of the channel names from the spec.
* Use pyhf.contrib.viz.brazil API for visualization.
* Deprecate use of pylab in notebook.

```


alexander-held (approved):
Looks good to me, I am fairly sure this fixed channel order is how this was always meant to be.

docs/examples/notebooks/multichannel-coupled-histo.ipynb
```diff
@@ -16,6 +16,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
+      "%pylab is deprecated, use %matplotlib inline and import the required libraries.\n",
```
alexander-held, line 19: Could consider replacing `%pylab inline` by
```python
import numpy as np
```
if you are touching this file already.

kratsg: I'll leave that up to @matthewfeickert to go through all notebooks and do this kind of update.

matthewfeickert: @kratsg yeah I can do that. We should also try to figure out a way to have these notebooks get moved to something that is either outside of the main repo or built out each time as I think that most of the size of the GitHub repo comes from the image diffs in notebooks, which feels bad. :grimacing:

matthewfeickert (approved):

PR #1978, opened 2022-09-01
feat: Update JSON Schema draft

kratsg:
# Pull Request Description

Update from Draft 6 to 2020-12 Draft instead. This also requires releasing a HiFa JSON v1.0.1 pointing at this new draft.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/schema/loader.py
+++ b/src/pyhf/schema/loader.py
@@ -3,6 +3,7 @@ import sys
 import json
 import pyhf.exceptions
 from pyhf.schema import variables
+from pyhf.typing import Schema
 
 # importlib.resources.as_file wasn't added until Python 3.9
 # c.f. https://docs.python.org/3.9/library/importlib.html#importlib.resources.as_file
@@ -12,7 +13,7 @@ else:
     import importlib_resources as resources
 
 
-def load_schema(schema_id: str):
+def load_schema(schema_id: str) -> Schema:
     """
     Get a schema by relative path from cache, or load it into the cache and return.
 
--- a/src/pyhf/schema/loader.py
+++ b/src/pyhf/schema/loader.py
@@ -39,9 +39,3 @@ def load_schema(schema_id: str):
             schema = json.load(json_schema)
             variables.SCHEMA_CACHE[schema['$id']] = schema
         return variables.SCHEMA_CACHE[schema['$id']]
-
-
-# pre-populate the cache to avoid network access
-# on first validation in standard usage
-# (not in pyhf.schema.variables to avoid circular imports)
-load_schema(f'{variables.SCHEMA_VERSION}/defs.json')
--- a/src/pyhf/schema/upgrader.py
+++ b/src/pyhf/schema/upgrader.py
@@ -1,9 +1,9 @@
 from pyhf.schema import variables
-from typing import Mapping
+from pyhf.typing import Workspace, PatchSet
 import copy
 
 
-def upgrade_workspace(spec: Mapping) -> Mapping:
+def upgrade_workspace(spec: Workspace) -> Workspace:
     """
     Upgrade the provided workspace specification to latest version.
 
@@ -26,7 +26,7 @@ def upgrade_workspace(spec: Mapping) -> Mapping:
     return spec
 
 
-def upgrade_patchset(spec: Mapping) -> Mapping:
+def upgrade_patchset(spec: PatchSet) -> PatchSet:
     """
     Upgrade the provided patchset specification to latest version.
 
--- a/src/pyhf/schema/upgrader.py
+++ b/src/pyhf/schema/upgrader.py
@@ -1,0 +1,49 @@
+from pyhf.schema import variables
+from typing import Mapping
+import copy
+
+
+def upgrade_workspace(spec: Mapping) -> Mapping:
+    """
+    Upgrade the provided workspace specification to latest version.
+
+    Args:
+        spec (dict): The specification to validate.
+        schema_name (str): The name of the schema to upgrade.
+
+    Returns:
+        upgraded_spec (dict): Upgraded workspace specification.
+
+    Raises:
+        pyhf.exceptions.InvalidSpecification: the specification is invalid
+    """
+
+    version = spec['version']
+
+    new_spec = copy.deepcopy(spec)
+    if version == '1.0.0':
+        new_spec['version'] = variables.SCHEMA_VERSION
+    return spec
+
+
+def upgrade_patchset(spec: Mapping) -> Mapping:
+    """
+    Upgrade the provided patchset specification to latest version.
+
+    Args:
+        spec (dict): The specification to validate.
+        schema_name (str): The name of the schema to upgrade.
+
+    Returns:
+        upgraded_spec (dict): Upgraded patchset specification.
+
+    Raises:
+        pyhf.exceptions.InvalidSpecification: the specification is invalid
+    """
+
+    version = spec['version']
+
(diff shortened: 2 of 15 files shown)


kratsg:
> @kratsg you have this slated for [`v0.7.1`](https://github.com/scikit-hep/pyhf/projects/19#card-85405165) at the moment. Does this mean that you feel we have a problem that needs to be patched? If not, we should probably place this in [`v0.8.0`](https://github.com/orgs/scikit-hep/projects/2).

It's going into the next release though. I don't know if we want to start juggling development HiFa v1.0.0 and v1.1.0 at the same time? But this will need to go in, so that we can deal with multiple schema versions.

matthewfeickert:
> It's going into the next release though. I don't know if we want to start juggling development HiFa v1.0.0 and v1.1.0 at the same time? But this will need to go in, so that we can deal with multiple schema versions.

Okay, then I've switched it to be v0.8.0. After the issues with getting v0.7.0 out and also accepting PRs I think it makes sense (for me to figure out how to do release branches) to start using patch releases for when we actually have a bug we need to go back and patch and start having other releases be minor (until we hit v1.0.0).

src/pyhf/schema/loader.py
```diff
@@ -54,9 +55,3 @@ def load_schema(schema_id: str):
             schema = json.load(json_schema)
             variables.SCHEMA_CACHE[schema['$id']] = schema
         return variables.SCHEMA_CACHE[schema['$id']]
-
-
-# pre-populate the cache to avoid network access
```
kratsg, line 59: NB: now works with local paths as expected (this is a change in `jsonschema` resolving). This was initially added back in #1917 by @lhenkelm , but it should not be needed anymore now.

src/pyhf/schema/upgrader.py
```diff
@@ -0,0 +1,61 @@
...
+from pyhf.typing import Workspace, PatchSet, SchemaVersion, UpgradeProtocol
+import copy
+
+
+class Upgrade_1_0_1:
+    """
+    Used for testing functionality of upgrade.
+    """
+
+    version: SchemaVersion = '1.0.1'
```
matthewfeickert, line 12:
```suggestion
class Upgrade(version: SchemaVersion = "1.0.1"):
    """
    Used for testing functionality of upgrade.
    """
```

Maybe I'm missing something about what the purpose of `upgrader` is, but is there a reason why we need to make this class schema `v1.0.1` hardcoded?

src/pyhf/schema/upgrader.py
```diff
@@ -0,0 +1,61 @@
...
+
+        new_spec = copy.deepcopy(spec)
+        if spec['version'] == '1.0.0':
+            new_spec['version'] = cls.version
+        return new_spec
+
+
+def upgrade(*, to_version: SchemaVersion | None = None) -> type[UpgradeProtocol]:
+    if to_version is None or to_version == '1.0.1':
+        return Upgrade_1_0_1
```
matthewfeickert, line 59:
```suggestion
def upgrade(*, to_version: SchemaVersion | None = None) -> type[UpgradeProtocol]:
    if to_version is None or to_version == "1.0.1":
        return Upgrade("1.0.1")
```

Follow up to the above.

tests/test_upgrade.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+
+
+def test_upgrade_bad_version(datadir):
+    with pytest.raises(ValueError):
+        pyhf.schema.upgrade(to_version='0.9.0')
+
+
+def test_upgrade_to_latest(datadir):
+    ws = json.load(open(datadir.joinpath("workspace_1.0.0.json"), encoding="utf-8"))
```
matthewfeickert, line 14:
```suggestion
    with open(datadir.joinpath("workspace_1.0.0.json"), encoding="utf-8") as read_file:
        ws = json.load(read_file)
```

Would be good to use context managers for this whole file.

matthewfeickert (commented):
Thanks @kratsg for this nice work. Overall this looks good. :+1: 

I've added a question and some small formatting suggestions, but as we need to fix up the coverage thing I could add those in this weekend once I am at CERN so that you don't waste time having to poke at `coverage`'s decisions. It is probably faster to locally run 

```console
$ nox --session tests --python 3.10 -- coverage
$ nox --session coverage
$ xdg-open ./htmlcov/index.html
```

to debug coverage than to have GHA do all of that and then have Codecov report it back to us.

If you can also fill out the PR body with some additional context that would be helpful, given that you've done a lot of nice work here (I can also summarize this later, but am not able to succinctly do so now given a lot was added to support this).

matthewfeickert:
(Ignore the docs failing as this is Springers fault: https://twitter.com/HEPfeickert/status/1600915112403304449?s=20&t=kjTW5TNbU_Jukxb0ALd0uQ. We can either fix this up locally in another PR or just wait for them to fix it.)

src/pyhf/schema/upgrader.py
kratsg: We need an `Upgrade_X_Y_Z` class for every single version we plan to support upgrading to. When we add a new version of the spec, we have to have a new upgrade class that knows how to upgrade from an old version to the version the class represents.

src/pyhf/schema/loader.py
kratsg: See #1980 as well.

src/pyhf/schema/upgrader.py
matthewfeickert: @kratsg Ah okay, so the real reason is that we can't deterministically know what version we're upgrading from in advance? 

Do we have a plan for how to handle upgrading multiple versions in one go? e.g. Let's say we have `v1.0.0`, `v1.1.0`, `v1.2.0, and `v2.0.0`. If at this point someone comes along with a `v1.0.0` schema and wants to move from `v1.0.0` to `v2.0.0` what's the plan for doing this?

tests/test_workspace.py
```diff
@@ -453,6 +453,7 @@ def test_combine_workspace_incompatible_poi(workspace_factory, join):
 def test_combine_workspace_diff_version(workspace_factory, join):
     ws = workspace_factory()
     ws.version = '1.0.0'
+    ws['version'] = '1.0.0'
```
matthewfeickert, line 456: I'm probably missing the point of this test, but if you put a `breakpoint` just after `ws = workspace_factory()`

```
(Pdb) ws.version
'1.0.0'
(Pdb) ws["version"]
'1.0.0'
```

so is the reason to explicitly set this a means to solidify the versions that will be used in the future?

matthewfeickert (commented):
@kratsg sorry just a few more questions to help me understand this better. As this will go in this weekend can I also ask you sometime before Monday to write up a summary in the PR body and the squash and merge summary commit? This is a huge amount of work (:bow:) and I want to make sure that we have a summary of this as I think if I had to try to sit down and reread all of this again late at night I might be pressed to do so without some accompanying text.

kratsg:
Look into how [nbformat](https://github.com/jupyter/nbformat/blob/590cf592d725c09394f9ce5c52db56876c783a18/nbformat/converter.py) handles converting between versions: (thanks @agoose77 )

src/pyhf/schema/upgrader.py
kratsg: I have ideas on how to do this with python's `mro`

tests/test_workspace.py
kratsg: This is just a test to enforce absolutely different versions.

PR #1985, opened 2022-09-05
feat: Add support for `_ModelConfig.set_poi(None)` to unset model POI

### LINKED ISSUE #1984: Can no longer set several model config attributes
### Summary

The changes in #1972 cause `model.config.poi_index` to not be writable anymore. `cabinetry` sets this value directly to support running `pyhf.infer.hypotest` with a configurable POI (without the need to re-build the model).

### OS / Environment

```console
n/a
```

### Steps to Reproduce

```python (paste below)
import pyhf

    signal=[24.0, 22.0], bkg=[50.0, 52.0], bkg_uncertainty=[3.0, 7.0]
)
model.config.poi_index = 1
```

### File Upload (optional)

_No response_

### Expected Results

assignment succeeds

### Actual Results

```console
  File "test.py", line 7, in <module>
    model.config.poi_index = ""
AttributeError: can't set attribute
```

### pyhf Version

```console
pyhf, version 0.7.0rc4.dev2
```

### Code of Conduct

kratsg:
# Pull Request Description

Resolves #1984 by allowing `_ModelConfig.set_poi(None)` to be allowed behavior to remove the POI from a model.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Allow for `_ModelConfig.set_poi(None)` as a way to unset the POI from a model.
* Add test of `model.config.set_poi` for value of the poi that are valid strings and None.
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -429,6 +429,8 @@ class _ModelConfig(_ChannelSummaryMixin):
         """
         Set the model parameter of interest to be model parameter ``name``.
 
+        If ``name`` is ``None``, this will reset the parameter of interest.
+
         Example:
             >>> import pyhf
             >>> model = pyhf.simplemodels.uncorrelated_background(
@@ -438,6 +440,11 @@ class _ModelConfig(_ChannelSummaryMixin):
             >>> model.config.poi_name
             'mu'
         """
+        if name is None:
+            self._poi_name = None
+            self._poi_index = None
+            return
+
         if name not in self.parameters:
             raise exceptions.InvalidModel(
                 f"The parameter of interest '{name:s}' cannot be fit as it is not declared in the model specification."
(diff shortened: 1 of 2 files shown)


src/pyhf/pdf.py
```diff
@@ -429,6 +429,8 @@ def set_poi(self, name):
         """
         Set the model parameter of interest to be model parameter ``name``.
 
+        If ``name`` is ``None``, this will reset the parameter of interest.
```
alexander-held, line 432: "Reset" could maybe be misunderstood as "set back to what was originally in the spec from which the model was built"? I cannot think of another word to prevent this though.

alexander-held: "unset" sounds very clear now, thanks!

matthewfeickert (approved):
LGTM. :+1: Thanks for being so fast on this @kratsg and for the quick iteration @alexander-held.

PR #1991, opened 2022-09-08
feat: Experimental implementation of custom modifiers

kratsg:
# Pull Request Description

Refer to #850.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/experimental/modifiers.py
+++ b/src/pyhf/experimental/modifiers.py
@@ -42,6 +42,8 @@ def make_builder(
     funcname: str, deps: list[str], newparams: dict[str, dict[str, Sequence[float]]]
 ) -> BaseBuilder:
     class _builder(BaseBuilder):
+        is_shared = False
+
         def __init__(self, config):
             self.builder_data = {'funcs': {}}
             self.config = config
--- a/src/pyhf/experimental/modifiers.py
+++ b/src/pyhf/experimental/modifiers.py
@@ -1,0 +1,160 @@
+from __future__ import annotations
+import pyhf
+from pyhf.parameters import ParamViewer
+from pyhf import get_backend
+from pyhf import events
+
+from typing import Sequence, Callable, Any
+
+
+class BaseApplier:
+    ...
+
+
+class BaseBuilder:
+    ...
+
+
+def _allocate_new_param(
+    p: dict[str, Sequence[float]]
+) -> dict[str, str | bool | int | Sequence[float]]:
+    return {
+        'paramset_type': 'unconstrained',
+        'n_parameters': 1,
+        'is_shared': True,
+        'inits': p['inits'],
+        'bounds': p['bounds'],
+        'is_scalar': True,
+        'fixed': False,
+    }
+
+
+def make_func(expression: str, deps: list[str]) -> Callable[[Sequence[float]], Any]:
+    def func(d: Sequence[float]) -> Any:
+        import numexpr as ne
+
+        return ne.evaluate(expression, local_dict=dict(zip(deps, d)))
+
+    return func
+
+
+def make_builder(
+    funcname: str, deps: list[str], newparams: dict[str, dict[str, Sequence[float]]]
+) -> BaseBuilder:
+    class _builder(BaseBuilder):
+        def __init__(self, config):
+            self.builder_data = {'funcs': {}}
+            self.config = config
+
+        def collect(self, thismod, nom):
+            maskval = True if thismod else False
+            mask = [maskval] * len(nom)
+            return {'mask': mask}
+
+        def append(self, key, channel, sample, thismod, defined_samp):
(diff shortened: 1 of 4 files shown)


alexander-held:
My reading of "contrib" is "quite stable API, but might get factored out eventually", while I would view "experimental" more like "anything goes, expect things to break". You could of course communicate how exactly you want the modules to be understood, but this is what I would think without further context.

matthewfeickert:
> do we want to differentiate between `experimental` and `contrib`?

Yes, I think that's a good idea if we really mean "experimental". I have the same take as @alexander-held does.

lukasheinrich:
Ok !

alexander-held:
Related to the discussion in https://github.com/scikit-hep/cabinetry/issues/382, it may be good to provide a suitable schema for this as well (perhaps same PR makes sense?).

rmnmllr:
If I may add a request, it would be great if `pyhf` could catch a wrongly typed function expression in the custom modifier. I stumbled over `srqt` instead of `sqrt` in the related cabinetry issue discussion https://github.com/scikit-hep/cabinetry/issues/382#issuecomment-1362881632 which produced an obscure error output.

matthewfeickert:
@kratsg while I will go back and review Issue #850 (I've forgotten all of it) can you also comment on what else would be needed for this PR before it would be ready for review?

pariRieck:
Hi, just to comment also here that on the side of an ATLAS analysis there is some urgency with this and a merge would help us.
We are looking for a way to incorporate different signal histograms into our statistical model and they scale differently with the parameter of interest, hence custom_modifiers are what we are looking for.
Apart from that, we should be fine.

jmontejo:
Hi,
just wanted to add a +1 of an ATLAS analysis (RPV-multijets) depending on this to leave HistFitter for pyhf.

Cheers
Javier

jmontejo:
Hi,
just wanted to add a +1 of an ATLAS analysis (RPV-multijets) depending on this to leave HistFitter for pyhf.

Cheers
Javier

src/pyhf/experimental/modifiers.py
```diff
@@ -0,0 +1,181 @@
...
+            nom = (
+                defined_samp["data"]
+                if defined_samp
+                else [0.0] * self.config.channel_nbins[channel]
+            )
+            moddata = self.collect(thismod, nom)
+            self.builder_data[key][sample]["data"]["mask"] += moddata["mask"]
+            if thismod:
+                if thismod["name"] != func_name:
+                    print(thismod)
```
matthewfeickert, line 87:
```suggestion
```

src/pyhf/experimental/modifiers.py
```diff
@@ -0,0 +1,181 @@
...
+            """
+            Returns:
+                modification tensor: Shape (n_modifiers, n_global_samples, n_alphas, n_global_bin)
+            """
+            if not self.param_viewer.index_selection:
+                return
+            tensorlib, _ = get_backend()
+            if self.batch_size is None:
+                deps = self.param_viewer.get(pars)
+                print("deps", deps.shape)
```
matthewfeickert, line 153:
```suggestion
```

src/pyhf/experimental/modifiers.py
```diff
@@ -0,0 +1,181 @@
...
+                results = tensorlib.astensor([f(deps) for f in self.funcs])
+                results = tensorlib.einsum(
+                    "msab,m->msab", self.custom_mod_mask, results
+                )
+            else:
+                deps = self.param_viewer.get(pars)
+                print("deps", deps.shape)
```
matthewfeickert, line 160:
```suggestion
```

matthewfeickert (commented):
@kratsg I assume that these prints were put in here for debugging? and that we can remove them and they aren't actually supposed to be in there for logging. Can you confirm?

matthewfeickert:
Relevant RTD build: https://pyhf--1991.org.readthedocs.build/en/1991/api.html#experimental

src/pyhf/experimental/modifiers.py
matthewfeickert: I've removed the prints in https://github.com/scikit-hep/pyhf/pull/1991/commits/bb302c2423f702a489a42c8311dc412388e42bd9 but if we need them back for debugging quickly just rever it.

alexander-held:
While giving this another try, I ran into something that is either a bug or a usage question. I also have some API feedback. I opened #2350 for this to not clutter this PR too much.

itsaklid:
Hello,

I wanted to inquire about the status of this PR. Are there any plans to keep pursuing this or the effort will be dropped in favor of #2579 ? 

I have used this branch for a fit with custom expressions but it seems that a feature outlined in #2595 is missing.

PR #2009, opened 2022-09-16
test: Update test_plot_results_no_axis baseline image

matthewfeickert:
# Description

matplotlib v3.6.0 results in a slightly different baseline image (the location of the $\mathrm{CL}_{s}$ `y-axis` label is shifted very slighlty) than matplotlib v3.5.x, so regenerate the baseline image using matplotlib v3.6.0 with

```
pytest --mpl-generate-path=tests/contrib/baseline tests/contrib/test_viz.py
```

Also mark the `test_plot_results_no_axis` test as `xfail` for Python 3.7 as `matplotlib` `v3.6.0` is Python 3.8+ and so the image is guaranteed to be different as Python 3.7 runtimes will install `matplotlib` `v3.5.x`.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* matplotlib v3.6.0 results in a slightly different baseline image than
  matplotlib v3.5.x, so regenerate the baseline image using matplotlib v3.6.0
  with `pytest --mpl-generate-path=tests/contrib/baseline tests/contrib/test_viz.py`.
* Mark the test_plot_results_no_axis test as xfail for Python 3.7 as matplotlib v3.6.0
  is Python 3.8+ and so the image is guaranteed to be different as Python 3.7 runtimes
  will install matplotlib v3.5.x.
```


matthewfeickert:
Henry has warned us before about checking the actual output of plots, but I'm not sure how to properly test visualization related functions otherwise. I'm open to suggestions though.

matthewfeickert:
Ah things get more complicated as `matplotlib` `v3.6.0` is Python 3.8+ only, and so our Python 3.7 tests fail on this as they are using `matplotlib` `v3.5.3`. So I guess we could get around this by just no longer testing the viz tests on Python 3.7?

tests/contrib/test_viz.py
```diff
@@ -67,6 +68,10 @@ def test_plot_results(datadir):
 
 
 @pytest.mark.mpl_image_compare
+@pytest.mark.xfail(
```
matthewfeickert, line 71: I'm going with `xfail` over `skip` here as skip just doesn't run the test at all, but `xfail` will guarantee that the behavior is failing as expected and so is a stronger test.

matthewfeickert:
As this is a fix for comparison of baseline images and necessary to get the test suite passing for the v0.7.0 release, I'm going to approve and merge this myself. As always, PRs approved by a single core dev can be reverted as needed by the rest of the dev team.

PR #2010, opened 2022-09-17
ci: Ignore Dependabot PRs to update tests/constraints.txt 

matthewfeickert:
# Description

Ignore all dependencies that are under the domain of pip in Dependabot's package ecosystems to avoid automatic security PRs that target `tests/constraints.txt`, which must remain static for lower bound constraints testing. This is designed to keep things like PR #2008 from ever happening.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Ignore all dependencies that are under the domain of pip in Dependabot's package
  ecosystems to avoid automatic security PRs that target tests/constraints.txt,
  which must remain static for lower bound constraints testing.
* Add default labels and reviewers to be applied to PRs from GHA updates from Dependabot.
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    labels:
+      - "github-actions"
+      - "dependencies"
+    reviewers:
+      - "matthewfeickert"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,11 @@ updates:
       - "dependencies"
     reviewers:
       - "matthewfeickert"
+
+  # Ignore all pip dependencies to avoid PRs to update tests/constraints.txt
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,3 +5,16 @@ updates:
...
+    reviewers:
+      - "matthewfeickert"
+
+  # Ignore all pip dependencies to avoid PRs to update tests/constraints.txt
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    ignore:
+      - dependency-name: "*"
```
matthewfeickert, line 20: We don't want things like PR #2008 to ever happen, as `tests/constraints.txt` is something that should only ever be updated when the lower bounds in `setup.cfg`/`setup.py` need to be updated. I _think_ based on https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore that this will ignore anything that is related to what Dependabot views as being in the Python ecosystem, but I'm not 100% sure on this. So if someone with more Dependabot experience can review this that would be great.

kratsg (approved):

PR #2011, opened 2022-09-20
fix: Use function scope to avoid altering hypotest_args fixture

matthewfeickert:
# Description

Needed for PR #1274 

* Use the 'function' scope for hypotest_args fixture. Fixing the POI in the model config changes it for the entire fixture if the fixture is 'module' scope, which would cause errors in other tests that later use it and defeats the point of the fixture.
* Remove unnecessary use of tmpdir fixture as no output is written.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use the 'function' scope for hypotest_args fixture. Fixing the POI in the model config
  changes it for the entire fixture if the fixture is 'module' scope, which would cause
  errors in other tests that later use it and defeats the point of the fixture.
* Remove unnecessary use of tmpdir fixture as no output is written.
```

### DIFF
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -1,13 +1,10 @@
-from copy import deepcopy
-
-import numpy as np
 import pytest
+import pyhf
+import numpy as np
 import scipy.stats
 
-import pyhf
 
-
-@pytest.fixture(scope='module')
+@pytest.fixture(scope="function")
 def hypotest_args():
     pdf = pyhf.simplemodels.uncorrelated_background(
         signal=[12.0, 11.0], bkg=[50.0, 52.0], bkg_uncertainty=[3.0, 7.0]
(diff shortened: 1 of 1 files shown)


tests/test_infer.py
```diff
@@ -1,8 +1,11 @@
...
+from copy import deepcopy
+
 import numpy as np
+import pytest
 import scipy.stats
 
+import pyhf
+
 
 @pytest.fixture(scope='module')
```
kratsg, line 10:
```suggestion
@pytest.fixture(scope='function')
```

kratsg (changes requested):
Just change the fixture scope to function instead and you don't need the deepcopy. This should remove the need for `deepcopy`.

tests/test_infer.py
matthewfeickert: That's fine. I hadn't done this to avoid having to have the fixture evaluate as often, but this is actually using the scoping for how it was intended, so that's good.

tests/test_infer.py
```diff
@@ -466,16 +466,16 @@ def test_toy_calculator(tmpdir, hypotest_args):
...
     addition of the return_expected keyword arg is as expected
     """
 
-    _, _, pdf = hypotest_args
-    pdf.config.param_set('mu').suggested_fixed = [True]
+    test_poi, data, model = hypotest_args
+    model.config.param_set("mu").suggested_fixed = [True]
     with pytest.raises(pyhf.exceptions.InvalidModel):
-        pyhf.infer.hypotest(*hypotest_args)
+        pyhf.infer.hypotest(test_poi, data, model)
```
matthewfeickert, line 478: Changing this still to make it more clear that the unpacked `model` is what is being passed in, as otherwise this might be confusing at first glance.

kratsg (approved):

kratsg:
The coverage failure/change here is rather suspicious / odd as none of the changes should impact coverage.

matthewfeickert:
> The coverage failure/change here is rather suspicious / odd as none of the changes should impact coverage.

Yeah I think Codecov is having issue. :/

PR #2017, opened 2022-09-21
test: Discover pytest fixtures during doctest

### LINKED ISSUE #2015: GitHub Actions only error: AttributeError("'NoneType' object has no attribute '_sorted_indices'")
### Summary

In PR #1274 there is a `AttributeError("'NoneType' object has no attribute '_sorted_indices'")` that is happening for Python 3.10 only for unclear reasons. This is blocking release `v0.7.0`. This is also not reproducible locally on my laptop.

### OS / Environment

```console
GitHub Actions on Python 3.10
```

### Steps to Reproduce

```console (paste below)
$ pytest src/pyhf/tensor/jax_backend.py
```

### File Upload (optional)

_No response_

### Expected Results

For the CI to pass

### Actual Results

```pytb
____________ [doctest] pyhf.tensor.jax_backend.jax_backend.astensor ____________
207 
208         Convert to a JAX ndarray.
209 
210         Example:
211 
212             >>> import pyhf
213             >>> pyhf.set_backend("jax")
UNEXPECTED EXCEPTION: AttributeError("'NoneType' object has no attribute '_sorted_indices'")
  File "/Users/runner/hostedtoolcache/Python/3.10.7/x64/lib/python3.10/doctest.py", line 1350, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest pyhf.tensor.jax_backend.jax_backend.astensor[1]>", line 1, in <module>
  File "/Users/runner/work/pyhf/pyhf/src/pyhf/events.py", line 161, in register_wrapper
    result = func(*args, **kwargs)
  File "/Users/runner/work/pyhf/pyhf/src/pyhf/tensor/manager.py", line 192, in set_backend
    events.trigger("tensorlib_changed")()
  File "/Users/runner/work/pyhf/pyhf/src/pyhf/events.py", line 70, in __call__
    func()(arg(), *args, **kwargs)
  File "/Users/runner/work/pyhf/pyhf/src/pyhf/tensor/common.py", line 33, in _precompute
    self.sorted_indices = tensorlib.astensor(self._sorted_indices, dtype='int')
AttributeError: 'NoneType' object has no attribute '_sorted_indices'
/Users/runner/work/pyhf/pyhf/src/pyhf/tensor/jax_backend.py:213: UnexpectedException
------------------------------ Captured log call -------------------------------
INFO     absl:xla_bridge.py:174 Remote TPU is not linked into jax; skipping remote TPU.
INFO     absl:xla_bridge.py:350 Unable to initialize backend 'tpu_driver': Could not initialize backend 'tpu_driver'
INFO     absl:xla_bridge.py:350 Unable to initialize backend 'cuda': module 'jaxlib.xla_extension' has no attribute 'GpuAllocatorConfig'
INFO     absl:xla_bridge.py:350 Unable to initialize backend 'rocm': module 'jaxlib.xla_extension' has no attribute 'GpuAllocatorConfig'
INFO     absl:xla_bridge.py:350 Unable to initialize backend 'tpu': module 'jaxlib.xla_extension' has no attribute 'get_tpu_client'
```

### pyhf Version

```console
PR #1274
```

### Code of Conduct

matthewfeickert:
# Description

* Resolves #2015
* Needed for PR #1274

Symlink the `tests/conftest.py` under the `src` tree to allow for the autouse pytest fixtures to be discovered and used during doctest. This is needed as the backends are not currently being automatically reset between doctests and the fixtures will not be discovered during doctest runs invoked with `pytest src/` as noted in the [pytest docs](https://docs.pytest.org/en/7.1.x/how-to/doctest.html#doctest-namespace-fixture):

> Note that like the normal `conftest.py`, the fixtures are discovered in the directory tree conftest is in. Meaning that if you put your doctest with your source code, the relevant `conftest.py` needs to be in the same directory tree. Fixtures will not be discovered in a sibling directory tree!

   - c.f. https://docs.pytest.org/en/7.1.x/how-to/doctest.html#doctest-namespace-fixture
   - https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html

Exclude `src/conftest.py` from the `MANIFEST.in` so as to not include it in the sdist build.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Symlink the tests/conftest.py under the src/ tree to allow for the
  autouse pytest fixtures to be discovered and used during doctest.
  This is needed as the backends are not currently being automatically
  reset between doctests and the fixtures will not be discovered during
  doctest runs invoked with `pytest src/` as noted in the pytest docs:

> Note that like the normal conftest.py, the fixtures are discovered in the
> directory tree conftest is in. Meaning that if you put your doctest with
> your source code, the relevant conftest.py needs to be in the same directory
> tree. Fixtures will not be discovered in a sibling directory tree!

   - c.f. https://docs.pytest.org/en/7.1.x/how-to/doctest.html#doctest-namespace-fixture
   - https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html

* Exclude src/conftest.py from the MANIFEST.in so as to not include it
  in the sdist build.
```

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,6 @@
 prune **
 graft src
-exlucde src/conftest.py
+exclude src/conftest.py
 
 include setup.py
 include setup.cfg
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,6 @@
 prune **
 graft src
+exlucde src/conftest.py
 
 include setup.py
 include setup.cfg
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -1,0 +1,1 @@
+../tests/conftest.py


kratsg:
Does this work if we move `conftest.py` up one level instead of having a symlink?

matthewfeickert:
> Does this work if we move `conftest.py` up one level instead of having a symlink?

I think so (testing now). Assuming so, you'd prefer to have `conftest.py` be a top level file? If so, that's fine with me.

kratsg:
> I think so (testing now). Assuming so, you'd prefer to have `conftest.py` be a top level file? If so, that's fine with me.

It feels a bit weird and hacky but I guess I see why. I wonder if `pytest tests/conftest.py src/` would fix this as well. I think I might prefer the symlink as it's more explicit, but the problem is we might forget why it's there later...

matthewfeickert:
> It feels a bit weird and hacky but I guess I see why. I wonder if `pytest tests/conftest.py src/` would fix this as well. I think I might prefer the symlink as it's more explicit, but the problem is we might forget why it's there later...

Yeah, it is not ideal I agree. Though turns out we don't want to move `conftest.py` to the top level as it causes additional problems (that I think are somewhat related to https://github.com/pytest-dev/pytest/issues/2269). So I think we'd have to either keep the symlink or during the doctest do some on the fly injection/copying (which I think isn't a good idea).

kratsg (approved):

PR #2027, opened 2022-09-23
feat: Make `par_names` a _ModelConfig property attribute

### LINKED ISSUE #1700: Make `model.config.par_names()` a property?
### Summary

Currently [`model.config.par_names()`](https://github.com/scikit-hep/pyhf/blob/767ed59d06274f810e47944f1fe4c95d94968543/src/pyhf/pdf.py#L304-L329) takes no arguments and returns a list of parameter names. The behavior slightly differs from `model.config.channels` and `model.config.samples`, which are properties. For consistency, it may be useful to turn `par_names` into a property as well.

### Additional Information

`model.config.par_names()` used to take an argument to set formatting, which required it to be a function. Since that has been dropped in the meantime, this opens up the possibility of using `@property`.

I don't know whether both `par_names` and `par_names()` can be supported simultaneously, which would allow to not break existing workflows with `par_names()`, but it may be better to only have a single API regardless.

### Code of Conduct

matthewfeickert:
# Description

Resolves #1700

Make `pyhf.pdf._ModelConfig.par_names()` a property attribute to match the `pyhf.pdf._ModelConfig.par_order` API.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Make pyhf.pdf._ModelConfig.par_names() property attribute pyhf.pdf._ModelConfig.par_names,
  like the the pyhf.pdf._ModelConfig.par_order API.
* Update all usage of `.par_names()` to `par_names`.
* Use unittest.mock.PropertyMock to properly mock par_names return.
```

### DIFF
--- a/src/pyhf/optimize/mixins.py
+++ b/src/pyhf/optimize/mixins.py
@@ -181,7 +181,7 @@ class OptimizerMixin:
 
         # handle non-pyhf ModelConfigs
         try:
-            par_names = pdf.config.par_names()
+            par_names = pdf.config.par_names
         except AttributeError:
             par_names = None
 
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -358,6 +358,7 @@ class _ModelConfig(_ChannelSummaryMixin):
         """
         return self.par_map[name]['slice']
 
+    @property
     def par_names(self):
         """
         The names of the parameters in the model including binned-parameter indexing.
@@ -370,7 +371,7 @@ class _ModelConfig(_ChannelSummaryMixin):
             >>> model = pyhf.simplemodels.uncorrelated_background(
             ...     signal=[12.0, 11.0], bkg=[50.0, 52.0], bkg_uncertainty=[3.0, 7.0]
             ... )
-            >>> model.config.par_names()
+            >>> model.config.par_names
             ['mu', 'uncorr_bkguncrt[0]', 'uncorr_bkguncrt[1]']
         """
         _names = []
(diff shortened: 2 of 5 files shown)


matthewfeickert:
@kratsg @lukasheinrich while adding the

```pycon
>>> import pyhf
>>> model = pyhf.simplemodels.correlated_background(
...     signal=[12.0, 11.0],
...     bkg=[50.0, 52.0],
...     bkg_up=[45.0, 57.0],
...     bkg_down=[55.0, 47.0],
... )
>>> model.config.par_order
['correlated_bkg_uncertainty', 'mu']
>>> model.config.par_names()
['correlated_bkg_uncertainty', 'mu']
```

example to PR #1705 I noted there's a bit of an inconsistency here in API. Can we slip this in under the wire for `v0.7.0`?

matthewfeickert:
@alexander-held This will hit `cabinetry` but can be easily patched with

```
git grep --name-only "par_names()" -- src/ | xargs sed -i 's/par_names()/par_names/g'
```

alexander-held:
See #1700, I assume that will still go into 0.7?

matthewfeickert:
> See #1700 

Nice. :sunglasses: 

> I assume that will still go into 0.7?

Unless people object to be messing around with the API this late. :grimacing:

matthewfeickert:
CI is failing as I forgot that

https://github.com/scikit-hep/pyhf/blob/cccbda16e56b34150b55489f2d172d492f879209/tests/test_optim.py#L579-L582

needs to get fixes now too.

matthewfeickert:
Relevant RTD build: https://pyhf--2027.org.readthedocs.build/en/2027/_generated/pyhf.pdf._ModelConfig.html#pyhf.pdf._ModelConfig.par_names

kratsg (approved):

PR #2044, opened 2022-10-10
feat: Drop Python 3.7 support

matthewfeickert:
# Description

Following Discussion #2042, drop Python 3.7 support to follow NEP 29 development recommendations. As a result, drop the requirement of `typing_extensions` as that is needed for Python 3.7 only and update the lower bound on SciPy to v1.3.2 &mdash; the first release with Python 3.8 wheels.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Drop Python 3.7 support to follow NEP 29 development recommendations.
   - c.f. https://github.com/scikit-hep/pyhf/discussions/2042
* Drop typing_extensions as a requirement as that is needed by Python 3.7 only.
* Update lower bound of the supported scipy versions to v1.3.2, the first scipy release
  with Python 3.8 wheels.
* Update pre-commit hooks to use Python 3.8.
```

### DIFF
--- a/src/pyhf/typing.py
+++ b/src/pyhf/typing.py
@@ -1,7 +1,15 @@
 import os
-from typing import Any, MutableSequence, Sequence, Tuple, Union
-
-from typing import Literal, Protocol, SupportsIndex, TypedDict
+from typing import (
+    Any,
+    Literal,
+    MutableSequence,
+    Protocol,
+    Sequence,
+    SupportsIndex,
+    Tuple,
+    TypedDict,
+    Union,
+)
 
 __all__ = (
     "PathOrStr",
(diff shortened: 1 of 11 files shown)


src/pyhf/typing.py
```diff
@@ -1,11 +1,7 @@
 import os
-import sys
 from typing import Any, MutableSequence, Sequence, Tuple, Union
 
-if sys.version_info >= (3, 8):
-    from typing import Literal, Protocol, SupportsIndex, TypedDict
-else:
-    from typing_extensions import Literal, Protocol, SupportsIndex, TypedDict
+from typing import Literal, Protocol, SupportsIndex, TypedDict
```
kratsg, line 4: this can be merged with above line

kratsg (approved):

PR #2049, opened 2022-10-22
chore: Set stacklevel to 2 for logging warning and debug

matthewfeickert:
# Description

Fill in later. c.f. https://youtu.be/CtFdXBEwYfk

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/cli/infer.py
+++ b/src/pyhf/cli/infer.py
@@ -130,7 +130,7 @@ def fit(
     else:
         with open(output_file, "w+", encoding="utf-8") as out_file:
             json.dump(result, out_file, indent=4, sort_keys=True)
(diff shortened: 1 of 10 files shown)


matthewfeickert:
I think this PR is actually not needed as `logging.warn` doesn't do the same action of showing where in the stack this problem comes up as composed to `warning.warn` where it does. So in this case something like

```python
import pyhf

pyhf.set_backend("numpy")
)
test_mu = 1.0
CLs_obs, CLs_exp = pyhf.infer.hypotest(
    test_mu, data, model, test_stat="q", return_expected=True
)
```

will log

```
```

regardless of what the `stacklevel` is and so we don't need this.

PR #2087, opened 2022-12-27
test: Add lower bound on supported numpy versions

matthewfeickert:
# Description

Set lower bound on numpy to v1.21.0 as this is the lower bound required by JAX v0.4.1 in addition to the lower bound of numpy>=1.14.5 by scipy v1.5.1 (the current lowest supported version). Newer versions of scipy place an upper bound on numpy for compatibility reasons and so this is only required for older scipy releases.

Example: `scipy` `v1.9.3` requires `numpy<1.26.0,>=1.18.5`.

This needs to be bumped again for JAX 
```
jax 0.4.1 depends on numpy>=1.20
The user requested (constraint) numpy==1.14.5
```

Need `numpy` `v1.21.0` for

```python
from numpy.typing import ArrayLike, DTypeLike, NBitBase, NDArray
```

to not give

```pytb
ImportError: cannot import name 'NDArray' from 'numpy.typing'
```

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
Set lower bound on numpy to v1.21.0 as this is the lower bound required by JAX v0.4.1
in addition to the lower bound of numpy>=1.14.5 by scipy v1.5.1 (the current lowest
supported version). Newer versions of scipy place an upper bound on numpy for
compatibility reasons and so this is only required for older scipy releases.
```

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -208,26 +208,3 @@ def build(session):
 
     session.install("build")
     session.run("python", "-m", "build")
-
-
-@nox.session(python="3.8", reuse_venv=True)
-def minimum(session: nox.Session):
-    """
-    Run tests with minimum supported dependencies.
-    """
-    session.install("--upgrade", "pip", "setuptools", "wheel")
-    session.install("--constraint", "tests/constraints.txt", ".[test]")
-    session.run(
-        "PYTHONWARNINGS='default'",
-        "pytest",
-        "--override-ini",
-        "filterwarnings=",
-        "--ignore",
-        "tests/contrib",
-        "--ignore",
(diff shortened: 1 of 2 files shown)


matthewfeickert:
This might take a bit to find a minimum version of numpy that works with JAX so I'm putting this into draft mode again.

matthewfeickert:
As other maintainers are on holiday, I'm going to approve and merge this myself. As always, PRs approved by a single core dev can be reverted as needed by the rest of the dev team.

matthewfeickert:
pre-commit is failing for known reasons that will get fixed in a maintenance PR later.

PR #1888, opened 2022-06-13
Chore: use Hatch build backend

agoose77:
# Description

Setuptools as a build backend is getting fairly long in the tooth. A few consequences of this are:
- deprecated APIs
- fragmentation of configuration (setup.py, setup.cfg, MANIFEST.in)
- additional configuration files (besides pyproject.toml)

The new selection of PEP517 builders means that we can simplify our build backend, whilst also making things future-proof!

## TODO:
- [ ] ensure wheel contents match (waiting on https://github.com/pypa/hatch/issues/276)

# Checklist Before Requesting Reviewer

- [ ] Tests are passing
- [ ] "WIP" removed from the title of the pull request
- [ ] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [ ] Summarize commit messages into a comprehensive review of the PR

PR #791, opened 2020-03-04
docs: Fix MANIFEST and put schemas back in documentation

kratsg:
# Pull Request Description

This resolves #686. The solution is to symlink the `src/pyhf/schemas/` to `docs/_extras/schemas` and tell sphinx to copy over everything in the `_extras/` path to the output (via `html_extra_path`).

Additionally, this will also fix the `MANIFEST.in` to be much stricter about which files to include. Previous behavior was including `validation/` and `binder/` and `docs/`. Current behavior does not include those directories (and the corresponding dotfiles we have in our top-level).

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR
```
* MANIFEST.in excludes all directories by default, and grafts in src/
* MANIFEST.in excludes all dotfiles
* Schemas are symlinked under docs/_extras/schemas to src/pyhf/schemas
* Add a check in the ci workflow to ensure the `schemas/` directory is being copied over
```

PR #2095, opened 2023-01-19
build: Switch to using Hatchling as build backend

matthewfeickert:
# Description

Resumes PR #1888 (which can't be reopened as `master` is now deleted).

* Migrate build system to Hatchling and add Hatch configuration in `pyproject.toml`.
   - Use `pip` `v21.2+`'s ability to handle recursive dependencies to define
     nested extras. c.f. https://hynek.me/articles/python-recursive-optional-dependencies/
   - Use `hatch-vcs` to dynamically determine the version information from Git.
   - Use `only-include` option with `[tool.hatch.build.targets.sdist]` to avoid problem with `hatchling` discovering the symlink of `src/pyhf/schemas` under `docs/` before `src/` and skipping including it in the sdist. c.f. PR #791 and https://github.com/pypa/hatch/issues/276
* Remove all `setuptools` related files: `setup.py`, `setup.cfg`, `MANIFEST.in`.
* Add `.flake8` config file as the config for `flake8` had previously been in `setup.cfg` as `flake8` doesn't support `pyproject.toml`.
* Remove use of `check-manifest` from packaging GitHub Actions workflow.
* Remove mention of `setup.cfg` from release checklist.
* Add Scikit-HEP admins info to maintainers metadata.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Migrate build system to Hatchling and add Hatch configuration in
  pyproject.toml.
   - Use pip v21.2+'s ability to handle recursive dependencies to define
     nested extras.
     c.f. https://hynek.me/articles/python-recursive-optional-dependencies/
   - Use hatch-vcs to dynamically determine the version information from Git.
   - Use 'only-include' option with `[tool.hatch.build.targets.sdist]` to avoid
     problem with hatchling discovering the symlink of src/pyhf/schemas under
     docs/ before src/ and skipping including it in the sdist.
     c.f. https://github.com/scikit-hep/pyhf/pull/791 and
          https://github.com/pypa/hatch/issues/276
* Remove all setuptools related files: setup.py, setup.cfg, MANIFEST.in.
* Add .flake8 config file as the config for flake8 had previously been in setup.cfg
  as flake8 doesn't support pyproject.toml.
* Remove use of check-manifest from packaging GitHub Actions workflow.
* Remove mention of setup.cfg from release checklist.
* Add Scikit-HEP admins info to maintainers metadata.

Co-authored-by: Angus Hollands <goosey15@gmail.com>
Co-authored-by: Ofek Lev <oss@ofek.dev>
```


matthewfeickert:
@agoose77 we would like to give you coauthorship on this PR as we couldn't reopen PR #1888, but I see that you have set all of your GitHub activity to private. Given Do you have a different account that you'd like us to associate with this?

agoose77:
@matthewfeickert oops! Try again. I was trying out fewer metrics on my profile, but it appears that there's a tangible inconvenience to it.

pyproject.toml
```diff
@@ -1,11 +1,169 @@
...
+    "sphinx_rtd_theme",
+    "nbsphinx!=0.8.8",  # c.f. https://github.com/spatialaudio/nbsphinx/issues/620
+    "ipywidgets",
+    "sphinx-issues",
+    "sphinx-copybutton>=0.3.2",
+    "sphinx-togglebutton>=0.3.0",
+    "ipython!=8.7.0",  # c.f. https://github.com/scikit-hep/pyhf/pull/2068
+]
+
+devleop = [
```
agoose77, line 114:
```suggestion
develop = [
```

pyproject.toml
```diff
@@ -1,11 +1,169 @@
...
+    "pyhf[test,docs]",
+    "tbump>=6.7.0",
+    "pre-commit",
+    "nox",
+    "codemetapy>=2.3.0",
+]
+
+[tool.hatch]
+version.source = "vcs"
+build.hooks.vcs.version-file = "src/pyhf/_version.py"
```
agoose77, line 124: I'd explode this, as it seems to be best practice.

```suggestion
[tool.hatch.version]
source = "vcs"

[tool.hatch.build.hooks.vcs]
version-file = "src/pyhf/_version.py"
```

agoose77:
Before merging this, we should compare the build metadata between the current wheel and a wheel produced with this PR.

pyproject.toml
```diff
@@ -1,11 +1,171 @@
...
+    "/README.rst",
+    "/pyproject.toml",
+    "/AUTHORS",
+    "/CITATION.cff"
+]
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/pyhf"]
+
+#FIXME: Decide if using hatch envs or staying with lots of extras
```
matthewfeickert, line 144: Need to decide if we can use hatch envs with the current workflows that we have around nox (which I don't want to change in this PR) or if we'll need to remove hatch envs for the time being and make an Issue to decide how to move forward using `nox` and `hatch` in the future.

agoose77: @matthewfeickert perhaps let's drop the goal of using envs in this PR, get it merged, and then re-open that discussion?

matthewfeickert: Yeah I think this is wise for now.

pyproject.toml
```diff
@@ -1,11 +1,143 @@
...
+    "tqdm>=4.56.0",  # for readxml
+    "numpy",  # compatible versions controlled through scipy
+]
+
+[project.scripts]
+pyhf = "pyhf.cli:cli"
+
+[project.urls]
+Documentation = "https://pyhf.readthedocs.io/"
+Homepage = "https://github.com/scikit-hep/pyhf"
```
agoose77, line 56: I probably added this, but I don't think we need it?

pyproject.toml
```diff
@@ -1,11 +1,143 @@
...
+maintainers = [ {name = "The Scikit-HEP admins", email = "scikit-hep-admins@googlegroups.com"} ]
+keywords = [
+    "fitting",
+    "jax",
+    "numpy",
+    "physics",
+    "pytorch",
+    "scipy",
+    "tensorflow",
+]
```
agoose77, line 26: This is properly comma-separated in the generated METADATA, whereas it does not appear so in `main`

pyproject.toml
```diff
@@ -1,11 +1,143 @@
...
+name = "pyhf"
+dynamic = ["version"]
+description = "pure-Python HistFactory implementation with tensors and autodiff"
+readme = "README.rst"
+license = { text = "Apache-2.0" }  # SPDX short identifier
+requires-python = ">=3.8"
+authors = [
+    { name = "Lukas Heinrich", email = "lukas.heinrich@cern.ch" },
+    { name = "Matthew Feickert", email = "matthew.feickert@cern.ch" },
+    { name = "Giordon Stark", email = "gstark@cern.ch" },
```
agoose77, line 15: This information also ends up in the wheel under `LICENSE-FILE` proposed metadata entry (PEP 639), due to AUTHORS existing. That's probably fine.

https://github.com/pypa/hatch/blob/8a9ba2091c2ccb49021de5d44595f535c36331c9/backend/src/hatchling/metadata/core.py#L712

pyproject.toml
```diff
@@ -1,11 +1,143 @@
...
+build-backend = "hatchling.build"
 
-[tool.setuptools_scm]
-write_to = "src/pyhf/_version.py"
-local_scheme = "no-local-version"
+[project]
+name = "pyhf"
+dynamic = ["version"]
+description = "pure-Python HistFactory implementation with tensors and autodiff"
+readme = "README.rst"
```
agoose77, line 9: Do we want to use https://github.com/hynek/hatch-fancy-pypi-readme#readme to remove the badges here? We use it in Awkward, for reference.

pyproject.toml
matthewfeickert: I think this is just the `url` field from `setup.cfg` getting remapped

https://github.com/scikit-hep/pyhf/blob/3eef1fff2e6d4ccf474bbe51e2211cf690752b82/setup.cfg#L6

I'm fine to keep it for the time being.

pyproject.toml
matthewfeickert: Which badges get removed? I haven't looked at how `hatch-fancy-pypi-readme` works yet but comparing https://github.com/scikit-hep/awkward/tree/b6e4c71029152fcd93c1b156d8d4dfb8227b1e24 and https://pypi.org/project/awkward/2.0.6/ it looks like the things stripped out is from "Installation for developers" downwards.

pyproject.toml
matthewfeickert: Yeah I think it is okay to have this duplicated. I really only cared about getting `AUTHORS` in the sdist.

pyproject.toml
matthewfeickert: I think that's okay though? 

https://github.com/scikit-hep/pyhf/blob/3eef1fff2e6d4ccf474bbe51e2211cf690752b82/setup.cfg#L11

shows up as

```
physics, fitting, numpy, scipy, tensorflow, pytorch, jax
```

on https://pypi.org/project/pyhf/0.7.0/ so I would assume that this would render the same with hatchling(?).

pyproject.toml
agoose77: I mean to say that it's now duplicated here, in `"Source Code"` too. Your call!

pyproject.toml
agoose77: Oh yes, I was making the observation that the old package's METADATA seems off-spec, but clearly nothing's broken due to it ;)

pyproject.toml
agoose77: Agreed - another "I noticed this whilst here"

pyproject.toml
matthewfeickert: Yeah we've already been duplicating it:

https://github.com/scikit-hep/pyhf/blob/3eef1fff2e6d4ccf474bbe51e2211cf690752b82/setup.cfg#L14

pyproject.toml
agoose77: You're totally right! I had a memory of removing the badges too, but I think Jim preferred them in there. 

We use the same mechanism though — one can add hidden markers to the `README.rst` / `README.md`, and treat them as boundaries of text chunks that are used to build the final README.

agoose77 (approved):

matthewfeickert:
Just to provide cross referencing, the discussion about how to use nox vs./and hatch for high level developer workflows is moved to Issue #2096. Thanks also to Angus for all of his time explaining different aspects on the IRIS-HEP Slack!

.flake8
```diff
@@ -0,0 +1,6 @@
+[flake8]
```
kratsg, line 1: I think we should use the default ones that come from scikit-hep cookie cutter: https://github.com/scikit-hep/cookie/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.flake8%5Esetuptools%2Cpybind11

pyproject.toml
```diff
@@ -1,11 +1,143 @@
...
+    "/src",
+    "/LICENSE",
+    "/README.rst",
+    "/pyproject.toml",
+    "/AUTHORS",
+    "/CITATION.cff"
+]
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/pyhf"]
```
kratsg, line 140: I don't think this is needed.

pyproject.toml
```diff
@@ -1,11 +1,143 @@
...
+    "pyhf[test,docs]",
+    "tbump>=6.7.0",
+    "pre-commit",
+    "nox",
+    "codemetapy>=2.3.0",
+]
+
+[tool.hatch.version]
+source = "vcs"
+
```
kratsg, line 122: you'll want

```
[tool.hatch.version.raw-options]
local_scheme = "no-local-version"
```

to match what we used to do with `setuptools_scm`.

kratsg (changes requested):

.flake8
matthewfeickert: This PR is just porting them over from `setup.cfg` so that they still exist. If we want to change them we should do that in another PR.

pyproject.toml
matthewfeickert: Hm. From local testing I agree and will remove

```toml
[tool.hatch.build.targets.wheel]
packages = ["src/pyhf"]
```

in this push that I'm about to do, but this was autogenerated for us and I'm a bit confused as I thought this was introduced partly because PR #1888 had problems. @agoose77 am I remembering wrong?

agoose77: IIRC, `packages` is similiar to `only-include`, i.e. it means Hatch won't go looking for things to include automatically. This behavior was causing problems with symlinks. I'd assume that removing this without setting `only-include` will cause that bug to reappear.

ofek: Hatch will automatically do a targeted inclusion https://hatch.pypa.io/latest/plugins/builder/wheel/#default-file-selection

If nothing matches, that last fallback mentioned will do an exhaustive search

matthewfeickert: Ah thanks! I remember what the bug was now: we were hitting a sitution in which our `src/pyhf/schemas` dir was symlinked and wasn't getting picked up and found for our sdist build and so never made it into the wheel.

This PR is currently broken as a result:

```console
$ docker run --rm -ti python:3.10 /bin/bash
root@2812f4d10f06:/# python -m venv venv && . venv/bin/activate
(venv) root@2812f4d10f06:/# python -m pip --quiet install --upgrade pip
(venv) root@2812f4d10f06:/# python -m pip --quiet install build
(venv) root@2812f4d10f06:/# git clone https://github.com/scikit-hep/pyhf.git && cd pyhf
...
(venv) root@2812f4d10f06:/pyhf# git checkout origin/build/migrate-to-hatch -b build/migrate-to-hatch 
...
(venv) root@2812f4d10f06:/pyhf# python -m build .
...
Successfully built pyhf-0.7.1.dev66.tar.gz and pyhf-0.7.1.dev66-py3-none-any.whl
(venv) root@2812f4d10f06:/pyhf# python -m tarfile --list dist/pyhf*.tar.gz | grep schemas  # no schemas dir!
(venv) root@2812f4d10f06:/pyhf# python -m zipfile --list dist/pyhf*.whl | grep schemas  # no schemas dir!
(venv) root@2812f4d10f06:/pyhf# cd ..
(venv) root@2812f4d10f06:/# python -m pip download --no-deps pyhf
Collecting pyhf
  Downloading pyhf-0.7.0-py3-none-any.whl (159 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 159.2/159.2 kB 5.0 MB/s eta 0:00:00
Saved /pyhf-0.7.0-py3-none-any.whl
Successfully downloaded pyhf
(venv) root@2812f4d10f06:/# python -m zipfile --list pyhf-0.7.0-py3-none-any.whl | grep schemas
pyhf/schemas/HistFactorySchema.dtd             2022-09-24 19:11:26         6521
pyhf/schemas/1.0.0/defs.json                   2022-09-24 19:11:26        13069
pyhf/schemas/1.0.0/jsonpatch.json              2022-09-24 19:11:26          180
pyhf/schemas/1.0.0/measurement.json            2022-09-24 19:11:26          184
pyhf/schemas/1.0.0/model.json                  2022-09-24 19:11:26          172
pyhf/schemas/1.0.0/patchset.json               2022-09-24 19:11:26          178
pyhf/schemas/1.0.0/workspace.json              2022-09-24 19:11:26          180
(venv) root@2812f4d10f06:/# 
```

I'll need to fix that.

agoose77: Yes, that was it. @ofek points out another part of the puzzle, which is whether one would normally need to define this config option. In our case, it's a symlink quirk that's nontrivial to fix. Well, it seemed nontrivial, but the back of my mind said it should be possible. Anyway, setting `packages` or `only-include` here should be perfectly workable.

matthewfeickert: Thank you again for PR https://github.com/pypa/hatch/pull/299 @ofek! :D

pyproject.toml
```diff
@@ -1,12 +1,146 @@
...
+[tool.hatch.build.hooks.vcs]
+version-file = "src/pyhf/_version.py"
+
+[tool.hatch.build.targets.sdist]
+exclude = [
+    "/src/conftest.py"
+]
+# Needed to properly include src/pyhf/schemas
+# c.f. https://github.com/pypa/hatch/pull/299
+only-include = [
```
matthewfeickert, line 135: Okay so with the addition of `only-include` the `schemas` dir is getting properly picked up by the sdist.

I'm a bit confused though given the docs https://hatch.pypa.io/latest/config/build/#packages

> The `packages` option is semantically equivalent to `only-include` (which takes precedence) except that the shipped path will be collapsed to only include the final component.
> 
> So for example, if you want to ship a package `foo` that is stored in a directory `src` you would do:
>
> ```toml
> [tool.hatch.build.targets.wheel]
> packages = ["src/foo"]
> ```

what "final component" means. Though there seems to be no difference in behavior between including or not including

```toml
[tool.hatch.build.targets.wheel]
packages = ["src/pyhf"]
```

pyproject.toml
matthewfeickert: https://github.com/scikit-hep/pyhf/pull/2095/files#r1081956320

pyproject.toml
agoose77: `packages` is like` only-include` combined with `sources`, where `sources` would be `sources = ["src"]` for this example. It means that `packages = ["src/this/that/foo/bar"]`  would behave like 
```toml
only-include = ["src/this/that/foo/bar"]
sources = ["src/this/that/foo"[
```
i.e. the final path component (`bar`) is the name of the package as included in the final wheel (because it's taken relative to `sources`).

matthewfeickert:
```console
$ rm -rf dist && python -m build . && python -m tarfile --list dist/pyhf*.tar.gz | sed 's|[^/]*/||' > /tmp/sdist-without.txt && python -m zipfile --list dist/pyhf*.whl | sed 's|[^/]*/||' > /tmp/wheel-without.txt
$ git revert 4a4da3e0
$ rm -rf dist && python -m build . && python -m tarfile --list dist/pyhf*.tar.gz | sed 's|[^/]*/||' > /tmp/sdist-with.txt && python -m zipfile --list dist/pyhf*.whl | sed 's|[^/]*/||' > /tmp/wheel-with.txt
$ diff /tmp/sdist-without.txt /tmp/sdist-with.txt  # nothing
$ diff /tmp/wheel-without.txt /tmp/wheel-with.txt  # nothing
```

matthewfeickert: > i.e. the final path component (`bar`) is the name of the package as included in the final wheel (because it's taken relative to `sources`).

So is the reason that removing

```toml
[tool.hatch.build.targets.wheel]
packages = ["src/pyhf"]
```

didn't have any effect is because we're using `only-include` for `[tool.hatch.build.targets.sdist]` and we only have one package directory in `src/`?

agoose77: OK, the original issue that we were fixing with `only-include` has been fixed (I think) by copying the schemas into the docs at docs time: https://github.com/scikit-hep/pyhf/commit/81a429ec8888263d69a97da40ff03da5ed4f6dda

So we don't need this explicitly. I found that reverting `9ca5b12` didn't leave the schemas out of the sdist or wheel for me. If you revert `9ca5b12` and keep `4a4da3e`, what do you see?

ofek: Configuration of build targets doesn't conflict with nor influence other targets. The reason omitting that from the wheel results in the same contents is because the file `src/pyhf/__init__.py` exists and therefore Hatchling determines that the default inclusion option for the wheel target should be as you had it `packages = ["src/pyhf"]`

- https://hatch.pypa.io/latest/plugins/builder/wheel/#default-file-selection
- https://github.com/pypa/hatch/blob/hatchling-v1.12.2/backend/src/hatchling/builders/wheel.py#L172-L200

agoose77: Yep, I think we're all on the same page! @ofek serious question, do you ever sleep? :eyes: ;)

matthewfeickert: > OK, the original issue that we were fixing with `only-include` has been fixed (I think) by copying the schemas into the docs at docs time

No, copying the schemas into the docs was added way back in PR #458.

> i.e., I think the default Hatch build configuration should pick everything up. But, if you want to use e.g. only-include, you can for other reasons

Unfortunately no. :/ If we look at the changes between now (`f3ebf0e6`) and before we started poking at `[tool.hatch.build.targets.wheel]` when @kratsg wondered if it was needed

```diff
$ git diff 1bf9c0e9 f3ebf0e6 
diff --git a/pyproject.toml b/pyproject.toml
index 90e2d9c7..73aebb4c 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -130,7 +130,9 @@ version-file = "src/pyhf/_version.py"
 exclude = [
     "/src/conftest.py"
 ]
-include = [
+# only-include needed to properly include src/pyhf/schemas
+# c.f. https://github.com/pypa/hatch/pull/299
+only-include = [
     "/src",
     "/LICENSE",
     "/README.rst",
@@ -139,9 +141,6 @@ include = [
     "/CITATION.cff"
 ]
 
-[tool.hatch.build.targets.wheel]
-packages = ["src/pyhf"]
-
 [tool.black]
 line-length = 88
 target-version = ['py38', 'py39', 'py310']
```

```
$ git reset --hard 1bf9c0e9
```

then we have

```toml
...

[tool.hatch.build.targets.sdist]
exclude = [
    "/src/conftest.py"
]
include = [
    "/src",
    "/LICENSE",
    "/README.rst",
    "/pyproject.toml",
    "/AUTHORS",
    "/CITATION.cff"
]

[tool.hatch.build.targets.wheel]
packages = ["src/pyhf"]

...
```

With this we are not getting the `schemas/` directory included in the sdist

```console
$ rm -rf dist && python -m build . && python -m tarfile --list dist/pyhf*.tar.gz | grep schemas  # nothing
```

If we now 

```
git cherry-pick 9ca5b123
```

to use `only-include` then the schemas are included

```console
$ rm -rf dist && python -m build . && python -m tarfile --list dist/pyhf*.tar.gz | grep schemas
...
pyhf-0.7.1.dev66/src/pyhf/schemas/HistFactorySchema.dtd 
pyhf-0.7.1.dev66/src/pyhf/schemas/1.0.0/defs.json 
pyhf-0.7.1.dev66/src/pyhf/schemas/1.0.0/jsonpatch.json 
pyhf-0.7.1.dev66/src/pyhf/schemas/1.0.0/measurement.json 
pyhf-0.7.1.dev66/src/pyhf/schemas/1.0.0/model.json 
pyhf-0.7.1.dev66/src/pyhf/schemas/1.0.0/patchset.json 
pyhf-0.7.1.dev66/src/pyhf/schemas/1.0.0/workspace.json
```

regardless of if we have 

```toml
[tool.hatch.build.targets.wheel]
packages = ["src/pyhf"]
```

or not

```console
$ git reset --hard f3ebf0e6
$ rm -rf dist && python -m build . && python -m tarfile --list dist/pyhf*.tar.gz | grep schemas
...
pyhf-0.7.1.dev68/src/pyhf/schemas/HistFactorySchema.dtd 
pyhf-0.7.1.dev68/src/pyhf/schemas/1.0.0/defs.json 
pyhf-0.7.1.dev68/src/pyhf/schemas/1.0.0/jsonpatch.json 
pyhf-0.7.1.dev68/src/pyhf/schemas/1.0.0/measurement.json 
pyhf-0.7.1.dev68/src/pyhf/schemas/1.0.0/model.json 
pyhf-0.7.1.dev68/src/pyhf/schemas/1.0.0/patchset.json 
pyhf-0.7.1.dev68/src/pyhf/schemas/1.0.0/workspace.json
```

So I'm a bit confused as to why it doesn't have any effect even when building without an sdist first

```
# This is without it
$ rm -rf dist && python -m build --wheel . && python -m zipfile --list dist/pyhf*.whl | grep schemas
* Creating virtualenv isolated environment...
* Installing packages in isolated environment... (hatch-vcs, hatchling>=1.0.0)
* Getting build dependencies for wheel...
* Building wheel...
Successfully built pyhf-0.7.1.dev68-py3-none-any.whl
pyhf/schemas/HistFactorySchema.dtd             2020-02-02 00:00:00         6521
pyhf/schemas/1.0.0/defs.json                   2020-02-02 00:00:00        13069
pyhf/schemas/1.0.0/jsonpatch.json              2020-02-02 00:00:00          180
pyhf/schemas/1.0.0/measurement.json            2020-02-02 00:00:00          184
pyhf/schemas/1.0.0/model.json                  2020-02-02 00:00:00          172
pyhf/schemas/1.0.0/patchset.json               2020-02-02 00:00:00          178
pyhf/schemas/1.0.0/workspace.json              2020-02-02 00:00:00          180
```

but if you would suggest putting it in just to guard in the future I'm okay with doing so if @kratsg doesn't have strong objections.

matthewfeickert: ...thought it might be worth noting that @henryiii's `cookie` doesn't use `tool.hatch.build` at all

```console
...
(venv) root@882d0e43004d:/# cookiecutter gh:scikit-hep/cookie
project_name [package]: 
org [Scikit-HEP]: 
url [https://github.com/Scikit-HEP/package]: 
full_name [My Name]: 
Select project_type:
1 - hatch
2 - setuptools
3 - setuptools621
4 - pybind11
5 - skbuild
6 - poetry
7 - flit
8 - pdm
9 - trampolim
10 - whey
11 - maturin
Choose from 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 [1]: 1
email [me@email.com]: 
project_short_description [A great package.]: 
year [2023]: 
Select license:
1 - BSD
2 - Apache
3 - MIT
Choose from 1, 2, 3 [1]: 
(venv) root@882d0e43004d:/# grep -r "tool.hatch" package/
package/pyproject.toml:[tool.hatch]
...
```

agoose77: You're right, I didn't interrogate the blame deeply enough. Probably just the move to nox registered in my mind as "added in this commit". OK.

I did some more digging, and indeed we still have the same problem as the original PR: `docs/_extras/schemas` is a symlink, so it would get resolved, and then the `pyhf/schemas` would be missing. Indeed, I only see the docs schemas when I delete `only-include`.

So, we need either `packages` or `only-include` here, unless we decide to delete the symlink in the repo. This would mean that the sdist would not include these schemas, though, which might not be desirable. 

@matthewfeickert can we delete these and have the `docs` target copy them too?

matthewfeickert: > If you don't pass any arguments it builds the source distribution then unpacks that to build the wheel whereas if you pass the flag to just build the wheel it will do so from the project root directly

Yes, this was clear to me. Sorry, I should have mentioned something like the last bit of my comment in https://github.com/scikit-hep/pyhf/pull/2095#discussion_r1082010630 earlier or made this more explicit than just grepping the sdist contents. :grimacing: 

> Configuration of build targets doesn't conflict with nor influence other targets. The reason omitting that from the wheel results in the same contents is because the file `src/pyhf/__init__.py` exists and therefore Hatchling determines that the default inclusion option for the wheel target should be as you had it `packages = ["src/pyhf"]`
> 
> * https://hatch.pypa.io/latest/plugins/builder/wheel/#default-file-selection
> * https://github.com/pypa/hatch/blob/hatchling-v1.12.2/backend/src/hatchling/builders/wheel.py#L172-L200

..and _this_ answers my confused questions about `[tool.hatch.build.targets.wheel]`. :+1: Sorry I should have read the docs much more carefully.

> When the **user has not set any [file selection](https://hatch.pypa.io/latest/config/build/#file-selection) options**, the [project name](https://hatch.pypa.io/latest/config/metadata/#name) will be used to determine the package to ship in the following heuristic order

...and this also motivates me to say that even though it isn't _necessary_ it is still worth including

```toml
[tool.hatch.build.targets.wheel]
packages = ["src/pyhf"]
```

as it means that `hatchling` doesn't have to do any additional logic checks to figure anything out. :tada: 

So I'm going to add that back in. :+1: 

Thanks very much for your patience @agoose77 and thank you very much for your time and help @ofek! :pray:

pyproject.toml
matthewfeickert: Moved to Issue #2097

matthewfeickert:
@kratsg this is ready for review again. I left some of the discussions open instead of resolving them in the event the context helps.

pyproject.toml
agoose77: @matthewfeickert I think you should still need that option. Hatch, for example, has `hatch build`, and PEP 517 says
> To ensure that wheels from different sources are built the same way, frontends may call build_sdist first, and then call build_wheel in the unpacked sdist. But if the backend indicates that it is missing some requirements for creating an sdist (see below), the frontend will fall back to calling build_wheel in the source directory.

I presume @ofek that `hatch build` is directly generating a wheel without an intermediate sdist, right? (If I drop the `packages` configuration for `targets.wheel`, the missing-schemas phenomenon (as a symptom for no `include-only`/`packages` ) appears again). Perhaps we should change that in Hatch, as Hatch is a PEP 517 frontend (technically, even if only `hatchling` is supported right now)? If you agree, I can open an issue and get back to it in a couple of months.

Regardless of whether hatch will render this unneeded, let's keep it.

matthewfeickert: > It doesn't proscribe that one _must_ build a wheel from an sdist.

Correct. At the end of my rambling comment https://github.com/scikit-hep/pyhf/pull/2095#discussion_r1082010630 I showed (at the "So I'm a bit confused as to why it doesn't have any effect even when building without an sdist first" part) that without building an sdist first 

```console
$ python -m build --wheel .
```

would correctly pickup the `schemas/` dir regardless of if 

```toml
[tool.hatch.build.targets.wheel]
packages = ["src/pyhf"]
```

was present or not thanks to hatchling's default file selection working as intended (https://github.com/scikit-hep/pyhf/pull/2095#discussion_r1081998703).

> Regardless of whether hatch will render this unneeded, let's keep it.

Agreed.

agoose77: I'm just being verbose here for ease. This might be your understanding too, but I want to be sure we are all on the same page! :)

Python `build --wheel` is doing the recommended thing of `sdist → wheel`. Hatch's default file inclusion logic always picks up the `pyhf/schemas` directory, but the symlinked schemas in `docs/_extras/` cause this to fail. Eliminating these from the SDist (manually, with `only-include` et al.) means that the wheel builder then succeeds with default inclusion logic, assuming that indeed we go `sdist → wheel`. 

Meanwhile, `hatch build --target wheel` doesn't do `sdist → wheel`, and therefore the wheel build suffers the same problem as the sdist; the schemas are found, but encountered under the wrong (excluded) path (`docs/_extras`)

So we definitely need an `only-include` like directive in both `targets.sdist` and `targets.wheel` whilst we have these duplicate schemas.

matthewfeickert: > Python `build --wheel` is doing the recommended thing of `sdist → wheel`.

I don't think that's exactly correct. From `python -m build --help`:

```
    By default, a source distribution (sdist) is built from {srcdir}
    and a binary distribution (wheel) is built from the sdist.
    This is recommended as it will ensure the sdist can be used
    to build wheels.

    Pass -s/--sdist and/or -w/--wheel to build a specific distribution.
    If you do this, the default behavior will be disabled, and all
    artifacts will be built from {srcdir} (even if you combine
    -w/--wheel with -s/--sdist, the wheel will be built from {srcdir}).
```

so my `python -m build --wheel .` is building from the srcdir, and not from sdist (the default behavior).

agoose77: Gah, then "`Python build --wheel` is _not_ doing the recommended thing of `sdist → wheel`" :laughing: 

Admission time ... I discovered my environment didn't fully upgrade; I was evaluating `hatch` behaviour in-part based upon CLI exploration. I upgraded `hatchling` to the newest version explicitly, and the different results that I was seeing evaporated. I had been confused that we seemed to be drawing different conclusions.

Now Ofek's points make more sense — for the SDist, we need to curtail the default search for the above reasons, but the wheel builder works fine due to the comment from Ofek that you linked!

So my view is different now; we don't need that setting, and we only added it because it fixes an annoying build problem. As long as hatchling's PEP 517 hook produces a good wheel when used without an intermediate sdist (which it does), it should be fine to remove it. So, your call, but I think you can safely remove it!

Sorry for the wild goose chase. I had thought I was being helpful, but with the above context I can see it was actually just confusing things!

matthewfeickert: Nothing to apologize about, you have been beyond helpful. :)

> Now Ofek's points make more sense — for the SDist, we need to curtail the default search for the above reasons, but the wheel builder works fine due to the comment from Ofek that you linked!

:heavy_check_mark: :+1: We are in agreement on understanding now!

> As long as hatchling's PEP 517 hook produces a good wheel when used without an intermediate sdist (which it does), it should be fine to remove it. So, your call, but I think you can safely remove it!

We _can_ remove it but keeping it in skips all of

https://github.com/pypa/hatch/blob/hatchling-v1.12.2/backend/src/hatchling/builders/wheel.py#L172-L200

which while I'm sure is only a few microseconds, seems like a "why not just leave in 2 lines?" question to me now.

ofek: Hatchling currently does not build the wheel from the source distribution. I plan to do that but I don't have time currently. It would actually be useful if you did open an issue so I can track it

kratsg (approved):

Issue #566, opened 2019-09-17
allow data-less workspaces


lukasheinrich:
# Description

some workspaces do not include data (think blinded workspaces) and instead just describe the pdf. These workspaces are still useful as they allow us to do expected limits etc..

kratsg:
This is related to #514. I have ideas on #514 though :)

alexander-held:
Hi, adding some more detail to this, since I was writing it up to open an issue before finding this one.

Workspaces without histograms specified in the `Data` field in a channel cannot be read by `pyhf xml2json`. Example:

```xml
<Data HistoName="" InputFile="" HistoPath=""/>
```

Such workspaces are commonly used before unblinding an analysis. It is possible to fit the Asimov dataset with them, since all the required information for that is in the remaining histograms. The code tries to interpret the `""` as a path to a histogram, and then fails with a `IsADirectoryError` via `uproot`.

Minimal example here: `/afs/cern.ch/user/a/alheld/public/pyhf_asimov`, run via `pyhf xml2json minimal_example/RooStats/minimal_example.xml`.

One of the following two options might be good:

- For this specific case of a missing observation histogram, fill all entries with `-1`. This is easy to catch by the user, could be accompanied by a warning, and allows the fit model to be built through `pyhf`. This behavior should probably not exist for other histograms that define the model and might be missing, since that would alter the model and could lead to unexpected behavior.

- Catch the issue and issue a more descriptive error message. The user would then need to build a different `.xml` input to be able to use `pyhf`. This requires manual intervention by the user, but could at least guide them to the right thing to do, as the error otherwise is not easy to understand.

PR #1931, opened 2022-08-10
fix: Coverage for importing XML channels with no data

kratsg:
# Description

Handle cases where channel is missing data in the XML configuration (associated with Asimov). c.f. Issue #1911

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR
```
* Raise RuntimeError for cases where channel is missing data in the XML configuration
(associated with Asimov).
   - c.f. Issue #1911
* Add tests and test XML files.
```

Issue #1911, opened 2022-07-02
Support asimov data in the HiFa spec


kratsg:
### Summary

As it turns out, the HistFactory XML spec [allows](https://github.com/scikit-hep/pyhf/blob/acde7f4ff8d0db2351f5d6e31ff5584e34da0cf0/src/pyhf/schemas/HistFactorySchema.dtd#L63-L76) for `<Channel />` with no `<Data />`. In those cases, it seems that [Asimov data](https://github.com/scikit-hep/pyhf/blob/acde7f4ff8d0db2351f5d6e31ff5584e34da0cf0/src/pyhf/schemas/HistFactorySchema.dtd#L78-L86) is used instead.

We don't currently do this in `pyhf`. Actually, we haven't run into a spec from a user yet where `<Data/>` was missing so we never caught the uncovered case that I found in `readxml.py` as part of #1909 while I'm adding typehints there. Specifically: https://github.com/scikit-hep/pyhf/blob/acde7f4ff8d0db2351f5d6e31ff5584e34da0cf0/src/pyhf/readxml.py#L229-L233 - `parsed_data = None` can happen which means that when we write out the `observations`, we will have `'data': None` which is not allowed by the HiFa JSON spec right now (but it technically should be).

In the situation when there's no data for a channel, we should probably generate asimov data for it, but this feels very "magical" to me, but maybe functionality users will want, even if it appears that most frameworks (exporting to XML+ROOT) are not using this right now.

### Additional Information

```python (paste below)

```

### Code of Conduct

- [X] I agree to follow the Code of Conduct

alexander-held:
See also #566 which includes an example. I have seen the `<Data>` tag being kept in these cases as well, without any path information provided.

When considering the addition of automatic Asimov dataset generation, there needs to be a method for picking the Asimov values of free-floating parameters (assuming that "pre-fit Asimov" would be used, and I think that generally makes sense here since data might not only be missing from some channels, but also from all of them so there is no way to perform a partial fit to actual data first).

kratsg:
For now, current solution is to raise a `RuntimeError`.

PR #2109, opened 2023-02-09
fix: Raise error when converting xml workspaces without data

alexander-held:
# Description

Raise an error when attempting to convert an xml workspace that contains no data, i.e. looks like this:
```xml
<Data HistoName="" InputFile="" HistoPath=""/>
```

Previously this would raise an `IsADirectoryError` via `uproot`, which is confusing to users. See #566 for context.

I have not added any new test for this, as I haven't seen anything specifically targeting that API at the moment. I confirmed it raises as expected, an example is linked in https://github.com/scikit-hep/pyhf/issues/566#issuecomment-700573969.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Raise NotImplementedError when attempting to convert an xml workspace that
  contains no data. The previous behavior was uproot raising IsADirectoryError,
  which is confusing.
* Add a test for raising NotImplementedError in tests/test_import.py and add the
  test files.
```

### DIFF
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -500,7 +500,7 @@ def test_import_noChannelDataPaths(mocker, datadir):
     with pytest.raises(NotImplementedError) as excinfo:
         pyhf.readxml.parse(basedir.joinpath("config/example.xml"), basedir)
     assert (
-        'conversion of workspaces without data is currently not supported, see issue #566'
+        "Conversion of workspaces without data is currently not supported.\nSee https://github.com/scikit-hep/pyhf/issues/566"
         in str(excinfo.value)
     )
 
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -497,12 +497,11 @@ def test_import_noChannelDataPaths(mocker, datadir):
     mocker.patch('pyhf.readxml.import_root_histogram', return_value=(_data, _err))
 
     basedir = datadir.joinpath("xmlimport_noChannelDataPaths")
-    with pytest.raises(NotImplementedError) as excinfo:
+    with pytest.raises(
+        NotImplementedError,
+        match="Conversion of workspaces without data is currently not supported.\nSee https://github.com/scikit-hep/pyhf/issues/566",
+    ):
         pyhf.readxml.parse(basedir.joinpath("config/example.xml"), basedir)
-    assert (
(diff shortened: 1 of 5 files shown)


alexander-held:
I realized that this affects lots of tests that probably relied on the function just doing nothing, so I'll look at a better way to implement.

matthewfeickert (changes requested):
Thanks @alexander-held. We should add a test that covers this use case so that we don't lose coverage, but overall this looks good. :+1:

alexander-held:
Do you just want a unit test or something more integration-like? I didn't find any related tests for this part actually, so I was not sure where to best add something.

matthewfeickert:
Just a unit test that covers the `raise NotImplementedError` line which codecov says isn't' fully covered at the moment.

alexander-held:
While looking where to put a test I came across `test_import_noChannelData` and #1911 / #1931 (cc @kratsg) which I had completely forgotten about. Not sure yet why this did not catch my example, will have a look.

tests/test_import.py
```diff
@@ -488,7 +488,21 @@ def test_import_noChannelData(mocker, datadir):
...
-        assert 'Channel channel1 is missing data. See issue #1911' in str(excinfo.value)
+    assert 'Channel channel1 is missing data. See issue #1911' in str(excinfo.value)
+
+
+def test_import_noChannelDataPaths(mocker, datadir):
+    _data = [0.0]
+    _err = [1.0]
+    mocker.patch('pyhf.readxml.import_root_histogram', return_value=(_data, _err))
+
+    basedir = datadir.joinpath("xmlimport_noChannelDataPaths")
```
alexander-held, line 499: This duplicates the setup used for #1911 (other than a one line change to the xml) which isn't really ideal, but I figured it's not really better to copy the files to a temporary directory and edit the file there either.

matthewfeickert:
@alexander-held sorry this hasn't been merged yet. It is late my time in Canada right now, but I have this on my list of things to do tomorrow during conference breaks.

matthewfeickert (approved):
LGTM. Thanks @alexander-held and sorry about the months of delay.

tests/test_import.py
kratsg: This is duplicating #1931, not #1911. It's better to just have this duplicated/hard-coded. I don't see why this is an issue.

kratsg (approved):

Issue #2096, opened 2023-01-19
Explore hatch vs. nox for managing high level developer workflows


matthewfeickert:
Need to decide if we can use hatch envs with the current workflows that we have around nox (which I don't want to change in this PR) or if we'll need to remove hatch envs for the time being and make an Issue to decide how to move forward using `nox` and `hatch` in the future.

_Originally posted by @matthewfeickert in https://github.com/scikit-hep/pyhf/pull/2095#discussion_r1080896032_

ofek:
Let me know if you have any questions! Environment management and configuration is probably the feature I put the most thought and design work into so I'm interested to see how this goes 🙂

matthewfeickert:
Based on https://github.com/scikit-hep/pyhf/pull/2118#issuecomment-1432584212 @kratsg and @agoose77 do you have any clever thoughts on how to determine if `tensorflow` or `tensorflow-macos` should be installed at runtime depending on if the user is on an Apple silicon Mac or not while using `hatch` environments?

ofek:
Do [environment markers](https://hatch.pypa.io/latest/config/dependency/#environment-markers) not work for this use case?

matthewfeickert:
> Do [environment markers](https://hatch.pypa.io/latest/config/dependency/#environment-markers) not work for this use case?

Yes, yes they will. Thanks @ofek! (I actually _do_ generally read the docs for projects, but I also usually post these GitHub comments as reminders to do so when it isn't after midnight local time for me).

The way that `hatchling` does this is actually super nice, so many thanks for all the work that went into this design. :+1:

matthewfeickert:
`hatchling` environment markers now enabling PR #2119. :rocket:

PR #2118, opened 2023-02-15
docs: Add FAQ on backend support on Apple silicon Macs 

### LINKED ISSUE #2111: Cannot resolve dependencies in "all" setup extra
### Summary

Installing the "all" setup extra (which I assume is the replacement for "complete", see #2110) currently results in `pip` struggling to find compatible versions of dependencies.

### OS / Environment

```console
python:3.9 via Docker
```

### Steps to Reproduce

```console
> docker run -it --rm python:3.9 /bin/bash
$ pip install pyhf[all]@git+https://github.com/scikit-hep/pyhf.git
```

### File Upload (optional)

_No response_

### Expected Results

I expect the resolution of compatible dependencies to happen reasonably quickly.

### Actual Results

pip keeps downgrading numpy/scipy versions.

### pyhf Version

```console
main@2b168b7
```

### Code of Conduct

matthewfeickert:
# Description

Superseded by PR #2119

---

Resolves #2111

At the moment, [TensorFlow does not have Apple silicon compatible wheels on PyPI](https://github.com/tensorflow/tensorflow/issues/57185) under the `tensorflow` project namespace, though it does under `tensorflow-macos`. As an alternative, note that conda-forge has `linux-64`, `osx-64`, and `osx-arm64` compatible releases for all of pyhf's backends.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* At the moment, TensorFlow does not have Apple silicon compatible
  wheels on PyPI. As an alternative, note that conda-forge has linux-64,
  osx-64, and osx-arm64 compatible releases for all of pyhf's backends.
   - c.f. https://github.com/tensorflow/tensorflow/issues/57185
```

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -33,15 +33,15 @@ For more information see :code:`pyhf cls --help`.
 Why can't ``pip`` find compatible backends on PyPI for Apple silicon Macs?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Unfortunately, at this time |tensorflow Issue 57185|_ of ``pyhf``'s backends support
-wheels for Apple silicon Macs (``osx-arm64``).
-So if you are installing wheels from PyPI on an ``osx-arm64`` machine, you will
-only be able to use a subset of all the backends.
+Unfortunately, at this time |tensorflow Issue 57185|_ of ``pyhf``'s backends
+support wheels for Apple silicon Macs (``osx-arm64``).
+If you are installing wheels from PyPI on an ``osx-arm64`` machine, you will only
+be able to use a subset of all the backends.
 
 An alternative would be to create a |micromamba|_ environment and install the
 backends individually from `conda-forge <https://prefix.dev/channels/conda-forge/>`__
-as all the backends have conda-forge releases compatible with ``linux-64`` and
-``osx-arm64``.
+as all the backends have conda-forge releases compatible with ``linux-64``,
+``osx-64``, and ``osx-arm64``.
 
 .. |tensorflow Issue 57185| replace:: not all
 .. _`tensorflow Issue 57185`: https://github.com/tensorflow/tensorflow/issues/57185
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -30,6 +30,25 @@ Use the :code:`--backend` option for :code:`pyhf cls` to specify a tensor backen
 The default backend is NumPy.
 For more information see :code:`pyhf cls --help`.
 
+Why can't ``pip`` find compatible backends on PyPI for Apple silicon Macs?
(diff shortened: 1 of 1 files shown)


matthewfeickert:
Relevant RTD build: https://pyhf--2118.org.readthedocs.build/en/2118/faq.html#why-can-t-pip-find-compatible-backends-on-pypi-for-apple-silicon-macs

alexander-held (commented):
You can also install `tensorflow-macos` (and `tensorflow-metal`) via `pip`, but I guess that's discussed in the issue you linked anyway.

matthewfeickert:
> You can also install `tensorflow-macos` (and `tensorflow-metal`) via `pip`, but I guess that's discussed in the issue you linked anyway.

Ah yeah I had forgotten about the Metal project to get TensorFlow to work on Mac GPUs. Thanks for the reminder, I'll mention something about them. Though I guess related to that, @alexander-held on your Apple silicon Mac are you able to do the following in your virtual environment that you tried to use for developing in Issue #2111?

```
# After cloning things locally
$ python -m pip install --upgrade tensorflow-macos
$ python -m pip install --upgrade '.[develop]'
```

matthewfeickert:
> on your Apple silicon Mac are you able to do the following in your virtual environment that you tried to use for developing in Issue https://github.com/scikit-hep/pyhf/issues/2111?

Ah, no I realize that this will fail regardless. Even though `tensorflow-macos` should install fine, we'll still run into trouble here as

```
python -m pip install --upgrade '.[develop]'
```

invokes the `develop` extra which includes the `backends` extra which includes the `tensorflow` extra

https://github.com/scikit-hep/pyhf/blob/29fb68ecfc8be3011fe78096c5da225ed4cfb794/pyproject.toml#L64-L67

and so it will still try to do a `pip` install of `tensorflow` as that's going to be in the requested list, which will fail.

Normally this is where we would have some code to dynamically change determine the package name to use here based off of `os.uname().machine` or something, but we can't do that with TOML. I think this is again where using `hatchling` forces us into using `hatch` to setup environment workflows rather than using extras. We have Issue #2096 for that but we haven't done further investigation of how to orchestrate this.

docs/faq.rst
```diff
@@ -30,6 +30,40 @@ Use the :code:`--backend` option for :code:`pyhf cls` to specify a tensor backen
...
+
+        python -m pip install pyhf tensorflow-macos tensorflow-probability
+
+Similarly, if you want to install the TensorFlow backend dependencies along with the
+other backends included in the ``[backends]`` extra you'll need to specify the extras
+as well
+
+    .. code-block:: console
+
+        python -m pip install 'pyhf[torch,jax,minuit]' tensorflow-macos tensorflow-probability
```
matthewfeickert, line 54: @alexander-held I think this should work to actually get things installed. Though I think it is going to be annoying to get a full developer environment setup at the moment given that the `test` extra also pulls in the `backends` extra and so will break here.

PR #2124, opened 2023-02-28
chore: Move to using Ruff for pre-commit

henryiii:
# Description

This moves the flake8 & pyupgrade checks into a single Ruff check, and moves the config to pyproject.toml. There are a lot of checks that could be added (see https://scikit-hep.org/developer/style#ruff), I've just stuck to replicating the current config. The one loss is flake8-encodings, someone could request this added to Ruff.

I've skipped the failing checks for notebooks, might be worth checking to see if they are valid.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Drop flake8, yesqa, pyupgrade, and nbqa-pyupgrade pre-commit hooks in favor of Ruff.
   - Remove .flake8 config file as no longer used.
   - Remove nbqa config options for pyupgrade in pyproject.toml and add ruff config options.
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,11 +26,6 @@ repos:
       # exclude generated files
       exclude: ^validation/|\.dtd$|\.xml$
 
--   repo: https://github.com/MarcoGorelli/absolufy-imports
-    rev: v0.3.1
-    hooks:
-    - id: absolufy-imports
-
 - repo: https://github.com/charliermarsh/ruff-pre-commit
   rev: "v0.0.253"
   hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,16 +26,16 @@ repos:
       # exclude generated files
       exclude: ^validation/|\.dtd$|\.xml$
 
--   repo: https://github.com/asottile/pyupgrade
-    rev: v3.3.1
-    hooks:
-    - id: pyupgrade
-      args: ["--py38-plus"]
-
 -   repo: https://github.com/MarcoGorelli/absolufy-imports
     rev: v0.3.1
     hooks:
     - id: absolufy-imports
+
+- repo: https://github.com/charliermarsh/ruff-pre-commit
+  rev: "v0.0.253"
+  hooks:
+    - id: ruff
+      args: ["--fix", "--show-fixes"]
 
 -   repo: https://github.com/psf/black
     rev: 23.1.0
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -48,18 +48,6 @@ repos:
     - id: blacken-docs
       additional_dependencies: [black==23.1.0]
 
--   repo: https://github.com/asottile/yesqa
-    rev: v1.4.0
-    hooks:
-    - id: yesqa
```
henryiii, line 54: Oops, forgot to enable the matching RUFF check.

kratsg: Is that RUF100 and PGH004 ? I couldn't figure it out... Ruff says it can replace `yesqa` but doesn't provide guidance on what that looks like.

src/pyhf/cli/spec.py
```diff
@@ -108,7 +108,7 @@ def inspect(workspace, output_file, measurement):
     )
 
     # summary
-    fmtStr = '{{0: >{0:d}s}}  {{1:s}}'.format(maxlen + len('Summary'))
+    fmtStr = '{{: >{:d}s}}  {{:s}}'.format(maxlen + len('Summary'))
```
henryiii, line 111: Oops, this is a Ruff mistake, should be reported! :)

henryiii: Oh, actually, I think it isn't. Only one of these is a substitution (the second 0), and the resulting string then will have a 0 and a 1 - which will then work. So Ruff is correct.

.pre-commit-config.yaml
henryiii: It's RUF100. PGH replaces pygrep-hooks, which looks for blanket noqa (no code).

henryiii:
I've opened an issue about flake8-encodings in  https://github.com/charliermarsh/ruff/issues/3278.

matthewfeickert:
Thanks for this PR @henryiii and sorry that I'm so slow in getting to it.

> I've skipped the failing checks for notebooks, might be worth checking to see if they are valid.

Which checks are these? It looks like everything is passing. Or am I missing something that you are now skipping in the PR?

kratsg (approved):

.pre-commit-config.yaml
```diff
@@ -26,16 +26,11 @@ repos:
...
-    hooks:
-    - id: pyupgrade
-      args: ["--py38-plus"]
-
--   repo: https://github.com/MarcoGorelli/absolufy-imports
-    rev: v0.3.1
-    hooks:
-    - id: absolufy-imports
+- repo: https://github.com/charliermarsh/ruff-pre-commit
+  rev: "v0.0.253"
```
matthewfeickert, line 30:
```suggestion
  rev: "v0.0.255"
```

just bumping to latest version.

matthewfeickert (approved):

pyproject.toml
```diff
@@ -222,12 +222,6 @@ precision = 1
 sort = "cover"
 show_missing = true
 
-[tool.nbqa.mutate]
-pyupgrade = 1
```
matthewfeickert, line 226: A note to self, the reason this is safe to remove is that the option was deprecated in `v1.0.0`. c.f. https://github.com/nbQA-dev/nbQA/blob/26f3ce04cd29bdfcda487fefdbae78925e80a5ff/nbqa/cmdline.py#L37

Issue #1759, opened 2022-02-04
TensorFlow v2.8.0 fails test_simple_tensor_op on macOS


matthewfeickert:
### Summary

[`tensorflow` `v2.8.0`](https://github.com/tensorflow/tensorflow/releases/tag/v2.8.0) is causing an `AssertionError` on macOS only because of floating point disagreement in 

https://github.com/scikit-hep/pyhf/blob/270d59d3900fdef7549015db45260e85f83e33b6/tests/test_tensor.py#L54

```pytb
=================================== FAILURES ===================================
______________________ test_simple_tensor_ops[tensorflow] ______________________

backend = (<pyhf.tensor.tensorflow_backend.tensorflow_backend object at 0x11ab80580>, None)

    def test_simple_tensor_ops(backend):
        tb = pyhf.tensorlib
        assert tb.tolist(tb.astensor([1, 2, 3]) + tb.astensor([4, 5, 6])) == [5, 7, 9]
        assert tb.tolist(tb.astensor([1]) + tb.astensor([4, 5, 6])) == [5, 6, 7]
        assert tb.tolist(tb.astensor([1, 2, 3]) - tb.astensor([4, 5, 6])) == [-3, -3, -3]
        assert tb.tolist(tb.astensor([4, 5, 6]) - tb.astensor([1])) == [3, 4, 5]
        assert tb.tolist(tb.sum(tb.astensor([[1, 2, 3], [4, 5, 6]]), axis=0)) == [5, 7, 9]
        assert tb.tolist(tb.product(tb.astensor([[1, 2, 3], [4, 5, 6]]), axis=0)) == [
            4,
            10,
            18,
        ]
        assert tb.tolist(tb.power(tb.astensor([1, 2, 3]), tb.astensor([1, 2, 3]))) == [
            1,
            4,
            27,
        ]
        assert tb.tolist(tb.divide(tb.astensor([4, 9, 16]), tb.astensor([2, 3, 4]))) == [
            2,
            3,
            4,
        ]
        assert tb.tolist(tb.sqrt(tb.astensor([4, 9, 16]))) == [2, 3, 4]
>       assert tb.tolist(tb.log(tb.exp(tb.astensor([2, 3, 4])))) == [2, 3, 4]
E       assert [1.9999999999999998, 3.0, 4.0] == [2, 3, 4]
E         At index 0 diff: 1.9999999999999998 != 2
E         Use -v to get the full diff

tests/test_tensor.py:54: AssertionError
[...]

PR #2142, opened 2023-03-22
test: Remove xfail test for floating point agreement on macOS

matthewfeickert:
# Description

Remove the xfail test as for TensorFlow v2.12.0+ the floating point agreement is restored and so the test is no longer needed. c.f. Issue #1759 for details.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Remove the xfail test as for TensorFlow v2.12.0+ the floating point
  agreement is restored and so the test is no longer needed.
   - c.f. https://github.com/scikit-hep/pyhf/issues/1759 for summary details.
```

### DIFF
--- a/tests/test_tensor.py
+++ b/tests/test_tensor.py
@@ -1,7 +1,9 @@
-import pytest
 import logging
+
 import numpy as np
+import pytest
 import tensorflow as tf
+
 import pyhf
 from pyhf.simplemodels import uncorrelated_background
 
--- a/tests/test_tensor.py
+++ b/tests/test_tensor.py
@@ -1,4 +1,3 @@
-from sys import platform
 import pytest
 import logging
 import numpy as np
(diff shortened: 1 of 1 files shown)


matthewfeickert:
Need to make sure this passes for the 'minimum supported dependencies' workflow before this can get merged. If not, then will have to keep the test and then add a skip to it depending on other information.

matthewfeickert:
> Need to make sure this passes for the 'minimum supported dependencies' workflow before this can get merged

This is good to go as we just test against Ubuntu for this workflow.

https://github.com/scikit-hep/pyhf/blob/89b3e402f27e98e56eb90cf6ac7ebaa014e4ec25/.github/workflows/lower-bound-requirements.yml#L15-L17

matthewfeickert:
@kratsg gave approval over chat, so I'm going to merge this.

PR #2154, opened 2023-03-31
ci: Add workflow input for bump version branch

matthewfeickert:
# Description

Add option in bump version workflow for the target branch to checkout and push the resulting tag back to. Also include changes that have the string `(backport):` in the commit message, but include them last in the list of changes.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add option in bump version workflow for the target branch to checkout
  and push the resulting tag back to.
* Add check that the target branch is either 'main' or a 'release/' branch.
  Though don't run this check if the 'force override check' option is
  enabled.
* Include changes that have the string '(backport):' in the commit
  message, but include them last in the list of changes.
```

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -20,6 +20,12 @@ on:
       new_version:
         description: 'New version to bump to'
         required: true
+        type: string
+      target_branch:
+        description: 'Branch to push tag to'
+        required: true
+        type: string
+        default: 'main'
       force:
         type: choice
         description: 'Force override check?'
@@ -241,5 +247,5 @@ jobs:
         if [ ${{ github.event.inputs.dry_run }} == 'true' ]; then
             echo "# DRY RUN"
         else
-            git push origin main --tags
+            git push origin ${{ github.event.inputs.target_branch }} --tags
         fi
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -49,6 +49,7 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v3
       with:
+        ref: ${{ github.event.inputs.target_branch }}
         fetch-depth: 0
         token: ${{ secrets.ACCESS_TOKEN }}
 
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -53,6 +53,17 @@ jobs:
         fetch-depth: 0
         token: ${{ secrets.ACCESS_TOKEN }}
 
+    - name: Check target branch is intended for release
+      if: github.event.inputs.force == 'false'
+      shell: bash
+      run: |
+        git rev-parse --abbrev-ref HEAD | grep 'main\|release/'
+        if [ $? -eq 1 ]; then
+            echo "ERROR: Branch $(git rev-parse --abbrev-ref HEAD) is not intended for release."
+            echo "       Releases are made only from main or release branches."
+            exit 1
+        fi
+
     - name: Verify new version bump step is valid
       if: github.event.inputs.force == 'false'
       id: script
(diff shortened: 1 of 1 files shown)


matthewfeickert:
Running this on the `release/v0.7.x` branch at 69239157fb8635bf320ce3336287cf8af4edce17 gives

```
v0.7.1          This is a patch release from v0.7.0 → v0.7.1.
    
    Changes:
      - ci: Run CI and packaging workflows on release branch PRs (#2151)
      - fix: Use unittest.mock.patch.object to mock attributes (#2144)
      - test: Remove xfail test for floating point agreement on macOS (#2142)
      - docs: Add matplotlib to piplite installs for JupyterLite example (#2135)
      - fix: Use `jax.errors` API as `jax._src` removed from public namespace (#2034)
      - fix: Skip ReadTheDocs verions during Sphinx link check (#2031)
      - docs(backport): Update JAX backend docstring types to jax.Array (#2152)
      - chore(backport): Update pre-commit hooks (#2150)
      - build(backport): Switch to using Hatchling as build backend (#2148)
```

:+1: This is all working as expected which is great for being able to make patch releases a more regular thing.

I now need to test the other deployment workflows ([publish-package](https://github.com/scikit-hep/pyhf/blob/bad73dc696196bfce6d02aceb37463ce0b400ac0/.github/workflows/publish-package.yml) and [docker](https://github.com/scikit-hep/pyhf/blob/bad73dc696196bfce6d02aceb37463ce0b400ac0/.github/workflows/docker.yml)) to make sure that they can handle getting a tag and not assuming that they should run on `main`. I'm pretty sure they don't handle this correctly yet, but that's something that can get fixed up. :+1:

.github/workflows/bump-version.yml
```diff
@@ -20,6 +20,12 @@ on:
       new_version:
         description: 'New version to bump to'
         required: true
+        type: string
+      target_branch:
+        description: 'Branch to push tag to'
+        required: true
+        type: string
+        default: 'main'
```
matthewfeickert, line 28: I actually think this isn't getting picked up correctly as a default, so I should check the GHA docs on this. :/

matthewfeickert: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs

matthewfeickert:
> I now need to test the other deployment workflows ([publish-package](https://github.com/scikit-hep/pyhf/blob/bad73dc696196bfce6d02aceb37463ce0b400ac0/.github/workflows/publish-package.yml) and [docker](https://github.com/scikit-hep/pyhf/blob/bad73dc696196bfce6d02aceb37463ce0b400ac0/.github/workflows/docker.yml)) to make sure that they can handle getting a tag and not assuming that they should run on `main`. I'm pretty sure they don't handle this correctly yet, but that's something that can get fixed up.

Good news it that the `actions/checkout` GitHub Actions does this all correctly already. :+1: If the `ref` option for `actions/checkout` isn't set then the reference will be set by the event that triggered the workflow. So for

https://github.com/scikit-hep/pyhf/blob/bad73dc696196bfce6d02aceb37463ce0b400ac0/.github/workflows/docker.yml#L3-L8

if the workflow gets triggered on a tag event then the setup logs for `actions/checkout` will include something like

```
Checking out the ref
  /usr/bin/git checkout --progress --force refs/tags/<tag name here>
  Note: switching to 'refs/tags/<tag name here>'.
  
  You are in 'detached HEAD' state. You can look around, make experimental
  changes and commit them, and you can discard any commits you make in this
  state without impacting any branches by switching back to a branch.
...
```

matthewfeickert:
After this is rebased and merged after PR #2155 and PR #2156 go in, _and_ after all of them are backported to `release/v0.7.x` _then_ we will have the ability to cut patch releases from `release/vX.Y.x` branches all through Github Actions. :+1:

matthewfeickert:
I'm going to approve and merge this myself. As always, PRs approved by a single core dev can be reverted as needed by the rest of the dev team.

matthewfeickert:
This needs to be backported to `release/v0.7.x`.

PR #2162, opened 2023-04-06
test: Use pytest.raises match argument to check Error messages

### LINKED ISSUE #2161: Use pytest.raises match over checking for string content manually
> These were previously not doing anything and have to be moved outside the `pytest.raises` to actually work. I checked other places in the tests where the same pattern is used and there it all seemed fine.
>
> There's also built-in functionality to do this quite conveniently:
> ```python
> with pytest.raises(RuntimeError, match="SigXsecOverSM"):
>     ...
> ```

_Originally posted by @alexander-held in https://github.com/scikit-hep/pyhf/pull/2109#discussion_r1107409334_

This is great and should be applied everywhere!

matthewfeickert:
# Description

Resolves #2161 

* This is a much simpler way and more maintainable way to check Error message content rather than manually checking captured output. In the cases where multiple strings need to be checked for still keep checking the values are in `str(exc_info.value)`.
* Fix checks in `tests/test_import.py` that were not properly checking the output.

Should go in before PR #2109 to allow that PR to use this as @alexander-held has already started to implement these changes there.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* This is a much simpler way and more maintainable way to check Error
  message content rather than manually checking captured output. In the
  cases where multiple strings need to be checked for still keep
  checking the values are in str(exc_info.value).
* Fix checks in tests/test_import.py that were not properly checking the
  output.

Co-authored-by: Alexander Held <45009355+alexander-held@users.noreply.github.com>
```

### DIFF
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -497,11 +497,10 @@ def test_import_missingPOI(mocker, datadir):
     mocker.patch('pyhf.readxml.import_root_histogram', return_value=(_data, _err))
 
     basedir = datadir.joinpath("xmlimport_missingPOI")
-    with pytest.raises(RuntimeError) as excinfo:
+    with pytest.raises(
+        RuntimeError, match="Measurement GaussExample is missing POI specification"
+    ):
         pyhf.readxml.parse(basedir.joinpath("config/example.xml"), basedir)
-        assert 'Measurement GaussExample is missing POI specification' in str(
-            excinfo.value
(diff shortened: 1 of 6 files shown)


matthewfeickert:
I'm going to approve and merge this myself. As always, PRs approved by a single core dev can be reverted as needed by the rest of the dev team.

PR #2183, opened 2023-04-21
ci: Use PyPI Trusted Publisher for publishing package

matthewfeickert:
# Description

Use the OpenID Connect (OIDC) standard to publish to PyPI and TestPyPI using PyPI's "Trusted Publisher" implementation to publish without using API tokens stored as GitHub Actions secrets. Use an optional GitHub Actions environment to further restrict publishing to selected branches (`main` and `release/*`) for additional security.
   - c.f. https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/
   - c.f. https://docs.pypi.org/trusted-publishers/

publish-package-environment

Thanks to @di and the PyPI team for making this happen and for the helpful [discussion on Twitter](https://twitter.com/HEPfeickert/status/1649181900521472001?s=20) and https://github.com/pypi/warehouse/issues/13463!

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Split the build and publish steps into two separate jobs. The 'build' job
  builds and checks the distributions and then uploads them as a job artifact.
  The 'publish' job downloads the required artifact from the 'build' job and
  the publishes them to TestPyPI or PyPI if the typical publishing requirements
  are met.
* Use the OpenID Connect (OIDC) standard to publish to PyPI and TestPyPI
  using PyPI's "Trusted Publisher" implementation to publish without
  using API tokens stored as GitHub Actions secrets. Use an optional
  GitHub Actions environment to further restrict publishing to selected
  branches ('main' and 'release/*') for additional security.
   - c.f. https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/
   - c.f. https://docs.pypi.org/trusted-publishers/
```

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -25,16 +25,9 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build-and-publish:
-    name: Build and publish Python distro to (Test)PyPI
+  build:
+    name: Build Python distribution
     runs-on: ubuntu-latest
-    # Mandatory for publishing with a trusted publisher
-    # c.f. https://docs.pypi.org/trusted-publishers/using-a-publisher/
-    permissions:
-      id-token: write
-    # Restrict to the environment set for the trusted publisher
-    environment:
-      name: publish-package
     steps:
     - uses: actions/checkout@v3
       with:
@@ -94,6 +87,36 @@ jobs:
 
     - name: List contents of wheel
       run: python -m zipfile --list dist/pyhf-*.whl
+
+    - name: Upload distribution artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: dist-artifact
+        path: dist
+
+  publish:
+    name: Publish Python distribution to (Test)PyPI
+    if: github.event_name != 'pull_request'
+    needs: build
+
+    runs-on: ubuntu-latest
+    # Mandatory for publishing with a trusted publisher
+    # c.f. https://docs.pypi.org/trusted-publishers/using-a-publisher/
+    permissions:
+      id-token: write
+    # Restrict to the environment set for the trusted publisher
+    environment:
+      name: publish-package
+
+    steps:
+    - name: Download distribution artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: dist-artifact
+        path: dist
+
+    - name: List all files
+      run: ls -lh dist
 
     - name: Publish distribution 📦 to Test PyPI
       # Publish to TestPyPI on tag events of if manually triggered
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -98,7 +98,6 @@ jobs:
     name: Publish Python distribution to (Test)PyPI
     if: github.event_name != 'pull_request'
     needs: build
-
     runs-on: ubuntu-latest
     # Mandatory for publishing with a trusted publisher
     # c.f. https://docs.pypi.org/trusted-publishers/using-a-publisher/
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -28,6 +28,13 @@ jobs:
   build-and-publish:
     name: Build and publish Python distro to (Test)PyPI
     runs-on: ubuntu-latest
+    # Mandatory for publishing with a trusted publisher
+    # c.f. https://docs.pypi.org/trusted-publishers/using-a-publisher/
+    permissions:
+      id-token: write
+    # Restrict to the environment set for the trusted publisher
+    environment:
+      name: publish-package
     steps:
     - uses: actions/checkout@v3
       with:
@@ -96,7 +103,6 @@ jobs:
         || (github.event_name == 'workflow_dispatch' && github.event.inputs.publish == 'true' && github.repository == 'scikit-hep/pyhf')
       uses: pypa/gh-action-pypi-publish@v1.8.5
       with:
-        password: ${{ secrets.test_pypi_password }}
         repository-url: https://test.pypi.org/legacy/
         print-hash: true
 
@@ -104,5 +110,4 @@ jobs:
       if: github.event_name == 'release' && github.event.action == 'published' && github.repository == 'scikit-hep/pyhf'
       uses: pypa/gh-action-pypi-publish@v1.8.5
       with:
-        password: ${{ secrets.pypi_password }}
         print-hash: true


matthewfeickert:
I tested this on a `release/debug-trusted-publisher` branch (to match the `release/*` branch name pattern of the `publish-package` environment) and it worked as expected

publish-workflow-success

noting

> Publish Python distribution to (Test)PyPI
> Attempting to perform trusted publishing exchange to retrieve a temporary short-lived API token for authentication against https://test.pypi.org/legacy/ due to __token__ username with no supplied password field

and published https://test.pypi.org/project/pyhf/0.7.1.dev86/ to TestPyPI. :+1: So I think this should be good to merge.

matthewfeickert:
cc @agoose77 @henryiii @jpivarski @HDembinski @eduardo-rodrigues as this might be of interest to them as well. :+1:

HDembinski:
Thanks for the ping, but what is the practical advantage over the current system? In both cases I need to change something exactly once in PyPI and in the repo settings. It is different things now, but it is the same amount of work. One has to do exactly once per repo, so anyway not a time-saver.

Or am I missing something?

matthewfeickert:
> Thanks for the ping, but what is the practical advantage over the current system?

It is a security measure and not a convince thing. c.f. https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/ and https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect

agoose77:
Thanks for sharing this @matthewfeickert. I saw the news the other day, but haven't gotten around to doing it yet. Nice to see that it's simple in practice! Furthermore, looking at your workflow reminds me that we should probably also push to test pypi on manual triggers, rather than just skipping the deploy step.

I think the main security difference between API tokens and OIDC-connect is that one cannot easily steal the API token and use it in a different context. This means improved security, both from tokens being leaked from GitHub (unlikely!) and tokens being leaked through other means (e.g. accidentally publishing to the repo).

webknjaz:
> One has to do this exactly once per repo, so anyway not a time-saver.

Those have unrestricted validity. OIDC-acquired tokens autoexpire in 15 minutes, which is safer. Plus they are no longer bound to user accounts so if a person loses access to a project, there's no need to remember to reprovision stuff, for example.

matthewfeickert:
> any reason you didn't add the required reviewers. You can use a team there too.

@webknjaz my understanding is that the adding [required reviewers to a GitHub Actions environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#required-reviewers) blocks actions in that environment from running until they have been approved by the reviewers. In the case of how we run the publish steps these are controlled either through the [bump version workflow](https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/.github/workflows/bump-version.yml) (for TestPyPI) or through creating a GitHub Release (for PyPI). As both of these actions need to be done by project maintainers having them do the action and then have to approve the action again seemed like too much duplicated approvals.

Am I wrong about how required reviewers work for actions in environments?

webknjaz:
I guess it's my personal preference to pause workflows, especially if the previous jobs take a long time and I want to have more control over cancelation without having to babysit the workflow as it reaches the said job.. Though, I also don't create git tags / GH releases before the publishing since the build/publish may fail, causing the need to cleanup stuff.

matthewfeickert:
> I guess it's my personal preference to pause workflows, especially if the previous jobs take a long time and I want to have more control over cancelation without having to babysit the workflow as it reaches the said job.. Though, I also don't create git tags / GH releases before the publishing since the build/publish may fail, causing the need to cleanup stuff.

These manually triggered workflows for us are quite short (the bump version workflow takes usually less than 30 seconds). We have a bit of a different publishing workflow (c.f. the full [release process for `v0.7.1`](https://youtu.be/ZV20tr3EpTw)) as we've tried to make it happen fully in GitHub Actions as possible such that any one of us maintainers can do it from anywhere and we also have a pure-Python project so we only have 1 sdist and 1 wheel to upload.

> I also don't create git tags / GH releases before the publishing since the build/publish may fail, causing the need to cleanup stuff.

It is true that the automation could fail, but we've yet to have it happen and it is pretty easy to delete a tag if needed.

webknjaz:
> it is pretty easy to delete a tag if needed

It's easy for you, sometimes. But by the time you do it, it may propagate into external systems (the ones that watch releases), local git repos of different people, forks made before you have a chance to clean up etc. These external side effects are harder to deal with, not talking about the convention that the tags should be immutable.

matthewfeickert:
(On phone sorry for typos)

> It's easy for you, sometimes. But by the time you do it, it may propagate into external systems (the ones that watch releases), local git repos of different people, forks made before you have a chance to clean up etc. These external side effects are harder to deal with, not talk

Hm. I know that you deal with a lot more real world infrastructure than we do, and so say these things because you have experience with them, so can you share some examples of these external systems that watch for **git tags** and not GitHub releases?

> These external side effects are harder to deal with

Are there side effects of an eager user or automated system that we don't control that grabs a yet unreleased tag (that is deleted 10 minutes later) with a `git fetch --all` (beyond that they might need to reset against the changes history later)? We try to make clear in the docs that we only provide support for released software.

> not talking about the convention that the tags should be immutable.

Is this actually convention? If you made a Git tag for a release and pushed it but then caught an error before any distribution was published should one leave a tag that is known to be bad in version control and then leave a gap in the release history (assuming you're using SemVer)? That feels SVN-like. I can understand this if there is a distribution published that is also getting yanked but I don't understand the reason to treat Git history as immutable before a distribution is published.

webknjaz:
(also from phone)

> external systems that watch for git tags and not GitHub releases?

I don't think that the watchers do that. But for said case, I consider git tags and GH releases to be equal. If an external system sees there's a new release, I don't think it'll clean it up on its side if you remove it. As for the Git tags, forks may get a copy of them and when you force-push them upstream, a fork would be stuck with the tags pointing to commits they pointed to at the time of the fork creation.

webknjaz:
> Are there side effects of an eager user or automated system that we don't control that grabs a yet unreleased tag

I'm not sure everyone can easily filter these out, though. And it's relatively easy to run a chain of reactions where the webhook processors wouldn't have enough context to make proper judgment.

webknjaz:
> Is this actually convention?

Yes, the reasoning is exactly the same as for any other public references. The rules are the same as with force-pushing public branches as `main`.

henryiii:
I've got an update for the dev pages at https://github.com/scikit-hep/scikit-hep.github.io/pull/328. I've got both methods in tabs. I've not used it yet (at PyCon), so would like reviews. :) 

AFAICT, it does not solve the "first push" problem - you need an existing project to use it.

di:
> AFAICT, it does not solve the "first push" problem - you need an existing project to use it.

https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/

henryiii:
Wow! :heart: Will need to update https://github.com/scikit-hep/scikit-hep.github.io/pull/328 then, thanks!

kratsg:
FYI, Zenodo is one of those systems that treats tags and releases differently. It sees tags as something you build more intermediate (almost akin to a release candidate) while the actual drafted release is what triggers the archiving functionality of Zenodo. This is probably going to have lots of pros and cons here, I imagine, but I'm somewhat less opinionated on whether workflows should be running on tags or releases.

webknjaz:
> I'm somewhat less opinionated on whether workflows should be running on tags or releases.

Not sure if this was clear but I did suggest using tags or releases as triggers. Rather vise versa — in my projects I use `workflow_run` to kick things off and tags/releases are results of a successful workflow run.
FWIW I'm not calling for immediate changes or how things are done here, just sharing what I got myself.

eduardo-rodrigues:
Thank you for sharing @matthewfeickert. I will look into this soon but not now as I'm loaded.

matthewfeickert:
I think things got a bit off topic from the focus of this PR because of me (my bad!). This PR is focused on adopting PyPI's trusted publishing, which it does, and so as @kratsg has signed off on these changes and no one else has raised concerns with the actual implementation in the workflow I'm going to merge this.

I think the discussions that were started here are quite valuable though and so I would invite those who are interested to continue them in the following GitHub Discussion: https://github.com/scikit-hep/pyhf/discussions/2185

Also to be clear, I'm quite thankful for @webknjaz's feedback and the continued advice that he's given pyhf over the years. I know that he was also taking time out of PyCon 2023 to discuss things here and for that alone I want to say a huge thank you.

---

I'll mention one more thing before moving over to https://github.com/scikit-hep/pyhf/discussions/2185 to discuss tags.

The only reason I didn't enable the [required reviewers](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#required-reviewers) option for the environment is that it would be a redundant check and I want to make sure that checks aren't something we get complacent about (i.e., I don't want them to be the GitHub equivalent of Microsoft Windows "Are you sure?" checks for things you literally just did).

I'll step through [how `pyhf` does publishing](https://youtu.be/ZV20tr3EpTw) to TestPyPI and PyPI to try to explain this more. It is worth noting that we run the distribution building steps of the workflow on each PR

https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/.github/workflows/publish-package.yml#L2-L11

and use that same workflow to build the distribution that will get published on releases.

### TestPyPI

Publishing to TestPyPI can be done either by workflow dispatch to manually check that a given commit on `main` will render out correctly

https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/.github/workflows/publish-package.yml#L14-L21

and if a tag is pushed

https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/.github/workflows/publish-package.yml#L2-L7

https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/.github/workflows/publish-package.yml#L91-L96

As we create the tag before anything is released as a published distribution we want to test that things get deployed to TestPyPI correctly before anything goes further.

As this PR is the first time that any GitHub Actions environment is created for publishing this means that at any point any member of the Scikit-HEP GitHub org could publish to TestPyPI using workflow dispatch and we rely on mutual trust to not have this happen.

**Perhaps this alone is a reason to restrict the members who can approve a publishing environment workflow as any Scikit-HEP admin who has their account compromised could then publish?**
This feels like something for general Scikit-HEP discussion as I don't think that any projects have protected GitHub Actions environments for deployment (`pyhf`, `pylhe`, `boost-histogram`, `awkward`, and `uproot5` all don't for example).

However, assuming that there no compromised account is going to try to push a tag directly (big assumption, but let's move forward), the only people that are allowed to run the bump version workflow to create a tag and push it back to GitHub are myself, @lukasheinrich, and @kratsg

https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/.github/workflows/bump-version.yml#L216-L223

(Here we could of course use an environment which is more standard)

### PyPI

Once a tag exists for a release and we've verified it looks fine on TestPyPI we deploy to PyPI by creating a GitHub release from that tag

https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/.github/workflows/publish-package.yml#L12-L13

https://github.com/scikit-hep/pyhf/blob/d6208cbabb8ab727f201e6e06673f938d77b14b2/.github/workflows/publish-package.yml#L103-L108

A GitHub release is only able to be made by a member of the Scikit-HEP organization.

---

So writing all of this out, @webknjaz your advice would be that every Scikit-HEP project should restrict their publishing workflows GitHub Action environments with [required reviewers](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#required-reviewers) as a security measure, correct?

webknjaz:
> as a security measure, correct?

@matthewfeickert yes, but I think that you can now have tag protection rules so maybe that'd help too. Or could be explored instead.
Also, I see that your workflow makes different assumptions than the ones I tend care about. And that's okay, it only means that you have better understanding of what you need/want within said constraints. I mostly share my habits/experiences to show different ways of doing things, trying to explain my motivation but it doesn't necessarily mean that you need to copy my approaches. I know that many people's preferences differ from mine.
In your case, it's probably fine to keep it as is, unless you switch to having a workflow that is similar to mine.
For me personally, the approvals are important because of two things:
1) security
2) ability to take time to double-check stuff

Regarding (2), having a combined workflow, publishing to PyPI and TestPyPI is happening in parallel but the PyPI one is paused. So if I want, I could inspect whatever is published to TestPyPI before letting PyPI publish proceed. And all of the post-publishing jobs depend on that PyPI one.
This also enables creating a draft release before publishing, uploading dists for preview there, and turning it into a real release after the PyPI job succeeds.
It is also important when the tests are taking a long time, allowing me to switch context and come back to finalizing it later, giving me time to reflect on whether I need to abort the release.

Your process is different so you don't need to care about these things. I don't think there's a compelling reason for you to enable approvers after your explanation.
But you may want to enable this temporarily, as an experiment, and disable after 2-3 releases, having experienced what it's like. I think, it's better to try it out for yourself rather than discuss theoretical implications. It may annoy you enough that you'll understand that you don't want it, or you might little it for whatever reason 🤷‍♂️

PR #2197, opened 2023-05-11
fix: Catch use of multi-component parameters as POI with error message

### LINKED ISSUE #2200: `shapefactor` as parameter of interest
### Discussed in https://github.com/scikit-hep/pyhf/discussions/2194

<div type='discussions-op-text'>

<sup>Originally posted by **TommyDESY** May  4, 2023</sup>
Hi all,

I'm currently setting up a fit with 5 templates/samples. I included as a PoI a `normfactor` (Unconstrained Normalisation) which didn't cause any particular issue. I managed to create my pyhf model and managed to plot pulls, errors and fit with Asimov data. 

However, what I actually want to do is use a `shapefactor` (Data-driven Shape). I use exactly the same code as previously but here an error arises. After building my pyhf spec I add the `shapefactor` as a modifier:

`sample['modifiers'].append({'name': 'mu', 'type': 'shapefactor', 'data': None})`

When I build my model with `model = pyhf.Model(pyhf_spec)`, I get an AssertionError from pyhf/pdf.py l. 468:

`    467         s = self.par_slice(name)`
`--> 468         assert s.stop - s.start == 1`

With a `normfactor`, this slice is indeed 1, but with `shapefactor` this slice has the same size as the number of bins in the fit (17 bins in lepton energy in my specific case).

I tried to remove this line from pdf.py but then this causes other errors when running the MLE fit. Therefore, it seems that this requirement is important. 

What am I missing exactly here ? Is there anything specific that needs to be added to the pyhf spec when considering data-driven shapes ? Or is this due to some know issue ?

Thank you for your help.</div>

alexander-held:
# Description

Resolves #2200.

This addresses #2194. When a parameter with multiple components (`shapefactor` / `shapesys` / `staterror`) is used as POI, the model construction currently fails at an `assert` (assuming Python is not run with `-O`). This PR turns the failure into an exception with an error message to provide more context.

For simple MLEs, bypassing the `assert` actually  makes things work (so one could argue to not catch this early and force the exception). However, this would cause even more confusing issues down the line once the POI starts mattering. As an example, a `hypotest` could result in `RuntimeWarning: invalid value encountered in subtract` within `scipy`. I do not see any benefit in allowing to bypass the exception to allow running things that do not depend on the POI definition though.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

(summary provided by @matthewfeickert)

```
* Raise exceptions.InvalidModel for multiple component parameter of interest.
  This guards against modifiers like 'shapefactor', 'shapesys', and 'staterror'
  from being used as POIs.
   - Remove use of 'assert' to check the same information.
* Add test to test_pdf.py to validate.
```

### DIFF
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -464,11 +464,10 @@ class _ModelConfig(_ChannelSummaryMixin):
             raise exceptions.InvalidModel(
                 f"The parameter of interest '{name:s}' cannot be fit as it is not declared in the model specification."
             )
-        if dict(self.modifiers).get(name, "") == "shapefactor":
-            # in the case of custom modifiers, the desired POI is in self.parameters but
-            # not in self.modifiers
+        if self.param_set(name).n_parameters > 1:
+            # multi-parameter modifiers are not supported as POIs
             raise exceptions.InvalidModel(
-                f"The parameter '{name:s}' is of type 'shapefactor' and thus cannot be used as parameter of interest."
+                f"The parameter '{name:s}' contains multiple components and is not currently supported as parameter of interest."
             )
         s = self.par_slice(name)
         assert s.stop - s.start == 1
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -469,10 +469,8 @@ class _ModelConfig(_ChannelSummaryMixin):
             raise exceptions.InvalidModel(
                 f"The parameter '{name:s}' contains multiple components and is not currently supported as parameter of interest."
             )
-        s = self.par_slice(name)
-        assert s.stop - s.start == 1
         self._poi_name = name
-        self._poi_index = s.start
+        self._poi_index = self.par_slice(name).start
 
     def _create_and_register_paramsets(self, required_paramsets):
         next_index = 0
--- a/src/pyhf/pdf.py
+++ b/src/pyhf/pdf.py
@@ -464,7 +464,9 @@ class _ModelConfig(_ChannelSummaryMixin):
             raise exceptions.InvalidModel(
                 f"The parameter of interest '{name:s}' cannot be fit as it is not declared in the model specification."
             )
-        if dict(self.modifiers)[name] == "shapefactor":
+        if dict(self.modifiers).get(name, "") == "shapefactor":
+            # in the case of custom modifiers, the desired POI is in self.parameters but
+            # not in self.modifiers
             raise exceptions.InvalidModel(
                 f"The parameter '{name:s}' is of type 'shapefactor' and thus cannot be used as parameter of interest."
             )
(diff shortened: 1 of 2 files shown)


alexander-held:
Not sure how to best handle the test failures related to custom modifiers: is this a sign that they should also go into the `modifiers` tuples? Otherwise we could of course check for the presence of the given parameter first.

src/pyhf/pdf.py
```diff
@@ -464,6 +464,12 @@ def set_poi(self, name):
             raise exceptions.InvalidModel(
                 f"The parameter of interest '{name:s}' cannot be fit as it is not declared in the model specification."
             )
+        if dict(self.modifiers).get(name, "") == "shapefactor":
```
kratsg, line 467: since the entire thing is built up, can't you go and look at the paramset instead of in modifiers? I would've thought we have information about the modifier type in paramset... but maybe not.

alexander-held: The type is not in there as far as I can see, but `self.param_set(name).n_parameters` (plus optionally `self.param_set(name).constrained`) could serve as a proxy. Having now looked at this, `shapesys` will suffer from the same issue with the assert so we can also switch to just raising for any cases of `n_parameters>1`.

src/pyhf/pdf.py
```diff
@@ -464,6 +464,11 @@ def set_poi(self, name):
             raise exceptions.InvalidModel(
                 f"The parameter of interest '{name:s}' cannot be fit as it is not declared in the model specification."
             )
+        if self.param_set(name).n_parameters > 1:
```
alexander-held, line 467: This now handles both `shapefactor` and `shapesys` if they act on channels with multiple bins. Both of these cases would fail the `assert` below. Setups like a `histosys` as POI seem to technically work, so even if they might not be useful in practice I guess there is no need to catch those here.

src/pyhf/pdf.py
matthewfeickert, line 1: > turns the failure into an exception with an error message to provide more context.

@alexander-held Given this is there any reason to not just remove the `assert` then with

```diff
-        s = self.par_slice(name)
-        assert s.stop - s.start == 1
         self._poi_name = name
-        self._poi_index = s.start
+        self._poi_index = self.par_slice(name).start
```

matthewfeickert (approved):
Thanks @alexander-held and LGTM. :+1: I just have one question on if we couldn't go further.

src/pyhf/pdf.py
alexander-held: Good point, I think the new exception should fully handle this already.

matthewfeickert (approved):
After the `assert` removal this all LGTM @alexander-held. :+1: If @kratsg also approves then once this is merged and I finish the other backports I'll release `pyhf` `v0.7.2` with this in it today.

kratsg (approved):

PR #2239, opened 2023-07-03
chore: target-version no longer needed by Black or Ruff

henryiii:
See https://github.com/scientific-python/cookie/issues/201.

Committed via https://github.com/asottile/all-repos

```
* Remove target-version from Black and Ruff metadata in pyproject.toml.
   - c.f. https://github.com/scientific-python/cookie/issues/201
```


matthewfeickert (approved):
LGTM, thanks @henryiii!

PR #2273, opened 2023-08-10
test: Add test for iterator exhaustion with `next`

matthewfeickert:
# Description

* Add test to `tests/contrib/test_viz.py` to cover the case where the wrong labels are somehow returned in `brazil.plot_results`, causing the legend label ordering logic to fail.
* Amends PR #2264 (this provides the coverage that was missing in https://github.com/scikit-hep/pyhf/pull/2264#pullrequestreview-1571224664).

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add test to tests/contrib/test_viz.py to cover the case where the
  wrong labels are somehow returned in brazil.plot_results, causing the
  legend label ordering logic to fail.
* Amends PR #2264
```

### DIFF
--- a/tests/contrib/test_viz.py
+++ b/tests/contrib/test_viz.py
@@ -181,3 +181,25 @@ def test_plot_results_components_data_structure(datadir):
         brazil.plot_results(
             data["testmus"], data["results"], test_size=0.05, ax=ax, components=True
         )
+
+
+def test_plot_results_wrong_axis_labels(datadir, mocker):
+    """
+    If the returned labels are different from the expected, then the hardcoded
+    values in label_part will be wrong, causing `next` to fail on the iterator
+    for label_idx.
+    """
+    data = json.load(datadir.joinpath("hypotest_results.json").open(encoding="utf-8"))
+
+    fig = Figure()
+    ax = fig.subplots()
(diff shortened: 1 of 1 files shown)


tests/contrib/test_viz.py
```diff
@@ -181,3 +181,25 @@ def test_plot_results_components_data_structure(datadir):
...
+    fig = Figure()
+    ax = fig.subplots()
+
+    get_legend_handles_labels = mocker.patch(
+        "matplotlib.axes._axes.Axes.get_legend_handles_labels",
+        return_value=(None, ["fail"]),
+    )
+
+    with pytest.raises(StopIteration):
+        brazil.plot_results(data["testmus"], data["results"], test_size=0.05, ax=ax)
```
matthewfeickert, line 203: This really just causes the error to cover the iterator exhaustion case for `next` and so is somewhat of a bad test, but we'd need to add a custom exception here otherwise and I'm not sure if that is worth it.

kratsg (approved):

kratsg:
I'm ok with the test looking like this, since this doesn't necessarily mean an impact in actual user-facing code.

Issue #1785, opened 2022-02-22
Documentation of default Minuit strategy choice


alexander-held:
### Summary

The documentation of `minuit_optimizer` lists `strategy` as a possible kwarg with default value of `None`, but it does not explain what that choice leads to. `pyhf` uses `0`/`1` depending on whether user-provided gradients are used (#1172 / #1183), and I believe it would be good to have that documented on the optimizer page. I'd be happy to submit a PR.

### Documentation Page Link

https://pyhf.readthedocs.io/en/v0.6.3/_generated/pyhf.optimize.opt_minuit.minuit_optimizer.html#pyhf.optimize.opt_minuit.minuit_optimizer

### Code of Conduct

- [X] I agree to follow the Code of Conduct

alexander-held:
Maybe I misunderstood the way `strategy` is supposed to be used, but it seems like `None` is actually not a supported setting:
```
import pyhf

model = pyhf.simplemodels.uncorrelated_background(
    signal=[12.0, 11.0], bkg=[50.0, 52.0], bkg_uncertainty=[3.0, 7.0]
)
data = [51, 48] + model.config.auxdata
pyhf.set_backend("numpy", "minuit")
pyhf.infer.mle.fit(data, model, strategy=None)
```
will fail:
```
Traceback (most recent call last):
  File "[...]/test.py", line 8, in <module>
    pyhf.infer.mle.fit(data, model, strategy=None)
  File "[...]/pyhf/src/pyhf/infer/mle.py", line 131, in fit
    return opt.minimize(
  File "[...]/pyhf/src/pyhf/optimize/mixins.py", line 185, in minimize
    result = self._internal_minimize(
  File "[...]/pyhf/src/pyhf/optimize/mixins.py", line 51, in _internal_minimize
    result = self._minimize(
  File "[...]/pyhf/src/pyhf/optimize/opt_minuit.py", line 117, in _minimize
    minimizer.strategy = strategy
  File "[...]/iminuit/minuit.py", line 203, in strategy
    self._strategy.strategy = value
TypeError: (): incompatible function arguments. The following argument types are supported:
    1. (arg0: iminuit._core.MnStrategy, arg1: int) -> None

Invoked with: <iminuit._core.MnStrategy object at 0x1420a0070>, None
```
`iminuit` only supports values of 0/1/2.

[...]

PR #2277, opened 2023-08-15
fix: Correct handling of strategy 0 for Minuit optimizer

wernerd-cern:
# Description
The relevant line is supposed to set the strategy of minuit to one of the supported values `0`, `1`, ~~or `2`~~ (Note from @matthewfeickert: `pyhf` doesn't support option `2` at this time). Since the strategy `0` evaluates to `False`, this strategy couldn't be set when creating the optimizer if `do_grad` was `False`

https://github.com/scikit-hep/pyhf/blob/f8f3c96e8be560c3e5712fe080f97e95702d4942/src/pyhf/optimize/opt_minuit.py#L104-L106

. The default value in the case of no specified strategy is `None`

https://github.com/scikit-hep/pyhf/blob/f8f3c96e8be560c3e5712fe080f97e95702d4942/src/pyhf/optimize/opt_minuit.py#L39

meaning that the statement `is not None` solves the issue with

```python
self.strategy if self.strategy is not None else not do_grad
```
.

Amends PR https://github.com/scikit-hep/pyhf/pull/1183.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Properly guard the `strategy` option for the Minuit optimizer in the case that
  `strategy` option of 0 is used.
  Without checking `self.strategy is not None` the `if self.strategy` for
  self.strategy==0 evaluates to truthiness of False, giving the wrong value of
  `not do_grad` for do_grad of False.
   - Amends PR https://github.com/scikit-hep/pyhf/pull/1183
* Use the corrected behavior to improve testing of default in test_minuit_strategy_global.
* Add Daniel Werner to contributors list.
```

### DIFF
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -182,7 +182,6 @@ def test_minuit_strategy_global(mocker, backend, strategy):
     m = pyhf.simplemodels.uncorrelated_background([50.0], [100.0], [10.0])
     data = pyhf.tensorlib.astensor([125.0] + m.config.auxdata)
 
-    do_grad = pyhf.tensorlib.default_do_grad
     pyhf.infer.mle.fit(data, m)
     assert spy.call_count == 1
     assert spy.spy_return.minuit.strategy == strategy
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -185,7 +185,7 @@ def test_minuit_strategy_global(mocker, backend, strategy):
     do_grad = pyhf.tensorlib.default_do_grad
     pyhf.infer.mle.fit(data, m)
     assert spy.call_count == 1
-    assert spy.spy_return.minuit.strategy == strategy if do_grad else 1
+    assert spy.spy_return.minuit.strategy == strategy
 
     pyhf.infer.mle.fit(data, m, strategy=0)
     assert spy.call_count == 2
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -31,3 +31,4 @@ Contributors include:
 - Jerry Ling
 - Nathan Simpson
 - Beojan Stanislaus
+- Daniel Werner
--- a/src/pyhf/optimize/opt_minuit.py
+++ b/src/pyhf/optimize/opt_minuit.py
@@ -102,7 +102,7 @@ class minuit_optimizer(OptimizerMixin):
         # 0: Fast, user-provided gradient
         # 1: Default, no user-provided gradient
         strategy = options.pop(
-            'strategy', self.strategy if self.strategy else not do_grad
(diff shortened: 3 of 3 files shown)


matthewfeickert:
Thanks for the PR @wernerd-cern. This should have been caught in the `tests/test_optim.py` tests, so we'll want to have a test added to this PR that covers this.

If you have time to add a test that will speed things up in getting this approved. If not, then the core devs can add one later this week and get this in.

kratsg (approved):

tests/test_optim.py
```diff
@@ -185,7 +185,7 @@ def test_minuit_strategy_global(mocker, backend, strategy):
     do_grad = pyhf.tensorlib.default_do_grad
```
kratsg, line 185: this drops the need for `do_grad`. We might need to remove the line entirely.

```suggestion
```

kratsg (approved):

matthewfeickert (approved):
Thanks for your PR @wernerd-cern &mdash; your contributions are appreciated! This all LGTM. :+1:

We'll go ahead and merge this in, and I'll work on getting a patch release out in hopefully the next day. Though given that this bug was found I think I should probably address Issue https://github.com/scikit-hep/pyhf/issues/1785 as well so that can go into the same patch release.

matthewfeickert:
For future us, here's a simple example of what was going wrong:

```python
# pr_2777_example.py

import itertools

strategies = [None, 0, 1]
grad_option = [False, True]

for strategy, do_grad in itertools.product(strategies, grad_option):
    print(f"\n# strategy: {strategy}")
    if strategy is not None:
        print(f"# do_grad: {do_grad}")
    else:
        print(f"# int(not do_grad): {int(not do_grad)}")

    print(f"v0.7.2:   {strategy if strategy else not do_grad}")
    print(f"PR #2277: {strategy if strategy is not None else not do_grad}")
    if strategy is None:
        print(
            f"PR #2277 cast to int: {strategy if strategy is not None else int(not do_grad)}"
        )
```

```console
$ python /tmp/pr_2777_example.py 

# strategy: None
# int(not do_grad): 1
v0.7.2:   True
PR #2277: True
PR #2277 cast to int: 1

# strategy: None
# int(not do_grad): 0
v0.7.2:   False
PR #2277: False
PR #2277 cast to int: 0

# strategy: 0
# do_grad: False
v0.7.2:   True
PR #2277: 0

# strategy: 0
# do_grad: True
v0.7.2:   False
PR #2277: 0

# strategy: 1
# do_grad: False
v0.7.2:   1
PR #2277: 1

# strategy: 1
# do_grad: True
v0.7.2:   1
PR #2277: 1
```

PR #1035, opened 2020-08-18
fix: Weakly reference callables for pub-sub model

kratsg:
# Description

(See this nice answer for other libs: https://stackoverflow.com/a/16192256/1532974)

Fix a source of memory leaks observed by SModelS developers reported in #934, #620. This leak occurs because even though certain objects go out of scope - garbage-collect in python isn't invoked because we kept rather strong references to the bound methods of these objects. Instead, use [`weakref.WeakMethod`](https://docs.python.org/3/library/weakref.html#weakref.WeakMethod) to keep weaker references that are auto-removed from events if the objects owning those methods go out of scope (allowing python's garbage collect to actually.. collect garbage).

This was not so easy to track down, so I thank @lukasheinrich for narrowing it down a lot more with a simple MWE and his latest theory.

```python
    a = pyhf.tensorlib.astensor([0.0])
    for _ in range(10000):
        _TensorViewer([a], names=['hello'], batch_size=None)
        test = pyhf.tensorlib.astensor([1]*100000)
```

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR
```
* Fix memory leaks seen when creating many objects that use pub-sub events API
* Store weaker references to bound methods/callables in the events API
```

PR #2310, opened 2023-08-31
fix: Skip callbacks with dead weakrefs while iterating over callbacks

### LINKED ISSUE #2311: weakrefs to dead objects occuring when changing backends in pyhf benchmark
### Summary

I try to perform a benchmark of `pyhf` using `pytest-benchmark` quite similarly to the benchmark in the `tests/benchmarks` directory.
A short example of such a simple benchmark is given below. To reproduce this bug, the python code needs to be saved in a file of the format `test_<name>.py` and executed via `pytest test_<name>.py`.
The bug occurs only sometimes when the backend is changed between different benchmarking cases. Since the occurence of the bug includes some amount of randomness, it may happen that it doesn't occur on the first try but that the benchmark must be executed multiple times. The full benchmark takes around 1 min on my machine.

The suspected origin of this bug is that every time the backend is changed, an event called `tensorlib_changed` is triggered that in turn leads to the execution of some `_precompute()` functions on different objects (e.g. a `TensorViewer` object as in the error message). The problem occurs, when the referenced object no longer exists, as all references to it have been removed. The reference used to call the function does not change this as it is a weakref.

A proposed solution can be found in PR #2310. 

### OS / Environment

```console
PRETTY_NAME="Ubuntu 22.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

### Steps to Reproduce

```python (paste below)
# content of test_benchmark.py
import pytest
import pyhf

@pytest.fixture(
    scope='function',
    params=[
        (pyhf.tensor.numpy_backend(), None),
        (pyhf.tensor.pytorch_backend(precision='64b'), None),
        (pyhf.tensor.tensorflow_backend(), None),
        (pyhf.tensor.jax_backend(), None),
        (
            pyhf.tensor.numpy_backend(poisson_from_normal=True),
        ),
    ],
    ids=['numpy', 'pytorch', 'pytorch64',
         'tensorflow',
         'jax', 'numpy_minuit'],
)
def backend(request):
    # get the ids of all the backends
    param_ids = request._fixturedef.ids
    # the backend we're using: numpy, tensorflow, etc...
    param_id = param_ids[request.param_index]
    # name of function being called (with params), the original name is .originalname
    func_name = request._pyfuncitem.name

    pyhf.set_backend(*request.param)

    yield request.param

def hypotest(data, pdf):
    return pyhf.infer.hypotest(1.0, data, pdf, test_stat="qtilde", return_expected=True)

bins = [1, 2, 4, 8, 16, 32]
bin_ids = [f'{n_bins}_bins' for n_bins in bins]

@pytest.mark.parametrize('n_bins', bins, ids=bin_ids)
def test_hypotest(benchmark, backend, n_bins):
    model = pyhf.simplemodels.uncorrelated_background(signal=[12.0]*n_bins, bkg=[50.0]*n_bins, bkg_uncertainty=[5.0]*n_bins)
    data = [51.0]*n_bins + model.config.auxdata

    assert benchmark(hypotest, data, model)
```

```console (paste below)
pytest test_benchmark.py
```

### File Upload (optional)

_No response_

### Expected Results

The expected behavior is to output the benchmarking results for all considered cases as it can be observed when executing `pytest` in `pyhf/tests/benchmarks/`.
This output should not show any "test failures" as no normal tests are performed but only functions that run without an error, when called outside of the benchmark.

### Actual Results

```console
_________________________ ERROR at setup of test_hypotest[jax-1_bins] _________________________

request = <SubRequest 'backend' for <Function test_hypotest[jax-1_bins]>>

    @pytest.fixture(
        scope='function',
        params=[
            (pyhf.tensor.numpy_backend(), None),
            (pyhf.tensor.pytorch_backend(precision='64b'), None),
            (pyhf.tensor.tensorflow_backend(), None),
            (pyhf.tensor.jax_backend(), None),
            (
                pyhf.tensor.numpy_backend(poisson_from_normal=True),
            ),
        ],
        ids=['numpy', 'pytorch', 'pytorch64',
             'tensorflow',
             'jax', 'numpy_minuit'],
    )
    def backend(request):
        # get the ids of all the backends
        param_ids = request._fixturedef.ids
        # the backend we're using: numpy, tensorflow, etc...
        param_id = param_ids[request.param_index]
        # name of function being called (with params), the original name is .originalname
        func_name = request._pyfuncitem.name
    
>       pyhf.set_backend(*request.param)

test_hypo_pyhf.py:29: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../pyhfDev/pyhf/src/pyhf/events.py:161: in register_wrapper
    result = func(*args, **kwargs)
../../pyhfDev/pyhf/src/pyhf/tensor/manager.py:193: in set_backend
    events.trigger("tensorlib_changed")()
../../pyhfDev/pyhf/src/pyhf/events.py:70: in __call__
    func()(arg(), *args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = None

    def _precompute(self):
        tensorlib, _ = get_backend()
>       self.sorted_indices = tensorlib.astensor(self._sorted_indices, dtype='int')
E       AttributeError: 'NoneType' object has no attribute '_sorted_indices'

../../pyhfDev/pyhf/src/pyhf/tensor/common.py:33: AttributeError
```

### pyhf Version

```console
pyhf, version 0.7.1.dev116
```

### Code of Conduct

wernerd-cern:
# Description

Resolves #2311 

The bug fixed by this PR occured in my work in the repository https://gitlab.cern.ch/dawerner/statanabenchmark.
I am currently working on a summer student project at CERN benchmarking statistical analysis tools in python including pyhf.
When performing these benchmarks, pyhf is called multiple times in a row with different backends. In some cases an error occurs then as objects are called through weakrefs that are already dead and thus `NoneType` objects.

When events (such as 'tensorlib_changed') are called, a bug occured previously where objects (i.e. the `TensorViewer`) are already dead. In this case an error occured when member functions where using attributes such as `self._sorted_indices` in `src/pyhf/tensor/common.py:33`.

This bug is fixed by catching the cases, in which the object is invalidated by earlier callbacks.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Check refs while processing callbacks in case a callee is de-ref'd during
  the callback process.
* Additionally flush after processing callbacks in case any callback de-ref's
  a callee.
* Note that no additional tests are added for this case as the problem arises from
  and intermediate callback triggering a garbage-collect. It is unclear how to force
  this scenario deterministically in testing.
* Add Jonas Rembser to contributors list.

Co-authored-by: Jonas Rembser <jonas.rembser@cern.ch>
```

### DIFF
--- a/src/pyhf/events.py
+++ b/src/pyhf/events.py
@@ -64,12 +64,19 @@ class Callables:
         self._callbacks = _callbacks
 
     def __call__(self, *args, **kwargs):
-        for func, arg in self.callbacks:
+        for func, arg in self._callbacks:
             # weakref: needs to be de-ref'd first before calling
             if arg is not None:
-                func()(arg(), *args, **kwargs)
+                arg_ref = arg()
+                if arg_ref is not None:
+                    func()(arg_ref, *args, **kwargs)
             else:
                 func()(*args, **kwargs)
+        # We have to flush after calling all the callbacks, not before. That's
+        # beacause, earlier callbacks might cause new dead arg weakrefs in
+        # later callbacks. So we check for dead weakrefs in each iteration
+        # and then we flush at the end.
+        self._flush()
 
     def __iter__(self):
         return iter(self.callbacks)
--- a/src/pyhf/events.py
+++ b/src/pyhf/events.py
@@ -73,7 +73,7 @@ class Callables:
             else:
                 func()(*args, **kwargs)
         # We have to flush after calling all the callbacks, not before. That's
-        # beacause, earlier callbacks might cause new dead arg weakrefs in
+        # because, earlier callbacks might cause new dead arg weakrefs in
         # later callbacks. So we check for dead weakrefs in each iteration
         # and then we flush at the end.
         self._flush()
--- a/src/pyhf/events.py
+++ b/src/pyhf/events.py
@@ -72,10 +72,9 @@ class Callables:
                     func()(arg_ref, *args, **kwargs)
             else:
                 func()(*args, **kwargs)
-        # We have to flush after calling all the callbacks, not before. That's
-        # because, earlier callbacks might cause new dead arg weakrefs in
-        # later callbacks. So we check for dead weakrefs in each iteration
-        # and then we flush at the end.
+        # flush after calling all the callbacks, not before, as earlier
+        # callbacks might cause new dead arg weakrefs in later callbacks.
+        # So check for dead weakrefs in each iteration and flush at the end.
         self._flush()
 
     def __iter__(self):
--- a/src/pyhf/events.py
+++ b/src/pyhf/events.py
@@ -72,9 +72,11 @@ class Callables:
                     func()(arg_ref, *args, **kwargs)
             else:
                 func()(*args, **kwargs)
-        # flush after calling all the callbacks, not before, as earlier
-        # callbacks might cause new dead arg weakrefs in later callbacks.
-        # So check for dead weakrefs in each iteration and flush at the end.
+        # Flush after calling all the callbacks, not before, as callbacks in the
+        # beginning of the iteration might cause new dead arg weakrefs in
+        # callbacks that are iterated over later.
+        # Checking for dead weakrefs in each iteration and flushing at the end
+        # avoids redundant dead weakref checking in subsequent calls.
         self._flush()
 
     def __iter__(self):
--- a/docs/contributors.rst
+++ b/docs/contributors.rst
@@ -32,3 +32,4 @@ Contributors include:
 - Nathan Simpson
 - Beojan Stanislaus
 - Daniel Werner
+- Jonas Rembser


matthewfeickert (changes requested):
@wernerd-cern Can you [please report what the actual bug is](https://github.com/scikit-hep/pyhf/blob/main/CONTRIBUTING.md#issues) and give a reproducer as well for documentation? It would be better to make an Issue that clearly documents the bug and then have this PR close it. This PR will need tests as well.

matthewfeickert:
For reviewer reference, the last time this code was edited was in PR #1035.

src/pyhf/events.py
```diff
@@ -64,12 +64,18 @@ def _flush(self):
         self._callbacks = _callbacks
 
     def __call__(self, *args, **kwargs):
-        for func, arg in self.callbacks:
+        for func, arg in self._callbacks:
```
matthewfeickert, line 67: @wernerd-cern Can you comment on why `self._callbacks` is necessary here? It seems like a different situation than with `_flush(self)` just above this function.

matthewfeickert (changes requested):
Seems like this is indeed needed, to be able to do what we do in our tests without providing all of `conftest.py`'s machinery. So this should go in and then a patch release made from it. :+1: 

This code doesn't have full coverage though, so appropriate tests will need to be added.

kratsg (changes requested):
This PR doesn't seem to introduce any new functionality that does not already exist. `__call__` here in the changes is including changes covered by the `_flush` function. The `_flush()` should be called before trying to access the callbacks to determine the valid callbacks first. So I'm not sure how you ran into a situation that's covered by this PR that isn't already covered by the existing code...

kratsg:
In addition, the case you're describing is covered by our tests right now.

matthewfeickert:
I'm going to put this into draft until we can determine if Issue #2310 is actually reproducible.

kratsg:
> I suspect the problem is that some callbacks delete objects that owns the arguments of other callbacks, which is why flushing only at the beginning is not sufficient.

Again, this doesn't explain or change the logic. All callbacks are called through this function. Anytime you trigger the callbacks, the `_flush()` function is called, whose logic was just copied into the `__call__()` part which is already called by accessing `self.callbacks` (this got changed in the PR to access `self._callbacks` instead directly).

Without the changes, the flow is

1. switch backend
2. trigger callbacks through `__call__(self)`
3. access `self.callbacks` which triggers `self._flush()`
4. `self._flush()` removes dead refs
5. return `self.callbacks` and trigger callbacks on individual refs

With the changes, the flow is

1. switch backend
2. trigger callbacks through `self.__call__()`
3. access `self._callbacks`
4. trigger callbacks on individual refs that are not dead
5. `self._flush()` removes dead refs

Either way, `__call__` only runs on non-dead-refs. In before these changes, if a callback removes a reference to a weak ref, it will get garbage collected the next time callbacks are triggered. Which is why I'm saying that the changes aren't making a difference. The only thing that could be happening is that garbage-collection runs in between steps 4 and 5 before these changes -- which isn't currently caught, and would explain why you'd only see it if you run heavily enough to force GC to not run as frequently as it might typically do so.

guitargeek:
Thanks for making the bullet point list, maybe it's now easier for me to explain what I think is the crucial difference.

To quote the list you made and annotate with what I think is the flaw in the logic:

* switch backend

* trigger callbacks through `__call__(self)`

* access `self.callbacks` which triggers `self._flush()`

* `self._flush()` removes dead refs

* return `self.callbacks` and *iteratively* trigger *many* callbacks on individual refs, *where some of the first callbacks in the loop might drop the references on the arguments of callbacks later in the loop. The earlier call to flush has no chance of flushing these*


After the change, the flush will be called after iterating over the callbacks, catching any new dead refs that were the results of the callbacks.

kratsg:
> After the change, the flush will be called after iterating over the callbacks, catching any new dead refs that were the results of calling the callbacks.

Yes, after iterating over all callbacks, not after each one... so before the change is identical to after the change. They will get caught by the next access for callbacks before the change (delayed), but that's still ok. This is why I'm still confused. I don't see a scenario not currently caught by the original code.

guitargeek:
Here is a scenario that would not be caught by the original code:

1. There are two callbacks in the list:
      * `func_1` with `arg_1`. Calling `func_1` will delete all references to `arg_2`
      * `func_2` with `arg_2`
      Let's assume both `arg_1` and `arg_2` are not dead initially
      
2. We are calling `Callables.__call__()`, flush does nothing as there are no dead refs

3. `func_1(arg_1)` is called and there are no references to `arg_2` anymore as a result

4. `func_2` is called, but `arg_2` is now dead, resulting in an error.

I didn't dig into the pyhf code deep enough to exactly tell you what `func_1`, `func_2`, `arg_1`, and `arg_2` are, but since this scenario was the *only* was I could think of to get dead references even after flushing, I assumed that this must be happening somewhere. And indeed, when @wernerd-cern implemented the new logic that catches this case, the errors in the reproducer in #2311 go away.

kratsg:
> 4\. `func_2` is called, but `arg_2` is now dead, resulting in an error.

Ok, but again, you've been stating that moving the flush to the end catches this case when the original code does not -- this is an incorrect statement.

The difference is that you're skipping deadrefs as you iterate through the callbacks depending on if the ordering of the callbacks causes some refs to go dead during processing.

I bet if you just simply add in logic to skip refs that are dead, and remove all other changes -- that you'll catch exactly this case. Moving around the flush will do nothing though.

guitargeek:
> Ok, but again, you've been stating that moving the flush to the end catches this case when the original code does not -- this is an incorrect statement.

No, I never intended to say that (sorry if it came across like that). The position of the flush actually doesn't matter, it's all about the explicit dead arg check in the loop. We only put the flush at the end now to be a bit more efficient: like this, the flush() will also flush the new dead refs that came from calling the callbacks, so that the next time `__call__` is called it has to iterate over less elements.

You might as well put the flush also back on top.

But I'm happy we are now on the same page about skipping dead refs :+1: Sorry for not explaining explicitly why we moved the flush at the end.

kratsg (approved):
Approved, but with the caveat that this is somewhat impossible to properly force a test force, as it is only catching the scenario of an intermediate callback triggering a garbage-collect.

guitargeek:
Thanks, @kratsg and @matthewfeickert !

matthewfeickert (approved):
> Approved, but with the caveat that this is somewhat impossible to properly force a test force, as it is only catching the scenario of an intermediate callback triggering a garbage-collect.

@kratsg if you are also happy to not attempt to figure out a way to deterministically test this then we can just note it in the commit message (I've updated it) and accept this.

Thanks @wernerd-cern and @guitargeek. I'll prepare pyhf `v0.7.4` and get that out before the end of the week.

guitargeek:
As pointed out by @kratsg, also the new title is maybe not completely appropriate because it's about the skipping of dead weakrefs and not the flushing. Maybe `fix: Skip callbacks with dead weakrefs while iterating over callbacks` would be more appropriate. But not that important after all :)

matthewfeickert:
@wernerd-cern @guitargeek [`pyhf` `v0.7.4`](https://pypi.org/project/pyhf/0.7.4/) is now out on PyPI with this fix. Thanks!

PR #2328, opened 2023-09-13
feat: Don't specify a default POI name

### LINKED ISSUE #2327: Default POI name for model construction
The current default for the POI name in model construction is `"mu"`. The safer choice here would be to go with `""`, i.e. no POI specified. This reduces the assumptions being made and is not going to cause failures for models where no "mu" parameter is defined. I would suggest updating the default accordingly.

https://github.com/scikit-hep/pyhf/blob/f771581ae387493776f2c40170f82b63d0fad317/src/pyhf/pdf.py#L771

matthewfeickert:
# Description

Resolves #2327

To avoid making assumptions for users about model behavior set the default POI name to `None`.

To avoid making behavior changes for `pyhf.simplemodels` add the default `poi_name` to the returned model.

This then requires explicilty setting `poi_name="mu"` in the tests for all explicity `pyhf.pdf.Model` constrcution.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Set the default model poi_name to None to avoid making assumptions
  for users.
* To maintain the pyhf.simplemodels users experience add poi_name="mu"
  as a default argument to be passed to the returned pyhf.pdf.Model
  object.
* Add poi_name="mu" to all pyhf.Model construction in the tests.
```

### DIFF
--- a/src/pyhf/simplemodels.py
+++ b/src/pyhf/simplemodels.py
@@ -79,7 +79,7 @@ def correlated_background(
             }
         ]
     }
-    return Model(spec, batch_size=batch_size, validate=validate)
+    return Model(spec, batch_size=batch_size, validate=validate, poi_name="mu")
 
 
 def uncorrelated_background(
@@ -146,7 +146,7 @@ def uncorrelated_background(
             }
         ]
     }
-    return Model(spec, batch_size=batch_size, validate=validate)
+    return Model(spec, batch_size=batch_size, validate=validate, poi_name="mu")
 
 
 # Deprecated APIs
--- a/src/pyhf/simplemodels.py
+++ b/src/pyhf/simplemodels.py
@@ -10,7 +10,7 @@ def __dir__():
 
 
 def correlated_background(
-    signal, bkg, bkg_up, bkg_down, batch_size=None, validate=True
+    signal, bkg, bkg_up, bkg_down, batch_size=None, validate=True, poi_name="mu"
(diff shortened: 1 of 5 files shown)


src/pyhf/simplemodels.py
```diff
@@ -79,7 +79,7 @@ def correlated_background(
             }
         ]
     }
-    return Model(spec, batch_size=batch_size, validate=validate)
+    return Model(spec, batch_size=batch_size, validate=validate, poi_name="mu")
```
matthewfeickert, line 82: If we're going to be changing the default behaivor of `pyhf.pdf.Model` should we just go all in and make sure that `pyhf.simplemodel` also requires users to set things? Or should we keep the `pyhf.simplemodel` experience the same?

alexander-held: I think it's fair to automatically set the POI for the `simplemodels` case. There is no ambiguity there for what else the POI should be.

matthewfeickert: We could elevate this to a user accesible argument that defaults to `mu` and pass that through here, but as you said, the POI for these models is meant to be `mu`.

matthewfeickert:
Relevant RTD build: https://pyhf--2328.org.readthedocs.build/en/2328/_generated/pyhf.simplemodels.uncorrelated_background.html

kratsg (approved):

PR #2363, opened 2023-10-25
ci: Add simplified CI for Windows

### LINKED ISSUE #2358: Add basic testing for Windows
Based off of the experience of Issue #2353, I was curious how much work it would take to add testing for Windows. The answer seems to be not much for Python 3.9+. For Python 3.8 this fails as there is no `jaxlib` for Python 3.8, but all the backends for Python 3.9+ support Windows wheels.

image

There's an additional caveat that many tests in [`tests/test_scripts.py`](https://github.com/scikit-hep/pyhf/blob/6203a8f6ab83b058e1b5b485eb08c97a66ea3268/tests/test_scripts.py) currently fail on Windows, and so the passing screen shot runs with `--ignore tests/test_scripts.py`, but all the Python tests pass.

I don't think that we should add the testing to the normal CI that runs on every PR as this would slow things down a bit, so I would propose using

```yaml
on:
  # Run daily at 1:23 UTC
  schedule:
  - cron:  '23 1 * * *'
  workflow_dispatch:
```

to have the Windows tests run as a nightly CRON job and then on deman with workflow dispatch (which is what I have in the passing tests above).

@kratsg @lukasheinrich what are your thoughts?

matthewfeickert:
# Description

Resolves #2358 

Add simplified CI for Windows. The workflow doesn't run on push or pull request events as it is slow but instead runs on a nightly schedule job and on demand with workflow dispatch. The workflow runs the normal test suite with the exception that it skips `tests/test_scripts.py` as these are currently broken on Windows. Python 3.8 is not added to the test matrix as there is no Python 3.8 version of `jaxlib`.

As this effectively adds testing support for Windows, additionally add operating system dependent PyPI trove classifiers.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add simplfied CI workflow for Windows.
   - Runs on nighly schedule and with workflow dispatch.
* Add operating system PyPI trove classifier as support is added for Windows.
```


pyproject.toml
```diff
@@ -32,6 +32,9 @@ classifiers = [
     "Environment :: WebAssembly :: Emscripten",
     "Intended Audience :: Science/Research",
     "License :: OSI Approved :: Apache Software License",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: MacOS",
+    "Operating System :: POSIX :: Linux",
```
matthewfeickert, line 37: Oh. @henryiii uses `Operating System :: OS Independent` here, so I guess that's maybe easier.
```suggestion
    "Operating System :: OS Independent",
```

kratsg (approved):

kratsg:
> The workflow runs the normal test suite with the exception that it skips `tests/test_scripts.py` as these are currently broken on Windows.

what's broken on windows here? It could just be a missing path.

matthewfeickert:
> what's broken on windows here? It could just be a missing path.

There we multiple errors but here's an example of one `FileNotFoundError`

```pytb
================================== FAILURES ===================================
___________________ test_import_prepHistFactory[inprocess] ____________________

E   FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_import_prepHistFactory_in0\\parsed_output.json'
All traceback entries are hidden. Pass `--full-trace` to see hidden and internal frames.

tmpdir = local('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_import_prepHistFactory_in0')
script_runner = <ScriptRunner inprocess>

    def test_import_prepHistFactory(tmpdir, script_runner):
        temp = tmpdir.join("parsed_output.json")
        command = f'pyhf xml2json validation/xmlimport_input/config/example.xml --basedir validation/xmlimport_input/ --output-file {temp.strpath:s} --hide-progress'
        ret = script_runner.run(shlex.split(command))
        assert ret.success
        assert ret.stdout == ''
        assert ret.stderr == ''
    
>       parsed_xml = json.loads(temp.read())

command    = 'pyhf xml2json validation/xmlimport_input/config/example.xml --basedir validation/xmlimport_input/ --output-file C:\\U...ata\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_import_prepHistFactory_in0\\parsed_output.json --hide-progress'
ret        = <pytest_console_scripts.RunResult object at 0x000002E66E873C40>
script_runner = <ScriptRunner inprocess>
temp       = local('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_import_prepHistFactory_in0\\parsed_output.json')
tmpdir     = local('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_import_prepHistFactory_in0')

tests\test_scripts.py:68: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\_pytest\_py\path.py:377: in read
    with self.open(mode) as f:
        mode       = 'r'
        self       = local('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_import_prepHistFactory_in0\\parsed_output.json')
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = local('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_import_prepHistFactory_in0\\parsed_output.json')
mode = 'r', ensure = False, encoding = None

    def open(self, mode="r", ensure=False, encoding=None):
        """Return an opened file with the given mode.
    
        If ensure is True, create parent directories if needed.
        """
        if ensure:
            self.dirpath().ensure(dir=1)
        if encoding:
            return error.checked_call(io.open, self.strpath, mode, encoding=encoding)
>       return error.checked_call(open, self.strpath, mode)
E       py.error.ENOENT: [No such file or directory]: open('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_import_prepHistFactory_in0\\parsed_output.json', 'r')

encoding   = None
ensure     = False
mode       = 'r'
self       = local('C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\test_import_prepHistFactory_in0\\parsed_output.json')

C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\_pytest\_py\path.py:[76](https://github.com/scikit-hep/pyhf/actions/runs/6635684924/job/18027015790#step:6:77)1: ENOENT
---------------------------- Captured stdout call -----------------------------
# Running console script: ['pyhf', 'xml2json', 'validation/xmlimport_input/config/example.xml', '--basedir', 'validation/xmlimport_input/', '--output-file', 'C:UsersrunneradminAppDataLocalTemppytest-of-runneradminpytest-0test_import_prepHistFactory_in0parsed_output.json', '--hide-progress']
# Script return code: 0
# Script stdout:

# Script stderr:
```

henryiii:
`tmp_path` is a modern replacement for `tmpdir`, by the way. And why are you assuming there's something in the `tmpdir`? Anyway, the problem is the string:

```
command = f'pyhf xml2json validation/xmlimport_input/config/example.xml --basedir validation/xmlimport_input/ --output-file {temp.strpath:s} --hide-progress'
        ret = script_runner.run(shlex.split(command))
```

Paths have backslashes on windows, which worries me a bit there, and https://docs.python.org/3/library/shlex.html#shlex.split says that posix=True, but you aren't running on a posix shell on Windows. Why are you putting a string in then splitting it up? And if you use `shlex.quote` in the runner to put it back together, that's 100% UNIX only.

matthewfeickert:
To make things compatible with Windows this is going to require some work, so I'm going to merge this in and then we can fix up the test suite in a follow up PR.

> `tmp_path` is a modern replacement for `tmpdir`, by the way.

Thanks and noted.

From the pytest docs (https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures):

> The `tmpdir` and `tmpdir_factory` fixtures are similar to `tmp_path` and `tmp_path_factory`, but use/return legacy [`py.path.local`](https://py.readthedocs.io/en/latest/path.html) objects rather than standard [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html#pathlib.Path) objects.
> ## Note
> These days, it is preferred to use tmp_path and tmp_path_factory.
> 
> In order to help modernize old code bases, one can run pytest with the legacypath plugin disabled:
> 
> ```
> pytest -p no:legacypath
> ```
>  This will trigger errors on tests using the legacy paths. It can also be permanently set as part of the [addopts](https://docs.pytest.org/en/7.1.x/reference/reference.html#confval-addopts) parameter in the config file.

We'll fix this up in another PR. :+1: 

> Anyway, the problem is the string:
> 
> ```python
> command = f'pyhf xml2json validation/xmlimport_input/config/example.xml --basedir validation/xmlimport_input/ --output-file {temp.strpath:s} --hide-progress'
> ret = script_runner.run(shlex.split(command))
> ```
> 
> Paths have backslashes on windows, which worries me a bit there, and https://docs.python.org/3/library/shlex.html#shlex.split says that `posix=True` is the default, but you aren't running on a posix shell on Windows. Why are you putting a string in then splitting it up? And if you use `shlex.quote` in the runner to put it back together, that's 100% UNIX only.

Looking at an example:

https://github.com/scikit-hep/pyhf/blob/6203a8f6ab83b058e1b5b485eb08c97a66ea3268/tests/test_scripts.py#L60-L70

`command` is a single string for convenience of readability for us as [`pytest-console-scripts`](https://github.com/kvas-it/pytest-console-scripts) expects a list of command fragrments that would be space seperated, similar to `nox`. This is not nice to look at once the commands get more than a few segements long.

Internally [`pytest-console-scripts` only uses `shlex.split`](https://github.com/kvas-it/pytest-console-scripts/blob/master/pytest_console_scripts/__init__.py#L159), and we could do something similar to what they do with

```python
import os
shlex.split(command, posix=os.name == 'posix')
```

Issue #2372, opened 2023-10-27
Python 3.12 support tracking


matthewfeickert:
At the moment all of `pyhf`'s base depenendies support Python 3.12 :tada: 

https://github.com/scikit-hep/pyhf/blob/6a347f4be87ece5fd14b726acd99efca3b5cc7f4/pyproject.toml#L46-L55

and JAX supports Python 3.12 as of [`jaxlib` `v0.4.17`](https://pypi.org/project/jaxlib/0.4.17/#files)

https://github.com/scikit-hep/pyhf/blob/6a347f4be87ece5fd14b726acd99efca3b5cc7f4/pyproject.toml#L76-L79
. However, both `torch` and `tensorflow` do not have releases with Python 3.12 wheels out as of 2023-10-27.

```console
root@39c7d1045ec7:/# python --version
Python 3.12.0
root@39c7d1045ec7:/# python -m pip index versions torch
WARNING: pip index is currently an experimental command. It may be removed/changed in a future release without prior warning.
ERROR: No matching distribution found for torch
root@39c7d1045ec7:/# python -m pip index versions tensorflow
WARNING: pip index is currently an experimental command. It may be removed/changed in a future release without prior warning.
ERROR: No matching distribution found for tensorflow
root@39c7d1045ec7:/#
```

matthewfeickert:
Though now that we have PR #2340 in, we can start bringing in the backends as they support the releases.

matthewfeickert:
Interestingly enough, apparently `scrapbook`

https://github.com/scikit-hep/pyhf/blob/6a347f4be87ece5fd14b726acd99efca3b5cc7f4/pyproject.toml#L90-L107

requires `pyarrow` and [`pyarrow` `v13.0.0`](https://pypi.org/project/pyarrow/13.0.0/) doesn't support Python 3.12 yet either.

matthewfeickert:
Also `iminuit` doesn't have Python 3.12 support yet either as of [`iminuit` `v2.24.0`](https://pypi.org/project/iminuit/2.24.0/).

https://github.com/scikit-hep/pyhf/blob/6a347f4be87ece5fd14b726acd99efca3b5cc7f4/pyproject.toml#L81

matthewfeickert:
Issues to get fixed before Python 3.12 support can be added:

* https://github.com/scikit-hep/pyhf/issues/2453
* https://github.com/scikit-hep/pyhf/issues/2454

[...]

PR #2373, opened 2023-10-27
build: Remove installation of backends from 'test' extra

matthewfeickert:
# Description

To be able to take advantage of PR #2340 in CI it requires the ability to select a subset of dependencies to install as well (c.f. Issue #2372). To do this, this PR decouples the installation of the test dependendies in the 'test' extra from the backends that are installed in the 'all' extra and now requires 'all' to be explicitly given.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Remove the 'all' extra from the 'test' extra installs and add 'all' to 
  the 'develop' extra.
* Add the 'all' extra to all install commands that also install the 'test' extra.
```

### DIFF
--- a/.github/workflows/ci-windows.yml
+++ b/.github/workflows/ci-windows.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip setuptools wheel
-        python -m pip install --upgrade '.[test]'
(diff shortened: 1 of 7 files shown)


matthewfeickert:
Merging this given no feedback, but we can iterate on revisions if there are future requests for changes.

PR #2383, opened 2023-11-17
chore: Use native Ruff Jupyter support

henryiii:
This is a quick update to pre-commit to use the native notebook support in Ruff rather than an adaptor.

I was looking at ruff-format, but can't use it due to the mix of single and double quotes. Ruff has options for both, but not a "let me mix and match" mode like black.

```
* Remove github.com/nbQA-dev/nbQA in favor of native Ruff Jupyter support.
* Add support for python, pyi, jupyter types to black-jupyter hook.
* Update github.com/astral-sh/ruff-pre-commit: v0.1.4 → v0.1.6.
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -45,7 +45,6 @@ repos:
     hooks:
     - id: black-jupyter
       types_or: [python, pyi, jupyter]
-      types_or: [python, pyi, jupyter]
 
 -   repo: https://github.com/adamchainz/blacken-docs
     rev: 1.16.0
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -37,6 +37,8 @@ repos:
     rev: 23.10.1
     hooks:
     - id: black-jupyter
+      types_or: [python, pyi, jupyter]
+      types_or: [python, pyi, jupyter]
```
matthewfeickert, line 41:
```suggestion
```

Seems there's a duplicated key here, so removing that.

matthewfeickert (changes requested):

matthewfeickert (approved):
Thanks @henryiii! LGTM now after that small fix.

.pre-commit-config.yaml
henryiii: Oops, was picking this out of other changes, and somehow it applied the patch twice.

PR #2385, opened 2023-11-30
feat: Allow skipping of validation when combining workspaces

lorenzennio:
# Description

It would be nice to include the option to avoid the validation step when combining two workspaces:
https://github.com/scikit-hep/pyhf/blob/a1a31f17967740c93244f5d9ad17aea1fbef6be3/src/pyhf/workspace.py#L709

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add validate arg to workspace.combine to allow for optionally skipping
  workspace validation during. Default is to validate.
* Add test for validate arg being used.
* Add Lorenz Gaertner to contributors list.
```

### DIFF
--- a/tests/test_workspace.py
+++ b/tests/test_workspace.py
@@ -800,7 +800,7 @@ def test_combine_workspace_without_validation(mocker, workspace_factory, join):
         modifiers={
             modifier: f"renamed_{modifier}"
             for modifier, _ in ws.modifiers
-            if not modifier == "lumi"
+            if modifier != "lumi"
         },
         measurements={
             measurement: f"renamed_{measurement}"


kratsg:
Is there a reason you want to combine workspaces that don't produce a valid workspace?

lorenzennio:
Yes, I am working with workspaces that contain instructions for a custom modifier. These don't pass the validation in the first place, but for single workspaces I can simply skip it. Now I want to combine two of them, for a global fit.

kratsg (approved):

src/pyhf/workspace.py
```diff
@@ -770,7 +773,7 @@ def combine(cls, left, right, join='none', merge_channels=False):
             'observations': new_observations,
             'version': new_version,
         }
-        return cls(newspec)
+        return cls(newspec, validate=validate)
```
matthewfeickert, line 776: This should get covered by a test. I can add that though unless someone beats me to it.

matthewfeickert (changes requested):

matthewfeickert (changes requested):
@lorenzennio Can you also [add your name](https://github.com/scikit-hep/pyhf/blob/a1a31f17967740c93244f5d9ad17aea1fbef6be3/CONTRIBUTING.md#pull-request-procedure) to `docs
/contributors.rst` below 

https://github.com/scikit-hep/pyhf/blob/a1a31f17967740c93244f5d9ad17aea1fbef6be3/docs/contributors.rst?plain=1#L35

?

PR #2390, opened 2023-12-07
feat: Improve warnings for qmu and qmu_tilde for the set POI bounds

lorenzennio:
# Description

When running 
```python
model = pyhf.simplemodels.uncorrelated_background(signal=[5.0, 10.0], bkg=[50.0, 60.0], bkg_uncertainty=[5.0, 12.0])
pyhf.infer.hypotest(1.0, [53.0, 65.0] + model.config.auxdata, model, test_stat="qmu")
```
we get a warning
```
```
and vice versa for POIs not bounded at 0. 

But this should also tell you to set `test_stat="qtilde"` or (`test_stat="q"`).

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add information to the user in the warning that provides them with the
  higher level pyhf.infer APIs kwarg to set the correct test statistic.
```

### DIFF
--- a/src/pyhf/infer/test_statistics.py
+++ b/src/pyhf/infer/test_statistics.py
@@ -134,7 +134,7 @@ def qmu(mu, data, pdf, init_pars, par_bounds, fixed_params, return_fitted_pars=F
     if par_bounds[pdf.config.poi_index][0] == 0:
         log.warning(
             'qmu test statistic used for fit configuration with POI bounded at zero.\n'
-            + 'Use the qmu_tilde test statistic (pyhf.infer.test_statistics.qmu_tilde) instead.'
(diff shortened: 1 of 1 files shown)


matthewfeickert (changes requested):
Thanks @lorenzennio. So this is good, but these log warnings are raised in the test stat APIs themselves, where there is no `test_stat` kwarg, as the API is already the test stat. So I would suggest on a newline (as this will get long otherwsise) noting that if the user is calling from the `pyhf.infer.mle` APIs or `pyhf.infer.hypotest` this can be set via kwarg.

matthewfeickert (approved):

matthewfeickert:
@lorenzennio This fix is now out in [`pyhf` `v0.7.6`](https://github.com/scikit-hep/pyhf/releases/tag/v0.7.6).

PR #2396, opened 2023-12-08
fix: Catch invalid model specification with duplicate channels, samples and modifiers.

### LINKED ISSUE #2059: Attaching modifiers twice to a sample
### Summary

When adding a modifier twice to a sample, only one instance is used. For `normfactor` modifiers this could in principle be something a user would want to do on purpose (as a simplified way to scale different samples by `mu` and `sqrt(mu)`, see also #850). It might be useful to catch such specifications and warn users about the current behavior.

### OS / Environment

```console
n/a
```

### Steps to Reproduce

```python (paste below)
import pyhf

spec = {
    "channels": [
        {
            "name": "SR",
            "samples": [
                {
                    "data": [25.0],
                    "modifiers": [
                        {"data": None, "name": "mu", "type": "normfactor"},
                    ],
                    "name": "Signal",
                },
            ],
        }
    ],
    "measurements": [
        {"config": {"parameters": [], "poi": "mu"}, "name": "minimal_example"}
    ],
    "observations": [{"data": [35.0], "name": "SR"}],
    "version": "1.0.0",
}


def fit(spec):
    ws = pyhf.Workspace(spec)
    model = ws.model()
    data = ws.data(model) + model.config.auxdata
    print(pyhf.infer.mle.fit(data, model))


# fit with one modifier
fit(spec)

# add a second POI
spec["channels"][0]["samples"][0]["modifiers"].append(
    {"data": None, "name": "mu", "type": "normfactor"}
)
fit(spec)
```

### File Upload (optional)

_No response_

### Expected Results

warning or correct application of two POIs on the sample

### Actual Results

```console
[1.40000145]
[1.40000145]
# second modifier is ignored
```

### pyhf Version

```console
pyhf, version 0.7.0
```

### Code of Conduct

lorenzennio:
# Description

Fixes #2059 but extends to channels and samples. We want to catch the case where we have multiple channels with the same name in the model specification, where the first of these would simply be ignored as of now. Similarly, this applies to samples with the same name in a given channel or modifiers with the same name **and** type in a given sample.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add _check_for_duplicates method to _ChannelSummaryMixin that validates
  channels, samples, and modifiers.
* Raise InvalidModel with informative message identifying duplicates.
* Add test fixtures: model_duplicate_channels.json, model_duplicate_samples.json,
  model_duplicate_modifiers.json
* Add parametrized test_schema_catch_duplicates test in test_schema.py
```


kratsg (approved):
LGTM

matthewfeickert (approved):
Thanks very much @lorenzennio, and apologies to you for having maybe the longest unmerged PR. 😬🙇

PR #2397, opened 2023-12-08
feat: New schema for HS3 Support

kratsg:
Runtime validations:
- check that `staterror` is only defined once per sample
- check that if `staterror` is defined, that the sample has defined `errors`
- check that `contents` and `errors` have the same shape if `errors` is defined

Accepted Defaults:
- axes has min=0, max=1 (or edges = [0, 1])
- parameter/parameters is just the name of the parameter(s)
- staterror data is moved to the sample `errors`

To Dos:
- [ ] drop luminosity when customizable constraints can be supported (#1829)
- [ ] support schema migration (#1978)

References:
- https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/discussions/30
- https://github.com/root-project/root/blob/master/tutorials/roofit/rf515_hfJSON.json

# Pull Request Description

Please first read [CONTRIBUTING.md](https://github.com/scikit-hep/pyhf/tree/main/CONTRIBUTING.md).

Please describe the purpose of this pull request in some detail and what the specific feature being added will do. Reference and link to any relevant issues or pull requests (such as the issue in which this feature was first suggested).

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/schemas/1.1.0/defs.json
+++ b/src/pyhf/schemas/1.1.0/defs.json
@@ -1,0 +1,333 @@
+{
+    "$schema": "http://json-schema.org/draft/2020-12/schema#",
+    "$id": "https://scikit-hep.org/pyhf/schemas/1.1.0/defs.json",
+    "definitions": {
+        "workspace": {
+          "type": "object",
+          "properties": {
+              "distributions": { "type": "array", "items": {"$ref": "#/definitions/distribution"}, "minItems": 1 },
+              "measurements": { "type": "array", "items": {"$ref": "#/definitions/measurement"}, "minItems": 1 },
+              "observations": { "type": "array", "items": {"$ref": "#/definitions/observation" }, "minItems": 1 },
+              "type": { "const": "histfactory_dist" },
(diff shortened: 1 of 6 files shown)


src/pyhf/schemas/1.1.0/defs.json
```diff
@@ -0,0 +1,406 @@
...
+            "type": "object",
+            "properties": {
+                "name": { "type": "string" },
+                "auxdata": { "type": "array", "items": {"type": "number"}, "minItems": 1 },
+                "factors": { "type": "array", "items": {"type": "number"}, "minItems": 1 },
+                "sigmas": { "type": "array", "items": {"type": "number"}, "minItems": 1}
+            },
+            "required": ["name"],
+            "additionalProperties": false
+        },
```
alexander-held, line 91: Is this the object that goes into `parameter_points` in the current HS3 draft? I don't quite follow how to map this to the document. I started looking at this as I as wondering if the format would allow to conveniently distinguish between initial parameter values for the purpose of minimization starting point and default values for the purpose of dataset generation.

PR #2467, opened 2024-04-18
fix: Cast non-hashable types to floats in toms748_scan

### LINKED ISSUE #2466: toms748_scan doesn't work with JAX backend
### Summary

Hello; perhaps this is known but I thought I'd file a bug report just in case. I was testing the [`upper_limits` API](https://scikit-hep.org/pyhf/_generated/pyhf.infer.intervals.upper_limits.toms748_scan.html#pyhf.infer.intervals.upper_limits.toms748_scan) and discovered that the example given in the documentation doesn't seem to work with the JAX backend. It fails with a complaint about an unhashable array type (see the traceback). If I switch to the numpy backend, as shown in the documentation, it runs fine.

I see this on both EL7 in an ATLAS environment (`StatAnalysis,0.3,latest`) and on my own desktop (Fedora 38); in both cases I have the same pyhf version (0.7.6) and I manually installed `jax[CPU] == 0.4.26` on top of that.

I should add that things work fine with JAX if I use the version of `upper_limits` where I pass in a range of mu values to scan-- so I guess maybe some extra type conversion is needed to go from the JAX array type to a list or something hashable?

### OS / Environment

```console
# Linux
$ cat /etc/os-release
NAME="Fedora Linux"
VERSION="38 (Thirty Eight)"
ID=fedora
VERSION_ID=38
VERSION_CODENAME=""
PLATFORM_ID="platform:f38"
PRETTY_NAME="Fedora Linux 38 (Thirty Eight)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:38"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f38/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=38
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=38
SUPPORT_END=2024-05-14
```

### Steps to Reproduce

Install pyhf and JAX through pip; then try to run the example in the documentation, but with the JAX backend instead of numpy:

```python
import numpy as np
import pyhf
pyhf.set_backend("JAX")
)
observations = [51, 48]
data = pyhf.tensorlib.astensor(observations + model.config.auxdata)
obs_limit, exp_limits = pyhf.infer.intervals.upper_limits.toms748_scan(
    data, model, 0., 5., rtol=0.01
)
```

### File Upload (optional)

_No response_

### Expected Results

Ideally the example would run without crashing (as it does with the numpy backend).

### Actual Results

```console
  File "/home/bjr/.local/lib/python3.11/site-packages/pyhf/infer/intervals/upper_limits.py", line 130, in toms748_scan
    toms748(f, bounds_low, bounds_up, args=(level, 0), k=2, xtol=atol, rtol=rtol)
  File "/usr/lib64/python3.11/site-packages/scipy/optimize/_zeros_py.py", line 1374, in toms748
    result = solver.solve(f, a, b, args=args, k=k, xtol=xtol, rtol=rtol,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/scipy/optimize/_zeros_py.py", line 1229, in solve
    fc = self._callf(c)
         ^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/scipy/optimize/_zeros_py.py", line 1083, in _callf
    fx = self.f(x, *self.args)
         ^^^^^^^^^^^^^^^^^^^^^
  File "/home/bjr/.local/lib/python3.11/site-packages/pyhf/infer/intervals/upper_limits.py", line 95, in f
    f_cached(poi)[0] - level
    ^^^^^^^^^^^^^
  File "/home/bjr/.local/lib/python3.11/site-packages/pyhf/infer/intervals/upper_limits.py", line 80, in f_cached
    if poi not in cache:
       ^^^^^^^^^^^^^^^^
TypeError: unhashable type: 'jaxlib.xla_extension.ArrayImpl'
```

### pyhf Version

```console
$ pyhf --version
pyhf, version 0.7.6
```

### Code of Conduct

matthewfeickert:
# Description

Resolves #2466 

As `toms748_scan` uses caching of results it requires that the results are hashable objects, which arrays are not. So to keep using this approach convert to floats as needed.

Add testing for all backends for the upper limits inference tests.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/infer/intervals/upper_limits.py
+++ b/src/pyhf/infer/intervals/upper_limits.py
@@ -15,7 +15,8 @@ def __dir__():
 
 def _interp(x, xp, fp):
     tb, _ = get_backend()
-    return tb.astensor(np.interp(x, xp.tolist(), fp.tolist()))
+    # xp has already been turned into a list at this point
+    return tb.astensor(np.interp(x, xp, fp.tolist()))
 
 
 def toms748_scan(
@@ -79,13 +80,19 @@ def toms748_scan(
 
     def f_cached(poi):
         if poi not in cache:
-            cache[poi] = hypotest(
+            # FIXME: scipy.optimize.toms748 still operates on floats,
+            # not any form of ndarray, so want everything in the
+            # cache to be a float.
+            # This may change with the Python array API standard
+            # in the future.
+            cls_obs, cls_exp_band = hypotest(
                 poi,
                 data,
                 model,
                 return_expected_set=True,
                 **hypotest_kwargs,
             )
+            cache[poi] = (float(cls_obs), [float(x) for x in cls_exp_band])
         return cache[poi]
 
     def f(poi, level, limit=0):
@@ -198,7 +205,9 @@ def linear_grid_scan(
     obs = tb.astensor([[r[0]] for r in results])
     exp = tb.astensor([[r[1][idx] for idx in range(5)] for r in results])
 
-    result_array = tb.concatenate([obs, exp], axis=1).T
+    # TODO: Can use `.T` after TensorFlow support is removed.
+    result_array = tb.transpose(tb.concatenate([obs, exp], axis=1))
+    result_array = tb.tolist(result_array)
 
     # observed limit and the (0, +-1, +-2)sigma expected limits
     limits = [_interp(level, result_array[idx][::-1], scan[::-1]) for idx in range(6)]
--- a/src/pyhf/infer/intervals/upper_limits.py
+++ b/src/pyhf/infer/intervals/upper_limits.py
@@ -92,6 +92,10 @@ def toms748_scan(
         # Use integers for limit so we don't need a string comparison
         # limit == 0: Observed
         # else: expected
+
+        # Arrays are not hashable types, so cast to float
+        poi = float(poi)
+
         return (
             f_cached(poi)[0] - level
             if limit == 0
--- a/tests/test_infer.py
+++ b/tests/test_infer.py
@@ -23,7 +23,7 @@ def check_uniform_type(in_list):
     )
 
 
-def test_toms748_scan(tmp_path, hypotest_args):
+def test_toms748_scan(backend, tmp_path, hypotest_args):
     """
     Test the upper limit toms748 scan returns the correct structure and values
     """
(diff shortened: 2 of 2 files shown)


tests/test_infer.py
```diff
@@ -72,18 +73,20 @@ def test_toms748_scan_bounds_extension(hypotest_args):
         data, model, 3, 5, rtol=1e-8
     )
 
-    assert observed_limit == pytest.approx(observed_limit_default)
+    # FIXME: Remove float cast after TensorFlow support removed
```
matthewfeickert, line 76: @kratsg Running the TensorFlow tests is _very_ slow and adds minutes to this run. As we're going to be dropping TensorFlow in the future, I think that we could probably just skip TensorFlow in these tests. Thoughts?

matthewfeickert:
CI is currently failing for Apple silicon macOS.

src/pyhf/infer/intervals/upper_limits.py
```diff
@@ -15,7 +15,8 @@ def __dir__():
 
 def _interp(x, xp, fp):
```
kratsg, line 16:
```suggestion
def _interp(x, xp: list[float], fp):
```

src/pyhf/infer/intervals/upper_limits.py
```diff
@@ -15,7 +15,8 @@ def __dir__():
...
     tb, _ = get_backend()
-    return tb.astensor(np.interp(x, xp.tolist(), fp.tolist()))
+    # xp has already been turned into a list at this point
```
kratsg, line 18: can drop this comment with a typehint above as suggested

PR #2473, opened 2024-05-03
ci: Add GitHub artifact attestations to package distribution

matthewfeickert:
# Description

* Add generation of GitHub artifact attestations to built sdist and wheel before upload.
  c.f.:
   - https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/
   - https://docs.github.com/en/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds
* Add verification of artifact attestation before publishing to PyPI using the `gh attestation verify` CLI API, added in `v2.49.0`.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add generation of GitHub artifact attestations to built sdist and wheel
  before upload.
  c.f.:
   - https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/
   - https://docs.github.com/en/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds
* Add verification of artifact attestation before publishing to PyPI
  using the 'gh attestation verify' CLI API, added in v2.49.0.
   - c.f. https://github.com/cli/cli/releases/tag/v2.49.0
```

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -1,10 +1,10 @@
 name: publish distributions
 on:
   push:
-    # branches:
-    # - main
-    # tags:
-    # - v*
+    branches:
+    - main
+    tags:
+    - v*
   pull_request:
     branches:
     - main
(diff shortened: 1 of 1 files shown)


matthewfeickert:
Currently blocked by https://github.com/actions/runner-images/issues/9784, without updating `ubuntu-latest` to `ubuntu-24.04`.

matthewfeickert:
So it seems that the most important takeaway is described https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/#a-more-secure-future, and while it seems that this should mean that the number of attestations generated isn't relevant as everything will be machine generated, it seems that it might(?) be useful to keep https://github.com/scikit-hep/pyhf/attestations containing mostly only releases.

@woodruffw @di, not to bug you with things that go beyond https://github.com/sigstore/sigstore-python, but is this the correct take (that we should only be publishing attestations for releases)?

matthewfeickert:
Also cc @sethmlarson given very helpful feedback on https://github.com/scientific-python/summit-2024/issues/9#issuecomment-2107536124.

woodruffw:
Thanks for the ping @matthewfeickert!

IMO, the number of attestations generated is not _currently_ relevant: these attestations don't appear on PyPI and are "generic" in nature (i.e. don't express a specific intent to build, release, etc.), so having multiple of them doesn't accomplish much at the moment.

However, to take a step back: we're building PEP 740 on the same foundations (Sigstore + `sigstore-python`), and it _will_ have support for multiple attestations, e.g. for distinguishing publish, release, and also third-party counter-attestations. That PEP uses the same underlying technologies as GitHub's Attestations feature but requires attestations to be uploaded to PyPI as well, which means we'll be building it into the standard PyPA tooling.

That's still a little out and not directly related to what you asked though 😅. But the TL;DR: IMO one GitHub attestation is fine, and in the future PEP 740 will *also* allow your tools to (automatically) load multiple attestations per-file onto PyPI.

.github/workflows/publish-package.yml
```diff
@@ -31,6 +31,11 @@ jobs:
   build:
     name: Build Python distribution
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      attestations: write
+      contents: read
```
matthewfeickert, line 37: Permissions added here given https://docs.github.com/en/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-binaries

matthewfeickert:
@meeseeksdev backport to release/v0.7.x

PR #2526, opened 2024-06-26
feat: Log warning when fitted parameter is at the bounds

### LINKED ISSUE #2525: Warn when parameter hits bounds
### Summary

Hey, 
I recently ran a lot of toys for a very large model and had a hard time debugging, as some nuisance parameters hit the parameter bounds for some of the toys.
It would be great if a warning could be added in that case. What do you think?

### Additional Information

A simple example:

```python (paste below)
import pyhf

bkg = [5]
sig = [3]

data = bkg

samples = [
    dict(name="signal", data=list(sig), modifiers=[dict(name="mu", type="normfactor", data=None)]),
    dict(name="background", data=list(bkg), modifiers=[]),
]
model = pyhf.Model(dict(channels=[dict(name="singlechannel", samples=samples)]))

init = [5]
bounds =[(3, 10)]
pyhf.infer.mle.fit(data, model, init_pars=init, par_bounds=bounds)
```

### Code of Conduct

kratsg:
# Pull Request Description

Resolves #2525.

When nuisance parameters hit their bounds during fitting (e.g. in a large-model toy run), there was previously no indication that this had happened, making it difficult to diagnose suspect fit results. This PR adds a `WARNING` log message from `_internal_postprocess` whenever a fitted parameter lands exactly on a bound, reporting the parameter index, the fitted value, and the bound range.

The implementation iterates over the free parameters in `fitresult.x` alongside their bounds (both already stripped of fixed parameters by `shim()` before optimization). Three architectural concerns (batching, fixed-parameter skipping, and backend compatibility) were investigated (using CLAUDE) and confirmed to be non-issues given the current design; a clarifying comment documents why the alignment is guaranteed.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add a WARNING log message in _internal_postprocess when a fitted
  parameter lands exactly on a bound. Reports the parameter index, 
  fitted value, and bound range.
* Add regression tests for lower bound, upper bound, and interior
  cases via _internal_postprocess directly with a synthetic
  OptimizeResult, making the check independent of optimizer
  floating-point behaviour.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Added clear warnings when fitted parameters reach, approach, exceed, or produce non-finite values relative to configured bounds.
  * Warnings identify affected parameters by name or index and include their values and limits.
  * Improved handling of fixed, stitched, unbounded, and one-sided-bound parameters.
  * Added support for optimizer bounds provided through SciPy’s standard bounds format.
  * Consolidated bound warnings so each fit reports them once.

* **Tests**
  * Added comprehensive coverage for boundary conditions across supported optimization scenarios.

### DIFF
--- a/src/pyhf/optimize/mixins.py
+++ b/src/pyhf/optimize/mixins.py
@@ -88,6 +88,8 @@ class OptimizerMixin:
         tensorlib, _ = get_backend()
 
         # TODO: check how to handle this for batching
+        # TODO: handle skipping fixed parameters
+        # TODO: handle various backends
         for par_index, (fitted_par, bound) in enumerate(zip(fitresult.x, par_bounds)):
             if fitted_par in bound:
                 log.warning(f'parameter at index {par_index} is at the bounds')
--- a/src/pyhf/optimize/mixins.py
+++ b/src/pyhf/optimize/mixins.py
@@ -90,9 +90,17 @@ class OptimizerMixin:
         # TODO: check how to handle this for batching
         # TODO: handle skipping fixed parameters
         # TODO: handle various backends
-        for par_index, (fitted_par, bound) in enumerate(zip(fitresult.x, par_bounds)):
-            if fitted_par in bound:
-                log.warning(f"parameter at index {par_index} is at the bounds")
+        for par_index, (fitted_par, (lower, upper)) in enumerate(
+            zip(fitresult.x, par_bounds)
+        ):
+            if fitted_par in (lower, upper):
+                log.warning(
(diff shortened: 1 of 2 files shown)


matthewfeickert:
@kratsg I think that we should probably look at the full state of what I've added from "fix: Correct at-bound warning alignment, tolerance, and fixed-parameter handling" onward and evaluate how much of this is needed and how much of this is my sloppy debugging of minimization issues that are outside the scope of this PR. Your input would be welcome here, but you should also feel free to just move any and all commits that I've made to another branch and revert things to a state that we can discuss where the PR scope should develop towards.

Issue #2554, opened 2024-11-01
Fix DeprecationWarning for PyTorch for NumPy 2.0


matthewfeickert:
With NumPy 2.0, PyTorch now has a

```
DeprecationWarning: In future, it will be an error for 'np.bool' scalars to be interpreted as an index
```

This started happening on 2024-10-24 when [TensorFlow v2.18.0](https://pypi.org/project/tensorflow/2.18.0/) was released with NumPy 2.0 support and NumPy 2.0 entered the CI system. The diff in the packages installed between the previous day and the first failing run is

| 2024-10-23 | 2024-10-24 | 
| -: | -: |
| **numpy                     1.26.4** |	**numpy                     2.0.2**  |
| protobuf                  4.25.5 |	protobuf                  5.28.3 |
| tensorboard               2.17.1 |	tensorboard               2.18.0 |
| tensorflow                2.17.0 |	tensorflow                2.18.0 |
| tf_keras                  2.17.0 |	tf_keras                  2.18.0 |
| tinycss2                  1.3.0  |	tinycss2                  1.4.0  |
| Werkzeug                  3.0.4  |	Werkzeug                  3.0.5  |

We should try to see if we can fix this behavior to avoid the `DeprecationWarning`. Worse case scenario we add it to the [existing `filterwarnings`](https://github.com/scikit-hep/pyhf/blob/e324ba383a0ca6adf7bc008ca5e948c1b69e06cd/pyproject.toml#L210).

PR #2566, opened 2024-12-21
chore: Drop Python 3.8

kratsg:
# Pull Request Description

Drop python 3.8 support (would be hard to fix most typing-related issues anyway). This will also fix maintenance issues with the CI we've been having hopefully.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,13 +30,13 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
         include:
           - os: macos-latest
-            python-version: '3.12'
+            python-version: '3.13'
           # Intel runner
           - os: macos-13
-            python-version: '3.12'
+            python-version: '3.13'
 
     steps:
     - uses: actions/checkout@v4
@@ -47,17 +47,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Install dependencies
-      if: matrix.python-version != '3.8'
       run: |
         python -m pip install uv
         uv pip install --system --upgrade ".[all,test]"
-
-      # c.f. https://github.com/astral-sh/uv/issues/2062
-    - name: Install dependencies (Python 3.8)
-      if: matrix.python-version == '3.8'
-      run: |
-        python -m pip install --upgrade pip
-        python -m pip install --upgrade ".[all,test]"
 
     - name: List installed Python packages
       run: python -m pip list
@@ -97,7 +89,7 @@ jobs:
         coverage xml
 
     - name: Report contrib coverage with Codecov
-      if: github.event_name != 'schedule' && matrix.python-version == '3.12' && matrix.os == 'ubuntu-latest'
+      if: github.event_name != 'schedule' && matrix.python-version == '3.13' && matrix.os == 'ubuntu-latest'
       uses: codecov/codecov-action@v5
       with:
         fail_ci_if_error: true
@@ -107,17 +99,17 @@ jobs:
 
     - name: Test docstring examples with doctest
       # TODO: Don't currently try to match amd64 and arm64 floating point for docs, but will in the future.
-      if: matrix.python-version == '3.12' && matrix.os != 'macos-latest'
+      if: matrix.python-version == '3.13' && matrix.os != 'macos-latest'
       run: coverage run --data-file=.coverage-doctest --module pytest src/ README.rst
 
     - name: Coverage report for doctest only
-      if: matrix.python-version == '3.12' && matrix.os != 'macos-latest'
+      if: matrix.python-version == '3.13' && matrix.os != 'macos-latest'
       run: |
         coverage report --data-file=.coverage-doctest
         coverage xml --data-file=.coverage-doctest -o doctest-coverage.xml
 
     - name: Report doctest coverage with Codecov
-      if: github.event_name != 'schedule' && matrix.python-version == '3.12' && matrix.os == 'ubuntu-latest'
+      if: github.event_name != 'schedule' && matrix.python-version == '3.13' && matrix.os == 'ubuntu-latest'
       uses: codecov/codecov-action@v5
       with:
         fail_ci_if_error: true
@@ -126,6 +118,6 @@ jobs:
         token: ${{ secrets.CODECOV_TOKEN }}
 
     - name: Run benchmarks
-      if: github.event_name == 'schedule' && matrix.python-version == '3.12'
+      if: github.event_name == 'schedule' && matrix.python-version == '3.13'
       run: |
         pytest --benchmark-sort=mean tests/benchmarks/test_benchmark.py
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,14 +58,14 @@ repos:
     hooks:
       - &mypy
         id: mypy
-        name: mypy with Python 3.8
+        name: mypy with Python 3.9
         files: src
         additional_dependencies:
           ['numpy', 'types-tqdm', 'click', 'types-jsonpatch', 'types-pyyaml', 'types-jsonschema', 'importlib_metadata', 'packaging']
-        args: ["--python-version=3.8"]
+        args: ["--python-version=3.9"]
       - <<: *mypy
-        name: mypy with Python 3.12
-        args: ["--python-version=3.12"]
+        name: mypy with Python 3.13
+        args: ["--python-version=3.13"]
 
 -   repo: https://github.com/codespell-project/codespell
     rev: v2.3.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ dynamic = ["version"]
 description = "pure-Python HistFactory implementation with tensors and autodiff"
 readme = "README.rst"
 license = { text = "Apache-2.0" }  # SPDX short identifier
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 authors = [
     { name = "Lukas Heinrich", email = "lukas.heinrich@cern.ch" },
     { name = "Matthew Feickert", email = "matthew.feickert@cern.ch" },
@@ -35,11 +35,11 @@ classifiers = [
     "Operating System :: OS Independent",
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: Implementation :: CPython",
     "Topic :: Scientific/Engineering",
     "Topic :: Scientific/Engineering :: Physics",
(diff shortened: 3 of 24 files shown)


src/pyhf/infer/calculators.py
```diff
@@ -120,7 +120,7 @@ def cdf(self, value):
             >>> import pyhf
             >>> pyhf.set_backend("numpy")
             >>> bkg_dist = pyhf.infer.calculators.AsymptoticTestStatDistribution(0.0)
-            >>> bkg_dist.cdf(0.0)
+            >>> print(bkg_dist.cdf(0.0))
```
matthewfeickert, line 123: Did something change RE: the way doctest works where the `print`s are now necessary? We had always treated this like REPL behavior, not script behavior.

kratsg: No, numpy changed. ubuntu-latest has numpy 2.x which gives `np.float64(0.0)` while macos-13 uses numpy 1.x which gives `0.0` -- we can't write a test that relies on `repr` for two diff major versions of numpy.

kratsg:
```python
>>> import numpy as np
>>> np.__version__
'2.0.2'
>>> np.float64(0.0)
np.float64(0.0)
>>> print(np.float64(0.0))
0.0
```

and

```python
>>> import numpy as np
>>> np.__version__
'1.26.4'
>>> np.float64(0.0)
0.0
>>> print(np.float64(0.0))
0.0
```

matthewfeickert: > we can't write a test that relies on `repr` for two diff major versions of numpy.

I understand what you mean, but I thought that we already did have ways to indicate to doc tests that it should only test on certain operating systems as we would already get different versions.

Let me take the afternoon today (once I'm free from running interviews for a search I'm on) to play with this and if I can't get it then we can just accept it and move on.

pyproject.toml
```diff
@@ -11,7 +11,7 @@ dynamic = ["version"]
 description = "pure-Python HistFactory implementation with tensors and autodiff"
 readme = "README.rst"
 license = { text = "Apache-2.0" }  # SPDX short identifier
-requires-python = ">=3.8"
+requires-python = ">=3.9"
```
matthewfeickert, line 14: I'm personally fine with jumping up to Python 3.9. I think though as the `0.7.6` release is still on Python 3.7 that regardless of features we should make a `0.8.0` release after this that is the same as what we have in the `0.7.x` series but just with non-EOL Python support. That will make multiple things easier to maintain and Python 3.9 is not new at all so I don't see that as an issue.

src/pyhf/modifiers/histosys.py
```diff
@@ -101,6 +101,8 @@ class histosys_combined:
     def __init__(
         self, modifiers, pdfconfig, builder_data, interpcode='code0', batch_size=None
     ):
+        default_backend = pyhf.default_backend
```
matthewfeickert, line 104: Is this needed just to deal with the PyTorch warnings from https://github.com/scikit-hep/pyhf/issues/2554?

kratsg: it's needed in order to type `_histosys_mask` correctly as a tensor (see the below lines where it is being used). You could drop it, but then the typehints won't work due to the changes in typing in numpy.

docs/conf.py
```diff
@@ -529,6 +529,15 @@ def setup(app):
...
     r'https://pyhf.readthedocs.io/en/.*',
+    # the following are 403s as they map to journals.aps.org or academic.oup.com
+    r'https://doi.org/10.1093/ptep/ptad144',
+    r'https://doi.org/10.1103/PhysRevD.104.055017',
+    r'https://doi.org/10.1103/PhysRevD.107.095021',
+    r'https://doi.org/10.1103/PhysRevD.108.016002',
+    r'https://doi.org/10.1103/PhysRevD.106.032005',
+    r'https://doi.org/10.1103/PhysRevLett.127.181802',
+    r'https://doi.org/10.1103/PhysRevLett.130.231801',
+    r'https://doi.org/10.1103/PhysRevLett.131.211802',
```
kratsg, line 533: @matthewfeickert i think this is for the broken docs

.github/workflows/ci.yml
```diff
@@ -30,13 +30,13 @@ jobs:
...
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
         include:
           - os: macos-latest
-            python-version: '3.12'
+            python-version: '3.13'
           # Intel runner
-          - os: macos-13
-            python-version: '3.12'
+          - os: macos-15-intel
+            python-version: '3.13'
```
matthewfeickert, line 39: Yeah, so having PyTorch fail here is just something that we'll have to get used to for a bit until we decide what to do. Is it possible to turn on allow failures but only for one set in the matrix? The full CI currently doesn't run because we currently fail out everything.

src/pyhf/modifiers/histosys.py
```diff
@@ -128,10 +130,13 @@ def __init__(
             ]
             for m in keys
         ]
-        self._histosys_mask = [
-            [[builder_data[m][s]['data']['mask']] for s in pdfconfig.samples]
-            for m in keys
-        ]
+        self._histosys_mask = default_backend.astensor(
```
matthewfeickert: I understand why we have these for PyTorch, but could we add a comment above all of them that this is for PyTorch only? Though being explicit about the `dtype` isn't a bad thing in general, I forget how much of penalty we get by calling `astensor` on stuff. I know that the backends are efficient about it, but I don't remember the Python overhead. If it is possible to drop this if we drop PyTorch that might be a good idea?

.ackrc
```diff
@@ -1,2 +1,8 @@
 --ignore-directory=docs/_generated
 --ignore-directory=htmlcov
+--ignore-directory=.mypy_cache
+--ignore-directory=.ipynb_checkpoints
+--ignore-directory=userHelp
+--ignore-file=ext:lock
+--ignore-file=ext:log
+--ignore-file=ext:xml
```
matthewfeickert: I had to look it up as we've never touched this file since you created this in PR https://github.com/scikit-hep/pyhf/pull/305, but I guess this is like a dotfile to `ack`. What is this file actually used for though. I guess we need it for something in the tests, but are these additions needed or just recommended by a tool?

pyproject.toml
```diff
@@ -51,7 +49,7 @@ dependencies = [
     "jsonschema>=4.15.0",  # for utils
     "pyyaml>=5.1",  # for parsing CLI equal-delimited options
     # c.f. https://github.com/scikit-hep/pyhf/issues/2593 for scipy v1.16.0 upper bound
-    "scipy>=1.5.2,<1.16.0",  # requires numpy, which is required by pyhf and tensorflow
+    "scipy>=1.5.2",  # requires numpy, which is required by pyhf
```
matthewfeickert: The cap on `scipy` comes from PR https://github.com/scikit-hep/pyhf/pull/2592. Did this get fixed in `scipy` `v1.16.1` or `v1.16.2`? If so, we should disallow `1.16.0`.

.ackrc
kratsg: just to grep for changes, I'll remove it.

kratsg:
> The other thing that I wanted to get your 👍 / 👎 on is that this is a pretty dang big PR (for good reason!). As we know that there's going to be some CI failures that we have to fix anyway, how do you feel about pulling all of the "drop TensorFlow" stuff out into a seperate PR so that this gets more atomic? Given that you've done all the work for this PR by yourself, I'm happy to volunteer to do that PR so that I don't give you Git surgery.

It's impossible. We've pushed this PR so far back that we have to drop pytorch AND tensorflow AND have these fixes just to have a working CI. I cannot tell you what to fix if you just drop one dependency without fixing the existing CI. You're stuck with this big PR I think.

matthewfeickert:
> It's impossible. We've pushed this PR so far back that we have to drop pytorch AND tensorflow AND have these fixes just to have a working CI. I cannot tell you what to fix if you just drop one dependency without fixing the existing CI. You're stuck with this big PR I think.

I agree it isn't possible to get the CI passing without a combination of things. But as we already have failing CI in some places I don't think it is a bad thing to say "we understand this" but let's break our fixes up into more focused PRs (which importantly will make cherry-picking things easier if needed).

matthewfeickert:
This is currently reverted atm(?), but for the

```yaml
          - os: macos-15-intel
            python-version: '3.13'
```

change to CI if we know that x86 macOS isn't going to be supported by most of our backends for Python 3.13, should we just test it at Python 3.12 or whatever the highest version where there are still wheels? That would make it pretty clear when we'll have to drop x86 macOS support, as the decision will have been made for us.

matthewfeickert:
@kratsg I think if you rebase this then it should be ready to go.

tests/test_backend_consistency.py
```diff
@@ -127,6 +127,10 @@ def test_hypotest_qmu_tilde(
     numpy_ratio = np.divide(test_statistic, test_statistic[0])
     numpy_ratio_delta_unity = np.absolute(np.subtract(numpy_ratio, 1))
 
+    # compare tensor libraries to each other
+    tensors_ratio = np.divide(test_statistic[0], test_statistic[1])
+    tensors_ratio_delta_unity = np.absolute(np.subtract(tensors_ratio, 1))
```
matthewfeickert: We removed this part in the PyTorch PR as there's no longer multiple tensor backends to compare to, we just compare to NumPy.

src/pyhf/modifiers/histosys.py
matthewfeickert: Do we still need this explicit `dtype=bool` casting bit for type hinting? I think we were only going to do this before because PyTorch raised an error.

kratsg:
> @kratsg Why are these additional import changes needed? PR #2625 is a subset of this (involves a cherry-pick+squash) and it doesn't require those changes.

because you didn't bump the minimum `mypy` version for python to 3.10. This changes enough in python typing (we shouldn't use 3.9 for typing).

src/pyhf/tensor/numpy_backend.py
```diff
@@ -384,7 +388,7 @@ def simple_broadcast(self, *args: Sequence[Tensor[T]]) -> Sequence[Tensor[T]]:
 
             >>> import pyhf
             >>> pyhf.set_backend("numpy")
-            >>> pyhf.tensorlib.simple_broadcast(
+            >>> list(pyhf.tensorlib.simple_broadcast(
```
kratsg:
```suggestion
            >>> pyhf.tensorlib.simple_broadcast(
```

.pre-commit-config.yaml
```diff
@@ -55,17 +55,18 @@ repos:
...
         files: src
         additional_dependencies:
           ['numpy', 'types-tqdm', 'click', 'types-jsonpatch', 'types-pyyaml', 'types-jsonschema', 'importlib_metadata', 'packaging']
-        args: ["--python-version=3.8"]
+        args: ["--python-version=3.10"]
       - <<: *mypy
-        name: mypy with Python 3.13
-        args: ["--python-version=3.13"]
+        name: mypy with Python 3.12
+        args: ["--python-version=3.12"]
```
kratsg:
```suggestion
        name: mypy with Python 3.13
        args: ["--python-version=3.13"]
```

src/pyhf/tensor/numpy_backend.py
```diff
@@ -205,9 +206,12 @@ def conditional(
...
         return true_callable() if predicate else false_callable()
 
-    def tolist(self, tensor_in: Tensor[T] | list[T]) -> list[T]:
+    def tolist(
+        self, tensor_in: Tensor[T] | list[T]
+    ) -> int | float | complex | list[T] | list[Any]:
         try:
-            return tensor_in.tolist()  # type: ignore[union-attr,no-any-return]
+            result = tensor_in.tolist()  # type: ignore[union-attr]
+            return cast(Union[int, float, complex, list[T], list[Any]], result)
```
matthewfeickert: PR https://github.com/scikit-hep/pyhf/pull/2625 is the required/sufficient of this PR to drop Python 3.8. What is the need for the `pyhf.tensor.manager` imports in this PR along with the casting working in `tolist`?

If this is a general improvement to typing that is needed to make something easier then we should factor that out from a Python support change.

pyproject.toml
```diff
@@ -10,8 +10,8 @@ name = "pyhf"
 dynamic = ["version"]
 description = "pure-Python HistFactory implementation with tensors and autodiff"
 readme = "README.rst"
-license = "Apache-2.0"
-requires-python = ">=3.8"
+license = { text = "Apache-2.0" }  # SPDX short identifier
```
matthewfeickert: This shouldn't get changed back. This was intentionally changed in PR https://github.com/scikit-hep/pyhf/pull/2623.

```suggestion
license = "Apache-2.0"
```

pyproject.toml
matthewfeickert, line 1: PR https://github.com/scikit-hep/pyhf/pull/2625 also removes `importlib_resources`, as that was for Pythons older than 3.9.

src/pyhf/tensor/numpy_backend.py
kratsg: This behavior is unreliable right now. Our `mypy` version is quite old... so I'd rather bump that up too... If you'd rathe rjust go with your PR instead because this one took 2+ months to deal with, we can do that.. but we still have many issues to deal with that are going to be hard to disentangle and not worth the time to do so when we're just trying to fell-swoop fix all issues.. It's not like we need to go back and undo the changes for typehints because they break runtime code (they won't).

matthewfeickert (approved):
Thanks @kratsg.

src/pyhf/tensor/numpy_backend.py
matthewfeickert: >  but we still have many issues to deal with that are going to be hard to disentangle and not worth the time to do so when we're just trying to fell-swoop fix all issues

@kratsg I don't think I understand the issues that exist for typing here, which is my fault. Can you elaborate them in an Issue so that we can get them addressed, regardless of if we get them into the `v0.8.0` release?

Issue #1422, opened 2021-04-29
jax is not happy with simplemodels?


kratsg:
```python
>>> import pyhf
>>> pyhf.set_backend("jax")
>>> m = pyhf.simplemodels.hepdata_like([10], [15], [5])
>>> pyhf.infer.mle.fit([12.5], m)
```

crashes like so

<img width="948" alt="Screen Shot 2021-04-29 at 1 06 17 PM" src="https://user-images.githubusercontent.com/761483/116590994-5e228180-a8ec-11eb-9005-dc3669fbe823.png">

with a possible hint?

> This error can occur when a JAX Tracer object is passed to a raw numpy function, or a method on a numpy.ndarray object. You might want to check that you are using `jnp` together with `import jax.numpy as jnp` rather than using `np` via `import numpy as np`. If this error arises on a line that involves array indexing, like `x[idx]`, it may be that the array being indexed `x` is a raw numpy.ndarray while the indices `idx` are a JAX Tracer instance; in that case, you can instead write `jax.device_put(x)[idx]`.

matthewfeickert:
This also fails

```python
import pyhf

if __name__ == "__main__":
    pyhf.set_backend("numpy")
    model = pyhf.simplemodels.hepdata_like([10], [15], [5])
    observations = [12.5]
    print(pyhf.infer.mle.fit(observations, model))
```

```pytb
Eval failed for data [12.5] pars: [1.0, 1.0]
Traceback (most recent call last):
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/pdf.py", line 722, in logpdf
    raise exceptions.InvalidPdfData(
pyhf.exceptions.InvalidPdfData: eval failed as data has len 1 but 2 was expected
Traceback (most recent call last):
  File "/tmp/fail.py", line 8, in <module>
    print(pyhf.infer.mle.fit(observations, model))
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/infer/mle.py", line 125, in fit
    return opt.minimize(
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/optimize/mixins.py", line 160, in minimize
    result = self._internal_minimize(**minimizer_kwargs, options=kwargs)
  File "/home/feickert/Code/GitHub/pyhf/src/pyhf/optimize/mixins.py", line 38, in _internal_minimize
    result = self._minimize(
[...]

PR #2580, opened 2025-03-21
fix: jax backend tolist for tracers in logging

pfackeldey:
# Description

This PR closes https://github.com/scikit-hep/pyhf/issues/1422

This failure is encountered during tracing time. JAX can not convert a tracer during `.tolist` which is why it fails with a different error. This PR checks if we're currently in tracing time and instead returns the tracer (or rather its abstract value for a little nicer representation).

The new error looks as expected like this now:
```python
>>> import pyhf
>>> pyhf.set_backend("jax")
>>> pyhf.simplemodels.uncorrelated_background([10], [15], [5])
>>> pyhf.infer.mle.fit([12.5], m)
>>> ... InvalidPdfData: "eval failed as data has len 1 but 2 was expected"
```
The logging will print:
```python
"Eval failed for data ShapedArray(float64[1]) pars: ShapedArray(float64[2])"
```
This is the most information available during tracing time (shape and dtype), there's not much more to give to the user.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

### DIFF
--- a/src/pyhf/tensor/jax_backend.py
+++ b/src/pyhf/tensor/jax_backend.py
@@ -2,6 +2,7 @@ from jax import config
 
 config.update('jax_enable_x64', True)
 
+import jax
 from jax import Array
 import jax.numpy as jnp
 from jax.scipy.special import gammaln, xlogy
@@ -12,6 +13,10 @@ import scipy.stats as osp_stats
 import logging
 
 log = logging.getLogger(__name__)
+
+
+def currently_jitting():
+    return isinstance(jnp.array(1) + 1, jax.core.Tracer)
 
 
 class _BasicPoisson:
@@ -184,6 +189,9 @@ class jax_backend:
         return true_callable() if predicate else false_callable()
 
     def tolist(self, tensor_in):
+        if currently_jitting():
+            # .aval is the abstract value and has a little nicer representation
+            return tensor_in.aval
         try:
             return jnp.asarray(tensor_in).tolist()
         except (TypeError, ValueError):


kratsg (approved):

kratsg:
this likely needs #2566 in

src/pyhf/tensor/jax_backend.py
```diff
@@ -14,6 +15,10 @@
 log = logging.getLogger(__name__)
 
 
+def currently_jitting():
```
matthewfeickert: Should this exist in the public API? If not, can this be renamed to `_currently_jitting`?

pfackeldey:
I've addressed your comments @matthewfeickert, hope _this_ test is what you were looking for.

tests/test_backends.py
```diff
@@ -84,3 +84,19 @@ def test_backend_array_type(backend):
 def test_tensor_array_types():
     # can't really assert the content of them so easily
     assert pyhf.tensor.array_types
+
+
+@pytest.mark.only_jax
+def test_jax_data_shape_mismatch_during_jitting(backend):
```
matthewfeickert, line 90: Thanks @pfackeldey! I went ahead and used the `pyhf` `conftest.py` tools that we've built to be able to scope this to the JAX backend only.

I also added some small notes on the test, but please revise them if they aren't fully correct.

src/pyhf/tensor/jax_backend.py
```diff
@@ -14,6 +15,10 @@
 log = logging.getLogger(__name__)
 
 
+def _currently_jitting():
```
matthewfeickert, line 18: Can you add a one line docstring here that just explains to the naive person how JAX is determining this? e.g. just noting what JAX `jit` is doing with the tracers.

matthewfeickert:
Peter, Can you also add the

> Summarize commit messages into a comprehensive review of the PR

part in the PR body? (Sorry, missed that till now.)

PR #2590, opened 2025-06-08
cast to list, as the default backend numpy casts the bools to np.bool…

WolfgangWaltenberger:
fix: we are casting the numpy array to a list. that way the booleans get cast from np.bool to python's native bool. i am doing
this because my torch backend has problems casting np.bool to torch.bool:

>In[0]: torch.as_tensor ( [False,np.bool_(True)], dtype=torch.bool )
> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> Cell In[6], line 1
> ----> 1 torch.as_tensor ( [False,np.bool_(True)], dtype=torch.bool )>
>
> TypeError: 'numpy.bool' object cannot be interpreted as an integer

Which happens for me at 

 File "/home/walten/.venvs/default/lib/python3.13/site-packages/pyhf/tensor/pytorch_backend.py", line 207, in astensor
    return torch.as_tensor(tensor_in, dtype=dtype)
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'numpy.bool' object cannot be interpreted as an integer

Feel free though to solve this weird casting issue differently. Why it seems to be a problem for me but not for you, eludes me.

Wolfgang

### DIFF
--- a/src/pyhf/modifiers/histosys.py
+++ b/src/pyhf/modifiers/histosys.py
@@ -67,6 +67,7 @@ class histosys_builder:
                 sample["data"]["mask"] = default_backend.concatenate(
                     sample["data"]["mask"]
                 )
+                sample["data"]["mask"] = sample["data"]["mask"].tolist()
                 sample["data"]["lo_data"] = default_backend.concatenate(
                     sample["data"]["lo_data"]
                 )


matthewfeickert:
@WolfgangWaltenberger I'm going to close this as this is a different problem with changes in recent versions of PyTorch that needs to be solved differently as part of Issue https://github.com/scikit-hep/pyhf/issues/2554. This should get addressed in a final patch release for the current series with parts of PR https://github.com/scikit-hep/pyhf/pull/2566. The simplest approach for now is to downgrade your versions of PyTorch. Feel free to comment on Issue https://github.com/scikit-hep/pyhf/issues/2554 with questions or follow up.

We're also going to be deprecating the PyTorch backend in the near future, so if you're actively relying on it please share your needs over in Discussion https://github.com/scikit-hep/pyhf/discussions/2470.

(Aside: While PRs are appreciated, we need them to not be from your default branch on your fork as we can't rebase them without closing the PR, so please always use a feature branch for PRs.)

PR #2608, opened 2025-09-30
chore: Clean up original code to get typehints passing

kratsg:
# Pull Request Description

This is pulled from #2566 to clean up the code to get the typehints passing.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Clean up source code to get mypy happy
```

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -33,7 +33,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v4.31.0
+      uses: github/codeql-action/init@v4.31.2
       with:
         languages: python
         # If you wish to specify custom queries, you can do so here or in a config file.
@@ -42,4 +42,4 @@ jobs:
         # queries: ./path/to/local/query, your-org/your-repo/queries@main
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v4.31.0
+      uses: github/codeql-action/analyze@v4.31.2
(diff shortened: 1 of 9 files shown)


src/pyhf/cli/infer.py
```diff
@@ -9,7 +9,8 @@
 from pyhf.infer import hypotest
 from pyhf.infer import mle
 from pyhf.workspace import Workspace
-from pyhf import get_backend, set_backend, optimize
+from pyhf.tensor.manager import get_backend, set_backend
```
matthewfeickert, line 12: For posterity can you remind me why we need to get things from `pyhf.tensor.manager`? That would be useful to note in the PR commit text.

kratsg: `mypy` didn't like it. I can undo the changes, but you won't see a problem here.. because we're masked by all the other bugs.

src/pyhf/readxml.py
```diff
@@ -5,7 +5,6 @@
     IO,
     Callable,
     Iterable,
-    List,
     MutableMapping,
```
matthewfeickert, line 8: We can do this because in Python 3.9+ (where we're going) typing has improved/been migrated in from `typing`?

kratsg: no, because python 3.8 has `from __future__ import annotations` which lets us do that, but ruff wasn't warning about this yet because it didn't look at the combination of a missing future annotations and the fact that the target was 3.8 - so it assumed it was 3.8, rather than targeting `< 3.9`.

src/pyhf/cli/infer.py
matthewfeickert: > mypy didn't like it. 

That's fine. Can you copy paste the `mypy` errors for future reference (no need to explain them)?

kratsg: I don't have access to the actions logs from that far back. it was essentially an import error for mypy.

PR #2634, opened 2025-11-10
chore: pytest log_level is better than log_cli_level

henryiii:
Committed via https://github.com/asottile/all-repos

See https://github.com/scientific-python/cookie/issues/679

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -171,7 +171,7 @@ addopts = [
     "--doctest-modules",
     "--doctest-glob='*.rst'",
 ]
-log_cli_level = "INFO"
+log_level = "INFO"
 testpaths = "tests"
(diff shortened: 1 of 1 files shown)


matthewfeickert (approved):
Thanks @henryiii. Do you have any references that we can both inform ourselves on and then also note here?

henryiii:
Yes, I paste in the reference link a few minutes after making the PRs because all-repos doesn't have a way to edit the pull request descriptions. So I open all the PRs and paste links.

Note that there's not a practical difference, since `log_cli_level` doesn't trigger live logging if it's in a config file, but this is more conceptually correct (and maps to the CLI better, as `--log-cli-level` does trigger live logging).

PR #2638, opened 2025-11-20
feat: Add Pixi workspace to pyproject.toml

matthewfeickert:
# Description

* Add Pixi workspace to `pyproject.toml` under `[tool.pixi]` table.
   - Add environments for all extras and dependency-groups.
   - Add Pixi tasks for all sessions in Nox.
   - Create conda specific environments so the only Python package is pyhf.
* Add Pixi lock file to .gitignore.
* Add .pixi/config to allow for insecure run-post-link-scripts.
* Add sphinx-autobuild as 'dev' dependency group requirement.
   - 'dev' is chosen over 'docs' as 'docs' should include only the required
      dependencies to build and deploy the docs to the website.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add Pixi workspace to pyproject.toml under [tool.pixi] table.
   - Add environments for all extras and dependency-groups.
   - Add Pixi tasks for all sessions in Nox.
   - Create conda specific environments so the only Python package is pyhf.
* Add Pixi lock file to .gitignore.
* Add .pixi/config to allow for insecure run-post-link-scripts.
* Add sphinx-autobuild as 'dev' dependency group requirement.
   - 'dev' is chosen over 'docs' as 'docs' should include only the required
      dependencies to build and deploy the docs to the website.
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -117,6 +117,7 @@ dev = [
     "tbump>=6.7.0",
     "pre-commit",
     "nox",
+    "sphinx-autobuild>=2025.8.25",
     { include-group = "test" },
     { include-group = "docs" },
 ]
@@ -390,10 +391,18 @@ cmd = "python -m http.server 8001 -d ./docs/_build/html"
 tbump = ">=6.7.0"
 pre-commit = ">=4.5.1"
 nox = ">=2025.11.12"
+# docs environment should contain only required dependencies for deploying docs
+sphinx-autobuild = ">=2025.8.25"
 
 [tool.pixi.feature.dev.tasks.lint]
 description = "Run pre-commit"
 cmd = "pre-commit run --all-files"
+
+[tool.pixi.feature.dev.tasks.watch-docs]
+description = "Build the pyhf documentation and auto-reload on changes"
+cmd = """
+sphinx-autobuild --builder html --nitpicky --show-traceback --open-browser ./docs/ ./docs/_build/html
+"""
 
 [tool.pixi.feature.dev.tasks.build-package]
 description = "Build a sdist and wheel"


pyproject.toml
```diff
@@ -280,3 +280,141 @@ flake8-tidy-imports.ban-relative-imports = "all"
...
+[tool.pixi.feature.dev.tasks.lint]
+description = "Run pre-commit"
+cmd = "pre-commit run --all-files"
+
+[tool.pixi.feature.dev.tasks.build-package]
+description = "Build a sdist and wheel"
+cmd = """
+rm -rf ./dist && \
+pixi exec python-build --installer uv . && \
+pixi exec twine check --strict ./dist/*
```
matthewfeickert, line 412: Doing `pixi exec` here are these are transitory processes that don't need to have their dependencies constrain the rest of the dev environment.

pyproject.toml
```diff
@@ -280,3 +280,141 @@ flake8-tidy-imports.ban-relative-imports = "all"
...
+"""
+
+[tool.pixi.environments]
+default = { solve-group = "default" }
+jax = { features = ["jax"], solve-group = "default" }
+jax-cpu = { features = ["jax-cpu"], solve-group = "default" }
+jax-gpu = { features = ["jax-gpu"], solve-group = "gpu" }
```
matthewfeickert, line 419: Add `jax-gpu` environment so can get a CUDA enabled `jaxlib` on demand.

matthewfeickert (commented):
This adds a fully conda-package-based Pixi workspace for `pyhf`, such that the only Python package is `pyhf` itself:

```console
$ pixi list -e dev | grep pypi
Environment: dev
pyhf                              0.7.1.dev336                                     pypi
```

It adds a suite of Pixi tasks as well

```console
$ pixi task list
Tasks that can run on this machine:
-----------------------------------
benchmarks, build-docs, build-package, doctest, lint, notebooks, serve, test, test-contrib
Task           Description
benchmarks     Run the benchmarks
build-docs     Build the pyhf documentation website
build-package  Build a sdist and wheel
doctest        Run the doctest tests
lint           Run pre-commit
notebooks      Run the test suite for the notebooks
serve          Serve the docs website
test           Run the test suite on the core library
test-contrib   Run the test suite for the contrib module
```

and the ability to get a CUDA accelerated `jaxlib` environment

```console
$ pixi list -e jax-gpu jax
Environment: jax-gpu
Name    Version  Build                            Size  Kind   Source
jax     0.7.2    pyhd8ed1ab_0                 1.78 MiB  conda  https://conda.anaconda.org/conda-forge
jaxlib  0.7.2    cuda129_py313h33cbcf4_202  166.01 MiB  conda  https://conda.anaconda.org/conda-forge
$ pixi run -e jax-gpu python
Python 3.13.11 | packaged by conda-forge | (main, Dec  6 2025, 11:24:03) [GCC 14.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import jax
>>> import pyhf
>>> jax.devices()
[CudaDevice(id=0)]
```

When Pixi is solving it will raise the following warnings

```
WARN The package `pytest-benchmark==5.2.3` does not have an extra named `histogram`
WARN The package `coverage==7.13.0` does not have an extra named `toml`
WARN The package `pytest-benchmark==5.2.3` does not have an extra named `histogram`
WARN The package `coverage==7.13.0` does not have an extra named `toml`
```

which can't be avoided as far as I know, given that it is trying to resolve the extras that exist in the `test` dependency-group

https://github.com/scikit-hep/pyhf/blob/86c214d0ce6be996e387e92ea1c221f612a3c507/pyproject.toml#L85

https://github.com/scikit-hep/pyhf/blob/86c214d0ce6be996e387e92ea1c221f612a3c507/pyproject.toml#L88

and can't.

matthewfeickert:
Note that as you can't specify default environments for tasks yet (c.f. https://github.com/prefix-dev/pixi/issues/5140) when you run

```console
pixi run test
```

you'll be asked to select an environment

```console
$ pixi run test
? The task 'test' can be run in multiple environments.

Please select an environment to run the task in: ›
❯ test
  dev

```

Currently the only way to avoid this is to specify the environment at runtime

```
pixi run --environment test test
```
.

pyproject.toml
```diff
@@ -278,3 +278,142 @@ flake8-tidy-imports.ban-relative-imports = "all"
...
+
+[tool.pixi.feature.docs.tasks.build-docs]
+description = "Build the pyhf documentation website"
+cmd = """
+sphinx-build -M html ./docs ./docs/_build -W --keep-going && \
+rsync -r ./docs/_build/html/_static ./docs/_build/html/docs && \
+rsync -r ./src/pyhf/schemas ./docs/_build/html/
+"""
+
+[tool.pixi.feature.docs.tasks.serve]
```
kratsg, line 385: I would also add a `[...tasks.watch]` that does `sphinx-autobuild --open-browser -n -T -b html docs docs/_build/html` which autoreloads when changes are made.

pyproject.toml
kratsg: Why not use `package` functionalities here with the `pixi-build` preview?

pyproject.toml
```diff
@@ -278,3 +278,142 @@ flake8-tidy-imports.ban-relative-imports = "all"
...
+[tool.pixi.feature.dev.tasks.lint]
+description = "Run pre-commit"
+cmd = "pre-commit run --all-files"
+
+[tool.pixi.feature.dev.tasks.build-package]
+description = "Build a sdist and wheel"
+cmd = """
+rm -rf ./dist && \
+pixi exec python-build --installer uv . && \
+pixi exec twine check --strict ./dist/*
```
kratsg, line 412: You're missing a `twine` dependency.

matthewfeickert: c.f. https://github.com/scikit-hep/pyhf/pull/2638#discussion_r2633027223

For a transitory dependency I don't see the point of adding them here.

pyproject.toml
matthewfeickert: Is that faster or more convenient? That seems like more configuration for no clear benefit.

kratsg: I've found it to be a little faster in `pyhs3`.

pyproject.toml
matthewfeickert: Done, but in `dev` environment.

pyproject.toml
matthewfeickert: Given that for this PR I wanted to get things working and in a single file, and https://github.com/scipp-atlas/pyhs3/pull/145 has shown us that _for now_ we would need to also use `pixi.toml` I'd prefer to skip this till `pixi-build` is fully out of preview.

matthewfeickert:
> Technically, you can do this by simply keeping the task to a single environment. There's some clever restructuring one can do if you don't need all envs all at once.

@kratsg indeed. We probably can try to do extra clever things moving forward, but I might suggest that we postpone until we can reevaluate the `pixi-build` backends.

Issue #2622, opened 2025-10-15
Find alternative to click-completion which is now in maintenance mode


matthewfeickert:
As https://github.com/click-contrib/click-completion 

> is in maintenance mode only. No new features will be added.

an alternative should be found so that it can be migrated away from. At the moment it is only used in the

https://github.com/scikit-hep/pyhf/blob/35acce8917513d316c8fee83f32239af7d63257c/pyproject.toml#L69

extra, so there is a low impact on pyhf as a project overall.

matthewfeickert:
Also noticed in https://github.com/scikit-hep/pyhf/pull/2638#issuecomment-3556488160:

> there is a [native `Click` `v8` completion](https://github.com/click-contrib/click-completion/blob/39e8b5453ee2a816b9fd86c563eeecc59060f7af/README.md#L16-L19). c.f. https://click.palletsprojects.com/en/stable/shell-completion/#shell-completion

matthewfeickert:
@kratsg I find https://click.palletsprojects.com/en/stable/shell-completion slightly confusing about what specifically would need to happen for `pyhf` to use this. Do you understand and can you give a simple example?

PR #2639, opened 2025-11-21
fix: Use Click v8.x native shell completion

matthewfeickert:
# Description

Resolves https://github.com/scikit-hep/pyhf/issues/2622

* Use click v8.x's native shell completions to produce pyhf shell completions for the CLI API.
   - c.f. https://click.palletsprojects.com/en/stable/shell-completion/
* Remove shellcomplete extra.
* Update tests for CLI API.
* Update documentation for producing shell completions for Bash, Zsh, and Fish shells.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Use click v8.x's native shell completions to produce pyhf shell completions for the CLI API.
   - c.f. https://click.palletsprojects.com/en/stable/shell-completion/
* Remove shellcomplete extra.
* Update tests for CLI API.
* Update documentation for producing shell completions.
```

### DIFF
--- a/src/pyhf/cli/complete.py
+++ b/src/pyhf/cli/complete.py
@@ -3,34 +3,18 @@
 import click
 
 
-@click.command(help="Generate shell completion code.", name="completions")
+@click.command(name="completions")
 @click.argument(
     "shell",
     required=False,
     type=click.Choice(["bash", "zsh", "fish"], case_sensitive=False),
 )
 def cli(shell):
-    """Generate shell completion code for various shells.
+    """
+    Generate shell completion code for various shells.
 
-    Supported shells: bash, zsh, fish
-
-    To enable completion, run the appropriate command for your shell:
-
-    \b
-    Bash:
-      mkdir -p ~/.completions
-      _PYHF_COMPLETE=bash_source pyhf > ~/.completions/pyhf-complete.sh
-      echo -e "\n. ~/.completions/pyhf-complete.sh" >> ~/.bashrc
-
-    \b
-    Zsh:
-      mkdir -p ~/.completions
-      _PYHF_COMPLETE=zsh_source pyhf > ~/.completions/pyhf-complete.zsh
-      echo -e "\n. ~/.completions/pyhf-complete.zsh" >> ~/.zshrc
-
-    \b
-    Fish:
-      _PYHF_COMPLETE=fish_source pyhf >> ~/.config/fish/completions/pyhf.fish
+    The necessary commands to enable completions for pyhf for the specified
+    shell will be printed to stdout.
     """
     if shell is None:
         click.echo(cli.get_help(click.Context(cli)))
@@ -61,5 +45,6 @@ def cli(shell):
     )
     click.echo(
         "and then source your shell configuration or restart your shell."
-        + "\nIt is recommended to press tab twice (<TAB><TAB>) to engage the completion during use."
+        + "\nPressing tab twice (<TAB><TAB>) will show all available subcommands."
+        + "\nOptions are only listed if at least a dash has been entered (-<TAB><TAB>)."
     )


matthewfeickert:
> If you generate it ahead of time, then it won't be as slow, but it's still going to be slow.

This is actually how I was testing it, I just wrote the docs to be more trivial. Though I guess we should give people the best shot at having things not be horribly slow, so I should revise this.

matthewfeickert:
From @henryiii: 

> Sounds like you need 3.15's lazy imports. :slightly_smiling_face: (Maybe https://github.com/scientific-python/lazy-loader would help in the interim)?

src/pyhf/cli/complete.py
```diff
@@ -1,30 +1,65 @@
...
+    }
+    click.echo(
+        click.style(
+            instructions[shell],
+            bold=True,
         )
+    )
+    click.echo(
+        "and then source your shell configuration or restart your shell."
+        + "\nIt is recommended to press tab twice (<TAB><TAB>) to engage the completion during use."
```
matthewfeickert: I've learned that while still a bit slow, some of my complaints were human user error.

e.g. I should be been doing double tab (`<TAB><TAB>`) all along.

```console
$ pyhf <TAB><TAB>
cls          combine      completions  contrib      digest       fit          inspect      json2xml     patchset     prune        rename       sort         xml2json
$ pyhf fit -<TAB><TAB>
--output-file  --measurement  -p             --patch        --value        --backend      --optimizer    --optconf      --help         -h 
```

src/pyhf/cli/complete.py
```diff
@@ -1,30 +1,62 @@
...
+
+    \b
+    Zsh:
+      mkdir -p ~/.completions
+      _PYHF_COMPLETE=zsh_source pyhf > ~/.completions/pyhf-complete.zsh
+      echo -e "\n. ~/.completions/pyhf-complete.zsh" >> ~/.zshrc
+
+    \b
+    Fish:
+      _PYHF_COMPLETE=fish_source pyhf >> ~/.config/fish/completions/pyhf.fish
```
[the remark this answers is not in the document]
kratsg: are we skipping this?

src/pyhf/cli/complete.py
kratsg: do we care if double-tabbing is slow? it can be slow for some applications, but as long as the normal single-tab to complete an argument like `--outpu<TAB>` is fast, then that's ok?

src/pyhf/cli/complete.py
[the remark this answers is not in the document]
matthewfeickert: @kratsg skipping the docstring? I'm not intending to skip anything, though I noticed that this isn't showing up now in https://pyhf--2639.org.readthedocs.build/en/2639/cli.html#pyhf-completions. Do you know why? If it isn't going to be visible to humans we can probably get rid of it.

src/pyhf/cli/complete.py
matthewfeickert: @kratsg Probably not, but `pyhf fi<TAB>` takes about a second to complete to `pyhf fit` for me. So none of these are super fast.

src/pyhf/cli/complete.py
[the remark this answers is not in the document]
kratsg: isn't it not visible because of the `\b` ? The commit suggestion from copilot hasn't been committed.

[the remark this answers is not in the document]
matthewfeickert: > isn't it not visible because of the `\b`?

idk, but I'll check.

> The commit suggestion from copilot hasn't been committed.

Yeah, because it is a bad suggestion, IMO.

src/pyhf/cli/complete.py
```diff
@@ -1,30 +1,50 @@
-'''Shell completions for pyhf.'''
+"""Shell completions for pyhf."""
 
 import click
 
-try:
-    import click_completion
 
-    click_completion.init()
+@click.command(name="completions")
```
matthewfeickert, line 6: I removed the `help='Generate shell completion code.'` as this was taking precedence over anything written in the docstring.

src/pyhf/cli/complete.py
```diff
@@ -1,30 +1,50 @@
...
-    @click.argument(
-        'shell',
-        required=False,
-        type=click_completion.DocumentedChoice(click_completion.core.shells),
-    )
-    def cli(shell):
-        '''Generate shell completion code for various shells.'''
-        click.echo(click_completion.core.get_code(shell, prog_name='pyhf'))
+    The necessary commands to enable completions for pyhf for the specified
+    shell will be printed to stdout.
```
matthewfeickert, line 17: As we want people to get these from the command anyway and not from copying and pasting from the docs, then I removed what the explicit logic is. I think that's fine?

src/pyhf/cli/complete.py
```diff
@@ -1,30 +1,50 @@
...
+    click.echo(
+        click.style(
+            instructions[shell],
+            bold=True,
         )
+    )
+    click.echo(
+        "and then source your shell configuration or restart your shell."
+        + "\nPressing tab twice (<TAB><TAB>) will show all available subcommands."
+        + "\nOptions are only listed if at least a dash has been entered (-<TAB><TAB>)."
```
matthewfeickert, line 49: Click's shell completions seem to heavily emphasize double tab (https://click.palletsprojects.com/en/stable/shell-completion/) so I'm only mentioning `<TAB><TAB>` here instead of single tab.

PR #2656, opened 2025-12-25
fix: skip excluded backends before test backend setup

### LINKED ISSUE #2654: Only load the necesssary backends when testing with `skip_backend` or `only_backend` markers
### Summary

I am currently packaging the `scikit-hep` stack for Guix, and I am facing troubles enabling the tests for the basic set of dependencies.

It looks like despite calling pytest with the `only_numpy` marker, it tries to load the other backends:

```
tests/conftest.py:63: in <module>
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
/gnu/store/jabv9192g9pi22w6cn6zwf57ajk8jm80-python-pyhf-0.7.6/lib/python3.11/site-packages/pyhf/tensor/__init__.py:54: in __getattr__
    raise exceptions.ImportBackendError(
E   pyhf.exceptions.ImportBackendError: ('There was a problem importing PyTorch. The pytorch backend cannot be used.', ModuleNotFoundError("No module named 'torch'"))
```

I did a quick review of the code in `conftest.py`, and indeed the [backend fixture](https://github.com/scikit-hep/pyhf/blob/b500270c6a7744d16b481dfda247c066ef3c2cb4/tests/conftest.py#L74)  calls `pyhf.set_backend(...)` for the whole parameter set, irrespective of whether they are excluded via `skip_$backend` or `only_$backend`.

It would be nice if such behaviour could be changed to only call `pyhf.set_backend("numpy")` if `-m only_numpy` is specified.

Cheers,  

### Additional Information

```python (paste below)

```

### Code of Conduct

- [x] I agree to follow the Code of Conduct

iamkrishpathak:
This PR updates the backend test fixture to respect only_* markers
before creating backend objects.

This avoids importing optional backends (e.g. JAX, PyTorch) when running
marker-restricted test suites such as `pytest -m only_numpy`, improving
test behavior in minimal environments.

Fixes #2654

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -136,6 +136,13 @@ def backend(request):
         request.node.add_marker(
             pytest.mark.xfail(reason=f"expect {func_name} to fail as specified")
         )
+    
+    # If this backend is not selected by only_* markers, do not even attempt setup
+    if only_backends and param_id not in only_backends:
+        pytest.skip(
+            f"skipping {func_name} as specified to only look at: {', '.join(only_backends)}"
+        )
+
 
     tensor_config, optimizer_config = request.param
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -136,13 +136,12 @@ def backend(request):
         request.node.add_marker(
             pytest.mark.xfail(reason=f"expect {func_name} to fail as specified")
         )
-    
+
     # If this backend is not selected by only_* markers, do not even attempt setup
     if only_backends and param_id not in only_backends:
         pytest.skip(
             f"skipping {func_name} as specified to only look at: {', '.join(only_backends)}"
         )
-
 
     tensor_config, optimizer_config = request.param
 


iamkrishpathak:
CI is currently failing during pytest session startup due to a DeprecationWarning
raised from matplotlib via pytest-mpl (pyparsing.oneOf), before any tests
execute.

This appears unrelated to the backend fixture change in this PR.
Happy to help address it if you’d like me to handle it here.

iamkrishpathak:
Got it, thanks! I’ll wait for that PR to land.

kratsg (approved):

matthewfeickert (changes requested):
Thanks for the PR @iamkrishpathak, but I don't think that this does what you expect it to do, and if it does, then I don't think that this is actually a helpful addition as it would skip most of the tests.

Example:

Let's start with a clean env

```
$ rm -rf .venv
$ uv venv
$ . .venv/bin/activate
$ uv pip install -e ".[all]" --group dev
```

if we then run with `only_numpy` we get

```console
$ pytest -m only_numpy
================================================================================= test session starts ==================================================================================
platform linux -- Python 3.13.7, pytest-9.0.2, pluggy-1.6.0
Matplotlib: 3.10.8
Freetype: 2.6.1
benchmark: 5.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/feickert/Code/GitHub/scikit-hep/pyhf
configfile: pyproject.toml
testpaths: tests
plugins: requests-mock-1.12.1, mpl-0.18.0, console-scripts-1.4.1, anyio-4.12.0, socket-0.7.0, benchmark-5.2.3, mock-3.15.1
collected 1160 items / 1151 deselected / 9 selected                                                                                                                                    

tests/test_pdf.py .ss.ss.ss                                                                                                                                                      [100%]

=============================================================================== short test summary info ================================================================================
SKIPPED [1] tests/test_pdf.py:169: skipping test_core_pdf_broadcasting[jax] as specified to only look at: numpy
SKIPPED [1] tests/test_pdf.py:169: skipping test_core_pdf_broadcasting[numpy_minuit] as specified to only look at: numpy
SKIPPED [1] tests/test_pdf.py:355: skipping test_pdf_integration_histosys[jax] as specified to only look at: numpy
SKIPPED [1] tests/test_pdf.py:355: skipping test_pdf_integration_histosys[numpy_minuit] as specified to only look at: numpy
SKIPPED [1] tests/test_pdf.py:501: skipping test_pdf_integration_shapesys[jax] as specified to only look at: numpy
SKIPPED [1] tests/test_pdf.py:501: skipping test_pdf_integration_shapesys[numpy_minuit] as specified to only look at: numpy
==================================================================== 3 passed, 6 skipped, 1151 deselected in 1.88s =====================================================================
```

which is covering **only 1 test file** and skipping the bulk of the tests. I understand that it is selecting only the tests that are able to be explicitly labeled as numpy only, but testing only a very tiny segment of the codebase isn't something we want to encourage.

Maybe there's a difference in behavior between the version of `pytest` that you used and the latest version? My guess is that yes, `pytest` probably broke along the way and we didn't notice it (sorry if so!) and that we actually need to fix the behavior of the `only` selectors.

I understand the desire to avoid having to install `jax` to be able to run tests. Though using a method that _excludes_ instead of _including_ seems preferable.

If we're missing something here (very possible) please let us know.

iamkrishpathak:
Thanks a lot for the detailed explanation — this is very helpful.

You’re absolutely right: running only 9 tests out of ~1160 is not something we should encourage, and I agree that the current behavior of `-m only_numpy` is far too restrictive to be useful in practice.

My original intent was to lower the barrier for contributors who don’t have JAX installed, but I see now that *including* only explicitly-marked tests is the wrong model here. An *exclusion-based* approach (e.g. “run everything except backend X”) makes much more sense and better matches real test coverage expectations.

Given your note about possible pytest/marker behavior drift, I suspect this might indeed be exposing a deeper issue with how the `only_*` selectors behave today.

Before making changes, I’d like to confirm the preferred direction:
- Would you like me to explore fixing the `only_*` marker semantics so they behave as exclusions rather than inclusions?
- Or should we move toward a new pattern like `not requires_jax` / backend-specific skip markers and deprecate `only_*` usage?
- Or would you prefer this PR be narrowed to documentation or test-selection guidance only?

Happy to rework or pivot the PR once I understand the intended direction. Thanks again for taking the time to explain this so clearly.

PR #2678, opened 2026-03-27
feat: add optimistix optimizer

pfackeldey:
# Description

This PR adds `optimistix` as a new optimizer backend. Optimistix implements JAX-native solvers and supports differentiation through minimization via the implicit function theorem, moving pyhf a step closer to being fully differentiable out of the box.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

---

```
* Add new `optimistix_optimizer` class wrapping `optx.minimise` 
  with DFP solver by default (maxiter=1000); fixed parameters are 
  represented as `None` pytree leaves so the minimizer skips
  gradient updates for them; returns a `scipy.optimize.OptimizeResult`
  including `optx_state`
* Add `optimistix` optional dependency in `pyproject.toml` constrained
  to `python_version >= '3.11'`; wire into `backends`, `all`, and `test`
  pixi environments
* Extend `tests/conftest.py` `backend` fixture to conditionally include
  a `jax_optimistix` parametrize entry when `optimistix` is importable, so 
  all existing parametrized tests exercise the new optimizer automatically
* Add targeted tests:`ValueError` on `solver=None`, `ValueError` when
  used with non-JAX backend, basic fit, fit with a fixed parameter, 
  and `return_result_obj` round-trip checking `optx_state` is present

Co-Authored-By: Giordon Stark <kratsg@gmail.com>
```

## Summary by CodeRabbit

* **New Features**
  * Added an Optimistix-based optimizer backend (Python 3.11+), including fixed-parameter handling and richer result metadata.
  * Added `empty` and `arange` array-creation utilities to tensor backends.
* **Bug Fixes**
  * Improved optional backend import errors when Optimistix is unavailable.
* **Documentation**
  * Expanded Sphinx cross-references to include Optimistix docs.
* **Tests**
  * Added/updated coverage and a new test marker related to the JAX+Optimistix combination.
* **Chores**
  * Updated optional dependency groups and tool configuration; improved local artifact ignore rules.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -361,7 +361,7 @@ requests = ">=2.22.0"
 iminuit = ">=2.7.0"
 
 [tool.pixi.feature.optimistix.dependencies]
-python = ">=3.11.*"
+python = ">=3.11"
 optimistix = ">=0.1.0"
 
 [tool.pixi.feature.test.dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,11 +70,12 @@ jax = [
 ]
 xmlio = ["uproot>=4.1.1"]  # c.f. PR #1567
 minuit = ["iminuit>=2.7.0"]  # c.f. PR #1895
+optimistix = ["optimistix>=0.1.0"]
 contrib = [
     "matplotlib>=3.0.0",
     "requests>=2.22.0",
 ]
-backends = ["pyhf[jax,minuit]"]
+backends = ["pyhf[jax,minuit,optimistix]"]
 all = ["pyhf[backends,xmlio,contrib]"]
 
 # Developer extras
@@ -359,6 +360,9 @@ requests = ">=2.22.0"
 [tool.pixi.feature.minuit.dependencies]
 iminuit = ">=2.7.0"
 
+[tool.pixi.feature.optimistix.dependencies]
+optimistix = ">=0.1.0"
+
 [tool.pixi.feature.test.dependencies]
 pytest = ">=6.0"
 scikit-hep-testdata = ">=0.4.11"
@@ -464,10 +468,11 @@ jax-cpu = { features = ["jax-cpu"], solve-group = "default" }
 jax-gpu = { features = ["jax-gpu"], solve-group = "gpu" }
 xmlio = { features = ["xmlio"], solve-group = "default" }
 minuit = { features = ["minuit"], solve-group = "default" }
+optimistix = { features = ["optimistix"], solve-group = "default" }
 contrib = { features = ["contrib"], solve-group = "default" }
-backends = { features = ["jax", "minuit"], solve-group = "default" }
-all = { features = ["jax", "minuit", "xmlio", "contrib"], solve-group = "default" }
+backends = { features = ["jax", "minuit", "optimistix"], solve-group = "default" }
+all = { features = ["jax", "minuit", "optimistix", "xmlio", "contrib"], solve-group = "default" }
 
-test = { features = ["jax", "minuit", "xmlio", "contrib", "test"], solve-group = "default" }
+test = { features = ["jax", "minuit", "optimistix", "xmlio", "contrib", "test"], solve-group = "default" }
 docs = { features = ["xmlio", "contrib", "docs"], solve-group = "default" }
 dev = { features = ["all", "test", "docs", "dev"], solve-group = "default" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -163,6 +163,7 @@ log_level = "INFO"
 testpaths = "tests"
 markers = [
     "fail_jax",
+    "fail_jax_optimistix",
     "fail_numpy",
     "fail_numpy_minuit",
     "only_jax",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,7 @@ jax = [
 ]
 xmlio = ["uproot>=4.1.1"]  # c.f. PR #1567
 minuit = ["iminuit>=2.7.0"]  # c.f. PR #1895
-optimistix = ["optimistix>=0.1.0"]
+optimistix = ["optimistix>=0.1.0;python_version>='3.11'"]
 contrib = [
     "matplotlib>=3.0.0",
     "requests>=2.22.0",
@@ -361,7 +361,7 @@ requests = ">=2.22.0"
 iminuit = ">=2.7.0"
 
 [tool.pixi.feature.optimistix.dependencies]
-optimistix = ">=0.1.0"
+optimistix = ">=0.1.0;python_version>='3.11'"
 
 [tool.pixi.feature.test.dependencies]
 pytest = ">=6.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -361,7 +361,8 @@ requests = ">=2.22.0"
 iminuit = ">=2.7.0"
 
 [tool.pixi.feature.optimistix.dependencies]
-optimistix = ">=0.1.0;python_version>='3.11'"
+python = ">=3.11.*"
+optimistix = ">=0.1.0"
 
 [tool.pixi.feature.test.dependencies]
 pytest = ">=6.0"
--- a/src/pyhf/optimize/opt_optimistix.py
+++ b/src/pyhf/optimize/opt_optimistix.py
@@ -1,0 +1,138 @@
+"""Optimistix Optimizer Class."""
+
+import optimistix as optx
+import scipy
+
+import pyhf
+from pyhf.optimize.mixins import OptimizerMixin
+
+
+class optimistix_optimizer(OptimizerMixin):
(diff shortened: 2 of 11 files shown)


src/pyhf/optimize/opt_optimistix.py
```diff
@@ -0,0 +1,134 @@
...
+
+        fixed_vals = fixed_vals or []
+        # for optimistix we re-express the func such that the x0 only contains
+        # floating point values for params (x0) that the optimizer should minimise.
+        # Fixed parameters are set to None (static pytree leaf-type) and thus won't receive
+        # a gradient update by the minimizer. We patch/set back their fixed values
+        # in the `wrapped_func` before forwarding them to the original `func`.
+        x0_floating = list(x0)
+        for idx, _ in fixed_vals:
+            x0_floating[idx] = None
```
lukasheinrich: is this a perf issue to use a python loop over e.g. masks? maybe not really .. just curious

src/pyhf/optimize/opt_optimistix.py
```diff
@@ -0,0 +1,134 @@
...
+            throw=throw,
+            has_aux=True,
+            options=options,
+        )
+
+        # patch fixed values back in:
+        x = list(res.value)
+        for idx, val in fixed_vals:
+            x[idx] = val
+
```
lukasheinrich, line 137: dito as above

src/pyhf/optimize/opt_optimistix.py
pfackeldey: For _very_ extreme cases where you'd have 10000 (fixed!) parameters this adds an overhead in the micro seconds regime. 
```python
In [1]: l = [[]] * 10000

In [2]: %timeit for _ in range(len(l)): l[_] = _
196 μs ± 130 ns per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
I'd argue that typically the number of fixed parameters is rather O(1), e.g. freezing one parameter for e.g. MINOS, so that's completely irrelevant (nano seconds regime). One could change this maybe to just `jax.tree` semantics, but calling into c++ is probably more overhead than using plain cpython for not too many fixed parameters (not too many ~= more than thousands).

Once this is code path becomes jit traceable, this code path will also only be executed once during tracing and then never again at runtime (because it involves no jax ops that are recorded in the JaxPr). 

So, bottom line: it is negligible, unless you have >>10k fixed params (will this ever happen?)

pyproject.toml
```diff
@@ -322,6 +323,9 @@ requests = ">=2.22.0"
 [tool.pixi.feature.minuit.dependencies]
 iminuit = ">=2.7.0"
 
+[tool.pixi.feature.optimistix.dependencies]
+optimistix = ">=0.1.0;python_version>='3.11'"
```
matthewfeickert: Is this (the `python_version` bit) valid Pixi syntax? 

I like the PR btw. 👍

pfackeldey: uhh, I was assuming so? :D do you know by chance how this constraint is expressed in 'pixi language'?

kratsg: @pfackeldey you can only put markers on the dependencies. if you want an optimistix feature, then you require something like

```
[tool.pixi.feature.optimistix.dependencies]
python = ">=3.11.*"
optimistix = ">=0.1.0"
```

src/pyhf/optimize/opt_optimistix.py
```diff
@@ -0,0 +1,134 @@
...
+import pyhf
+from pyhf.optimize.mixins import OptimizerMixin
+import scipy
+import optimistix as optx
+
+
+class optimistix_optimizer(OptimizerMixin):
+    """
+    Optimizer that uses :func:`scipy.optimize.minimize`.
+    """
```
kratsg, line 13: This was copied from the scipy optimizer — needs updating to reference optimistix.

tests/conftest.py
```diff
@@ -83,8 +83,9 @@ def reset_backend():
             ("numpy_backend", dict(poisson_from_normal=True)),
             ("minuit_optimizer", dict()),
         ),
+        (("jax_backend", dict()), ("optimistix_optimizer", dict())),
     ],
-    ids=['numpy', 'jax', 'numpy_minuit'],
+    ids=['numpy', 'jax', 'numpy_minuit', 'jax_optimistix'],
```
kratsg: The cleanest approach is to conditionally add the backend to the fixture params in `conftest.py` based on whether optimistix is importable:

```python
_backend_params = [
   (("numpy_backend", dict()), ("scipy_optimizer", dict())),
   (("jax_backend", dict()), ("scipy_optimizer", dict())),
   (("numpy_backend", dict(poisson_from_normal=True)), ("minuit_optimizer", dict())),
]
_backend_ids = ["numpy", "jax", "numpy_minuit"]

try:
   import optimistix  # noqa: F401
   _backend_params.append((("jax_backend", dict()), ("optimistix_optimizer", dict())))
   _backend_ids.append("jax_optimistix")
except ImportError:
   pass
```

Then use `params=_backend_params, ids=_backend_ids` in the `@pytest.fixture`. This way the backend simply doesn't appear in the test matrix when optimistix isn't installed, no explicit skip markers needed.

kratsg (changes requested):

tests/conftest.py
pfackeldey: I added your suggestion, I was thinking about the right place for a `pytest.importorskip`, but that makes it easier :+1:

src/pyhf/optimize/opt_optimistix.py
kratsg: this requires updating the docs configuration to pull in the sphinx file for optimistix (grok for intersphinx)

src/pyhf/optimize/opt_optimistix.py
alexander-held: Fixed parameters are the way to do MC stats uncertainty pruning at the moment, but even for that we should generally be at less than O(1k) I'd say. It's not unthinkable to get that above 10k but optimizing for that seems premature.

kratsg:
@pfackeldey it looks like we're seeing significantly longer run times for some of the tests with optimistix. If you look at the `public_api` tests, the hypotest we run with jax_optimistix takes significantly longer -- any ideas why?

pfackeldey:
@kratsg thanks for pushing this PR forward!! I didn't have the capacity last week to do so.

I need to check, maybe a recompilation+retracing issue? I'll investigate...

src/pyhf/optimize/opt_optimistix.py
```diff
@@ -0,0 +1,154 @@
...
+        self.solver = kwargs.pop("solver", optx.DFP(rtol=1e-6, atol=1e-6))
+        if self.solver is None:
+            msg = f"{type(self).__name__} needs a solver"
+            raise ValueError(msg)
+        self.options = kwargs.pop("options", None)
+        self.throw = kwargs.pop("throw", True)
+
+        # the default for maxiter is too high for optimistix, so we'll lower it here:
+        self.maxiter = kwargs.pop("maxiter", 1000)
+        super().__init__(*args, **kwargs)
```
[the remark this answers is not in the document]
pfackeldey: Do you (@kratsg & @matthewfeickert) want me to catch minimization errors and raise from it a `pyhf.exceptions.FailedMinimization`?

src/pyhf/optimize/opt_optimistix.py
```diff
@@ -0,0 +1,154 @@
...
+        return optx.minimise
+
+    def _minimize(
+        self,
+        minimizer,
+        func,
+        x0,
+        do_grad=False,  # noqa: ARG002
+        bounds=None,  # noqa: ARG002
+        fixed_vals=None,
```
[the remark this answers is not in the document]
pfackeldey: optimistix doesn't support bounds (yet). It is currently in the process of implemented bounded minimizers. In a separate PR one could implement the bound transforms that Minuit does (they're agnostic to the solver in principle). But then we're starting to re-implement a lot of functionality from https://github.com/MoAly98/everwillow/.

src/pyhf/optimize/opt_optimistix.py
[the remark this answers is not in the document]
kratsg: We do since we'd want to be able to raise the exception through the mixin properly and attach a minimization state. This allows users to catch the exception on their end and get the state of the minimization for debugging. But I think it would already have been done for you: https://github.com/scikit-hep/pyhf/blob/24b10139f1aeca4ec465c4bf07f25d30b2b62b2b/src/pyhf/optimize/mixins.py#L34-L69

matthewfeickert:
The failures are due to the new `matplotlib` release (`v3.11.0`). We'll fix this and then rebase this.

pfackeldey:
Once it's rebased the PR is ready from my side @matthewfeickert @kratsg :)

Issue #1413, opened 2021-04-23
pyhf.optimize not introspectable with inspect.getsource et al.


lhenkelm:
# Description

Trying to use inspect (from the stdlib) to inspect the source of pyhf.optimize raises an AttributeError.

# Expected Behavior

The inspect functions should work and return the source code.

# Steps to Reproduce
With pyhf 0.6.1 and python 3.7.6:
```python
import pyhf
import inspect
# this works when you replace pyhf.optimize with sys.modules[type(pyhf.optimize).__module__]
for l in inspect.getsourcelines(pyhf.optimize)[0]: print(l) 
```

# Checklist

- [ ] Run `git fetch` to get the most up to date version of `master`
- [x] Searched through existing Issues to confirm this is not a duplicate issue
- [x] Filled out the Description, Expected Behavior, Actual Behavior, and Steps to Reproduce sections above or have edited/removed them in a way that fully describes the issue

NB: I am not sure how this relates to #735 and #1402 . Would using `def __getattr__(): <...>` fix this?

matthewfeickert:
@lhenkelm Can you give a lot more context? Why do you need to do this?

lhenkelm:
"need" is a  strong word :-)  . Here is the story: 
I was working interactively, and trying to figure out the knobs I could play with for optimizers.
So I do `help(pyhf.optimize)`:
```
Help on _OptimizerRetriever in module pyhf.optimize object:

class _OptimizerRetriever(builtins.object)
 |  Methods defined here:
 |  
 |  __getattr__(self, name)
 |  
 |  ----------------------------------------------------------------------
 |  Data descriptors defined here:
 |  
 |  __dict__
 |      dictionary for instance variables (if defined)
 |  
 |  __weakref__
 |      list of weak references to the object (if defined)
```
Which tells me nothing. But never fear, introspection is here: 
```python
import inspect
for l in inspect.getsourcelines(pyhf.optimize)[0]: print(l)
```
yielding
```
TypeError                                 Traceback (most recent call last)
<lots of traceback>
TypeError: module, class, method, function, traceback, frame, or code object was expected, got NoneType
```
Aha. Now putting on the hacker hat, I come up with
```python
import sys
actual_module = sys.modules[type(pyhf.optimize).__module__]
for l in inspect.getsourcelines(actual_module)[0]: print(l)
```
And finally get the source [not shown for clarity], but as you know,
[...]

PR #1403, opened 2021-04-12
chore: Use module level __dir__ to restrict public API views

matthewfeickert:
# Description

Resolves #1402 

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Use module level __dir__ in combination with __all__ to restrict the public API import views to useful imports
* c.f. https://snarky.ca/lazy-importing-in-python-3-7/
```

PR #2688, opened 2026-04-03
style: Add extended Ruff rules from Scientific Python repo review

kratsg:
# Pull Request Description

This PR adds the additional extended rules for `ruff` from scientific-python/cookie: https://github.com/scientific-python/cookie/blob/10e253aaaa7f22418b95d0e446cea216acd4d5a7/%7B%7Bcookiecutter.project_name%7D%7D/pyproject.toml#L284-L329 .

The notebooks and validation code need updating / additional love.

- [ ] Tests are passing

# Before Merging

For the PR Assignees:

```
* Add additional Ruff rules recommended by the scientific-python/cookie
  / Scientific Python repo review.
   - Add ARG, BLE, C4, DTZ, EM, EXE, FA, FLY, FURB, G, ICN, ISC, LOG,
     NPY, PD, PERF, PGH, PIE, PL, PT, PTH, PYI, Q, RET, RSE, SIM, SLOT,
     T10, T20, TC, TRY, YTT.
   - Add per-file T20 ignores for notebooks, tests, noxfile, and validation
     scripts where print is legitimate. PLC0415, TC001, TC003 are ignored
     pending further review. PLR09, PLR2004 ignored per convention.
   - Noteable recurring fixes:
      * logging-f-string (G004)
      * raw-string-in-exception (EM101)
      * error-instead-of-exception (TRY400)
   - Amends PR https://github.com/scikit-hep/pyhf/pull/2686
```


src/pyhf/cli/infer.py
```diff
@@ -124,9 +125,9 @@ def fit(
     if output_file is None:
         click.echo(json.dumps(result, indent=4, sort_keys=True))
     else:
-        with open(output_file, "w+", encoding="utf-8") as out_file:
+        with Path(output_file).open("w+", encoding="utf-8") as out_file:
             json.dump(result, out_file, indent=4, sort_keys=True)
-        log.debug(f"Written to {output_file:s}")
+        log.debug("Written to %s", output_file)
```
matthewfeickert, line 130:
```suggestion
        log.debug(f"Written to {output_file:s}")
```

Why is this not using fstrings? This should be fixed.

src/pyhf/cli/infer.py
```diff
@@ -239,6 +240,6 @@ def cls(
...
```
matthewfeickert, line 245: Revert again.

matthewfeickert (changes requested):
I'm going to review this later given that it is massive and is going to have a lot of repeated things, but Claude Code decided to abandon f-string use for some reason which I do not want to do unless there is a very strong reason for it.

I marked two revision areas but it happens a lot more.

---

To be clear, I **love** this PR in general though. :D

src/pyhf/cli/infer.py
kratsg: That's not claude code -- that's the style recommendation. This defers evaluation of strings inside log statements until log time.  https://docs.astral.sh/ruff/rules/logging-f-string/

src/pyhf/cli/infer.py
kratsg: see https://docs.astral.sh/ruff/rules/logging-f-string/

src/pyhf/cli/infer.py
matthewfeickert: In retrospect that should have been obvious as this is a formatter PR. :confounded:

src/pyhf/contrib/viz/brazil.py
```diff
@@ -319,10 +325,11 @@ def plot_results(test_pois, tests, test_size=0.05, ax=None, **kwargs):
     no_cls = kwargs.pop("no_cls", False)
 
     if plot_components and len(tests[0]) != 3:
-        raise ValueError(
+        msg = (
             f"The components of 'tests' should have len of 3 to visualize the CLs components but have len {len(tests[0])}."
-            + "\n'tests' should have format of: [CLs_obs, [CLsb, CLb], [CLs_exp band]]"
+            "\n'tests' should have format of: [CLs_obs, [CLsb, CLb], [CLs_exp band]]"
         )
+        raise ValueError(msg)
```
matthewfeickert, line 332: Do you know what rule causes this? More just curious.

src/pyhf/contrib/cli.py
```diff
@@ -70,10 +70,9 @@ def download(archive_url, output_directory, verbose, force, compress):
 
         if verbose:
             file_list = [str(file) for file in list(Path(output_directory).glob("*"))]
-            print("\n".join(file_list))
+            click.echo("\n".join(file_list))
     except AttributeError:
-        log.error(
+        log.exception(
```
matthewfeickert, line 75: Shouldn't this still be an error? What rule is moving it to exception?

kratsg: https://docs.astral.sh/ruff/rules/error-instead-of-exception/

src/pyhf/contrib/viz/brazil.py
kratsg: https://docs.astral.sh/ruff/rules/raw-string-in-exception/

src/pyhf/contrib/viz/brazil.py
```diff
@@ -17,21 +25,10 @@ def __dir__():
...
-            "one_sigma_band",
-            "two_sigma_band",
-            "test_size",
-            "clsb",
-            "clb",
-            "axes",
-        ),
-    )
-):
+# python 3.10
```
matthewfeickert:
```suggestion
# TODO: Once pyhf is Python 3.10+ use slots
```

To help make these easier to find for when we drop Pythons we've generally added 'TODO's to them.

src/pyhf/__init__.py
```diff
@@ -2,7 +2,7 @@
 # ruff: isort: off
 from pyhf._version import version as __version__
 from pyhf.tensor import BackendRetriever as tensor
-from pyhf.optimize import OptimizerRetriever as optimize  # noqa
+from pyhf.optimize import OptimizerRetriever as optimize
```
matthewfeickert:
```suggestion
# Intentionally keep optimize out of __all__
# c.f. https://github.com/scikit-hep/pyhf/pull/1403
from pyhf.optimize import OptimizerRetriever as optimize  # noqa
```

We intentionally didn't include `optimize` in `__all__` here https://github.com/scikit-hep/pyhf/pull/1403#issuecomment-827893970 given user confusion in https://github.com/scikit-hep/pyhf/issues/1413.

src/pyhf/__init__.py
```diff
@@ -22,6 +22,7 @@
     "infer",
     "interpolators",
     "modifiers",
+    "optimize",
```
matthewfeickert:
```suggestion
```

Same reason as above.

matthewfeickert (changes requested):
Thanks @kratsg. This is all great. The one thing is that we had previously intentionally kept `optimize` out of `__all__` given user confusion and I think we should keep that.

My suggested changes are probably not the right ones though as I'm assuming that Ruff introduced them all.

matthewfeickert (approved):
I pushed my suggested revisions, but please review and revert them if you don't like them. Other than that I'm happy. 👍 

This will be a great PR for moving forward contributions where there's small stuff that Ruff can handle for us now!

PR #2690, opened 2026-04-04
style: Apply Ruff rules to most notebooks

kratsg:
# Pull Request Description

See #2689 for remaining outdated notebooks.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Apply Ruff rules ARG001, F401, F811, F821, F841, NPY002, PERF401,
  PTH123, SIM115 for notebooks.
* Amends PR https://github.com/scikit-hep/pyhf/pull/2688.
```


matthewfeickert (changes requested):
@kratsg These changes delete some of the output images that we had in the notebooks. If we had a system that built all of the notebooks and kept the images for the docs that would be okay, but we don't AFAIK.

pyproject.toml
```diff
@@ -248,6 +248,10 @@ ignore_errors = true
 
 [tool.ruff]
 line-length = 88
+extend-exclude = [
+  "docs/examples/notebooks/histogrammar.ipynb",
+  "docs/examples/notebooks/StatError.ipynb"
+]
```
matthewfeickert, line 254: Why are these getting excluded?

matthewfeickert (changes requested):
In some of the notebooks there are changes which result in code getting commented out but not removed. For example, in `docs/examples/notebooks/learn/TensorizingInterpolations.ipynb` there is now the comment

```python
# allset_all_histo_nom = histogramssets[:, :, 1]
```

If this can be commented out we can remove it, yes? Is there any reason to keep this?

Also, a curiosity/question, but not a problem: I'm very happy to see the use of `np.random.default_rng` now, but is there a linting rule in Ruff that sets all seeds to `42`?

```python
np.random.default_rng(seed=42)
```

That seems like a "not great" lint/formatting rule to have, as people can very reasonably want different seeds (that don't have to be tied to Hitchhiker's Guide to the Galaxy).

Also in `docs/examples/notebooks/XML_ImportExport.ipynb` we have the `import pyhf` there at the top even though we do everything through ipython magics at the command line to verify that people have things installed to be able to run the notebook. Can we keep that with a some sort of `#noqa` ignore?

pyproject.toml
matthewfeickert: Oh, because these are the outdated ones you identified in Issue https://github.com/scikit-hep/pyhf/issues/2689? Can you add a comment here that we should either revise them or remove them?

matthewfeickert:
> I don't understand this comment.

@kratsg sorry. What I mean is the diff of `docs/examples/notebooks/XML_ImportExport.ipynb` shows this

image

or

```diff
$ git diff upstream/main -- docs/examples/notebooks/XML_ImportExport.ipynb
diff --git a/docs/examples/notebooks/XML_ImportExport.ipynb b/docs/examples/notebooks/XML_ImportExport.ipynb
index d80cde21..9b3d942f 100644
--- a/docs/examples/notebooks/XML_ImportExport.ipynb
+++ b/docs/examples/notebooks/XML_ImportExport.ipynb
@@ -13,8 +13,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# NB: python -m pip install pyhf[xmlio]\n",
-    "import pyhf"
+    "# NB: python -m pip install pyhf[xmlio]"
    ]
   },
   {
```

As [`docs/examples/notebooks/XML_ImportExport.ipynb`](https://github.com/scikit-hep/pyhf/blob/30732e8b411a44d67d4d2a6417ae98630f6ebac9/docs/examples/notebooks/XML_ImportExport.ipynb) is a notebook that is all about the CLI AP, and so all of it is executed with ipython magics in the notebooks cells, my memory is that we had the `import pyhf` at the top as a check to make sure that you had `pyhf` installed before you started executing things at the CLI.

Can/should we keep this?

docs/examples/notebooks/learn/TensorizingInterpolations.ipynb
matthewfeickert, line 1: Thanks @kratsg. I think my final question is still https://github.com/scikit-hep/pyhf/pull/2690#pullrequestreview-4058899749, which is if we are commenting out things like 

```python
# allset_all_histo_nom = histogramssets[:, :, 1]
```

in `docs/examples/notebooks/learn/TensorizingInterpolations.ipynb`, then can we not just delete that line? Were these lines not doing anything to begin with?

kratsg: they were unused.

matthewfeickert (approved):
LGTM. Thanks a bunch @kratsg!

PR #2513, opened 2024-06-05
docs: Change docs theme to PyData Sphinx theme

melissawm:
# Description
This is a simple conversion of the documentation pages to the PyData Sphinx Theme using the default configurations for the theme.

From the documentation for the theme (https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/readthedocs.html#version-switcher) the version switcher from readthedocs should still work after this is done (but won't show up before the PR is merged). If you want to use the version switcher of the PyData Sphinx Theme instead, it would require having urls for all of the previous versions of the library and one json file mapping urls to versions. This would also enable using the Version warning banner from the PyData theme itself: https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/announcements.html#version-warning-banners

Closes #2512 

# Checklist Before Requesting Reviewer

- [N/A] Tests are passing
- [X] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Switch docs theme to use the PyData Sphinx theme which is used across the
  broader Scientific Python community ubiquitously.
   - Replace sphinx-rtd-theme with pydata-sphinx-theme in 'docs' extra.
* Reorder the navigation bar to keep API reference sections visible by default.
* Remove 'View me on GitHub' ribbon.
   - The ribbon will cause issues with the theme and isn't needed anymore.
* Add Melissa Weber Mendonça to contributors list.
```

PR #2694, opened 2026-04-11
docs: Show development version warning using theme

### LINKED ISSUE #2516: Use PyData Sphinx Theme docs version switcher
> From the documentation for the theme (https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/readthedocs.html#version-switcher) the version switcher from readthedocs should still work after this is done (but won't show up before the PR is merged). If you want to use the version switcher of the PyData Sphinx Theme instead, it would require having urls for all of the previous versions of the library and one json file mapping urls to versions. This would also enable using the Version warning banner from the PyData theme itself: https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/announcements.html#version-warning-banners
>
> From https://pyhf.readthedocs.io/en/latest/ seems like we'll need the to use the PyData Sphinx Theme version switcher.

_Originally posted by @matthewfeickert in https://github.com/scikit-hep/pyhf/issues/2513#issuecomment-2154298776_

kratsg:
# Pull Request Description

~~Use the [PyData Sphinx Theme version switcher](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html) and [version warning banner](https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/announcements.html#version-warning-banners) instead of the custom hand-rolled solution currently in the docs.~~

Show the development version warning banner using the [PyData Sphinx Theme announcement banners](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/announcements.html) instead of custom CSS.

Closes #2516.

- [x] Tests are passing

# Before Merging

For the PR Assignees:

---

```
* Show the development version warning banner using the PyData Sphinx Theme
  announcement banners instead of the custom CSS.
   - c.f. https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/announcements.html
   - Set a 'is_development_build' config value to show the development warning
     banner when not build on ReadTheDocs or when the ReadTheDocs version is 'latest'.
* Remove docs/_static/css/custom.css and docs/_static/js/custom.js.
* Unshallow the ReadTheDocs checkout to get tag information to properly display
  the development version.

Assisted-by: ClaudeCode:claude-opus-5[1m]
```

## Summary by CodeRabbit

* **Documentation**
  * Improved development-build notices across the documentation site using the site’s announcement banner.
  * Removed outdated inline version-warning messages and related styling.
  * Development notices now appear only on development builds and link users toward the stable version where applicable.
* **Chores**
  * Improved Read the Docs checkout configuration to make complete history and tag information available during builds.


docs/_static/switcher.json
```diff
@@ -0,0 +1,142 @@
+[
+  {
+    "name": "v0.8.0 (dev)",
+    "version": "latest",
+    "url": "https://pyhf.readthedocs.io/en/latest/"
+  },
+  {
+    "name": "v0.7.6 (stable)",
+    "version": "stable",
+    "url": "https://pyhf.readthedocs.io/en/stable/"
```
matthewfeickert: From https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/announcements.html#version-warning-banners we must set a "preferred" version for `"show_version_warning_banner"` to work.

```suggestion
    "url": "https://pyhf.readthedocs.io/en/stable/",
    "preferred": true
```

matthewfeickert (changes requested):
One minor fix.

docs/_static/switcher.json
```diff
@@ -0,0 +1,143 @@
...
+    "name": "v0.7.6 (stable)",
+    "version": "stable",
+    "url": "https://pyhf.readthedocs.io/en/stable/",
+    "preferred": true
+  },
+  {
+    "name": "v0.7.6",
+    "version": "v0.7.6",
+    "url": "https://pyhf.readthedocs.io/en/v0.7.6/"
+  },
```
matthewfeickert: @kratsg comparing with https://github.com/pydata/pydata-sphinx-theme/blob/d17eb2e1b7438164b92d40b947e01813a105fac3/docs/_static/switcher.json#L6-L21 as an example and noting from https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/announcements.html#version-warning-banners

> This functionality relies on the [version switcher](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html#version-dropdowns) to determine the version number of the latest stable release. It will only work if your version switcher .json has exactly one entry with property "preferred": true and your entries have version properties that are parsable by the [compare-versions node module](https://www.npmjs.com/package/compare-versions)
> ...
> If the active version compares less than the preferred version, the announcement will inform the user that they are viewing an older version of the documentation and provide a link to the preferred version. If the version compares greater than the preferred version (or if the version match contains the strings “dev”, “rc” or “pre”), the announcement will say they are viewing an unstable development version instead.

then I think we need to remove `0.7.6` to keep the version switcher from viewing `0.7.6` as somehow "older" than the stable version.

```suggestion
```

matthewfeickert (changes requested):
@kratsg one comment / suggested change that I want your approval on, but then this is good to do.

docs/_static/switcher.json
matthewfeickert: > but then this is good to do.

Hm. We might need more work though as the [ReadTheDocs build](https://pyhf--2694.org.readthedocs.build/en/2694/) (and locally for me) isn't rendering any dropdown menu correctly. There is just an empty box where the options should be.

image

matthewfeickert (changes requested):
Actually, I don't think we _want_ to have the PyData Sphinx Theme version switcher enabled at all, as the ReadTheDocs website (https://pyhf.readthedocs.io/) has its own working version switcher from ReadTheDocs (c.f. https://github.com/scikit-hep/pyhf/pull/2513#issue-2336739074) and the GitHub Pages website (https://scikit-hep.org/pyhf/) is always going to be the development version as it is deployed from the `HEAD` of `main`.

So we don't want for there to be a version switcher there, we just want to retain a banner notifying people they are on a development version. Though ideally we want to do this without still needing the `docs/_static/css/custom.css`.

matthewfeickert (approved):

Issue #2706, opened 2026-06-12
pyhf review from Claude Fable


henryiii:
:robot: _AI text below_ :robot:

Review complete. I fanned out subagents across the five major subsystems, then personally verified every load-bearing claim against the source — discarding several false positives along the way (e.g. "`_TensorViewer` caches the backend" — no, its `pyhf.default_backend` use is intentional setup-time index math with a proper `tensorlib_changed` subscription; and "ToyCalculator re-fits redundantly" — the pre-fits generate toys, the per-toy fits compute the test statistic; both are required). Everything below is verified unless marked otherwise. Nothing was modified.

## Bugs

**Likely worth fixing**

1. **`src/pyhf/optimize/mixins.py:87,105` — latent `NameError` in `_internal_postprocess`.** `num_fixed_pars` is defined only inside `if uncertainties is not None:` but used unconditionally in the `if correlations is not None:` block. Today minuit sets both `unc` and `corr` so it never fires, but any result with correlations and no uncertainties crashes. Hoist the assignment above both blocks.

2. **`src/pyhf/events.py:51-75` — dead plain-function callbacks are called as `None(...)`.** For bound methods, `__call__` checks the `__self__` weakref and `_flush()` prunes dead ones. But plain functions are stored as `(weakref.ref(callback), None)`: `_flush` keeps every `arg is None` entry forever, and `__call__` line 75 invokes `func()(*args)` without checking, so a garbage-collected function (e.g. a locally-defined subscriber, common in tests) raises `TypeError: 'NoneType' object is not callable` on the next `set_backend`.

3. **`src/pyhf/parameters/utils.py:46-62` — two defects in `reduce_paramsets_requirements`.** (a) After `combined_paramset[k].pop()`, the `continue` for `"undefined"` leaves an *empty set* as the value for every attribute the modifier doesn't support; downstream only survives because paramset constructors test truthiness (`if sigmas:`). (b) The `elif` ordering compares user lists against the literal string `"undefined"`, so configuring an unsupported attribute (e.g. `sigmas` on a normsys parameter) with anything other than exactly 9 elements yields *"expected 9"* instead of the intended *"does not use the sigmas attribute"* error. Check `default_v == "undefined"` first and `del` the key on skip.

[...]

PR #2711, opened 2026-06-12
fix: Close file handles in readxml filecache and warn on inactive StatError

henryiii:
:robot: _AI text below_ :robot:

Part of the code review in #2706.

## Summary

- `src/pyhf/readxml.py`: `modtag.attrib["Activate"]` raised a raw `KeyError` when a `<StatError>` element lacked the attribute, aborting the whole conversion with no indication of which element was at fault. Changed to `modtag.attrib.get("Activate", "False") == "True"`, consistent with all sibling attribute accesses.
- `src/pyhf/readxml.py`: That default alone would trade a loud failure for a silent one: `HistFactorySchema.dtd` declares `Activate` as `#REQUIRED` (ROOT's `hist2workspace` DTD-validates and rejects XML that omits it, while `readxml` does not validate), and silently dropping the modifier understates the MC statistical uncertainty in the resulting model — and since `staterror` is channel-shared, it also distorts the constraint for sibling samples. A `StatError` that is not activated (missing `Activate` or not `"True"`) is therefore now handled by a dedicated branch that warns with the sample name, channel name, and `Activate` value, instead of falling through to the generic "not considering modifier tag" warning that names neither.
- `src/pyhf/readxml.py` `clear_filecache`: the global `__FILECACHE__` stores open uproot file handles, and clearing the dict previously abandoned them without closing. Each handle is now closed individually in a `try`/`except OSError` that logs a warning naming the file, so one failing close neither aborts the loop nor passes silently. The cache is emptied in place with `.clear()` so external references don't retain closed handles, and the function gained a docstring documenting the close behavior.
- `tests/test_import.py`: regression tests pin both fixes — `test_process_sample_staterror_inactive` (no `staterror` modifier and the targeted warning for missing/`"False"` `Activate`) and `test_clear_filecache_closes_handles` (all handles closed even when one `close()` raises, cache emptied in place). Reverting either fix now fails the suite.
- ~~`src/pyhf/cli/infer.py` (two sites) and `src/pyhf/cli/rootio.py` (one site): `json.loads(click.open_file(...).read())` leaked file handles. Replaced with explicit `with click.open_file(...) as f: json.load(f)` context managers, matching the style used everywhere else in those files.~~

## Test plan

- [x] `pytest tests/test_import.py -q` — 24 tests pass, including the two new regression tests
- [x] Full core suite via `pixi run test` — 1124 passed (the single `tests/test_regression.py` failure is pre-existing on the base commit and specific to the local environment)
- [x] Branch coverage on `readxml.py` confirms the inactive-`StatError` warning branch and the `OSError` path in `clear_filecache` are exercised
- [x] `prek run --all-files` (pre-commit / ruff / mypy) — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* readxml.py: Use .get("Activate", "False") on StatError to avoid a
  KeyError when the attribute is absent.
* readxml.py: As HistFactorySchema.dtd declares Activate as #REQUIRED
  (and hist2workspace validates and rejects XML that omits it), add a
  dedicated branch for StatError modifiers that are not activated that
  warns with the sample name, channel, and Activate value instead of
  the generic "not considering modifier tag" message, as silently
  dropping the modifier would understate the MC statistical
  uncertainty in the resulting model.
* readxml.py: Close cached uproot file handles in clear_filecache()
  individually, logging a warning naming the file on OSError so that
  one failing close doesn't leak the remaining handles, and empty the
  cache in place so external references don't retain closed handles.
* Add regression coverage for both behaviors: a StatError with
  missing Activate or Activate="False" produces no staterror modifier
  and emits the targeted warning, and clear_filecache() closes every
  cached handle even when one close raises OSError.

Assisted-by: ClaudeCode:claude-fable-5
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-cache cleanup to safely close cached files, even when a close operation fails.
  * XML parsing now handles missing `Activate` attributes by treating them as disabled instead of raising an error.
  * Inactive statistical error modifiers are now skipped with a warning.

### DIFF
--- a/src/pyhf/cli/infer.py
+++ b/src/pyhf/cli/infer.py
@@ -99,10 +99,10 @@ def fit(
     with click.open_file(workspace, "r", encoding="utf-8") as specstream:
         spec = json.load(specstream)
     ws = Workspace(spec)
-    patches = [
-        json.loads(click.open_file(pfile, "r", encoding="utf-8").read())
-        for pfile in patch
-    ]
+    patches = []
+    for pfile in patch:
+        with click.open_file(pfile, "r", encoding="utf-8") as f:
+            patches.append(json.load(f))
 
     model = ws.model(
         measurement_name=measurement,
@@ -195,10 +195,10 @@ def cls(
 
     ws = Workspace(spec)
 
-    patches = [
-        json.loads(click.open_file(pfile, "r", encoding="utf-8").read())
-        for pfile in patch
-    ]
+    patches = []
+    for pfile in patch:
+        with click.open_file(pfile, "r", encoding="utf-8") as f:
+            patches.append(json.load(f))
     model = ws.model(
         measurement_name=measurement,
         patches=patches,
--- a/src/pyhf/cli/rootio.py
+++ b/src/pyhf/cli/rootio.py
@@ -94,7 +94,8 @@ def json2xml(workspace, output_dir, specroot, dataroot, resultprefix, patch):
     with click.open_file(workspace, "r", encoding="utf-8") as specstream:
         spec = json.load(specstream)
         for pfile in patch:
-            patch = json.loads(click.open_file(pfile, "r", encoding="utf-8").read())
+            with click.open_file(pfile, "r", encoding="utf-8") as pf:
+                patch = json.load(pf)
             spec = jsonpatch.JsonPatch(patch).apply(spec)
         Path(output_dir).joinpath(specroot).mkdir(parents=True, exist_ok=True)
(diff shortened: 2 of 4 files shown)


matthewfeickert:
@kratsg this is ready for review now.

@henryiii I made some additional changes and added tests. If you want to give feedback that's welcome.

matthewfeickert (approved):

kratsg (approved):

PR #2716, opened 2026-06-12
refactor: Migrate from jsonschema.RefResolver to referencing

### LINKED ISSUE #2121: drop refresolver in favor of the new referencing library
### Summary

See the comment here: https://github.com/orgs/python-jsonschema/discussions/995#discussioncomment-5047193

### Additional Information

_No response_

### Code of Conduct

### LINKED ISSUE #2139: jsonschema v4.18.0 will introduce breaking changes
Using [`jsonschema` `4.18.0a1`](https://pypi.org/project/jsonschema/4.18.0a1/) in the release candidate tests causes the tests to fail as the following `DeprecationWarning` is introduced when `jsonschema.RefResolver` is used.

```
DeprecationWarning: jsonschema.RefResolver is deprecated as of v4.18.0, in favor of the https://github.com/python-jsonschema/referencing library, which provides more compliant referencing behavior as well as more flexible APIs for customization. A future release will remove RefResolver. Please file a feature request (on referencing) if you are missing an API for the kind of customization you need.
```

This will require new releases of `pyhf` to fix to avoid causing future problems for `cabinetry` and other downstream users.

henryiii:
:robot: _AI text below_ :robot:

Part of the code review in #2706.

* Resolves #2139
* Resolves https://github.com/scikit-hep/pyhf/issues/2121

## Summary

`jsonschema.RefResolver` has been deprecated since jsonschema 4.18. This migrates pyhf's schema validation to the [`referencing`](https://github.com/python-jsonschema/referencing) library:

- `src/pyhf/schema/validator.py` now builds a `referencing.Registry` with a `retrieve` callback (`_retrieve_schema`) that loads referenced schemas from disk, and passes `registry=...` to the `jsonschema` validator instead of the deprecated `resolver=...`.
- The pyhf schemas carry a top-level `$ref` (e.g. `model.json` -> `defs.json#/definitions/model`). Under draft-06, a sibling `$ref` suppresses `$id`, so jsonschema cannot infer the root schema's base URI. The resolver base URI is therefore anchored explicitly to the schema `$id` so relative references like `defs.json` resolve correctly.
- A new `loader.read_schema` reads a schema directly from disk **without** consulting or mutating `SCHEMA_CACHE`, so resolving `$ref`s does not pollute the cache. This preserves the existing caching semantics, including the `pyhf.schema(new_path)` custom-schema-path context manager and the `InvalidSpecification` exception behavior.
- Bumped `jsonschema>=4.15.0` -> `jsonschema>=4.18.0` and added `referencing>=0.28.4` as an explicit dependency in both the project and pixi dependency tables (it is now imported directly).
- Removed the now-unneeded `RefResolver` `DeprecationWarning` filter. Since pytest runs with `filterwarnings=error`, any remaining deprecation path would now fail the suite.

## Test plan

- [x] `uv pip install --group dev --editable .`
- [x] `pytest tests/test_schema.py` (70 passed)
- [x] `pytest tests/test_workspace.py tests/test_pdf.py` (308 passed, 6 skipped)
- [x] `pytest tests/test_validation.py tests/test_patchset.py tests/test_public_api.py` (all pass)
- [x] `pytest --doctest-modules src/pyhf/schema/` (3 passed)
- [x] `prek -a` (lint clean)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
* Replace the deprecated jsonschema.RefResolver with a referencing.Registry
  whose retrieve callback loads schemas from pyhf.schema.path, and pass
  'registry=' to the validator instead of the deprecated 'resolver=' argument.
   - c.f. https://python-jsonschema.readthedocs.io/en/stable/referencing/
* Enter validation through the absolute form of a schema's top-level $ref
  (e.g. defs.json#/definitions/model), resolved against the directory of
  the requested version under pyhf.schema.path rather than the schema's own
  $id, as the RefResolver base_uri was. Under draft-06 a sibling $ref
  suppresses $id, so jsonschema cannot infer the root base URI from the
  schema itself, and anchoring at the on-disk directory means a stale or
  copy-pasted $id cannot redirect defs.json to another version. Entering at
  the referenced fragment keeps the tensor-aware validator class, which
  referencing the whole document by $id would discard.
* Key SCHEMA_CACHE by path relative to pyhf.schema.path instead of by $id so
  that lookups hit for both the bundled schemas and custom schemas with
  relative $ids. The previous lookup key was built with pathlib, which
  collapsed 'https://' to 'https:/', so the cache never hit.
* Route $ref retrieval through load_schema so that referenced schemas such
  as defs.json are cached after the first load.
* Raise 'pyhf.exceptions.SchemaNotFound' from 'pyhf.schema.validate' when a $ref
  cannot be resolved, naming the unresolved reference.
* Bump the minimum jsonschema to v4.18.0, add referencing as an explicit
  dependency with its version controlled by jsonschema, and drop the
  RefResolver DeprecationWarning filter.
* Add tests for _retrieve_schema, schema cache hits, and unresolvable
  references.

Assisted-by: ClaudeCode:claude-fable-5-1
```

## Improvements

- Improved JSON Schema validation for nested and relative `$ref` references.
- Increased reliability when loading and caching bundled or custom schemas.
- Improved resolution of referenced schema resources based on the requested schema location.

## Bug Fixes

- Improved error messages when referenced schema files cannot be resolved.
- Ensured consistent schema cache lookups across validation and direct schema reads.

## Tests

- Expanded coverage for schema caching, referenced resources, and resolution errors.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,8 @@ classifiers = [
 dependencies = [
     "click>=8.0.0",  # for console scripts
     "jsonpatch>=1.15",
-    "jsonschema>=4.15.0",  # for utils
+    "jsonschema>=4.18.0",  # for utils; 4.18.0 for referencing-based ref resolution
+    "referencing>=0.28.4",  # for schema $ref resolution (replaces jsonschema.RefResolver)
     "pyyaml>=5.1",  # for parsing CLI equal-delimited options
     # c.f. https://github.com/scikit-hep/pyhf/issues/2593 for excluded v1.16.x versions
     "scipy>=1.5.4,!=1.16.0,!=1.16.1,!=1.16.2",  # requires numpy, which is required by pyhf
@@ -179,7 +180,6 @@ filterwarnings = [
     'ignore:divide by zero encountered in (true_)?divide:RuntimeWarning',  #FIXME: pytest tests/test_tensor.py::test_pdf_calculations[numpy]
     'ignore:[A-Z]+ is deprecated and will be removed in Pillow 10:DeprecationWarning',  # keras
     "ignore:ml_dtypes.float8_e4m3b11 is deprecated.",  #FIXME: Can remove when jaxlib>=0.4.12
-    "ignore:jsonschema.RefResolver is deprecated as of v4.18.0, in favor of the:DeprecationWarning",  # Issue #2139
     "ignore:Skipping device Apple Paravirtual device that does not support Metal 2.0:UserWarning",  # Can't fix given hardware/virtualized device
     "ignore:jax.xla_computation is deprecated. Please use the AOT APIs:DeprecationWarning",  # jax v0.4.30
     "ignore:'MultiCommand' is deprecated and will be removed in Click 9.0. Use 'Group' instead.:DeprecationWarning",  # Click
@@ -330,7 +330,8 @@ pyhf = { path = ".", editable = true }
 [tool.pixi.dependencies]
 click = ">=8.0.0"
 jsonpatch = ">=1.15"
-jsonschema = ">=4.15.0"
+jsonschema = ">=4.18.0"
+referencing = ">=0.28.4"
 pyyaml = ">=5.1"
 scipy = ">=1.5.4,!=1.16.0,!=1.16.1,!=1.16.2"
 rich = ">=12.3.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ dependencies = [
     "click>=8.0.0",  # for console scripts
     "jsonpatch>=1.15",
     "jsonschema>=4.18.0",  # for utils; 4.18.0 for referencing-based ref resolution
-    "referencing>=0.28.4",  # for schema $ref resolution (replaces jsonschema.RefResolver)
+    "referencing",  # for schema $ref resolution, version controlled by jsonschema
     "pyyaml>=5.1",  # for parsing CLI equal-delimited options
     # c.f. https://github.com/scikit-hep/pyhf/issues/2593 for excluded v1.16.x versions
     "scipy>=1.7.2,!=1.16.0,!=1.16.1,!=1.16.2",  # requires numpy, which is required by pyhf
@@ -334,7 +334,7 @@ pyhf = { path = ".", editable = true }
 click = ">=8.0.0"
 jsonpatch = ">=1.15"
 jsonschema = ">=4.18.0"
-referencing = ">=0.28.4"
+referencing = "*"
 pyyaml = ">=5.1"
 scipy = ">=1.7.2,!=1.16.0,!=1.16.1,!=1.16.2"
 rich = ">=12.3.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,8 @@ classifiers = [
 dependencies = [
     "click>=8.0.0",  # for console scripts
     "jsonpatch>=1.15",
-    "jsonschema>=4.15.0",  # for utils
+    "jsonschema>=4.18.0",  # for utils; 4.18.0 for referencing-based ref resolution
+    "referencing>=0.28.4",  # for schema $ref resolution (replaces jsonschema.RefResolver)
     "pyyaml>=5.1",  # for parsing CLI equal-delimited options
     # c.f. https://github.com/scikit-hep/pyhf/issues/2593 for excluded v1.16.x versions
     "scipy>=1.7.2,!=1.16.0,!=1.16.1,!=1.16.2",  # requires numpy, which is required by pyhf
@@ -179,7 +180,6 @@ filterwarnings = [
     'ignore:divide by zero encountered in (true_)?divide:RuntimeWarning',  #FIXME: pytest tests/test_tensor.py::test_pdf_calculations[numpy]
     'ignore:[A-Z]+ is deprecated and will be removed in Pillow 10:DeprecationWarning',  # keras
     "ignore:ml_dtypes.float8_e4m3b11 is deprecated.",  #FIXME: Can remove when jaxlib>=0.4.12
-    "ignore:jsonschema.RefResolver is deprecated as of v4.18.0, in favor of the:DeprecationWarning",  # Issue #2139
     "ignore:Skipping device Apple Paravirtual device that does not support Metal 2.0:UserWarning",  # Can't fix given hardware/virtualized device
     "ignore:jax.xla_computation is deprecated. Please use the AOT APIs:DeprecationWarning",  # jax v0.4.30
     "ignore:'MultiCommand' is deprecated and will be removed in Click 9.0. Use 'Group' instead.:DeprecationWarning",  # Click
@@ -333,7 +333,8 @@ pyhf = { path = ".", editable = true }
 [tool.pixi.dependencies]
 click = ">=8.0.0"
 jsonpatch = ">=1.15"
-jsonschema = ">=4.15.0"
+jsonschema = ">=4.18.0"
+referencing = ">=0.28.4"
 pyyaml = ">=5.1"
 scipy = ">=1.7.2,!=1.16.0,!=1.16.1,!=1.16.2"
 rich = ">=12.3.0"
--- a/src/pyhf/schema/validator.py
+++ b/src/pyhf/schema/validator.py
@@ -5,11 +5,13 @@ from collections.abc import Mapping
 from pathlib import Path
 
 import jsonschema
+from referencing import Registry, Resource
+from referencing.jsonschema import DRAFT6
 
 import pyhf.exceptions
 from pyhf import tensor
 from pyhf.schema import variables
-from pyhf.schema.loader import load_schema
+from pyhf.schema.loader import load_schema, read_schema
 
 
 def _is_array_or_tensor(_checker, instance):
@@ -34,6 +36,25 @@ def _is_number_or_tensor_subtype(checker, instance):
     if is_number:
         return True
     return isinstance(instance, (numbers.Number, *tensor.array_subtypes))
+
+
+def _retrieve_schema(uri: str) -> Resource:
+    """
+    A ``referencing`` retrieve callback that loads a pyhf schema by its URI.
+
+    Cross-schema ``$ref``\\ s (e.g. ``defs.json``) are resolved against the
+    referring schema's ``$id``. For the bundled schemas these are absolute URIs
+    under :data:`pyhf.schema.variables.SCHEMA_BASE`; for custom schemas (see
+    :class:`pyhf.schema.Schema`) they are paths relative to
+    :attr:`pyhf.schema.path`. In both cases stripping the base leaves the disk
+    path relative to :attr:`pyhf.schema.path`.
+
+    The schema is read directly from disk via :func:`read_schema` so that
+    resolving references does not pollute :data:`pyhf.schema.variables.SCHEMA_CACHE`.
+    """
+    schema_id = uri.removeprefix(variables.SCHEMA_BASE)
+    schema = read_schema(schema_id)
+    return Resource.from_contents(schema, default_specification=DRAFT6)
 
 
 def validate(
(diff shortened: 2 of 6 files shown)


matthewfeickert:
@kratsg as our schema expert can you review this?

pyproject.toml
```diff
@@ -45,7 +45,8 @@ classifiers = [
 dependencies = [
     "click>=8.0.0",  # for console scripts
     "jsonpatch>=1.15",
-    "jsonschema>=4.15.0",  # for utils
+    "jsonschema>=4.18.0",  # for utils; 4.18.0 for referencing-based ref resolution
+    "referencing>=0.28.4",  # for schema $ref resolution (replaces jsonschema.RefResolver)
```
matthewfeickert: As [`jsonschema` `v4.18.0` added `referencing>=0.28.4` as a dependency](https://github.com/python-jsonschema/jsonschema/compare/v4.17.3...v4.18.0) we don't have a real need to give the lower bounds ourselves a `jsonschema` enforces that for us, in a similar way to how `scipy` enforces a lower bound for `numpy`.

_Unless_ we know that independently that the behavior of

```python

from referencing import Registry, Resource
from referencing.jsonschema import DRAFT6
```

was added in `referencing` `v0.28.4`.

```suggestion
    "jsonschema>=4.18.0",  # for utils; 4.18.0 for referencing-based ref resolution
    "referencing",  # for schema $ref resolution, version controlled by jsonschema
```

matthewfeickert (changes requested):

src/pyhf/schema/validator.py
```diff
@@ -93,7 +115,14 @@ def validate(
...
 
-    validator = Validator(schema, resolver=resolver, format_checker=None)
+    validator = Validator(schema, registry=registry, format_checker=None)
+
+    # The pyhf schemas carry a top-level ``$ref`` (e.g. ``model.json`` points at
+    # ``defs.json#/definitions/model``). Under draft-06 a sibling ``$ref``
+    # suppresses ``$id``, so jsonschema cannot infer the base URI of the root
+    # schema and relative references like ``defs.json`` fail to resolve. Anchor
+    # the resolver at the schema ``$id`` so those references resolve correctly.
+    validator._resolver = registry.resolver(base_uri=schema_id)
```
kratsg: I partially agree with coderabbit above but would like to take this further. There is a public API for this we should probably use rather than messing with internals. That is we should validate through a ref wrapper `Draft6Validator({"$ref": schema_id}, registry=registry, ...)`

kratsg (changes requested):
I think this isn't fully yet finished. There's some things I note which I'll summarize as bulleted points just for conciseness.

- `registry=` signature is in `jsonschema 4.18.0`, which itself requires `referencing>=0.28.4`. As @matthewfeickert noted, this is handled so we don't need to manually pin the lower-bound ourselves
- after validation, only the top-level schema is in `SCHEMA_CACHE`, but **not** `defs.json` which is a regression imo. Since `defs.json` is usually the largest file we'd want to cache and the other schemas are trivially small, I'm not sure there's a benefit for `SCHEMA_CACHE` if we can't cache `defs.json` after first load
- `validator._resolver = registry.resolver(base_uri=schema_id)` writes a private attrs field and this should instead be fixed to use the public api
- `read_schema` bypasses the cache, and `SCHEMA_CACHE` lookup in `load_schema` doesn't seem to be hitting the cache as `Path()` transforms `https://` to `https:/`. This is also a problem on `main` now I believe, but after this PR is merged in, then `SCHEMA_CACHE` is effectively an appendix that needs to be removed ;) 
- a custom schema with an absolute `$id` not under `SCHEMA_BASE` falls through `removeprefix` and raises a confusing/misleading `SchemaNotFound`
- `read_schema` and `_retrieve_schema` don't have explicit unit tests

pyproject.toml
matthewfeickert: Turns out the APIs pyhf uses date from referencing 0.2.0 to 0.12.0, and jsonschema 4.18.0's 0.28.4 floor is only a type-annotation release that jsonschema happened to track on its own release day.

**`referencing` APIs used by pyhf: first shipping versions**

| Symbol used by pyhf | First tag | Commit | Date |
|---|---|---|---|
| `Registry`, `Registry.with_resource`, `Registry.resolver` | v0.2.0 | [fc128a2](https://github.com/python-jsonschema/referencing/commit/fc128a24e82441e7e9026bde66b6a8fcfd52b743) "Cowboy code a conservative initial implementation" | 2022-12-11 |
| `default_specification=` keyword | v0.7.4 | [4dc3668](https://github.com/python-jsonschema/referencing/commit/4dc36683) "Combine specification_for until it needs to be separate" | 2022-12-22 |
| `referencing.jsonschema.DRAFT6` | v0.8.0 | [7f28317](https://github.com/python-jsonschema/referencing/commit/7f283175) "Dedupe by having one concrete implementation of Specification" | 2022-12-26 |
| `Resource.from_contents` | v0.9.0 | [9d4b2b7](https://github.com/python-jsonschema/referencing/commit/9d4b2b72) "Add a mini-implementation of Resource.id" | 2023-01-05 |
| `referencing.exceptions.Unresolvable` | v0.9.0 | [ad7e614](https://github.com/python-jsonschema/referencing/commit/ad7e6142) "Raise Unresolvable when reference lookup fails" | 2023-01-10 |
| `Registry(retrieve=...)` and `referencing.exceptions.NoSuchResource` | v0.12.0 | [93e5985](https://github.com/python-jsonschema/referencing/commit/93e59859) "Add a way to dynamically retrieve resources in a Registry" | 2023-02-08 |

The retrieve-callback flow is the last of these to arrive, so `referencing` v0.12.0 is the true runtime floor for the code in this PR (verified in a venv with `referencing==0.12.0`).

`referencing` v0.28.4 is a single commit, [de0d258](https://github.com/python-jsonschema/referencing/commit/de0d2583) "More explicit field aliasing for silly type checker behavior" (2023-05-23), whose changelog entry is "Fix a type annotation." It adds `alias="retrieve"` to the attrs field so static type checkers recognise the `retrieve=` keyword; at runtime that keyword already worked because attrs strips the leading underscore from `_retrieve`.

`jsonschema` pins `referencing>=0.28.4` via [d00d732](https://github.com/python-jsonschema/jsonschema/commit/d00d732) "Update requirements." (2023-05-23, same day as the `referencing` v0.28.4 tag), the last of fourteen floor bumps made during the 4.18.0 development cycle that simply tracked the latest `referencing` release.

`referencing`'s `CHANGELOG.rst` only starts at v0.24.0 and the early GitHub releases are auto-generated compare links (e.g. [v0.11.1...v0.12.0](https://github.com/python-jsonschema/referencing/compare/v0.11.1...v0.12.0)), so the commits above are the primary record.

matthewfeickert:
@kratsg I think that your review comments from https://github.com/scikit-hep/pyhf/pull/2716#pullrequestreview-4938021539 are addressed now. I'll take another look after dinner but I did a pass on everything you raised so far.

---

Claude Code generated text follow (sorry for not reducing it myself):

All points from https://github.com/scikit-hep/pyhf/pull/2716#pullrequestreview-4938021539 are addressed in three commits on the branch: `c330c1f7` (refactor), `17691f33` (tests), `195d46bd` (build). Item by item:

## 1. `referencing` lower bound

Done. `referencing` is now unpinned in both `[project].dependencies` and `[tool.pixi.dependencies]` and left to `jsonschema>=4.18.0` to control. `tests/constraints.txt` keeps `referencing==0.28.4` so the lower-bound job exercises the floor jsonschema implies. For the record, everything pyhf imports from `referencing` dates from v0.2.0 to v0.12.0; v0.28.4 itself is a one-line type-annotation fix that jsonschema happened to track on its release day.

## 2. `defs.json` not cached after validation

Done. `_retrieve_schema` now goes through `load_schema`, so `defs.json` (and anything else a schema references) is cached in `SCHEMA_CACHE` after the first load. `read_schema` is removed, so there is a single read path again. Covered by `test_validate_caches_referenced_schema`, which starts from an empty cache and asserts both `1.0.0/model.json` and `1.0.0/defs.json` are present afterwards.

## 3. Private `validator._resolver` write

Done, only the public API is used now. One wrinkle with the literal `Draft6Validator({"$ref": schema_id}, registry=registry)` form you suggested: when jsonschema descends through that `$ref` into the whole `model.json` document, `Validator.evolve` calls `validator_for()` on the resolved contents and, because the document carries `$schema`, swaps in the stock `Draft6Validator`. That discards the tensor-aware type checker from `validators.extend`, and `test_schema_tensor_type_allowed[*]` fails with `array([10.]) is not of type 'array'`. So validation instead enters through the absolute form of the schema's own top-level `$ref` (`defs.json#/definitions/model` resolved against the schema `$id`). The fragment has no `$schema`, so the extended validator survives, and the resolver is anchored at the referenced schema's URI. Draft-06 ignores the siblings of `$ref`, so this is equivalent to validating against the schema itself; `ValidationError.path` and `schema_path` are byte-identical to before. A `$ref`-less custom schema falls back to native `$id` handling. The reasoning is in a comment above the `root = ...` line in `validator.py`.

## 4. `read_schema` bypasses the cache and `Path()` breaks the lookup

Done. `SCHEMA_CACHE` is now keyed by the path relative to `pyhf.schema.path` (e.g. `"1.0.0/defs.json"`) instead of by `$id`. That is the only key that hits for both the bundled schemas (absolute `$id` under `SCHEMA_BASE`) and custom schemas such as `tests/test_schema/customschema` (relative `$id`s), and it removes the `Path(SCHEMA_BASE)` juggling entirely. `test_load_schema_cache_hit` is the regression test; it fails on `main`. I kept `SCHEMA_CACHE` rather than removing it because point 2 wants `defs.json` cached, and with the retrieve callback going through `load_schema` the cache is the natural place for that. The `SCHEMA_CACHE.clear()` in `pyhf.schema(path)` is now load-bearing (two search paths could both have a `1.1.0/defs.json`) and has a comment saying so. Replacing the dict with a module-level `Registry` could be a follow-up if you'd prefer that shape.

## 5. Custom schema with an absolute `$id` not under `SCHEMA_BASE`

Done. `_retrieve_schema` raises `referencing.exceptions.NoSuchResource` for any absolute URI (scheme or netloc present) that is not under `SCHEMA_BASE`, since nothing under `pyhf.schema.path` can satisfy it. `validate()` then catches `referencing.exceptions.Unresolvable` and raises `pyhf.exceptions.SchemaNotFound` naming the unresolved reference and the two accepted forms (an `$id` under `SCHEMA_BASE`, or a path relative to `pyhf.schema.path`). For a reference that does map onto `pyhf.schema.path` but whose file is missing, the original `SchemaNotFound` from `load_schema` is left to propagate so the cause chain still shows the disk path that was tried. Covered by `test_validate_unresolvable_ref[foreign_id]` and `test_validate_unresolvable_ref[missing_defs]`.

## 6. No unit tests for `read_schema` and `_retrieve_schema`

Done. `read_schema` no longer exists. New tests in `tests/test_schema.py`: `test_load_schema_cache_hit`, `test_retrieve_schema_bundled`, `test_retrieve_schema_foreign_absolute_id` (parametrized over `https://example.com/...` and `//example.com/...`), `test_retrieve_schema_missing`, `test_validate_caches_referenced_schema`, and `test_validate_unresolvable_ref` (two cases). The existing custom-schema tests now assert the exact cache keys (`1.1.0/workspace.json` and `1.1.0/defs.json`) instead of a count.

## Verification

`tests/test_schema.py` (78 passed) plus `test_workspace.py`, `test_patchset.py`, `test_validation.py`, `test_public_api_repr.py`, `test_pdf.py` (386 passed, 6 pre-existing backend skips) and the schema doctests all pass in the pixi `test` environment, and all pre-commit hooks pass. Since the lower-bound workflow only runs on schedule, I also ran `tests/test_schema.py` on Python 3.9 with `jsonschema==4.18.0` and `referencing==0.28.4`: 76 passed, with the only two errors being the `jax` backend not installed in that venv.

src/pyhf/schema/validator.py
```diff
@@ -93,9 +115,33 @@ def validate(
...
+    # ``$ref``, so this is equivalent to validating against the schema itself.
+    # (Referencing the document by its ``$id`` would instead make jsonschema
+    # re-select the stock Draft6Validator from the document's ``$schema`` and
+    # drop the tensor-aware type checker, c.f. jsonschema.validators.validator_for.)
+    root = (
+        {"$ref": urljoin(schema["$id"], schema["$ref"])} if "$ref" in schema else schema
+    )
+    validator = Validator(
+        root, registry=Registry(retrieve=_retrieve_schema), format_checker=None
+    )
```
kratsg: there's a regression here. in the way we did things with pyhf and `RefResolver`, one first defaulted to on-disk locations of the schema before doing the `$id` resolution shown here. Basically, the physical directory was the base URI that was preloaded and used. This was the request for some people who are running pyhf in areas with limited/restricted egress or otherwise "offline" mode, after shipping the preinstalled `pyhf` package around.

Example is

`1.1.0/model.json` with a $id pointing at 1.0.0, (e.g. I copy-pasted from an older version without updating it):

```json
  {
      "$schema": "http://json-schema.org/draft-06/schema#",
      "$id": "1.0.0/model.json",
      "$ref": "defs.json#/definitions/model"
  }
```

and if I have the `1.1.0/defs.json` use (e.g. some change in the schema like a `must_have_marker` requirement or something like that)

```json
  {
      "$schema": "http://json-schema.org/draft-06/schema#",
      "$id": "1.1.0/defs.json",
      "definitions": {
          "model": {
              "type": "object",
              "properties": { "must_have_marker": {"type": "string"} },
              "required": ["must_have_marker"],
              "additionalProperties": false
          }
      }
  }
```

while

`1.0.0/defs.json` use

```json
  {
      "$schema": "http://json-schema.org/draft-06/schema#",
      "$id": "1.0.0/defs.json",
      "definitions": {
          "model": {
              "type": "object",
              "properties": { "must_have_marker": {"type": "string"} },
              "additionalProperties": true
          }
      }
  }
```

the problem is that on current main, `pyhf.schema.validate({}, "model.json", version="1.1.0")` fails validation (correctly) while it passes in this PR. So we've got a regression and can't catch newer schema versions.

kratsg (changes requested):

kratsg (approved):

PR #2741, opened 2026-08-11
ci: Harden semantic PR check against pull_request_target risks

matthewfeickert:
# Description

* While the `pull_request_target trigger` is required by https://github.com/amannn/action-semantic-pull-request to validate PRs from forks with the workflow definition taken from the default branch, its elevated `GITHUB_TOKEN` is hardened.
   -  Pin `amannn/action-semantic-pull-request` to the full commit SHA so a moved tag can't run unreviewed code with the elevated token.
   -  Removed `'statuses: write'` job permission as unused in `v6+` of the action, and so use `'pull-requests: read'`.
   -  Set the workflow-level default permissions to deny-all (`'{}'`). The only job declares its own permissions block, so this only guards any future jobs added without one.
   -  Note why the trigger is acceptable here (no checkout or execution of PR-controlled content) and that future revisions must not change this.
* GitHub actions have also hardened pull_request_target in general so that it
  is no longer a direct security threat.
   - https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/

Assisted-by: ClaudeCode:claude-fable-5

- [x] Tests are passing

# Before Merging

For the PR Assignees:

```
ci: Harden semantic PR check against pull_request_target risks 

* While the pull_request_target trigger is required by
  amannn/action-semantic-pull-request to validate PRs from forks with the
  workflow definition taken from the default branch, its elevated
  GITHUB_TOKEN is hardened.
   -  Pin amannn/action-semantic-pull-request to the full commit SHA so a
      moved tag can't run unreviewed code with the elevated token.
   -  Removed 'statuses: write' job permission as unused in v6+ of the action,
      and so use 'pull-requests: read'.
   -  Set the workflow-level default permissions to deny-all ('{}'). The only
      job declares its own permissions block, so this only guards any
      future jobs added without one.
   -  Note why the trigger is acceptable here (no checkout or execution of
      PR-controlled content) and that future revisions must not change this.
* GitHub actions have also hardened pull_request_target in general so that it
  is no longer a direct security threat.
   - https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/

Assisted-by: ClaudeCode:claude-fable-5
```

## Summary by CodeRabbit

* **Chores**
  * Improved pull request validation workflow security.
  * Restricted workflow permissions to the minimum required access.
  * Pinned the validation action to a specific version for more predictable execution.
  * Added documentation clarifying security considerations.
  * Reduced the risk of unintended access during automated validation.

### DIFF
--- a/.github/workflows/semantic-pr-check.yml
+++ b/.github/workflows/semantic-pr-check.yml
@@ -1,5 +1,11 @@
 name: Semantic Pull Request
 
+# SECURITY: pull_request_target is required by amannn/action-semantic-pull-request
+# so the check works on PRs from forks and uses this workflow definition from the
+# default branch (PR authors can't alter the check). It runs with an elevated
+# GITHUB_TOKEN, which is safe here ONLY because this workflow never checks out or
+# executes anything from the PR and token permissions are restricted below.
+# c.f. https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/
 on:
   pull_request_target:
     types:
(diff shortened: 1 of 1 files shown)


matthewfeickert:
So, while I think we all agree that `pull_request_target` is worth avoiding, I think(?) this is safe (enough) as `pull_request_target` executes the _main repository's default branch's copy of the workflow_, not the PR's copy. For `pull_request_target` when an attacker's PR fires the opened/edited/synchronize events, GitHub runs the version of `semantic-pr-check.yml` that lives on the `main` branch of https://github.com/scikit-hep/pyhf/, and the elevated `GITHUB_TOKEN` is issued to _that_ definition. A possible attacker's modified copy in their PR is never executed. An attacker would need to get changes to `.github/workflows/semantic-pr-check.yml` merged to be able to carry out a follow up attack, but that would require a maintainer to accept those changes.

matthewfeickert:
@agriyakhetarpal pointed out on the Scientific Python Discord that

> for the vast majority of use cases, if I understand correctly, `pull_request_target` is no longer a problem since `actions/checkout@v7` as `v7` blocks checkouts from forks entirely for runs triggered by it (and similarly for `workflow_run`). The change has been (or is being?, I am not sure) backported to even as old as `actions/checkout@v4`. The only ways this trigger can cause trouble are if you run something weird that bypasses `actions/checkout` and checks out code using manual Git commands, or something else that elevates access, or suchlike. See also: https://github.com/zizmorcore/zizmor/issues/2134

PR #2693, opened 2026-04-09
ci: Add static analysis of GitHub Actions with zizmor

matthewfeickert:
# Description

Resolves #2685 

# Checklist Before Requesting Reviewer

- [x] Tests are passing
- [x] "WIP" removed from the title of the pull request
- [x] Selected an Assignee for the PR to be responsible for the log summary

# Before Merging

For the PR Assignees:

- [x] Summarize commit messages into a comprehensive review of the PR

```
* Add zizmor as a pre-commit hook.
   - Use hook's default persona of 'regular' instead of 'pedantic' for the time being.
     c.f. https://docs.zizmor.sh/usage/#using-personas
* Apply zizmor fixes to workflows.
   - https://docs.zizmor.sh/audits/#dependabot-cooldown
   - https://docs.zizmor.sh/audits/#artipacked
   - https://docs.zizmor.sh/audits/#secrets-outside-env
   - https://docs.zizmor.sh/audits/#template-injection

Co-authored-by: Lucas Colley <lucas.colley8@gmail.com>
```

## Chores

* Improved CI/CD security by preventing credentials from persisting after checkout.
* Added a seven-day cooldown for automated dependency updates.
* Enhanced workflow security validation and scanning configuration.
* Improved release and container workflow handling while preserving existing behavior.
* Added clearer descriptions to dependency testing jobs.
* Added automated pre-commit checks for workflow security configuration.

PR #2744, opened 2026-08-13
ci: harden GitHub Actions security

kratsg:
## Summary

CI security hardening pass over `.github/` (workflows, dependabot) and the pre-commit/uv/pixi config. No source, lint, or test code changed.

- Fully pinned every GitHub Action to a commit SHA with a version comment (via `actions-up`, `--min-age=7`).
- Froze every pre-commit hook rev to its SHA (via `prek auto-update --freeze --cooldown-days 7`); no hook versions changed, only SHA-pinned.
- Added a `zizmor` pre-commit hook (pedantic persona, scoped to `.github`) plus `.github/zizmor.yml`.
- Fixed everything `zizmor --persona=pedantic` reported:
  - `persist-credentials: false` added to checkout steps that don't need to push.
  - Inline explanatory comments added to all job/workflow `permissions:` entries.
  - `concurrency:` group added to every workflow missing one (`bump-version.yml`, `merged.yml`, `lower-bound-requirements.yml`, `scorecard.yml`).
  - Explicit `name:` added to previously-anonymous jobs.
  - `${{ github.event.* }}` / `${{ steps.*.outputs.* }}` values that were interpolated directly into `run:` shell blocks are now passed through `env:` instead (template-injection): `bump-version.yml`, `docker.yml`, `publish-package.yml`.
  - Fixed an always-true `if:` in `bump-version.yml` (`if: ${{ github.event.inputs.dry_run }} == 'false'` mixed `${{ }}` expansion with a string comparison). Corrected to `if: github.event.inputs.dry_run == 'false'`. **Behavior change**: previously the "Update the Git tag annotation" step ran even during dry runs (it only mutates the local runner's tag, never pushes — the actual `git push` step already gated correctly on `dry_run`), so the practical impact was minimal, but please double check this is the intended fix.
  - Corrected a garbled version comment in `scorecard.yml` (`actions/upload-artifact@043fb46...  # v4.3.7.0.1` — the SHA was already current for `v7.0.1`, only the comment was wrong, presumably left over from an earlier bad update).
- `.github/dependabot.yml`: renamed the `github-actions` group from `actions` to `github-actions`, switched its schedule from weekly to monthly, and added a 7-day cooldown to both update blocks.
- Added `exclude-newer = "7 days"` under `[tool.uv]` and `exclude-newer = "7d"` under `[tool.pixi.workspace]` in `pyproject.toml` so both resolvers skip packages published in the last 7 days.

## Major version changes

- `scorecard.yml`: `actions/checkout` v4.1.1 → v7.0.1, and `github/codeql-action/upload-sarif` v3.30.8 → v4.37.7. Both now match the SHAs already pinned everywhere else in the repo (this workflow appears to not have been kept in sync by dependabot).

## Held-back items (not changed, need a maintainer decision)

- **`docker.yml` build/deploy isolation**: the skill's guidance is to run deploy jobs separately from build jobs (build artifact → upload → separate job downloads → deploys), which `docs.yml` and `publish-package.yml` already do. `docker.yml`'s single `docker` job builds *and* pushes to Docker Hub/GHCR in the same job (with `packages: write` and Docker Hub credentials present for the whole job, including the pre-push test build). Splitting this out safely means restructuring the multi-arch (`linux/amd64,linux/arm64`) buildx push flow around artifact-transferred images or digest/manifest-list assembly, which is easy to get subtly wrong and risks breaking the Docker publishing pipeline. Left as-is; flagging for a dedicated follow-up rather than guessing at a fix.

## zizmor ignores added (`.github/zizmor.yml`)

- `excessive-permissions` on `scorecard.yml:22` (`permissions: read-all`) — this is OSSF Scorecard's own recommended starter-workflow default; the `analysis` job below it already scopes its actual write permissions down to `security-events` and `id-token`.
- `dangerous-triggers` on `semantic-pr-check.yml:9` (`pull_request_target`) — already justified by the in-file comment from #2741: required for `amannn/action-semantic-pull-request` to check fork PR titles, never checks out or executes PR content, permissions restricted to `pull-requests: read`.

## Verification

- `uvx zizmor --persona=pedantic .github` → **no findings** (2 ignored, 6 suppressed, both documented above).
- `prek run --all-files` → **all hooks pass**, including the new `zizmor` hook.
- Every `uses:` in `.github/workflows/` is pinned to a 40-character SHA with a version comment; spot-checked each SHA against the action repo's tags via the GitHub API.

## Summary by CodeRabbit

* **Chores**
  * Improved automation reliability by pinning workflow tools to immutable versions.
  * Added concurrency controls to reduce redundant or conflicting workflow runs.
  * Strengthened credential handling and dependency update safeguards.
  * Added automated security scanning for workflow configuration.
  * Configured package managers to use recently published dependency versions only.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,9 +4,9 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
     groups:
-      actions:
+      github-actions:
         patterns:
           - "*"
     labels:
@@ -14,6 +14,8 @@ updates:
       - "dependencies"
     reviewers:
       - "matthewfeickert"
+    cooldown:
+      default-days: 7
 
   # Ignore all pip dependencies to avoid PRs to update tests/constraints.txt
   - package-ecosystem: "pip"
@@ -22,3 +24,5 @@ updates:
       interval: "weekly"
     ignore:
       - dependency-name: "*"
+    cooldown:
+      default-days: 7
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -33,11 +33,17 @@ on:
         description: 'Perform a dry run to check'
         default: true
 
+concurrency:
+  # Queue rather than cancel-in-progress: this job pushes tags, and an in-flight
+  # push should not be interrupted by a subsequent trigger.
+  group: ${{ github.workflow }}-${{ github.ref }}
+
 permissions:
   contents: read
 
 jobs:
   bump-version:
+    name: Bump version
     permissions:
       contents: write  # for Git to git push
     runs-on: ubuntu-latest
@@ -46,11 +52,12 @@ jobs:
     steps:
     # Use GitHub PAT to authenticate so other workflows trigger
     - name: Checkout code
-      uses: actions/checkout@v7
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       with:
         ref: ${{ github.event.inputs.target_branch }}
         fetch-depth: 0
         token: ${{ secrets.ACCESS_TOKEN }}
+        persist-credentials: false
 
     - name: Check target branch is intended for release
       if: github.event.inputs.force == 'false'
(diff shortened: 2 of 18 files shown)


matthewfeickert (changes requested):
@kratsg this is a duplicate of PR https://github.com/scikit-hep/pyhf/pull/2693. We already did this, it just needs your review.

kratsg:
closing in favor of https://github.com/scikit-hep/pyhf/pull/2693

